PR #2, opened 2014-07-11
Update base_contact.py

TheCloneMaster:
context is now received as a frozendict, so pop is not allowed on it directly.
We need to make a copy of it first, so 

```
    if context is None:
        context = {}
```

is not enought...


rvalyi:
:+1:  you are right

bwrsandman:
:+1:

pedrobaeza:
:+1:

TheCloneMaster:
I am starting to LOVE  GIT !!

PR #15, opened 2014-07-28
[ADD] partner_relations

hbrunn:
## Introduction

This addon aims to provide generic means to model relations between partners.

Examples would be 'is sibling of' or 'is friend of', but also 'has contract X
with' or 'is assistant of'. This way, you can enode your knowledge about your
partners directly in your partner list.
## Usage

Before being able to use relations, you'll have define some first. Do that in
Sales / Configuration / Address Book / Partner relations. Here, you need to
name both sides of the relation: To have an assistant-relation, you would name
one side 'is assistant of' and the other side 'has assistant'. This relation
only makes sense between people, so you would choose 'Person' for both partner
types. For the relation 'is a competitor of', both sides would be companies,
while the relation 'has worked for' should have persons on the left side and
companies on the right side. If you leave this field empty, the relation is
applicable to all types of partners.

If you use categories to further specify the type of partners, you could for
example enforce that the 'is member of' relation can only have companies with
label 'Organization' on the left side.

Now open a partner and choose relations as appropriate in the 'Relations' tab.
## Searching partners with relations

Searching for relations is integrated transparently into the partner search
form. To find all assistants in your database, fill in 'is assistant of' and
autocomplete will propose to search for partners having this relation. Now if
you want to find Anna's assistant, you fill in 'Anna' and one of the proposals
is to search for partners having a relation with Anna. This results in Anna's
assistant(s), as you searched for assistants before.

By default, only active, not expired relations are shown. If you need to find
partners that had some relation at a certain date, fill in that date in the
search box and one of the proposals is to search for relations valid at that
date


pedrobaeza:
There are some PEP8 errors related to this module:

```
./partner_relations/model/res_partner.py:36:9: E265 block comment should start with '# '
./partner_relations/model/res_partner.py:93:17: E265 block comment should start with '# '
./partner_relations/model/res_partner.py:193:9: E265 block comment should start with '# '
./partner_relations/model/res_partner.py:194:9: E265 block comment should start with '# '
./partner_relations/model/res_partner.py:206:9: E265 block comment should start with '# '
./partner_relations/model/res_partner_relation_type_selection.py:58:9: E265 block comment should start with '# '
./partner_relations/model/res_partner_relation_type_selection.py:59:9: E265 block comment should start with '# '
./partner_relations/model/res_partner_relation_type_selection.py:60:9: E265 block comment should start with '# '
./partner_relations/model/res_partner_relation_type_selection.py:94:17: E265 block comment should start with '# '
./partner_relations/model/res_partner_relation_type_selection.py:125:9: E265 block comment should start with '# '
```

partner_relations/__init__.py
```diff
@@ -0,0 +1,21 @@
...
+#    This program is distributed in the hope that it will be useful,
+#    but WITHOUT ANY WARRANTY; without even the implied warranty of
+#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#    GNU Affero General Public License for more details.
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+import model
```
pedrobaeza: Use relative import: `from . import model`

partner_relations/model/__init__.py
```diff
@@ -0,0 +1,25 @@
...
+#    This program is distributed in the hope that it will be useful,
+#    but WITHOUT ANY WARRANTY; without even the implied warranty of
+#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#    GNU Affero General Public License for more details.
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+import res_partner
```
pedrobaeza: Use relative imports

partner_relations/model/res_partner.py
```diff
@@ -0,0 +1,238 @@
...
+from openerp.osv.expression import is_leaf
+from openerp.tools import DEFAULT_SERVER_DATE_FORMAT
+
+
+class ResPartner(orm.Model):
+    _inherit = 'res.partner'
+
+    def _get_relation_ids(
+            self, cr, uid, ids, dummy_name, dummy_arg, context=None):
+        if context is None:
```
pedrobaeza: You don't need to initialise context here

partner_relations/model/res_partner.py
```diff
@@ -0,0 +1,238 @@
...
+            context = {}
+
+        #TODO: do a permission test on returned ids
+        cr.execute(
```
pedrobaeza, line 35: Can you made it via ORM?

ploegvde:
Holger, I tested the module functional and my compliments! It is working perfect.

pedrobaeza:
@hbrunn Some remarks:
- You need to add to travis.yml the repository dependency for _web_m2x_options_.
- Why don't call it _base_partner_relations_ to follow current conventions?

hbrunn:
Pedro, thanks for your suggestions! I wanted to take this as an opportunity to upgrade the travis script to the current version from maintainer-quality-tools, but didn't manage to do so up to now.
As for the name: Do we have such a convention and why? I always experience 'base_'-prefixes as rather arbitrary...

pedrobaeza:
It's not a written convention, but to try to be as homogeneus as we can. My rule is: first put the affected area (sale, stock, purchase, or this case, base), then the model or subarea involved (order, invoice, etc, - for this case - partner), and the rest, a little explanation of the modification. That's why _base_partner_relations_ suits. Besides, we have already modules _base_partner_merge_, _base_partner_sequence_. There are another without the base_ prefix also, but I think we have to tend to homogenize all with the base_ prefix.

Do you need help with Travis configuration?

hbrunn:
no thanks, only time ;-)

hbrunn:
with #31, this works fine.

Concerning the rename: There's a very selfish argument that we use this module a lot internally and the rename would give me a nightmare of pulling it through all customer installations.

But this is not a very strong argument. I'm also generally opposed to the 'base_'-prefix in most cases because it seems very redundant to me. A bit like prefixing every module with 'odoo_'... I seriously think it should be 'partner_merge' and 'partner_sequence'. It's not interesting that the partner model happens to be defined in the base module, it's interesting that it does stuff with partners.

'sale_', 'stock_' and 'purchase_' are different beasts in my opinion (totally agree with those prefixes) because they are about an area that happens to be defined by a specific module.

pedrobaeza:
OK, you have convinced me! Let's go to partner_xxxx notation!

:+1:, as I have seen you have updated Travis in #31

bealdav:
About Name, 
Why this module name is in plural form. 98% of the modules are in singular form. Think 'is collection of' product_product for example.
It's like sql tables, why 'res_users' and 'res_groups' whereas it's a collection of ? It seems easier to remember if all are in singular, don't you think ? For new devs, why product_images and product_product?

There is a lot of cleaning since than OCA operates : branch names, flake 8, unit test, nice jobs.
But when new modules comes in OCA, the name is the most important thing because it's not easy to change? What do you think (you and the community) to have an official written rule for this ?

@hbrunn Yes it'll be difficult to change for you now, but impossible to change after years

Thanks for your answers

bealdav:
Tips to help module name change in db
http://stackoverflow.com/a/23036421/2819823

hbrunn:
Yes, please let's codify something, there's too much subjectivity in choosing names and without guidelines to refer to, it usually ends up in me liking jam for breakfast and someone else honey.

As for the names: I completely agree with object names and tables being in singular form (have a look at the code, it's `res.partner.relation`, `res.partner.relation.type` etc.). Module names however, I rather consider namespaces, and there the plural often is the way to go, cf. `from collections import...`, `itertools`, `functools` etc.

bealdav:
OK, you're right "there's too much subjectivity in choosing names and without guidelines to refer to".
May be it could be a good idea if OCA define written rules about module names included base or not base

Thanks for your reply

partner_relations/model/res_partner_relation_type.py
```diff
@@ -0,0 +1,48 @@
...
+            'Inverse name', size=128, required=True, translate=True),
+        'contact_type_left': fields.selection(
+            _get_partner_types, 'Left partner type'),
+        'contact_type_right': fields.selection(
+            _get_partner_types, 'Right partner type'),
+        'partner_category_left': fields.many2one(
+            'res.partner.category', 'Left partner category'),
+        'partner_category_right': fields.many2one(
+            'res.partner.category', 'Right partner category'),
+        }
```
veloutin: pep8 E123 closing bracket does not match indentation of opening bracket's line

max3903:
If you have suggestions to the OCA conventions, send an email to support@odoo-community.org

http://odoo-community.org/page/website.how-to

moylop260:
Hello team,
Hi I took the liberty of creating a build in our runbot instance (sorry), for those who want to do functional testing of this feature directly connecting with admin/admin in blue button.

[runbot-7.0-partner-relations](http://odoorunbot.vauxoo.com/runbot/team/partner-contact-10#7_0-partner-relations)

partner_relations/model/res_partner_relation_all.py
```diff
@@ -0,0 +1,128 @@
...
+#    GNU Affero General Public License for more details.
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+from openerp.osv.orm import Model
+from openerp.osv import fields
+from openerp.tools import drop_view_if_exists
+from res_partner_relation_type_selection import ResPartnerRelationTypeSelection
```
moylop260: Fix relative import, please
-from res_partner_relation_type_selection import ResPartnerRelationTypeSelection
+from .res_partner_relation_type_selection import ResPartnerRelationTypeSelection

moylop260:
Can you add data demo, please? 
This can help to see real configuration as you expected that should be used.

hbrunn:
Thanks for the hints! (travis fails for unrelated reasons)

max3903:
:+1:

bwrsandman:
:+1:

foutoucour:
:+1:

PR #19, opened 2014-08-04
Add res_bank_account module

bwrsandman:
Add res_bank_account module: add the title bank account, observation and active boolean fields.It allows to manage several bank accounts for each contact

Port of https://code.launchpad.net/~savoirfairelinux-openerp/partner-contact-management/partner-contact-management-base_contact_add_res_bank_account/+merge/204047
## Previous reviews:

@yvaucher:

> Description seems inaccurate to me, OpenERP already allows to manage multiple bank account per partner.
> 
> What is the plus-value when installing this module?
> I suppose this is to add details on bank accounts to describe for what they should be used.
> Why should I need a title_bank_account and an observation field on my res_partner_bank
> Wouldn't one field be enough?
> If I needed more info on res.partner.bank, I would simply add a 'description' field as on other models
> 
> title_bank_account seems useless to me as it isn't even in tree views.
> 
> Active field already exists on res.partner.bank model. By the way you overide the select=True parameter on it.
> 
> And finally this may be a matter of taste but I would name this module res_partner_bank_imp instead.
> 
> Thus sorry but I'll disaprove this one in that state.

Ana Juaristi Olalde:

@yvaucher

> title_bank_account seems useless to me as it isn't even in tree views.
> 
> This is a subjective opinion. If someone needs 2 fields, he adds 2 fields. That's all :)
> 
> @ehdem
> My oppinion is Need fixing because:
> 1. <field name="bank_ids" position="replace">
>    If you replace bank_ids with something hardcoded on your module view, you will not see new "future" core improvements on that view.
> 2. I agree with @yannick in --> Active field already exists on res.partner.bank model. By the way you overide the select=True parameter on it.

@yvaucher:

> Thanks for the review Ana
> 
> My point about title_bank_account is that most of Models have standard `name` and `description` fields.
> I agree having a second field description is quiet useful to add note on the object when you open it.
> 
> However as res.partner.bank is already a bank account, naming a field xxx_bank_account seams also redundant. Thus I think it is badly named.
> 
> Still, module description doesn't match what the module does. This is more about adding fields for informations than allowing to manage multiple bank accounts. Or am I missing something?

### DIFF
--- a/res_bank_account/__openerp__.py
+++ b/res_bank_account/__openerp__.py
@@ -1,0 +1,52 @@
+# -*- encoding: utf-8 -*-
+##############################################################################
+#
+#    OpenERP, Open Source Management Solution
+#    This module copyright (C) 2013 Savoir-faire Linux
+#    (<http://www.savoirfairelinux.com>).
+#
+#    This program is free software: you can redistribute it and/or modify
+#    it under the terms of the GNU Affero General Public License as
+#    published by the Free Software Foundation, either version 3 of the
+#    License, or (at your option) any later version.
+#
+#    This program is distributed in the hope that it will be useful,
+#    but WITHOUT ANY WARRANTY; without even the implied warranty of
+#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#    GNU Affero General Public License for more details.
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+
+{
+    'name': 'Bank Account',
+    'version': '0.1',
+    'author': 'Savoir-faire Linux',
+    'maintainer': 'Savoir-faire Linux',
+    'website': 'http://www.savoirfairelinux.com',
+    'category': 'MISC',
+    'description': """
+Bank Account
+============
+
+
+Contributors
+------------
+* El Hadji Dem (elhadji.dem@savoirfairelinux.com)
+""",
+    'depends': [
+        'account',
+    ],
+    'external_dependencies': {},
+    'data': [
+        'res_bank_view.xml',
+    ],
+    'demo': [],
+    'test': [],
+    'installable': True,
+    'active': False,
+}
+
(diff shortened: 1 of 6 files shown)


nbessi:
I wonder if it is the correct repostiory to get this module?

res_bank_account/__openerp__.py
```diff
@@ -0,0 +1,54 @@
...
+    'category': 'Customer Relationship Management',
+    'description': """
+Bank Account
+============
+
+This module allows to manage multiple bank accounts for a specific contact.
+
+It adds a bank account object linked to a bank with a name for the account, a
+description and wheter it is active or not.
+
```
hbrunn, line 38: I think this description is quite misleading, because it sounds like having multiple accounts is not possible without it. Maybe better describe the fields added and what they are good for?

res_bank_account/res_bank.py
```diff
@@ -0,0 +1,42 @@
...
+class res_partner_bank(orm.Model):
+    """Bank account"""
+    _inherit = 'res.partner.bank'
+    _columns = {
+        'title_bank_account': fields.char(
+            'Title bank account',
+            help="Name for the bank account.",
+        ),
+        'observation': fields.text(
+            'Observation',
```
hbrunn, line 35: from the French translation, wouldn't 'note' or 'comment' (that from my experience the most common names for remark fields) be a more appropriate name?

res_bank_account/res_bank.py
```diff
@@ -0,0 +1,42 @@
...
+
+from openerp.osv import orm, fields
+
+
+class res_partner_bank(orm.Model):
+    """Bank account"""
+    _inherit = 'res.partner.bank'
+    _columns = {
+        'title_bank_account': fields.char(
+            'Title bank account',
```
hbrunn, line 31: how does this relate to the 'name' fields already present in this model?

hbrunn:
@nbessi To me, it makes sense to put the module here. But I must admit I don't really understand what use case it fixes. Is that maybe something country specific?

bwrsandman:
This module turns out to be more of a client-specific alteration than something that can be used by the community at large.

Thank you for the review @hbrunn

PR #20, opened 2014-08-04
Base contact by function

bwrsandman:
- [ADD] added base_contact_function module: It allow to manage contacts by function"
- [ADD] Added base_contact_function_partner_firstname module: It allows to add firstname before name.Name is considered like the lastname and we have firstname and lastname on partner view and contact view."
  The base_contact_function_partner_firstname module depends on partner_firstname and base_contact_function module.
## Previous reviews:

@StefanRijnhart

> I have a hard time trying to find out what this module actually does. The manifest description is very terse. The name of the module reminds me of base_contact, which allowed multiple connections between the same person and a company. Is this module similar? The description should at least explain what 'institutions' are (a concept that this module introduces).
> 
> Terminology is off at times, it seems. l.130: 'acronym': is this abbreviation, or maybe stick to the standard 'code' field name and label?
> 
> Organism: I'm guessing you mean organization. But in any case, changing the 'company' terminology should be split off as a separate module.
> 
> The translations seem to contain obsolete terms, e.g. l.241 and other occurrences of 'passport'.

@bwrsandman

> This is for cases where one partner (Mike Fletcher President of Agrolait) can be part of another company (Mike Fletcher VP of AsusTEK) in a different function.
> 
> The module helps track these functions.


base_contact_by_functions/__init__.py
```diff
@@ -0,0 +1,27 @@
...
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+
+from . import res_partner
+from . import functions
+from . import res_institutions
+
+# vim:expandtab:smartindent:tabstop=4:softtabstop=4:shiftwidth=4:
```
nbessi: vim line

base_contact_by_functions/__openerp__.py
```diff
@@ -0,0 +1,56 @@
...
+    'version': '0.1',
+    'author': 'Savoir-faire Linux',
+    'maintainer': 'Savoir-faire Linux',
+    'website': 'http://www.savoirfairelinux.com',
+    'category': 'Customer Relationship Management',
+    'description': """
+Contacts by Functions
+=====================
+
+This module allows you to manage contacts by functions
```
nbessi: This should be completed with how to use it

base_contact_by_functions/__openerp__.py
```diff
@@ -0,0 +1,56 @@
...
+        'res_partner_view.xml',
+        'security/ir.model.access.csv',
+    ],
+    'demo': [],
+    'test': [],
+    'installable': True,
+    'active': False,
+}
+
+# vim:expandtab:smartindent:tabstop=4:softtabstop=4:shiftwidth=4:
```
nbessi: vim line

base_contact_by_functions/functions.py
```diff
@@ -0,0 +1,52 @@
...
+    _columns = {
+        'name': fields.char(
+            'Name',
+            required=True,
+            select=True,
+            help='Name of function.',
+        ),
+        'acronym': fields.char(
+            'Acronym',
+            size=50,
```
nbessi: maybe add a select=True on acronym?

base_contact_by_functions/functions_view.xml
```diff
@@ -0,0 +1,38 @@
...
+        </tree>
+      </field>
+    </record>
+     <!-- Form Views functions-->
+    <record model="ir.ui.view" id="functions_form_view">
+      <field name="name">Functions Form View</field>
+      <field name="model">functions</field>
+      <field name="type">form</field>
+      <field name="arch" type="xml">
+        <form string="FunctionsForm" version="7.0" >
```
nbessi: FunctionsForm -> Functions. It think that it is shown in View

base_contact_by_functions/functions.py
bwrsandman: Done

base_contact_by_functions/__openerp__.py
bwrsandman: ping @giotta

nbessi:
:+1:

base_contact_by_functions/functions.py
```diff
@@ -0,0 +1,53 @@
...
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+
+from openerp.osv import fields, orm
+
+
+class functions(orm.Model):
+    _description = 'Functions'
+    _name = 'functions'
```
gurneyalex: I'm concerned this may be a bit too generic and confusing (function fields...). Also the convention nowadays is to use a singular form for Model `_name`. 

I suggest `"res.partner.function"` for `_name`.

base_contact_by_functions/functions.py
```diff
@@ -0,0 +1,53 @@
...
+    _columns = {
+        'name': fields.char(
```
gurneyalex: should this field be translatable?

base_contact_by_functions/functions.py
```diff
@@ -0,0 +1,53 @@
...
+            'Name',
+            required=True,
+            select=True,
+            help='Name of function.',
+        ),
+        'acronym': fields.char(
```
gurneyalex: should this field be translatable?

base_contact_by_functions/res_institutions.py
```diff
@@ -0,0 +1,70 @@
...
+        ),
+    }
+    _sql_constraints = [
+        ('institution_name_unique', 'unique(name)', 'Name already exists'),
+    ]
+
+
+class res_institutions_functions(orm.Model):
+    _description = 'Institutions functions'
+    _name = 'res.institution.functions'
```
gurneyalex: same remark as above about singular form for `_name`

base_contact_by_functions/res_partner.py
```diff
@@ -0,0 +1,135 @@
...
+        ),
+    }
+
+
+class res_category_functions(orm.Model):
+    """
+    Adds this class to link category and functions
+    """
+    _description = 'Categories functions'
+    _name = 'res.category.functions'
```
gurneyalex: same remark as above regarding singular form for `_name`

base_contact_by_functions_partner_firstname/tests/test_partner.py
```diff
@@ -0,0 +1,51 @@
...
+        }, context=self.context)
+        self.vals = {
+            'name': 'test_partner',
+            'type': 'contact',
+            'contact_id': contact_id,
+        }
+
+    def test_create_partner(self):
+        cr, uid, vals, context = self.cr, self.uid, self.vals, self.context
+        self.partner_model.create(cr, uid, vals, context=context)
```
gurneyalex: could we have a couple of assertions checking that both function and first name are working properly?

base_contact_by_functions/tests/test_institutions.py
```diff
@@ -0,0 +1,103 @@
...
+    This is a base class for institution test cases.
+    Inherit from this and setup values.
+    """
+
+    def setUp(self, vals={}):
+        """
+        Setting up institution.
+        """
+        # Default test values
+        self.vals = {'name': 'This is a test institution name',
```
gurneyalex: please update the test with a unicode string for name, containing non ascii characters.

base_contact_by_functions/tests/test_institutions.py
```diff
@@ -0,0 +1,103 @@
...
+        for field in self.vals:
+            val = institution_obj[field]
+            if type(val) == browse_record:
+                self.assertEquals(self.vals[field], val.id,
+                                  "IDs for %s don't match: (%i != %i)" %
+                                  (field, self.vals[field], val.id))
+            else:
+                self.assertEquals(str(self.vals[field]), str(val),
+                                  "Values for %s don't match: (%s != %s)" %
+                                  (field, str(self.vals[field]), str(val)))
```
gurneyalex: str(val) is wrong, it will crash with a unicode string. Please fix the test to use unicode strings everywhere.

base_contact_by_functions/tests/test_institutions.py
```diff
@@ -0,0 +1,103 @@
...
+        for field in self.vals:
+            val = institution_obj[field]
+            if type(val) == browse_record:
+                self.assertNotEqual(self.vals[field], val.id,
+                                    "IDs for %s don't match: (%i != %i)" %
+                                    (field, self.vals[field], val.id))
+            else:
+                self.assertNotEqual(str(self.vals[field]), str(val),
+                                    "Values for %s don't match: (%s != %s)" %
+                                    (field, str(self.vals[field]), str(val)))
```
gurneyalex: use Unicode strings

base_contact_by_functions/tests/test_partner_category.py
```diff
@@ -0,0 +1,112 @@
...
+        for field in self.vals:
+            val = partner_category_obj[field]
+            if type(val) == browse_record:
+                self.assertEquals(self.vals[field], val.id,
+                                  "IDs for %s don't match: (%i != %i)" %
+                                  (field, self.vals[field], val.id))
+            else:
+                self.assertEquals(str(self.vals[field]), str(val),
+                                  "Values for %s don't match: (%s != %s)" %
+                                  (field, str(self.vals[field]), str(val)))
```
gurneyalex: use unicode strings

base_contact_by_functions/tests/test_partner_category.py
```diff
@@ -0,0 +1,112 @@
...
+    This is a base class for partner_category test cases.
+    Inherit from this and setup values.
+    """
+
+    def setUp(self, vals={}):
+        """
+        Setting up partner_category.
+        """
+        # Default test values
+        self.vals = {'name': 'This is a test partner_category name',
```
gurneyalex: please update the test with a unicode string with non ascii chars in name.

base_contact_by_functions/tests/test_partner_category.py
```diff
@@ -0,0 +1,112 @@
...
+        for field in self.vals:
+            val = partner_category_obj[field]
+            if type(val) == browse_record:
+                self.assertNotEqual(self.vals[field], val.id,
+                                    "IDs for %s don't match: (%i != %i)" %
+                                    (field, self.vals[field], val.id))
+            else:
+                self.assertNotEqual(str(self.vals[field]), str(val),
+                                    "Values for %s don't match: (%s != %s)" %
+                                    (field, str(self.vals[field]), str(val)))
```
gurneyalex: use unicode strings

base_contact_by_functions_partner_firstname/tests/test_partner.py
```diff
@@ -0,0 +1,51 @@
...
+        self.partner_model = self.registry("res.partner")
+        # Get context
+        self.context = self.user_model.context_get(self.cr, self.uid)
+        # Create values for test, contact also created
+        contact_id = self.partner_model.create(self.cr, self.uid, {
+            'name': 'test_contact',
+            'title': 1,
+        }, context=self.context)
+        self.vals = {
+            'name': 'test_partner',
```
gurneyalex: please test with a unicode string containing non-ascii characters for name (e.g. u"André Barnabé")

base_contact_by_functions/tests/test_functions.py
```diff
@@ -0,0 +1,101 @@
...
+        for field in self.vals:
+            val = function_obj[field]
+            if type(val) == browse_record:
+                self.assertEquals(self.vals[field], val.id,
+                                  "IDs for %s don't match: (%i != %i)" %
+                                  (field, self.vals[field], val.id))
+            else:
+                self.assertEquals(str(self.vals[field]), str(val),
+                                  "Values for %s don't match: (%s != %s)" %
+                                  (field, str(self.vals[field]), str(val)))
```
gurneyalex: don't cast potentially Unicode objects to strings. This will crash if there is a non-ascii character.

base_contact_by_functions/tests/test_functions.py
```diff
@@ -0,0 +1,101 @@
...
+        for field in self.vals:
+            val = function_obj[field]
+            if type(val) == browse_record:
+                self.assertNotEqual(self.vals[field], val.id,
+                                    "IDs for %s don't match: (%i != %i)" %
+                                    (field, self.vals[field], val.id))
+            else:
+                self.assertNotEqual(str(self.vals[field]), str(val),
+                                    "Values for %s don't match: (%s != %s)" %
+                                    (field, str(self.vals[field]), str(val)))
```
gurneyalex: fix str() call on unicode object.

base_contact_function_partner_firstname/tests/test_partner.py
```diff
@@ -0,0 +1,57 @@
...
+        self.registry('ir.model').clear_caches()
+        self.registry('ir.model.data').clear_caches()
+        # Get registries
+        self.user_model = self.registry("res.users")
+        self.partner_model = self.registry("res.partner")
+        # Get context
+        self.context = self.user_model.context_get(self.cr, self.uid)
+        # Create values for test, contact also created
+        contact_id = self.partner_model.create(self.cr, self.uid, {
+            'name': 'Astérix',
```
gurneyalex: s/'Astérix'/u'Astérix'/

base_contact_function_partner_firstname/tests/test_partner.py
```diff
@@ -0,0 +1,57 @@
...
+            'title': 1,
+        }, context=self.context)
+        self.vals = {
+            'name': 'Obélix',
```
gurneyalex: same thing here

base_contact_function/tests/test_functions.py
```diff
@@ -0,0 +1,103 @@
...
+    """
+
+    def setUp(self, vals={}):
+        """
+        Setting up function.
+        """
+        # Default test values
+        self.vals = {
+            'name': 'This is a test function name with uniödé',
+            'acronym': 'This is a test function acronym with uniödé',
```
gurneyalex: you need the u"" string form for the values. I agree this work because odoo does some processing to try to process utf-8 encoded string objects (what you are passing here) and decode the utf-8, but you really should be using unicode here.

base_contact_function_partner_firstname/tests/test_partner.py
bwrsandman: Why? This case catches more unicode errors.

base_contact_function/tests/test_functions.py
bwrsandman: I am intentionally not using unicode to catch more encoding errors. I.e. using `unicode` instead of `ustr`

base_contact_function_partner_firstname/tests/test_partner.py
gurneyalex: On 03/09/2014 13:52, Sandy wrote:

> In base_contact_function_partner_firstname/tests/test_partner.py:
> 
> > +
> > +class test_partner_contact_id(TransactionCase):
> > +
> > -    def setUp(self):
> > -        super(test_partner_contact_id, self).setUp()
> > -        # Clean up registries
> > -        self.registry('ir.model').clear_caches()
> > -        self.registry('ir.model.data').clear_caches()
> > -        # Get registries
> > -        self.user_model = self.registry("res.users")
> > -        self.partner_model = self.registry("res.partner")
> > -        # Get context
> > -        self.context = self.user_model.context_get(self.cr, self.uid)
> > -        # Create values for test, contact also created
> > -        contact_id = self.partner_model.create(self.cr, self.uid, {
> > -            'name': 'Astérix',
> 
> Why? This case catches more unicode errors.
> 
> —
> Reply to this email directly or view it on GitHub
> https://github.com/OCA/partner-contact/pull/20/files#r17043924.
> 
> I won't fight over this. It's just a better practice IMO, always using
> unicode strings where the API mandates unicode, rather than utf-8
> encoded string objects. This makes your source code the same regardless
> of the encoding declaration at the top of the file.

## 

Alexandre Fayolle
Chef de Projet
Tel : + 33 (0)4 79 26 57 94

Camptocamp France SAS
Savoie Technolac, BP 352
73377 Le Bourget du Lac Cedex
http://www.camptocamp.com

bwrsandman: I would agree with you if it was in the code, but this is in the unittests

gurneyalex:
:+1:

base_contact_function/res_partner.py
```diff
@@ -0,0 +1,97 @@
...
+        'other_contact_history_ids': fields.function(
+            _get_history_lines,
+            relation="res.partner",
+            method=True,
+            type="one2many",
+        ),
+        # Replace company by Organism
+        'use_parent_address': fields.boolean(
+            'Use Organism Address',
+            help="Select this if you want to set organism's address "
```
hbrunn: Does 'organism' have a meaning that's not mentioned here: http://www.oxforddictionaries.com/definition/english/organism ?

hbrunn:
Please be much more extensive in the addon's description. Especially, you should mention here how institutions, functions and categories work together.

bwrsandman:
@hbrunn addon's description provided by @giotta

bwrsandman:
Inconsistent fail between Odoo and OCB

base_contact_function/res_partner_category.py
```diff
@@ -0,0 +1,57 @@
...
+                    context=None, limit=100):
+        if not args:
+            args = []
+        if not context:
+            context = {}
+        if name:
+            name = name.split(' / ')[-1]
+            ids = self.search(cr, user, [('name', operator, name)] + args,
+                              limit=limit, context=context)
+            if ids:
```
hbrunn, line 49: this whole section doesn't respect the limit parameter, that will yield too many results when there are a lot of subcategories

base_contact_function/res_partner_category.py
```diff
@@ -0,0 +1,57 @@
...
+                child_ids = self.search(
+                    cr, user, [('parent_id', 'child_of', ids)],
+                    context=context)
+                if child_ids:
+                    ids.extend(child_ids)
```
hbrunn, line 54: the childof operator also returns the id itself, so add them only if not included already

hbrunn:
@bwrsandman the difference is caused by ocb using browse() in name_get and odoo using read(). The latter discards double ids and gives a random(?) order, the former delivers all ids in the order of the id list.

So to fix the test, we shouldn't pass double ids to name_get as pointed out in my comment above, and we should only check for set equality (= no order)

bwrsandman:
@hbrunn I have fixed the limits and the duplicates problem.

I don't understand this issue with id order, though. In this implementation, I used `set` to remove duplicates, this, however is unordered. Is this good?

hbrunn:
I personally think that every name_get should return its list sorted alphabetically, but not all name_gets do that.

As for the issue at hand: The test fails because it checks for list equality (https://github.com/OCA/partner-contact/pull/20/files#diff-482719082e50caa32544883db16a8259R156) and there, order matters. Given that I don't think we care about the lists' orders, I think you'd better use assertItemsEqual: https://docs.python.org/2/library/unittest.html#unittest.TestCase.assertItemsEqual

bwrsandman:
Thanks @hbrunn.

As for the order, I thought odoo relied on the defined `_order`.

bwrsandman:
>  The Travis CI build passed

:raised_hands:

hbrunn:
yay! Concerning `_order`: `search()` uses it, and `browse()` returns records in the order it got in the id list, but `read()` returns the records as they come from the database.

Anyway: :+1:

nbessi:
:+1:

PR #38, opened 2014-09-24
[ADD] Add base_phone module;it validates phone numbers

ehdem:
[ADD] Add base_phone module;it validates phone numbers using the _phonenumbers_ Python library.

This module comes from https://launchpad.net/openerp-asterisk-connector/7.0.

It was done by Alexis de Lattre from Akretion alexis.delattre@akretion.com


alexis-via:
As I already explained to @ehdem via a private mail, I am very much against moving my module base_phone to the partner-contact branch (this initiative was taken without asking my opinion beforehand).

One of the reasons why I am against this move is the following: this module is the base module of the openerp-telephony modules that I maintain on https://launchpad.net/openerp-asterisk-connector (which I plan to move to github and rename openerp-telephony, cf https://launchpad.net/openerp-asterisk-connector/+announcement/12914). All the other modules of the project depend on this module. At least, I need to keep the 7 modules independant from asterisk (base_phone, base_phone_popup, crm_claim_phone, crm_phone, event_phone, hr_phone, hr_recruitment_phone) all at the same place.

Imagine if we put the module "base_phone" in OCA/partner-contact, the modules "hr_phone" and "hr_recruitment_phone" in OCA/hr, the modules "crm_claim_phone" and "crm_phone" in OCA/crm, it will become completely impossible for me to maintain these modules in separate branches. This is very easy to understand by looking at the code : the method _get_phone_fields() on the object phone.common is used by 6 of the 7 asterisk-independant modules ; if I need to change the signature of this method (this method is quite new, I introduced it 2 months ago !), I would have to make several Pull Requests on several different github projects, that would be a nightmare.

Dear OCA reviewers, please reject this pull request.

hbrunn:
fair enough, i think

PR #42, opened 2014-11-03
Fix override of customer address fields to use function

nbessi:


yvaucher:
I prefer that as it was preventing from creating database with demodata if module was present on system :+1:

pedrobaeza:
:+1:

lepistone:
:+1: thanks

PR #43, opened 2014-11-04
update travis config

gurneyalex:
- test on odoo and OCB
- use UNIT_TEST=1


bwrsandman:
Could you also remove .coveragerc as it is supplied by the tests

gurneyalex:
builds with this config take way to much time, and we are finding this difficult to cope with in OCA/vertical-ngo and we are reverting the change. I'm withdrawing this PR for now.

PR #44, opened 2014-11-05
Fix look of Contact Page for partner First Name

bwrsandman:
Remove the "Is a company?" option in Partner's Contact Form as these are never companies
## Vanilla

capture d ecran de 2014-11-05 17 42 04
capture d ecran de 2014-11-05 17 41 52
## Current implementation

capture d ecran de 2014-11-05 17 41 29
capture d ecran de 2014-11-05 17 41 15
## Fixed look

capture d ecran de 2014-11-05 17 57 31
capture d ecran de 2014-11-05 17 57 43

### DIFF
--- a/partner_firstname/partner_view.xml
+++ b/partner_firstname/partner_view.xml
@@ -44,7 +44,6 @@
         <xpath expr="//form[@string='Contact']/sheet/div/label" position="replace">
             <div class="oe_edit_only">
                 <field name="is_company" on_change="onchange_type(is_company)" invisible="1"/>
-                <label for="name"/> (
             </div>
         </xpath>
         <xpath expr="//form[@string='Contact']/sheet/div/h1/field[@name='name']" position="attributes">
--- a/partner_firstname/partner_view.xml
+++ b/partner_firstname/partner_view.xml
@@ -43,9 +43,8 @@
         </xpath>
         <xpath expr="//form[@string='Contact']/sheet/div/label" position="replace">
             <div class="oe_edit_only">
+                <field name="is_company" on_change="onchange_type(is_company)" invisible="1"/>
                 <label for="name"/> (
-                <field name="is_company" on_change="onchange_type(is_company)" class="oe_inline"/>
-                <label for="is_company" string="Is a Company?"/>)
             </div>
         </xpath>
         <xpath expr="//form[@string='Contact']/sheet/div/h1/field[@name='name']" position="attributes">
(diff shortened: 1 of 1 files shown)


partner_firstname/partner_view.xml
```diff
@@ -33,19 +33,19 @@
...
         </xpath>
         <xpath expr="//form[@string='Contact']/sheet//field[@name='category_id']" position="attributes">
             <attribute name="style"/>
         </xpath>
         <xpath expr="//form[@string='Contact']/sheet/div/label" position="replace">
             <div class="oe_edit_only">
-                <label for="name"/> (
-                <field name="is_company" on_change="onchange_type(is_company)" class="oe_inline"/>
-                <label for="is_company" string="Is a Company?"/>)
+                <field name="is_company" on_change="onchange_type(is_company)" invisible="1"/>
```
pedrobaeza, line 48: Better to use position="attributes" for the same effect.

bwrsandman, line 48: What do you mean?

pedrobaeza, line 48: Instead of making a replace, you can use:

```
<xpath expr="//form[@string='Contact']/sheet/div/label" position="attributes">
    <attribute name="invisible">1</attribute>
</xpath>
```

bwrsandman, line 48: That's true, I guess... but I don't really see why that would block the PR.
Both methods do the same thing, this one just doesn't mess with the diff as much.
Can you explain why it is better to use your method? I'd be happy to change it is there is a good reason, otherwise, I don't see a point in changing it.

pedrobaeza, line 48: With your method, you have to define again all attributes of the field. It can be that across versions, these attributes change (a new onchange, groups, etc). With mine, you only overwrite desired attribute, keeping the rest unchanged.

bwrsandman, line 48: 1. It's not my method, it was there before. :P
2.  I don't think you understand what I am doing here, I am not refining a field with the same field invisible. I am replacing a **label** (for `name`, which has no place within a lastname enabled system) with the **field** `is_company`.
   If I was replacing a field with the same field but with invisible, I would agree with you, but this is not that case.
3. `is_company` is there and invisible because there are other fields requesting it to know if they should be invisible.

pedrobaeza, line 48: OK, I see. We can let it as is.

pedrobaeza:
:+1:

hbrunn:
:+1:

yvaucher:
:+1: 
Thanks for this improvement

PR #45, opened 2014-11-05
Put contact type under name and add label in edit mode

bwrsandman:
Categories not having a label and being in a strange spot
## Current

capture d ecran de 2014-11-05 18 18 25
capture d ecran de 2014-11-05 18 18 16
## Fixed

capture d ecran de 2014-11-05 18 25 58
capture d ecran de 2014-11-05 18 25 48

### DIFF
--- a/base_contact/base_contact_view.xml
+++ b/base_contact/base_contact_view.xml
@@ -161,8 +161,11 @@
                     </group>
                 </page>
             </page>
-            <xpath expr="//field[@name='child_ids']/form//field[@name='name']/.." position="before">
-                <field name="contact_type" readonly="0" on_change="onchange_contact_type(contact_type)"/>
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Seems OK, although I don't use this module. Too much complexity in the migration path from 6.1...

:+1:

foutoucour:
The layout seems broken now. Maybe add the label to tag or take category down to next group to order a bit the layout

bwrsandman:
> Maybe add the label to tag or take category down to next group to order a bit the layout

That would be hard to have it behave with other modules such as: #44 
I'm trying to be the most unobtrusive to the current code.

foutoucour:
ok, sounds good then
:+1:

hbrunn:
:+1:

Issue #46, opened 2014-11-07
partner_address_street3


alexis-via:
In the module partner_address_street3:

In model/partner.py, there is:

```
    def _commercial_fields(self, cr, uid, context=None):
        fields = super(res_partner, self
                       )._commercial_fields(cr, uid, context=context)
        fields.append('street3')
        return fields
```

Street3 is not a commercial field, so this code should not exists.

BUT street3 is an address field, that needs to be synced when you activate "Use parent address" ; for that, the following code is required:

```
    def _address_fields(self, cr, uid, context=None):
        '''Add street3 in sync of address when you say "Use company address"'''
        res = super(res_partner, self)._address_fields(
            cr, uid, context=context)
        res.append('street3')
        return res
```

I'll do a PR to fix this when I have time.

nbessi:
Fix here https://github.com/OCA/partner-contact/pull/47

PR #47, opened 2014-11-07
Fix street3 hook override.

nbessi:
Street3 is not part of commercial fields

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,8 @@ python:
   - "2.7"
 
 env:
-  - VERSION="8.0"
+  - VERSION="8.0" EXCLUDE="partner_address_street3"
+  - VERSION="8.0" INCLUDE="partner_address_street3"
 
 virtualenv:
   system_site_packages: true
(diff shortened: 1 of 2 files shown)


nbessi:
fixes #46

lepistone:
thanks :+1:

pedrobaeza:
:+1:

nbessi:
Ah it seems the addons should be tested aside as it interfer with partner first name

alexis-via:
:+1:  Thanks for the fix Nicolas.

yvaucher:
Travis fails on tests of partner_firstname

alexis-via:
I don't understand what is blocking this merge. If travis fails in partner_firstname, it's a problem that needs to be fixed in partner_firstname, but it's not linked to this bug fix.

pedrobaeza:
I can confirm that the failing test is not for this change, so I merge.

alexis-via:
@pedrobaeza Thank you Pedro !

PR #48, opened 2014-11-10
[ADD] Module to split partner street name and street number

StefanRijnhart:


pedrobaeza:
Hi, @StefanRijnhart,

Thanks for the module. I see this as a first approximation to the problem, but the question is very deep, because you can have block, door, and so on (even depending on locales). I know that this is a complete change of focus and approach, but maybe as you're introducing this split, we can conceive it as a framework like address format depending on country. What do you think?

Regards.

StefanRijnhart:
Out of scope of this module, and I cannot afford to embark on such a big project just for this functionality: I have nothing to win and no-one to bill. If you think the current scope of the module is too restricted, i could move the module to the l10n_netherlands project. You see, the only reason I am submitting this here is because our Dutch zipcode lookup module which I will be proposing there depends on it.

partner_street_number/model/res_partner.py
```diff
@@ -0,0 +1,148 @@
...
+import re
+import logging
+from openerp.osv import orm, fields
+from openerp import SUPERUSER_ID
+
+
+class ResPartner(orm.Model):
+    _inherit = 'res.partner'
+
+    def _auto_init(self, cr, context=None):
```
lmignon: In V8 This should be done via installation hooks  https://github.com/odoo/odoo/commit/4105b5f0285821d8056c721b5fd50b9cae902988
Since the PR is for 7.0, you can use migration scripts  with a test on

``` python
if not version:
   # it's a fresh install
```

pedrobaeza:
No, @StefanRijnhart, as I said, it's OK for a first version. If anybody wants to work in the general approach (I can't do it also), he can do it afterwards or tell it here.

hbrunn:
The first person wanting to hook into this would have to have _get_street call a function that does the actual concatenation (pretty much like _prepare_name_custom does it for partner names: https://github.com/OCA/partner-contact/blob/8.0/partner_firstname/partner.py#L45) - then all other modules can hook in there. Meanwhile, I'm fine :+1:

partner_street_number/model/res_partner.py
```diff
@@ -0,0 +1,148 @@
+# -*- coding: utf-8 -*-
+##############################################################################
+#
+#    OpenERP, Open Source Management Solution
+#    This module copyright (C) 2013 Therp BV (<http://therp.nl>).
```
pedrobaeza: You should put 2014

StefanRijnhart: Done

StefanRijnhart:
@hbrunn good idea. Code has been changed to feature an inheritable function that delivers the field value.

partner_street_number/model/res_partner.py
StefanRijnhart: @lmignon what is the advantage of moving this to a migration script? I can see that it would work, but it remains a hack either way.

pedrobaeza:
:+1:

partner_street_number/model/res_partner.py
lmignon: @StefanRijnhart IMO, the advantage is to split the logic required for the installation / migration outside the business logic.

StefanRijnhart: OK, done that. Seemed to me like the pre-script is not called at installation time, but I did not need it because having a stored function field leaves the street column intact.

lmignon:
@StefanRijnhart Thank you for the changes. IMO, it really improves the readability of the code. Apart from the 2 pep8 errors to fix, :+1:

StefanRijnhart:
Ah yes, sorry about the style errors. Fixed and rebased. Thanks for the review all of you.

PR #50, opened 2014-11-18
Add user rights to limit access to certain users (and give all rights to others)

mdietrichc2c:
The goal is, rather than give access to all users to the merge and refuse most cases if they're not the administrator (which is irrelevant later as SUPERUSER_ID is used), to give to a certain group 'Partner Merge' merging rights as if they were the administrator, and hide/deny merges otherwise.

### DIFF
--- a/base_partner_merge/__openerp__.py
+++ b/base_partner_merge/__openerp__.py
@@ -10,7 +10,7 @@ backport module, to be removed when we switch to saas2 on the private servers
         'base',
     ],
     'data': [
-        'security/ir.model.access.csv',
+        'security/security.xml',
         'base_partner_merge_view.xml',
     ],
     'installable': True,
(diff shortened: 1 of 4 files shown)


base_partner_merge/base_partner_merge.py
```diff
@@ -340,6 +340,16 @@ def _merge(self, cr, uid, partner_ids, dst_partner=None, context=None):
 
         partner_ids = proxy.exists(cr, uid, list(partner_ids),
                                    context=context)
+
+        if not self.pool['res.users'].has_group(
+            cr, uid, 'base_partner_merge.group_partner_merge'
+        ):
+            raise orm.except_orm(
+                _('Error'),
+                _("You do not belong in the 'Partner Merge' group. "
```
pedrobaeza: s/in/to

pedrobaeza:
A little typo, but seems fine. :+1: waiting for Travis answer.

mdietrichc2c:
@pedrobaeza thanks for the review, message fixed.

fclementic2c:
:+1:

base_partner_merge/__openerp__.py
```diff
@@ -10,7 +10,7 @@
         'base',
     ],
     'data': [
-        'security/ir.model.access.csv',
```
bwrsandman, line 13: Why is this removed while the file still exists?

PR #53, opened 2014-11-25
[IMP] partner_relations: merge from development

hbrunn:
This branch fixes two functional issues
- better check for overlapping relations
- edit relations within a popup

And improves some poorly phrased dutch translations

### DIFF
--- a/partner_relations/view/res_partner.xml
+++ b/partner_relations/view/res_partner.xml
@@ -1,6 +1,7 @@
 <openerp>
     <data>
         <record id="view_res_partner_filter" model="ir.ui.view">
+            <field name="name">partner_relations.view_partner_filter</field>
             <field name="inherit_id" ref="base.view_res_partner_filter" />
             <field name="model">res.partner</field>
             <field type="xml" name="arch">
@@ -15,6 +16,7 @@
             </field>
         </record>
         <record id="view_partner_form" model="ir.ui.view">
+            <field name="name">partner_relations.view_partner_form</field>
             <field name="inherit_id" ref="base.view_partner_form" />
             <field name="model">res.partner</field>
             <field type="xml" name="arch">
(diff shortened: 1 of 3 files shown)


partner_relations/view/res_partner.xml
```diff
@@ -79,6 +80,53 @@
...
+                                    <field
+                                        name="partner_id_display"
+                                        required="True"
+                                        attrs="{
+                                            'readonly': [('type_selection_id','=',False)],
+                                        }"
+                                        options="{'create': false, 'create_edit': false}"
+                                        widget="many2one_clickable"
+                                    />
+                                    <field
```
pedrobaeza: I think this style is not very readable. Don't you prefer for these cases with just one attribute to put them in only line?

pedrobaeza:
One minor style issue. The rest LGTM.

partner_relations/view/res_partner.xml
hbrunn: I leave this case to @NL66278

NL66278: Fields with really one attribute, that is only the name, I prefer on one line. When there are multiple attributes I prefer multiple lines, first to keep the line length within limits, secondly for easier adding, copying and deleting attributes. It certainly helps when using line oriented editors like vim.

pedrobaeza: I agree with you on multi-attributes tags, but for one attribute it's very ugly.

NL66278: @hbrunn Holger, I cannot change your pull request. Is it a problem for you to change the lines where the one-attribute elements are spread out unnecesarrily?

hbrunn: not at all, it's my pleasure @pedrobaeza the lines are fixed even though this doesn't show up here

pedrobaeza:
Thanks for the changes.

:+1:

guewen:
:+1:

NL66278:
:+1:  LGTM

PR #55, opened 2014-11-26
[ADD] partner notify on creation

max3903:


pedrobaeza:
I'm seeing some PRs for the same purpose, but on different objects.

Don't you think you can develop a general mechanism that allows to send a template on creation of an specific model to specific recipients? This way, you can use the same mechanism for partner, product, anything... creation notification.

max3903:
https://www.odoo.com/forum/help-1/question/sending-notification-when-creating-confirmation-validation-31499

PR #58, opened 2014-12-05
Update better_zip.py

ghost:
Bugfixing => name_get, cast to str.

### DIFF
--- a/base_location/better_zip.py
+++ b/base_location/better_zip.py
@@ -4,6 +4,7 @@
 #    Author: Nicolas Bessi. Copyright Camptocamp SA
 #    Contributor: Pedro Manuel Baeza <pedro.baeza@serviciosbaeza.com>
 #                 Ignacio Ibeas <ignacio@acysos.com>
+#                 Cesar Andres <cesar-andres.sanchez@brain-tec.ch>
 #
 #    This program is free software: you can redistribute it and/or modify
 #    it under the terms of the GNU Affero General Public License as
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Why do you need the casting?

ghost:
@pedrobaeza : this method collects all the information to be shown in a list (name)  and later it executes:  ", ".join(name).

The problem is that the information in this list may contain Boolean values.  When we  try to perform the operation it crashes. 

Test example:
", ".join(['anything', False])

Solution: To cast to str before executing join.

pedrobaeza:
But the only condition to appear False is because city field is empty (the rest are checked), and this field is mandatory (https://github.com/OCA/partner-contact/blob/8.0/base_location/better_zip.py#L35), so there's no such possibility.

ghost:
@pedrobaeza : I agree with you

pedrobaeza:
OK, so can I close this PR?

ghost:
Thank you I did it

2014-12-05 13:37 GMT+01:00 Pedro M. Baeza notifications@github.com:

> OK, so can I close this PR?
> 
> —
> Reply to this email directly or view it on GitHub
> https://github.com/OCA/partner-contact/pull/58#issuecomment-65784392.

## 

_César Andrés PhD Computer Science brain-tec AG | Überlandstrasse 10,
CH-3900 Brig Tel: +41 27 552 02 20 | Dir: +41 27 552 02 11_
http://www.brain-tec.ch/   cesar-andres.sanchez@brain-tec.ch
http://www.brain-tec.ch/

PR #59, opened 2014-12-05
Update better_zip.py

ghost:
Bugfixing name => cast to str all elements if exists.

### DIFF
--- a/base_location/better_zip.py
+++ b/base_location/better_zip.py
@@ -4,6 +4,7 @@
 #    Author: Nicolas Bessi. Copyright Camptocamp SA
 #    Contributor: Pedro Manuel Baeza <pedro.baeza@serviciosbaeza.com>
 #                 Ignacio Ibeas <ignacio@acysos.com>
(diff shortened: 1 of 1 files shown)


pedrobaeza:
The same question as the other PR.

ghost:
@pedrobaeza : Same answer as the other PR.

base_location/better_zip.py
```diff
@@ -51,6 +52,7 @@ def name_get(self, cursor, uid, ids, context=None):
                 name.append(bzip.state_id.name)
             if bzip.country_id:
                 name.append(bzip.country_id.name)
+            name = [str(x) for x in name if x]
```
bwrsandman: It would be better to use ustr from openerp.tools

ghost: @bwrsandman 
Hello, I did another commit with this suggestion, 

thank you!

pedrobaeza:
As said in #58, this doesn't apply, so I close it.

PR #62, opened 2014-12-12
Porting a commit which was never ported from launchpad

bwrsandman:
From http://bazaar.launchpad.net/%7Epartner-contact-core-editors/partner-contact-management/7.0/revision/41
[ADD] base_partner_help: Module for additional methods for manage partner. For now, there is a method to deal with addresses (get_split_address).

### DIFF
--- a/base_partner_helper/__init__.py
+++ b/base_partner_helper/__init__.py
@@ -1,0 +1,22 @@
+# -*- coding: utf-8 -*-
+##############################################################################
+#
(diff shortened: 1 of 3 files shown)


yvaucher:
Thanks  a lot @bwrsandman 
Can you do the PEP8 fixes too ?

pedrobaeza:
This one is already ported to Github, but renamed to _partner_helper_: https://github.com/OCA/partner-contact/tree/8.0/partner_helper, so I close it.

PR #63, opened 2014-12-13
IMP base_location module

jjscarafia:
- Migrate to v8 api
- Rename name field to zip
- Create new computed stored field 'name' that use zip, city, state and country in order to allow search using '%' from the m2o widget

### DIFF
--- a/base_location/better_zip.py
+++ b/base_location/better_zip.py
@@ -19,64 +19,43 @@
 #    along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #
 #
-from openerp.osv import orm, fields
+from openerp import models, fields, api
 
 
-class BetterZip(orm.Model):
+class BetterZip(models.Model):
 
     " City/locations completion object"
 
     _name = "res.better.zip"
     _description = __doc__
-    _order = "priority"
 
-    _columns = {'priority': fields.integer('Priority', deprecated=True),
-                'name': fields.char('ZIP'),
-                'city': fields.char('City', required=True),
-                'state_id': fields.many2one('res.country.state', 'State'),
-                'country_id': fields.many2one('res.country', 'Country'),
-                'code': fields.char('City Code', size=64,
-                                    help="The official code for the city"),
-                }
+    priority = fields.Integer('Priority', deprecated=True)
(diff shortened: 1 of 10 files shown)


base_location/better_zip.py
```diff
@@ -19,64 +19,43 @@
...
-            context = {}
-        ids = []
-        if name:
-            ids = self.search(
-                cr, uid, [('name', 'ilike', name)] + args, limit=limit)
-        if not ids:
-            ids = self.search(
-                cr, uid, [('city', operator, name)] + args, limit=limit)
-        return self.name_get(cr, uid, ids, context=context)
+    priority = fields.Integer('Priority', deprecated=True)
```
pedrobaeza, line 32: This field can be removed.

base_location/better_zip.py
```diff
@@ -19,64 +19,43 @@
...
+    name = fields.Char('Name', compute='_get_name', store=True)
+    zip = fields.Char('ZIP')
```
pedrobaeza, line 34: If you change the fields, you should put oldname attribute for making an automatic migration.

base_location/better_zip.py
```diff
@@ -19,64 +19,43 @@
...
+    @api.depends(
+        'zip',
+        'city',
+        'state_id',
+        'state_id.name',
+        'country_id',
+        'country_id.name',
+        )
+    def _get_name(self):
+        self.name = ('(%s) %s, %s, %s') % (
```
pedrobaeza, line 51: This cannot be done this way, because there will be odd strings if fields aren't filled (`() 14353, ,` for example).

pedrobaeza:
Thinking deeper about it, we can't change field names in the middle of a version, because there are a lot of modules that depends on this to introduce data via XML or SQL (for now, I'm thinking in spanish, brazilian and italian toponyms at least), so we must stick to the same nomenclature.

But the migration to new API is OK.

jjscarafia:
Thanks Pedro for the suggestions. So, what do you think about keeping same
field names and call the new computed one as  "complete_name", and set it
on _recname?

pedrobaeza:
It can be an option indeed.

pedrobaeza:
Superseeded by #81

PR #66, opened 2014-12-17
partner_firstname Fix tests with stock

bwrsandman:
There is an edge case where partner_firstname fails to work:

When creating a user, there can be a non-null constraint issue.

The most common error is when running odoo tests, there are some tests in stock which fail.


pedrobaeza:
:+1:

eLBati:
:+1:

guewen:
:+1:

PR #67, opened 2014-12-23
Spanish translation

antespi:
Adding Spanish translation


pedrobaeza:
You should squash all the commits

PR #70, opened 2014-12-30
Initial translation

rschnapka:


bwrsandman:
What's the difference with #71

rschnapka:
was my first commit - accidentally. I am trying to get german translation into the modules. This commit only should have had one added file in base_contact... I seem to have messed that up :-( Working on it... thank for the attention anyway.

PR #71, opened 2014-12-30
Initial german translation

rschnapka:
i18n/de.po for german translation


pedrobaeza:
Seems good technically speaking, but no opinion about the translation itself.

:+1:

bwrsandman:
:+1:

tafaRU:
:+1: from a technical point of view. So also considered the previous approves I proceed with the merge.

PR #75, opened 2015-01-01
[FIX] Make sure we create a new partner on import.

joaoalf:
When we try to import a funtion partner with a cantact_id field pointing to an existing contact, the create method will just return the existing contact id instead of creating a new entry in the database but the import will think that the importation went ok since no exception is triggered.

This PR removes the code that causes this behaviour.

### DIFF
--- a/base_contact_function_partner_firstname/res_partner.py
+++ b/base_contact_function_partner_firstname/res_partner.py
@@ -57,9 +57,6 @@ class res_partner(orm.Model):
             vals['firstname'] = contact_info.firstname
             vals['title'] = contact_info.title.id or ''
 
-        if ('contact_id'in vals and ('reset_password' in context or
-                                     'install_mode' in context)):
-            return vals['contact_id']
         res = super(res_partner, self).create(cr, user, vals, context=context)
         return res
 


base_contact_function_partner_firstname/res_partner.py
```diff
@@ -57,9 +57,6 @@ def create(self, cr, user, vals, context=None):
             vals['firstname'] = contact_info.firstname
             vals['title'] = contact_info.title.id or ''
 
-        if ('contact_id'in vals and ('reset_password' in context or
-                                     'install_mode' in context)):
-            return vals['contact_id']
```
bwrsandman, line 62: What does this statement do anyways? It looks like it prevents creation to be done are returns an existing object.
I think you said `"install_mode"` is in context while importing with the import button in the web interface.

joaoalf, line 62: The CVS importing routine sets the install_mode on triggering this behaviour.

guewen:
Could we have more context? I'm not sure to get it

bwrsandman:
Just from looking at this, we can see that statement returns an existing id in the create function but only in install mode. Looks like a hack to prevent failing tests, but this has unintended side effects such as, not actually creating an object.

The particular case being when importing a csv full of partners via the interface.

bwrsandman:
:+1:  Used in prod

base_contact_function_partner_firstname/res_partner.py
pedrobaeza, line 62: Isn't better to just assign that contact_id?

guewen:
:+1:

PR #79, opened 2015-01-13
[UPD] Hide the firstname in kanban view

ehdem:
The name field contains the firstname and the lastname, Don't display firstname field once again.


pedrobaeza:
That makes sense. Previously name == last_name. That's why it was two times.

:+1:

foutoucour:
:+1:

hbrunn:
:+1:

PR #84, opened 2015-01-26
[FIX] allow to change partner_id_display

hbrunn:


pedrobaeza:
:+1:

hbrunn:
I added a commit that also fixes a logic error

guewen:
:+1:

hbrunn:
and I added another commit that allows searching for translated relation names (res.partner.relation.type selection is a view on res.partner.relation.type, so we don't get openerp's translation mechanism on that model)

yvaucher:
:+1:

PR #86, opened 2015-02-02
[ADD] partner_email_validation

veloutin:
This module performs simple validation on email addresses. Emails are checked for:
- Having a mailbox and domain part: mailbox@domain
- Having a seemingly valid domain
- (sub.)*domain.tld where tld has at least two letters
- @test and @localhost
- Containing letters or numbers, with a few special characters:
- ".-_+" for mailbox part
- "-." for domain part
- Quoted mailboxes "Foo Bar"@spam.eggs are not supported as of now.

No email is sent for validation.

### DIFF
--- a/partner_email_validation/__init__.py
+++ b/partner_email_validation/__init__.py
@@ -22,5 +22,4 @@
 
 from . import (
     partner,
-    tests,
 )
--- a/partner_email_validation/__init__.py
+++ b/partner_email_validation/__init__.py
@@ -1,0 +1,26 @@
+# -*- encoding: utf-8 -*-
+##############################################################################
+#
+#    OpenERP, Open Source Management Solution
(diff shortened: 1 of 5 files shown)


pedrobaeza:
Why are you importing tests?

veloutin:
@pedrobaeza I suppose it was the bad habit of importing everything to make sure odoo doesn't forget about it :) I removed the import.

StefanRijnhart:
Did you consider using an existing, dedicated library, such as https://pypi.python.org/pypi/validate_email?

rafaelbn:
Hi, very interesting module. @veloutin do you think you could continue with @StefanRijnhart comment? Let me know please

@antespi

pedrobaeza:
As there's no answer, I close. Feel free to reopen in case there's interest.

PR #91, opened 2015-02-19
Fix encoding line to be ascii readable

bwrsandman:
Ironically, the encoding line was not ascii and could not be encoded by code analyzers such as sphinx


pedrobaeza:
:+1:

yvaucher:
:+1:  I merge as it's trivial

yvaucher:
Just curious how someone could have added an overscore there ^^

pedrobaeza:
Yeah, in what keyboard is that present??!!

PR #92, opened 2015-02-19
Fix added pep8 conventions

bwrsandman:
Put boolean operators at end of lines instead of beggining of newlines
Put lambda definition in fields.function def
Convert lambda into def

### DIFF
--- a/base_partner_merge/base_partner_merge.py
+++ b/base_partner_merge/base_partner_merge.py
@@ -123,8 +123,8 @@ class MergePartnerAutomatic(orm.TransientModel):
             context = {}
         res = super(MergePartnerAutomatic, self
                     ).default_get(cr, uid, fields, context)
-        if (context.get('active_model') == 'res.partner'
-                and context.get('active_ids')):
+        if (context.get('active_model') == 'res.partner' and
+                context.get('active_ids')):
             partner_ids = context['active_ids']
             res['state'] = 'selection'
             res['partner_ids'] = partner_ids
@@ -212,8 +212,8 @@ class MergePartnerAutomatic(orm.TransientModel):
                              '%(column)s IN %%s') % query_dic
                     cr.execute(query, (dst_partner.id, partner_ids,))
 
-                    if (column == proxy._parent_name
-                            and table == 'res_partner'):
+                    if (column == proxy._parent_name and
+                            table == 'res_partner'):
                         query = """
                             WITH RECURSIVE cycle(id, parent_id) AS (
                                     SELECT id, parent_id FROM res_partner
@@ -316,8 +316,8 @@ class MergePartnerAutomatic(orm.TransientModel):
 
         values = dict()
         for column, field in columns.iteritems():
-            if (field._type not in ('many2many', 'one2many')
-                    and not isinstance(field, fields.function)):
+            if (field._type not in ('many2many', 'one2many') and
+                    not isinstance(field, fields.function)):
                 for item in itertools.chain(src_partners, [dst_partner]):
                     if item[column]:
                         values[column] = write_serializer(column,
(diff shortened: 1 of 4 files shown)


base_partner_merge/base_partner_merge.py
```diff
@@ -381,8 +381,10 @@ def _merge(self, cr, uid, partner_ids, dst_partner=None, context=None):
...
         _logger.info("dst_partner: %s", dst_partner.id)
 
-        call_it = lambda function: function(cr, uid, src_partners,
-                                            dst_partner, context=context)
+        def call_it(function):
+            return function(
+                cr, uid, src_partners, dst_partner, context=context
+            )
 
         call_it(self._update_foreign_keys)
```
pedrobaeza: You can do directly:

```
self._update_foreign_keys(cr, uid, src_partners, dst_partner, context=context)
self._update_reference_fields(cr, uid, src_partners, dst_partner, context=context)
```

partner_firstname/partner.py
```diff
@@ -68,10 +68,10 @@ def _write_name(
         In addition an heuristic avoids to keep a firstname without a
         non-blank lastname
         """
-        field_value = (field_value
-                       and not field_value.isspace()
-                       and field_value
-                       or False)
+        field_value = (
+            field_value and not field_value.isspace() and field_value or
+            False
```
yvaucher: Could you write?

```
field_value if field_value and not field_value.isspace() else False
```

partner_relations/model/res_partner_relation.py
```diff
@@ -87,8 +87,8 @@ def get_values(this, dummy_field_names, dummy_arg, context=None):
                 ((this.type_id.id) * 10) + (on_right_partner and 1 or 0))
             # partner_id_display
             values['partner_id_display'] = (
-                on_right_partner and this.left_partner_id.id
-                or this.right_partner_id.id
+                on_right_partner and this.left_partner_id.id or
+                this.right_partner_id.id
```
yvaucher: We can use `if else` here

yvaucher:
Just 2 little non blocking remarks otherwise :+1:

hbrunn:
`./partner_relations/model/res_partner_relation.py:104:17: F812 list comprehension redefines 'self' from line 76`

I think you shouldn't touch the list comprehension in the last file. And what's wrong with `this` style-wise anyway?

bwrsandman:
It's defined in [PEP 8](https://www.python.org/dev/peps/pep-0008/#function-and-method-arguments):

> ## Function and method arguments 
> 
>  Always use `self` for the first argument to instance methods.
> 
> Always use `cls` for the first argument to class methods.
> 
> If a function argument's name clashes with a reserved keyword, it is generally better to append a single trailing underscore rather than use an abbreviation or spelling corruption. Thus `class_` is better than `clss` . (Perhaps better is to avoid such clashes by using a synonym.)

base_partner_merge/base_partner_merge.py
```diff
@@ -381,12 +381,15 @@ def _merge(self, cr, uid, partner_ids, dst_partner=None, context=None):
...
 
-        call_it = lambda function: function(cr, uid, src_partners,
-                                            dst_partner, context=context)
-
-        call_it(self._update_foreign_keys)
-        call_it(self._update_reference_fields)
-        call_it(self._update_values)
+        self._update_foreign_keys(
+            cr, uid, src_partners, dst_partner, context=context
+        )
```
lmignon, line 386: Why to you put the last ')' on a new line? Sometimes it's at the end of  line sometimes on a new line...  (I know it's a kind a nit-picking :smirk: )

lmignon:
:+1:  (code review)

base_partner_merge/base_partner_merge.py
bwrsandman, line 386: The reason I do it is for symmetry and simpler future diffs.

If the `(` has nothing to the right of it, then I put nothing at the left of `)`
It also saves a character on the argument line.

Additionally, if the argument list got longer than 80 chars, I would do two things, put a comma at the end of the last arg and put the closing parens on its own line:

``` python
foo(
    arg1,
    arg2,
    arg3,
    arg4,
    ...
    argN,
)
```

Notice the comma after `argN`, this means that if `argN+1,` is added, the diff only has one line.

``` diff
 foo(
     ...
     argN,
+    argN+1,
 )

```

Without the comma or with a closing parens, the diff would be three lines

``` diff
 foo(
     ...
-    argN)
+    argN,
+    argN+1)


```

PR #94, opened 2015-03-03
fix travis (flake8)

gurneyalex:


yvaucher:
Does it superseeds this ? https://github.com/OCA/partner-contact/pull/92

gurneyalex:
na.... I was just in travis fixing berserk mode, and did not check for existing PR (just using the mails from travis as feedback...)

PR #97, opened 2015-03-16
Add demo data

bwrsandman:
- Brussels Belgium city

In some modules which make reference to this module, I will be trying to add some demo data.
Having one demo better zip helps make that possible.

### DIFF
--- a/base_location/demo/better_zip.xml
+++ b/base_location/demo/better_zip.xml
@@ -1,0 +1,12 @@
+<?xml version = "1.0" encoding="utf-8"?>
+<openerp>
+  <data>
+
+    <record id="demo_brussels" model="res.better.zip">
+      <field name="name">1000</field>
+      <field name="city">Brussels</field>
+      <field name="country_id" ref="base.be"/>
+    </record>
+
+  </data>
+</openerp>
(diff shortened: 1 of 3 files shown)


bwrsandman:
If there are more demo datas which would be nice to put, please provide me some examples and I'll gladly include them.

pedrobaeza:
`demo/better_zip.xml` file not included

bwrsandman:
Thanks. I went too fast.

base_location/demo/better_zip.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version = "1.0" encoding="utf-8"?>
+<openerp>
+  <data>
+
+    <record id="demo_brussels" model="res.better.zip">
+      <field name="city">Brussels</field>
```
pedrobaeza, line 7: No zip code?

bwrsandman, line 7: Using it for City management rather than ZIP codes. I don't know any Belgian ZIP codes but I can add one.

pedrobaeza, line 7: Well, it's for providing a full example record.

bwrsandman:
I assume 1000 is the zip code for Brussels.

yvaucher:
:+1:

pedrobaeza:
:+1:

PR #98, opened 2015-03-17
[ADD] Add res_company_logo_secondary module.

foutoucour:
The module associates a secondary logo and name to the company.

The case the module required the module was because the company had some alternate logo and name.

### DIFF
--- a/res_company_logo_secondary/README.rst
+++ b/res_company_logo_secondary/README.rst
@@ -1,0 +1,29 @@
+Company Logo Secondary
+======================
+
+The module associates a secondary logo and name to the company.
+
+The case the module required the module was because the company had some
+alternate logo and name.
+
+Credits
+=======
+
+Contributors
+------------
+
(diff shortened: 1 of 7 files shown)


yvaucher:
Warning There are 2 PR for a module of the same purpose: #98 #108

The other one could be more generic.

I see in header that copyright is for Nicolas Bessi.

And this remind me of some code to add company logos for headers in `report_webkit`:
https://github.com/odoo/odoo/blob/8.0/addons/report_webkit/header.py

yvaucher:
Both modules are on different odoo version but it would make migration easier if we agree on one.

ghost:
@foutoucour What is the use case behind your module ? Is it to display a different logo in different types of document ?

foutoucour:
I agree @yvaucher . I would vote for @dufresnedavid module which seems to be more advanced than this one.

But the backport will be a bit tricky ;)

gurneyalex:
closing the PR as per author request and merging #108

PR #101, opened 2015-03-25
protect import of external dependencies

### LINKED ISSUE #100: unprotected import in base partner geoname
Hello, 

https://github.com/OCA/partner-contact/blob/8.0/base_location_geonames_import/wizard/geonames_import.py#L30

import is not protected and may break server startup .

gurneyalex:
Odoo won't install an addon if the external dependencies are not met.
However, the python modules of the addons are imported at startup, and the
lack of an external dependency for an external addon will cause a crash,
therefore the import needs to be in a try..except block.

closes #100

### DIFF
--- a/base_location_geonames_import/wizard/geonames_import.py
+++ b/base_location_geonames_import/wizard/geonames_import.py
@@ -27,10 +27,14 @@ from openerp.exceptions import Warning
 import requests
 import tempfile
 import StringIO
-import unicodecsv
 import zipfile
 import os
 import logging
+
+try:
+    import unicodecsv
+except ImportError:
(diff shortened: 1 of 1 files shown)


lmignon:
:cry:  It's the kind of hack I hate... We really need a proper dependency manager. Nevertheless, it's the only way to avoid problems at startup. :+1:

rdeheele:
:+1:

pedrobaeza:
:+1:

nbessi:
@lmignon I agree with you, this topic should be properly discussed during Odoo days.
I guess this kind of issue have to be manged at core level of Odoo server.

PR #102, opened 2015-03-25
7.0 - [IMP] Add unique constraint on res.better.zip

ghost:
This pull request adds a uniq constraint on res.better.zip to prevent users from creating multiple times the same location.

The migration script is currently in a PR to OpenUpgrade. It has unittests to prove that it works.
https://github.com/OpenUpgrade/OpenUpgrade/pull/266

### DIFF
--- a/base_location/__openerp__.py
+++ b/base_location/__openerp__.py
@@ -20,7 +20,7 @@
 #
 ##############################################################################
 {'name': 'Location management (aka Better ZIP)',
- 'version': '0.3.1',
+ 'version': '1.0',
  'depends': ['base'],
  'author': "Camptocamp,Odoo Community Association (OCA)",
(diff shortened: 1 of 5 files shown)


pedrobaeza:
Thanks for the work on this. Only a minor issue at first sight: the migration script should be done before update (pre-migration), not after, because this way, constraint is not going to be applied.

ghost:
@pedrobaeza I tested the migration and the constraint is applied even if the script is post-migration. A pre-migration script would be much more complex, because the orm could not be used to search an unlink res.better.zip records.

pedrobaeza:
OK then

bwrsandman:
:+1: used in prod

yvaucher:
:+1:

PR #103, opened 2015-03-26
Partner relations Port to v8

bwrsandman:


partner_relations/__openerp__.py
```diff
@@ -0,0 +1,95 @@
...
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+{
+    "name": "Partner relations",
+    "version": "1.1",
+    "author": "Therp BV,Odoo Community Association (OCA)",
+    "complexity": "normal",
+    "description": """
+Introduction
```
foutoucour: please user OCA template for descriptions:
https://github.com/OCA/maintainer-tools/blob/master/template/module/README.rst

partner_relations/__openerp__.py
```diff
@@ -0,0 +1,95 @@
...
+is to search for partners having a relation with Anna. This results in Anna's
+assistant(s), as you searched for assistants before.
+
+By default, only active, not expired relations are shown. If you need to find
+partners that had some relation at a certain date, fill in that date in the
+search box and one of the proposals is to search for relations valid at that
+date.""",
+    "category": "Customer Relationship Management",
+    "depends": [
+        'base',
```
foutoucour: Base is not necessary in v8.
All modules depend from base now.

partner_relations/__openerp__.py
```diff
@@ -0,0 +1,95 @@
...
+    ],
+    "demo": [
```
foutoucour: demo section does not exist anymore in v8. It should be in data section

partner_relations/model/res_partner.py
```diff
@@ -0,0 +1,316 @@
...
+
+        Don't count inactive relations.
+        """
+        self.relation_count = len([r for r in self.relation_ids if r.active])
+
+    def _get_relation_ids_select(self, cr, uid, ids, field_name, arg,
+                                 context=None):
+        '''return the partners' relations as tuple
+        (id, left_partner_id, right_partner_id)'''
+        cr.execute(
```
foutoucour: Could you comment why you use a query instead of ORM here.
I am sure there is a good reason ;)

bwrsandman, line 50: I am sure, but that is not my code. Maybe @hbrunn can clarify that point.

hbrunn, line 50: there was a time when it was relevant if we found the relation as a left partner or a right partner. Meanwhile, the magic for this is done in `res.partner.relation`, so probably this can be done in the ORM by now. There might be weirdnesses casued by https://github.com/odoo/odoo/pull/5807 though, as both `left_partner_id` and `right_partner_id` are auto_joined.

bwrsandman:
There might be a lot of dead code, I need that reviewed.

Also, @alexis-via please let me know what can be done to make this more compatible with the other partner_relation

partner_relations/model/res_partner_relation_all.py
```diff
@@ -0,0 +1,199 @@
...
+
+    @api.multi
+    def name_get(self):
+        return {
+            this.id: '%s %s %s' % (
+                this.this_partner_id.name,
+                this.type_selection_id.name_get()[0][1],
+                this.other_partner_id.name,
+            )
+            for this in self.browse(cr, uid, ids, context=context)}
```
RoelAdriaans: Old style browse is used here, gives an error because cr,uid,ids are unknown.

partner_relations/__openerp__.py
```diff
@@ -0,0 +1,51 @@
...
+        'view/res_partner_relation.xml',
+        'view/res_partner.xml',
+        'view/res_partner_relation_type.xml',
+        'view/menu.xml',
+        'security/ir.model.access.csv',
+    ],
+    "js": [
+    ],
+    "css": [
+    ],
```
hbrunn, line 43: those two can go away

partner_relations/model/res_partner_relation.py
```diff
@@ -0,0 +1,366 @@
...
+    )
+
+    right_contact_type = fields.Selection(
+        lambda s: s.env['res.partner.relation.type']._get_partner_types(),
+        'Right Partner Type',
+        compute='_get_partner_type_any',
+        store=True,
+    )
+
+    any_partner_id = fields.Many2many(
```
hbrunn, line 99: From my understanding, this field is used to allow to search for a partner and see all of her relations. Does this actually work? It's not stored and I don't see a search function

hbrunn:
Thanks! :+1:, we'll add tests and some rewrites of legacy code later, I'd say

bwrsandman:
@hbrunn I factored out the `id * 10` padding to a variable which can be changed to 2 eventually

I however, noticed that the constraints on left and right partners don't exist in tree view, only in the smart button view and this will have to be done.

antespi:
@bwrsandman i've PR https://github.com/savoirfairelinux/partner-contact/pull/1 to your PR the Spanish translation, please merge it, in order to append to this PR
:+1:

partner_relations/__openerp__.py
```diff
@@ -0,0 +1,51 @@
...
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+{
+    "name": "Partner relations",
+    "version": "1.1",
+    "author": "Therp BV,Odoo Community Association (OCA)",
+    "complexity": "normal",
+    "category": "Customer Relationship Management",
+    "depends": [
```
antespi: Does not depend on base addon?

partner_relations/model/res_partner.py
```diff
@@ -0,0 +1,316 @@
...
+                                ('record_type', '=', 'b'),
+                            ],
+                            context=context))
+
+                if not relation_type_selection_ids:
+                    result = AND([result, [FALSE_LEAF]])
+
+                for relation_type_selection_id in relation_type_selection_ids:
+                    type_id, is_inverse = (
+                        relation_type_selection_id.get_type_from_selection_id()
```
antespi: AttributeError: 'int' object has no attribute 'get_type_from_selection_id'
relation_type_selection_ids is a list of integers (old API search method returns a list of integers), so you need to use browse, isn't it?

partner_relations/__openerp__.py
antespi, line 27: Does not depend on base addon?

partner_relations/view/res_partner_relation.xml
```diff
@@ -0,0 +1,97 @@
...
+            </field>
+        </record>
+
+        <record id="tree_res_partner_relation" model="ir.ui.view">
+            <field name="model">res.partner.relation</field>
+            <field name="arch" type="xml">
+                <tree
+                    string="Partner Relations"
+                    colors="gray:date_end and date_end &lt; current_date or not active;blue:date_start &gt; current_date"
+                    editable="bottom"
```
antespi: IMHO, this should be `editable="top"`, because if there are lots of relations, when user clicks to create a new one, apparently nothing happends, user must go to the end of the screen to see an editable line.

partner_relations/view/res_partner_relation_all.xml
```diff
@@ -0,0 +1,92 @@
...
+<openerp>
+    <data>
+
+        <record id="tree_res_partner_relation_all" model="ir.ui.view">
+            <field name="model">res.partner.relation.all</field>
+            <field name="arch" type="xml">
+                <tree
+                    string="Partner Relations"
+                    colors="gray:(date_end and date_end &lt; current_date) or not active;blue:date_start &gt; current_date"
+                    editable="bottom"
```
antespi: IMHO, this should be `editable="top"`, because if there are lots of relations, when user clicks to create a new one, apparently nothing happends, user must go to the end of the screen to see an editable line.

partner_relations/__openerp__.py
foutoucour, line 27: Base dependency is not required anymore in v8. It is implicitly added actually.

rafaelbn:
Hi @hbrunn , this module was merged. Could you remove label "needs review" even though it's true, it has a bug.

rafaelbn:
https://github.com/OCA/partner-contact/pull/170#issuecomment-141930996

PR #104, opened 2015-03-30
Better partner_firstname.

yajo:
Pros:
- Cleaner v8 API code.
- Cleaner tests.
- Less overload of framework methods.
- Instead of having a required last name field, now it's required to
  provide either the last or the first name.

Cons:
- Broken backwards compatibility when copying a partner. The "(copy)"
  suffix is added to the first name instead of the last name. However,
  the user is expected to replace that word for the real one, so this
  should not be a big problem.
- Translations must be updated.


yajo:
Sorry because I sent the pull request and later I saw some mistakes. Now it's ready for review. I already sent the signed CLA.

The point in upgrading this module is that I'm also creating a `partner_lastname2` module that I need and I think it can be useful for other users of this repo. I'll post it too ASAP.

gurneyalex:
@Yajo could you tell me under which name the CLA has been sent? I'm not able to identify you from the github login only. Thanks

Yajo:
@gurneyalex Jairo Llopis

partner_firstname/__openerp__.py
```diff
@@ -15,35 +17,34 @@
...
-#
-##############################################################################
 
 {
-    'name': 'Partner first name, last name',
+    'name': 'Partner first name and last name',
     'description': """
-This module splits first name and last name for non company partners
-====================================================================
+Split first name and last name for non company partners
```
yvaucher: Can you move description under README.rst ? using OCA template: https://github.com/OCA/maintainer-tools/blob/master/template/module/README.rst

partner_firstname/__openerp__.py
```diff
@@ -15,35 +17,34 @@
...
+The field *name* becomes a stored function field concatenating the last name
+and the first name.
 
 Contributors
-============
-Jonathan Nemry <jonathan.nemry@acsone.eu>
-Olivier Laurent <olivier.laurent@acsone.eu>
+~~~~~~~~~~~~
+
+- Jonathan Nemry <jonathan.nemry@acsone.eu>
```
yvaucher: see OCA template, use \* and no blank line between contributors

yvaucher: Can you also add Nicolas Bessi (@nbessi) on top of contributors? He is the author of the module so he deserve to be in the readme file

partner_firstname/__openerp__.py
```diff
@@ -15,35 +17,34 @@
 #
 #    You should have received a copy of the GNU Affero General Public License
 #    along with this program.  If not, see <http://www.gnu.org/licenses/>.
-#
-##############################################################################
 
 {
-    'name': 'Partner first name, last name',
+    'name': 'Partner first name and last name',
```
yvaucher, line 22: Would be great to add a `summary` key here.

Your title in description would fix perfectly
"Split first name and last name for non company partners"

partner_firstname/__openerp__.py
yajo: Thanks! I did not know about the README feature, it's great!

Also I created https://github.com/OCA/maintainer-tools/pull/55 to improve that template.

partner_firstname/models.py
```diff
@@ -0,0 +1,97 @@
...
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+from openerp import api, fields, models
+from . import exceptions
+
+
+class ResPartner(models.Model):
+    """Adds last name and first name; name becomes a stored function field."""
```
pedrobaeza: Put this as `_description` variable so that Odoo can be aware of.

partner_firstname/models.py
```diff
@@ -0,0 +1,97 @@
...
+        compute="_name_compute",
+        inverse="_name_inverse",
+        required=False,
+        store=True)
+
+    @api.one
+    @api.depends("firstname", "lastname")
+    def _name_compute(self):
+        """Write the 'name' field according to splitted data."""
+
```
pedrobaeza: There's no need of the space here.

partner_firstname/models.py
```diff
@@ -0,0 +1,97 @@
...
+                                          self.firstname) if p))
+
+    @api.one
+    def _name_inverse(self):
+        """Try to reverse the effect of _compute_name_custom.
+
+        - If the partner is a company, save it in the first name.
+        - Otherwise, make a guess.
+        """
+
```
pedrobaeza, line 53: The same. Empty lines inside methods are waste LOC (lines of code). You already can use comments as separators (as they are here)

partner_firstname/models.py
```diff
@@ -0,0 +1,97 @@
...
+            self.lastname, self.firstname = parts
+
+    @api.one
+    @api.constrains("firstname", "lastname")
+    def _check_name(self):
+        """Ensure at least one name is set."""
+
+        if not (self.firstname or self.lastname):
+            raise exceptions.EmptyNames(self)
```
pedrobaeza: Creating a new Exception is not really needed, because it doesn't require anything but a text, that you can pass as argument

partner_firstname/models.py
```diff
@@ -0,0 +1,97 @@
...
+
+    @api.model
+    def _firstname_install(self):
```
pedrobaeza: You should put this as post_init_hook

partner_firstname/res_user.py
```diff
@@ -1,43 +0,0 @@
...
-##############################################################################
-from openerp.osv import orm
-from openerp.tools.translate import _
-
-
-class ResUsers(orm.Model):
-
-    _inherit = 'res.users'
-
-    def copy_data(self, cr, uid, _id, default=None, context=None):
```
pedrobaeza, line 34: Isn't this code needed?

partner_firstname/models.py
Yajo: This is not the description of the model, but the docstring of what the submodule is doing in this class.

pedrobaeza: Ah, OK, I didn't read it, but I see this kind of docstring so redundant...

partner_firstname/models.py
yajo: IMHO it is always practical to have 1 exception per error. It makes it easier to work with them in submodules and tests.

In fact, [one test I added uses this exception](https://github.com/OCA/partner-contact/pull/104/files#diff-7dde7b0bda5956cf9df1a6948c2fab9fR97) to ensure the correct error is raised. If the ORM raises any other exception during the test and I do what you say, that test will pass when it should not.

partner_firstname/models.py
yajo: I did not find that in the docs and don't know what it is. Do you have any docs?

partner_firstname/models.py
pedrobaeza: Well, it's very unlikely that you need to filter specifically by that error, but no special problem.

partner_firstname/models.py
pedrobaeza: No doc, but directly a use case:
- https://github.com/odoomrp/odoomrp-wip/blob/8.0/quality_control_stock/__openerp__.py#L44
- https://github.com/odoomrp/odoomrp-wip/blob/8.0/quality_control_stock/__init__.py#L10

For being trigerred, you need to remove the module from the module list (because a bug), update the list, and start a fresh install of the module.

partner_firstname/res_user.py
yajo, line 34: That code made that when copying a record it was stored with:
- lastname = "Previous lastname (copy)"
- firstname = "Previous firstname"

Removing that code makes it be copied as:
- lastname = "Previous lastname"
- firstname = "Previous firstname (copy)"

The default behavior seems good enough to me. Nothing of value is added with that code, as the user is expected to change those fields after copying, and I don't think its that important that the word "(copy)" appears in the last name instead of the first.

See [the test for that](https://github.com/OCA/partner-contact/pull/104/files#diff-7dde7b0bda5956cf9df1a6948c2fab9fR40). I removed it just for KISS.

pedrobaeza, line 34: OK, I agree.

partner_firstname/models.py
yajo: OK, I understand it's something that gets executed after the install.

However I don't find it necessary. [See the README](https://github.com/OCA/partner-contact/pull/104/files#diff-ff69505d7ff341d798fb0a454d2bf31cR23). This module assumes that you saved your contacts in the form "Lastname Firstname", but what about people that saves them as "Firstname Lastname"? Or what about those having "Firstname Secondfirstname Lastname"? Or "Firstname Lastname1 Lastname2"? That would lead to errors, and anyway the user will be forced to fulfill them manually the next time he needs to edit that contact.

IMHO It's better to remove a feature than to have it badly implemented, so I removed it to KISS again.

partner_firstname/models.py
yajo: Actually [I will need it for the next PR](https://github.com/OCA/partner-contact/pull/105/files#diff-707f7a50bf3ebd6c95af51564f66788cR97).

yajo:
What do the coverage messages mean? Is there anything left to be done? Thanks.

pedrobaeza:
It means the quantity of the repo code that it's been tested automatically. The ideal is 100%.

More reviews are needed to be merged.

partner_firstname/exceptions.py
```diff
@@ -0,0 +1,36 @@
...
+
+class PartnerNameError(exceptions.ValidationError):
+    def __init__(self, record, value=None):
+        self.record = record
+        self._value = value
+        self.name = _("Error(s) with partner %d's name.") % record.id
+
+    @property
+    def value(self):
+        raise NotImplementedError()
```
bwrsandman: What is the point of subclassing if it is only used once? Can we merge these two Exceptions?

partner_firstname/models.py
```diff
@@ -0,0 +1,92 @@
...
+        Before installing the module, field ``name`` contains all full names.
+        When installing it, this method parses those names and saves them
+        correctly into the database. This can be called later too if needed.
+        """
+        # Find records with empty firstname and lastname
+        records = self.search([("firstname", "=", False),
+                               ("lastname", "=", False)])
+
+        # Force calculations there
+        records._write_name()
```
eLBati: Sorry, what is `_write_name()` ?

bwrsandman:
Could we add a test for [_install_firstname](https://coveralls.io/files/733787334#L80) and for the [exceptions' value functions](https://coveralls.io/files/733787333)?

There are also a few branches in [_name_inverse](https://coveralls.io/files/733787334#L45) which could use a test case.

bwrsandman:
:+1: 
Overall, this looks like a very good improvement.
Great job!

partner_firstname/exceptions.py
yajo: It's a common pattern I follow for modules: do a base exception for that module and subclass it.  It helps with submodules and tests. However, seems like after finishing only one possible exception was left, so I will merge it as you say.

partner_firstname/models.py
yajo: It's a bug, I'll fix it.

yajo:
> Could we add a test for _install_firstname

Hardly, the method's effect can be checked only by creating a partner, then installing the module and then checking the partner's firstname and lastname, because after installing the module all that happens automatically with the constraints.

On the other hand, to run tests, you have to install the module first.

Suggestions are welcome, however.

> and for the exceptions' value functions?

After the last update, it's not necessary now.

> There are also a few branches in _name_inverse which could use a test case.

I'm working on all of it, I'll update soon. Thanks everyone!

yajo:
Everything fixed AFAIK.

partner_firstname/models.py
```diff
@@ -0,0 +1,92 @@
...
+        compute="_name_compute",
+        inverse="_name_inverse",
+        required=False,
+        store=True)
+
+    @api.one
+    @api.depends("firstname", "lastname")
+    def _name_compute(self):
+        """Write the 'name' field according to splitted data."""
+        self.name = " ".join((p for p in (self.lastname,
```
nbessi: Just a detail but u" ".join seems better

partner_firstname/models.py
```diff
@@ -0,0 +1,92 @@
...
+
+    @api.one
+    def _name_inverse(self):
+        """Try to reverse the effect of :meth:`._check_name`.
+
+        - If the partner is a company, save it in the first name.
+        - Otherwise, make a guess.
+        """
+        # Remove unneeded whitespace
+        clean = " ".join(self.name.split(None))
```
nbessi: same here

partner_firstname/views/res_partner.xml
```diff
@@ -0,0 +1,113 @@
...
+
+You should have received a copy of the GNU Affero General Public License
+along with this program.  If not, see <http://www.gnu.org/licenses/>.
+-->
+
+<openerp>
+<data>
+<record id="partner_simple_form" model="ir.ui.view">
+    <field name="name">Add firstname and lastname</field>
+    <field name="model">res.partner</field>
```
nbessi, line 7: Quick remark, changing the indentation to 4 makes the diff unpractical. No harm so

partner_firstname/tests/test_partner_firstname.py
```diff
@@ -25,129 +24,73 @@
...
-        self.assertFalse(
-            vals['firstname'],
-            'Update of the user firstname failed with wrong firstname')
+        super(PartnerFirstnameCase, self).setUp()
+        self.check_fields = True
+        self.create_original()
+
+    def create_original(self):
+        self.original = self.env["res.partner"].create({
+            "lastname": "lastname",
```
nbessi: I would strongly advise to use some unicode string with special char in tests case. I should prevent infamous encoding error regressions.

nbessi:
Some details but it is globally OK.

Thanks

partner_firstname/tests/test_partner_firstname.py
yajo: Good point.

partner_firstname/views/res_partner.xml
yajo, line 7: Sorry about that, I felt the need to give it a more proper format, for easier further maintenance.

yajo:
I rebased, I think that will fix Travis errors. Also, I applied the fixes suggested by @nbessi.

yvaucher:
Thanks for the changes, looking good. :+1: 

@Yajo @gurneyalex is it ok for the CLA question? (https://github.com/OCA/partner-contact/pull/104#issuecomment-87712835)

partner_firstname/__init__.py
```diff
@@ -15,8 +17,5 @@
 #
 #    You should have received a copy of the GNU Affero General Public License
 #    along with this program.  If not, see <http://www.gnu.org/licenses/>.
-#
-##############################################################################
 
-from . import partner
-from . import res_user
+from . import models
```
Olivier-LAURENT: * Making this it seems you used `git rm/git add` commands. Would not be better to use `git mv` to avoid to lose the file history?;
* in addition it would be better to rename the file `partner.py` to `models/res_partner.py`

partner_firstname/__openerp__.py
```diff
@@ -15,35 +17,19 @@
...
     'author': "Camptocamp,Odoo Community Association (OCA)",
     'maintainer': 'Camptocamp, Acsone',
     'category': 'Extra Tools',
     'website': 'http://www.camptocamp.com, http://www.acsone.eu',
     'depends': ['base'],
     'data': [
-        'partner_view.xml',
-        'res_user_view.xml',
+        'views/res_partner.xml',
+        'views/res_user.xml',
```
Olivier-LAURENT: Idem: these 2 files have also been moved/renamed with `git rm/git add` commands

partner_firstname/models.py
```diff
@@ -0,0 +1,92 @@
...
+    @api.one
+    @api.depends("firstname", "lastname")
+    def _name_compute(self):
+        """Write the 'name' field according to splitted data."""
+        self.name = u" ".join((p for p in (self.lastname,
+                                           self.firstname) if p))
+
+    @api.one
+    def _name_inverse(self):
+        """Try to reverse the effect of :meth:`._check_name`.
```
Olivier-LAURENT: Did you mean `_name_compute` in this comment ?

partner_firstname/models.py
```diff
@@ -0,0 +1,92 @@
...
+
+        - If the partner is a company, save it in the first name.
```
Olivier-LAURENT: Did you mean **last** _name_ in this comment ?

partner_firstname/models.py
```diff
@@ -0,0 +1,92 @@
...
+        else:
+            # Company name goes to the lastname
+            if self.is_company:
+                parts = [clean, False]
+
+            # Guess name splitting
+            else:
+                parts = clean.split(" ", 1)
+                while len(parts) < 2:
+                    parts.append(False)
```
Olivier-LAURENT: It would be better to externalize these 3 lines in another method allowing other customizations to subclass it together with `_name_compute` to implement another name format convention

partner_firstname/views/res_partner.xml
```diff
@@ -0,0 +1,113 @@
...
+(at your option) any later version.
+
+This program is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU Affero General Public License for more details.
+
+You should have received a copy of the GNU Affero General Public License
+along with this program.  If not, see <http://www.gnu.org/licenses/>.
+-->
```
Olivier-LAURENT: Why adding such header in xml files? Usually it is intended for pyhton files.

partner_firstname/views/res_partner.xml
```diff
@@ -0,0 +1,113 @@
...
+
+<openerp>
+<data>
+<record id="partner_simple_form" model="ir.ui.view">
```
Olivier-LAURENT: Renaming an xml_id may often crash existing projects using the module at update time. Currently was it necessary to do this ?

partner_firstname/views/res_partner.xml
```diff
@@ -0,0 +1,113 @@
...
+                           attrs="{'required': [('firstname', '=', False)]}"/>
+                    <field name="firstname"
+                           attrs="{'required': [('lastname', '=', False)]}"/>
+                </group>
+            </xpath>
+        </data>
+    </field>
+</record>
+
+<record id="partner_form" model="ir.ui.view">
```
Olivier-LAURENT: Idem regarding xml id renaming

partner_firstname/views/res_user.xml
```diff
@@ -0,0 +1,49 @@
...
+(at your option) any later version.
+
+This program is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU Affero General Public License for more details.
+
+You should have received a copy of the GNU Affero General Public License
+along with this program.  If not, see <http://www.gnu.org/licenses/>.
+-->
```
Olivier-LAURENT: Idem regarding xml files header

partner_firstname/__init__.py
pedrobaeza: Sometimes this happens although you make `git mv`. Diff automatic detection...

partner_firstname/views/res_user.xml
```diff
@@ -0,0 +1,49 @@
...
+
+<openerp>
+<data>
+<record id="users_form" model="ir.ui.view">
```
Olivier-LAURENT: Idem regarding xml id renaming

partner_firstname/__init__.py
bwrsandman: Git doesn't track moves the way bzr does. It makes no difference. Git tries to match the correct files. It isn't always right.

partner_firstname/views/res_partner.xml
bwrsandman: Use a migration script in the case where do you want to keep the rename.

hbrunn: @bwrsandman I think @Olivier-LAURENT's point is that other (probably customer-specific) modules depending on partner_firstname break if you do this. Given that this is a major overhaul, those modules probably will need to be adapted anyway, but I also don't see the benefit in this id change.
Depending on how the depending modules' overrides are crafted, they might survive survive the changes and only depend on this form to be sure to be applied after this did its work. And then, it indeed causes unnecessary work.

yajo:
@Olivier-LAURENT Thanks for your reviews. All fixed now.

yajo:
@yvaucher @gurneyalex About the CLA question, I'm Jairo Llopis, and I already sent that.

partner_firstname/models.py
Olivier-LAURENT: Sorry @Yajo I understand regarding your last version I was not clear enough. My remark was intended for the **last** 3 lines, i.e., the code implementing the inverse logic of `_name_compute()`, expecting a new version like this:
`parts = self._name_split()`
and leaving unchanged the rest of your `_name_inverse()` method.

Olivier-LAURENT:
Hi @Yajo In addition just another little remark.

Previously when duplicating a partner regarding the added suffix `(copy)` the result was:
- for a company partner (`name=X`):
  - lastname: `X (copy)`
  - firstname: _empty_
- for a non company partner (`lastname=X; firstname=Y`):
  - lastname: `X (copy)`
  - firstname: `Y`

Your PR replaces this heuristic by another for a non company partner:
- lastname: `X`
- firstname: `Y (copy)`

introducing different conventions depending on the type of the partner.

Does this change provide a real benefit ?

partner_firstname/models.py
yajo: Sorry but I see no benefit on doing that instead of just changing `_name_inverse()` to match the new behavior. Odoo model methods commonly change records instead of returning values.

Suppose I want to add a `lastname2` field (which actually will be my next PR). I'd have to change 2 methods now instead of just 1.

yajo:
This is Odoo's default behavior. The real point is: does the old behavior provide a real benefit? Because there was an ugly and unnecessary method for doing just that that now is removed. See https://github.com/OCA/partner-contact/pull/104#discussion-diff-28058229.

antespi:
:+1: tested on production instance

antespi:
@Yajo I've PR https://github.com/grupoesoc/partner-contact/pull/1 to you to be able to select names order ('firstname lastname' or 'lastname firstname'), please review, thanks

yajo:
@antespi Wow you were faster than me in testing on production! :smile:

To others... seems like runbots hate me :confused::
- Travis says: [_No output has been received in the last 10 minutes, this potentially indicates a stalled build or something wrong with the build itself._](https://travis-ci.org/OCA/partner-contact/jobs/68791095#L384). My bug or travis'?
- Coveralls says: [_coverage decreased (-0.91%) to 47.65%_](https://coveralls.io/builds/2929547), but then _changed files: 0_. :confounded: My bug or coveralls'?

rafaelbn:
@Yajo we have in several production enviroment firstname's module with a couple of issues that are correctect with your fix, that's why! :-D
:+1:

yajo:
@antespi @rafaelbn Glad to hear that. If so, please update to last version, it fixes some critical usability bugs I found today.

yajo:
[Coveralls fails because the code is shorter](https://github.com/OCA/maintainer-quality-tools/issues/208#issuecomment-118274393), so it should be considered OK.

pedrobaeza:
:+1:

yajo:
Sorry to bother you again, but I found a bug with the onchange method that I'm fixing right now. Maybe you want to roll this back before it's too late.

pedrobaeza:
Please make it the PR fast and I'll fast track it.

houzefa-abba:
Hey,

Indeed the new version no longer exhibits this issue. =)

yajo:
I could not fix it in time, I'll patch it this monday. Sorry for that, the bug was just in the very one line that is not tested :grin:. Is there a way to unit test onchange methods?

hbrunn:
@Yajo say

```
with self.env.do_in_onchange():
    self.env['my.model'].new(whatever_values).my_onchange_method()
```

yajo:
@hbrunn You guys are awesome. You'll have the fix tomorrow.

PR #106, opened 2015-03-30
[ADD] Partner NACE addon

antespi:
# NACE Activities in Partner

This module adds the concept of NACE activity to the partner. Allows you to
select in partner form:
- Main NACE activity in a dropdown (many2one)
- Secondary NACE activities in a multi label input (many2many)

This addon is inspired in OCA/community-data-files/l10n_eu_nace, but it does
not use partner categories to assign NACE activities to partner.

Applies only to partners marked as companies

After installation, you must click at import wizard to populate NACE items
in Odoo database in:
Sales > Configuration > Address Book > Import NACE Rev.2 from RAMON

This wizard will download from Europe RAMON service the metadata to
build NACE database in Odoo in all installed languages.

If you add a new language (or want to re-build NACE database), you should call
import wizard again.

Only Administrator can manage NACE activity list (it is not neccesary because
it is an European convention) but any registered user can read them,
in order to allow to assign them to partner object.


partner_nace/README.rst
```diff
@@ -0,0 +1,34 @@
+NACE Activities in Partner
+==========================
+
+This module adds the concept of NACE activity to the partner. Allows you to
```
pedrobaeza: Explain what is NACE

partner_nace/README.rst
```diff
@@ -0,0 +1,34 @@
...
+
+* Main NACE activity in a dropdown (many2one)
+* Secondary NACE activities in a multi label input (many2many)
+
+This addon is inspired in OCA/community-data-files/l10n_eu_nace, but it does
+not use partner categories to assign NACE activities to partner.
+
+Applies only to partners marked as companies
+
+After installation, you must click at import wizard to populate NACE items
```
pedrobaeza, line 32: Put it on Configuration section

partner_nace/README.rst
```diff
@@ -0,0 +1,34 @@
+NACE Activities in Partner
```
pedrobaeza, line 4: Use full OCA's template from here: https://github.com/OCA/maintainer-tools/blob/master/template/module/README.rst

partner_nace/__openerp__.py
```diff
@@ -0,0 +1,55 @@
...
+{
+    # Addon information
+    'name': 'NACE Activities',
+    'category': 'Localisation/Europe',
+    'version': '1.0',
+    'depends': [
+        'base',
+    ],
+    'external_dependencies': {},
+    # Views templates, pages, menus, options and snippets
```
pedrobaeza: Avoid these comments

partner_nace/__openerp__.py
```diff
@@ -0,0 +1,55 @@
...
+    'data': [
+        'views/res_partner_nace_view.xml',
+        'views/res_partner_view.xml',
+        'wizard/nace_import_view.xml',
+        'security/ir.model.access.csv',
+    ],
+    # Qweb templates
+    'qweb': [
```
pedrobaeza: Remove empty sections

partner_nace/__openerp__.py
```diff
@@ -0,0 +1,55 @@
...
+    ],
+    # Your information
+    'author': 'Antiun Ingeniería, SL',
+    'maintainer': 'Antiun Ingeniería, SL',
```
pedrobaeza: Maintainer should be OCA (or remove this useless key)

partner_nace/models/res_partner_nace.py
```diff
@@ -0,0 +1,49 @@
...
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+
+from openerp import models, fields
+
+
+class ResPartnerNace(models.Model):
+    _name = 'res.partner.nace'
+    # _order = "parent_left"
```
pedrobaeza: Remove these commented lines, but why don't use it?

partner_nace/wizard/nace_import.py
```diff
@@ -0,0 +1,282 @@
...
+from collections import OrderedDict
+
+logger = logging.getLogger(__name__)
+
+
+class NaceImport(models.TransientModel):
+    _name = 'nace.import'
+    _description = 'Import NACE activities from European RAMON service'
+    _parents = [False, False, False, False]
+    _available_langs = {
```
pedrobaeza, line 40: Why do you need to harcode this list?

partner_nace/wizard/nace_import.py
```diff
@@ -0,0 +1,282 @@
...
+            nace.write(data)
+        else:
+            nace = nace_model.create(data)
+        if level >= 1 and level <= 4:
+            self._parents[level - 1] = nace.id
+        return nace
+
+    @api.model
+    def translate_nace(self, node, lang):
+        translation_model = self.env['ir.translation']
```
pedrobaeza, line 217: You can avoid this writing directly to parent model (res.partner.nace) with lang as context key.

partner_nace/models/res_partner_nace.py
antespi: It decrease drastically performance when update NACE items and there are useless at this moment. It could be added when needed in the future.
I'll remove comments in the next commit

partner_nace/wizard/nace_import.py
antespi, line 40: Because there is not a direct relation between Odoo language codes and NACE languages. If you notice, NACE language code sometimes refers to language and sometimes to country. For example, Norwegian Bokmål or Slovenian

partner_nace/models/res_partner_nace.py
pedrobaeza: This is because you have to pass a context variable called `defer_parent_store_computation`. See this PR for a similar case: https://github.com/OCA/server-tools/pull/100.

partner_nace/wizard/nace_import.py
pedrobaeza, line 40: Ahm, right. And European NACE service includes all latin-spanish variants?

partner_nace/wizard/nace_import.py
antespi, line 217: OK, I didn't known. In next revision we could change this. Now, this implementation is fully tested and running well.

partner_nace/wizard/nace_import.py
antespi, line 40: No, but with this map, import proccess will download ES (NACE language code) even if you have installed es_AR language, for instance.

pedrobaeza, line 40: Well, strange use case, but I'm not going to block.

antespi:
Travis and runbot are ok, any other comments from reviewers?
Thanks

yvaucher:
@antespi Shouldn't you use data from https://github.com/OCA/community-data-files/tree/8.0/l10n_eu_nace

What do you add to this module?

antespi:
@yvaucher Two main reasons:
- I need a partner field, not use res.partner.category. 
  - NACE texts are so long
  - NACE items are related between them via parent
  - NACE items have rich descriptions to indicate user when used them.
- This addo downloads NACE items from RAMON online service, and translates them in all available languages (all languages of European Community countries) through a wizard.

I check l10n_eu_nace addon first (I mentioned it in README) in order to use and go, but did not fit our needs.

yvaucher:
@antespi Ok so this could replace community-data-files ?

It seems better to rely on a web service than on static datas.

For long text I see there is a code which doesn't mean much.

We should keey only one of those 2 modules

@clonedagain review, opinion would be great

antespi:
@yvaucher, @clonedagain, I think this addon is compatible with l10n_eu_nace. Admins could select one of them. Maybe we could explain differences in both READMEs, so admin can decide which approach fits better.

clonedagain:
TL;DR: :-1: for the current proposal, @antespi thanks but please integrate your work in existing OCA code and provide a migration path.

The improved features look nice, however I think it's a bad idea to have a new module in OCA for them, it's too much trouble for such a trivial matter.

The existing NACE module uses partner categories because several community modules worked this way.
If you think it's wrong please just make a proposal to change the existing module.
I for one have no preferences. If others agree with you I'd not block, as long as migration scripts are provided.

Regarding the main/secondary activities, in the NAF module (French implementation of NACE) I've added an extra field to the partner form for the primary activity, and left the secondary activities as categories: https://github.com/OCA/l10n-france/blob/8.0/l10n_fr_naf_ape/partner.py#L29
So if you'd please just patch the existing modules (NACE and NAF) to add a "primary activity" field in the NACE module directly, I'd approve the changes.

Regarding the direct download of data it's nice to have, however the data is - on purpose - going to change very rarely (every ten years or so, I think) and we already have translations.
It seems possible to adapt your code to the existing module. Be careful to keep existing user data if you remove the existing data file.

This is also probably the wrong project, being EU-centric localization data. There is talk of creating a l10n-european-union project (or maybe renaming community-data-files to a new name).

antespi:
Hi @clonedagain, thanks for your feedback.
We do not have enought resources to improve OCA/community-data-files/l10n_eu_nace addon and we have already partner_nace addon in production in one customer.
I think that your approach is the best solution and i understand that OCA does not accept two addons with similar features. Is better to improve the addons already exist and provide a migration script.
My suggestion then is: close this PR, move this addon as is to antiun/antiun-odoo-addons/partner_nace and open an issue to explain the pending work to do in OCA/community-data-files repo. So, if everyone aggre with this, close this PR and i'll do the rest.

pedrobaeza:
OK for me.

antespi:
more opinions?

hbrunn:
agreed, seems like the right way to do it. Thanks for your understanding!

antespi:
Thanks all for your comments and review.
I've just move this addons to our public repo: https://github.com/Antiun/antiun-odoo-addons/commit/59b062c1296c8b2d41ec3b7d7ec0e0e206e27761
I've just create an issue in community-data-files repo: https://github.com/OCA/community-data-files/issues/5

PR #107, opened 2015-03-31
[ADD] base_location_nuts addon

antespi:
# NUTS Regions

This module allows to import NUTS locations.

Creates two new fields in Partner object:
- Region (res.partner.region): Classification over state, automatically
  calculated when state is selected
- Substate (res.partner.substate): Classification above state, user must select
  one from available for selected state

You need to install another addon (one for each country) in order to use
these NUTS, for example:
- l10n_es_location_nuts :
  - Spanish Provinces (NUTS level 4) as Partner State
  - Spanish Autonomous communities (NUTS level 3) as Partner Substate
  - Spanish Regions (NUTS level 2) as Partner Region
- l10n_de_location_nuts :
  - German states (NUTS level 2) as Partner State
  - German regions (NUTS level 3) as Partner Substate

After installation, you must click at import wizard to populate NUTS items
in Odoo database in:
Sales > Configuration > Address Book > Import NUTS 2013

This wizard will download from Europe RAMON service the metadata to
build NUTS in Odoo. Each localization addon (l10n_es_location_nuts,
l10n_de_location_nuts, ...) will inherit this wizard and
relate each NUTS item with states. So if you install a new localization addon
you must re-build NUTS clicking this wizard again.

Only Administrator can manage NUTS list (it is not neccesary because
it is an European convention) but any registered user can read them,
in order to allow to assign them to partner object.


base_location_nuts/__openerp__.py
```diff
@@ -0,0 +1,55 @@
...
+    # Your information
+    'author': 'Antiun Ingeniería S.L.',
+    'maintainer': 'Antiun Ingeniería S.L.',
+    'website': 'http://www.antiun.com',
+    'license': 'AGPL-3',
+    # Technical options
+    'demo': [],
+    'test': [],
+    'installable': True,
+    # 'auto_install':False,
```
OSguard: remove comment code

base_location_nuts/models/res_partner_nuts.py
```diff
@@ -0,0 +1,48 @@
...
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+
+from openerp import models, fields
+
+
+class ResPartnerNuts(models.Model):
+    _name = 'res.partner.nuts'
+    # _order = "parent_left"
```
OSguard: remove comment code

base_location_nuts/wizard/nuts_import.py
```diff
@@ -0,0 +1,223 @@
...
+                logger.debug("xpath = '%s', not found" % field_xpath)
+            if field_required and not value:
+                raise Warning(
+                    _('Value not found for mandatory field %s' % k))
+            item[k] = value
+        return item
+
+    def _download_nuts(self):
+        url_base = 'http://ec.europa.eu'
+        url_path = '/eurostat/ramon/nomenclatures/index.cfm'
```
OSguard, line 103: can we store this in ir.config.value whitch create by install?

base_location_nuts/wizard/nuts_import.py
```diff
@@ -0,0 +1,223 @@
...
+        url_params = {
+            'TargetUrl': 'ACT_OTH_CLS_DLD',
+            'StrNom': 'NUTS_2013',
+            'StrFormat': 'XML',
+            'StrLanguageCode': 'EN',
```
OSguard, line 108: should the language depend on the installation?

OSguard:
overall cool stuff, not tested yet

bealdav:
Good job thanks.
May be could you consider module name convention
http://odoo-community.org/page/how-to

Maybe res.partner.nuts could also be res.partner.nut

pedrobaeza:
@bealdav, NUTS is an acronym, not a plural: http://en.wikipedia.org/wiki/Nomenclature_of_Territorial_Units_for_Statistics, so it doesn't apply to remove the 's'.

bealdav:
Oops, sorry for my lack of culture, never heard this term. It was not :chestnut: then ... :laughing:

base_location_nuts/wizard/nuts_import.py
antespi, line 103: This URL is not user configurable, because there is only one in the world.

base_location_nuts/wizard/nuts_import.py
antespi, line 108: No, regions are in theirs own language. But name of regions are translateable to allow user to change the name in its language.
RAMON export file in English, German and French, but most regions have the same name.

base_location_nuts/wizard/nuts_import.py
OSguard, line 103: 1) URLs are data, storing data in code is bad design
2) there is no guarantee that that url never changes
3) maybe you don't want load data from external server, so you cache the data somewhere on your server and then load the data to odoo, then you have also a different url (in the most case this depends on it policy in a organisation)

antespi, line 103: 1) I agree, i've just moved this data to a class attribute.
2) There is no garantee that XML format never change too. When change we will update addon properlly
3) This is a use case not covered by this addon, someone can create a custom addon that inherits NaceImport class and change _url_base and _url_path attributes with its custom values.

IMHO, in this case, add a new config param adds complexity to Odoo configuration and most of admins will not have to change it. More opinions?

OSguard:
Runbot returns:
2015-04-06 09:12:14,897 11692 ERROR 2926052-107-2e410f-all openerp.models: The column parent_id on object res.partner.nuts must be set as ondelete='cascade' or 'restrict'

antespi:
@pedrobaeza I've already done in https://github.com/Antiun/partner-contact/commit/1a3d11e8406c475ef7e0c80e5a28d217f3f541d8

pedrobaeza:
`on_delete="set null"` is redundant. It's the default behaviour.

antespi:
Travis and runbot are ok, any other comments from reviewers?
Thanks

pedrobaeza:
:+1:

antespi:
any news?

hbrunn:
One hint and a conceptual question:
hint: You could have the wizard autostart if you create an appropriate `ir.actions.todo` for it with state `automatic`
conceptual questions: From my understanding, different l10n_\* modules attach different sematics to this fields. Is this really a good idea? How would I fill in a spanish customer's region with the german localization installed?

base_location_nuts/models/res_partner.py
```diff
@@ -0,0 +1,71 @@
...
+            d[k] = r
+        else:
+            d[k] = u[k]
+    return d
+
+
+class ResPartner(models.Model):
+    _inherit = 'res.partner'
+
+    region = fields.Many2one(comodel_name='res.partner.nuts',
```
hbrunn, line 24: can we put domains for country_id on this fields?

antespi:
@hbrunn, thanks for your review:
- Hint: You're right, thanks, i've just fix this
- Question: That's right, I've just PRs two samples: Spainsh and German, very diferent between them:
  - https://github.com/OCA/l10n-spain/pull/150
  - https://github.com/OCA/l10n-germany/pull/11

The problem i couldn't fix is that Spanish translation sets labels according to Spanish nuts but what happens when user introduces a german address? Labels doesn't mean what data means. I don't know how to change labels/placeholder in on_change (v8 flavor) method, do you?

hbrunn:
you could have function fields that you use as labels (so you'll have to write the html code for the label yourself in the view). Then you wouldn't really need the localization modules, if you extend res.country with fields telling what level means what in the respective country, and if they apply at all. Based on that, you can hide the fields not relavant for the country selected

antespi:
@hbrunn sorry for my late answer and thanks for your review. I'm following your instructions to provide a correct label depending on partner country.

In the other hand, we need localization modules for these reasons:
- We have to relate state_ids with NUTS items. Each localization defines an addons to add state_ids and each one use its nomenclature. Not always match with NUTS items. We can not use an extension of res.country, we need a map, look at spainsh one for example: https://github.com/Antiun/l10n-spain/blob/pr-add_l10n_es_location_nuts/l10n_es_location_nuts/wizard/nuts_import.py
- In one country state_ids match with NUTS level 4 (spain) and other match with NUTS level 2 (germany). Locallization addon define this match and implement onchange rules (if apply)
- Each localization team must be responsible of these relations, if they change toponyms addons must change it nuts addons properly.

antespi:
@hbrunn i've just finished the fix you commented. Now when country changes, labels changed dinamically, and localization addons can change labels properly. There are specific labels for each country, and each addon can translate them to each language.

For testing, I recommend to install:
- [Spain localization](https://github.com/OCA/l10n-spain/pull/150)
- [German localization](https://github.com/OCA/l10n-germany/pull/11)

rafaelbn:
:+1:

rafaelbn:
Hi @OSguard @bealdav @pedrobaeza @hbrunn , please could you review this PR, changes are done from June and module it's working fine in production environment. Thanks

base_location_nuts/README.rst
```diff
@@ -0,0 +1,85 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+    :alt: License: AGPL-3
+
+NUTS Regions
```
pedrobaeza, line 5: Put top line

base_location_nuts/README.rst
```diff
@@ -0,0 +1,85 @@
...
+you must re-build NUTS clicking this wizard again.
+
+
+Usage
+=====
+
+Only Administrator can manage NUTS list (it is not neccesary because
+it is an European convention) but any registered user can read them,
+in order to allow to assign them to partner object.
+
```
pedrobaeza: Put "Try me on runbot" button

base_location_nuts/models/res_partner.py
pedrobaeza, line 24: This question is not asked

pedrobaeza:
I re-check for new conventions on README, and few remarks. Honor that remarks and answer the question from @hbrunn, and I'll merge it.

base_location_nuts/models/res_partner.py
```diff
@@ -0,0 +1,80 @@
...
+                             string="Region")
+    substate = fields.Many2one(comodel_name='res.partner.nuts',
+                               string="Substate")
+    lbl_region = fields.Char(compute='_labels_get')
+    lbl_substate = fields.Char(compute='_labels_get')
+
+    @api.one
+    @api.depends('country_id')
```
lmignon, line 32: Sorry but I don't understand the meaning of this `api.depends` since your labels are computed by the translation tools.
Even if you change the country, your labels will stay the same since they depend of the language of the connected user. Maybe I miss something.

antespi, line 32: @lmignon each country define what 'region' and 'substate' fields mean. For example:
- Spain:
  - substate = 'Autonomous Community' 
  - region = 'Region'
- Germany
  - substate = 'Government region'
  - region = 'District'

This depends on NUTS classification and which NUTS level is related with Odoo res.country.state. Each localization project (l10n-spain, l10n-germany, ...) creates their res.country.state objects

After this, you can translate labels, of course. For example, in spanish will be:
- España (Spain):
  - substate = 'Comunidad autónoma' 
  - region = 'Región'
- Alemania (Germany)
  - substate = 'Región de gobierno'
  - region = 'Distrito'

pedrobaeza, line 32: But what you will get is the same label for all customers, independently of their country. Ideally, you should get the correct label for the country language. This is similar to the address layout depending on the selected country.

antespi, line 32: Maybe, you have to see these other PRs to understand this:
- https://github.com/OCA/l10n-spain/pull/150 : `l10n_es_location_nuts` addon setting labels for matching NUTS levels to spanish states defined by `l10n_es_toponyms` addon
- https://github.com/OCA/l10n-germany/pull/11 : `l10n_de_location_nuts` addon setting labels for matching NUTS levels to german states defined by `l10n_de_country_states` addon

This is not only country dependant, but also localization dependant, that's why this is not a country config parameter

@pedrobaeza, you'll will have diferent labels in different partners if country is different. In examples:
- Partner A: Germany country, labels 'Government region' and 'Region'
- Partner B: Spain country, labels 'Autonomous Community' and 'Region'

base_location_nuts/__openerp__.py
```diff
@@ -0,0 +1,42 @@
...
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+
+{
+    'name': 'NUTS Regions',
+    'category': 'Localisation/Europe',
+    'version': '1.0',
```
pedrobaeza: Put new version numbering scheme

base_location_nuts/models/res_partner.py
pedrobaeza, line 32: Put a comment on the method then indicating that each localization module will inherit this method to add their specific labels.

antespi:
@pedrobaeza, remarks and version number fixed, squash and rebase. Thanks for your review

base_location_nuts/models/res_partner.py
antespi, line 24: I'm testing to see if we can do what @hbrunn is asking. This configuration must be compatible with onchages defined in localizations addons (l10n_es_location_nuts, l10n_de_location_nuts, etc ...)

rafaelbn:
This modules has been in customer's production environment for over 3 months. Tested! :+1: 
Any technical remarks?

PR #111, opened 2015-04-15
Slovene translations added

sysadminmatmoz:


pedrobaeza:
Technically :+1:

yvaucher:
Technically :+1:

tafaRU:
Technically LGTM too, so :+1:

tafaRU:
Just noticed po headers still refer to OpenERP instead Odoo, but it's quite irrilevant as error.
@yvaucher @pedrobaeza what's your opinion?

pedrobaeza:
That comments are ignored, so it's only for reference. We know all what OpenERP is :wink:

tafaRU:
@pedrobaeza thanks for your clarification. I proceed with the merge then.

Issue #112, opened 2015-04-17
Can not create user after installing partner_firstname


NL66278:
After installing partner_firstname I get an invalid field error for field name, when trying to create a new user.

NL66278:
Solved by: https://github.com/OCA/partner-contact/pull/113

PR #113, opened 2015-04-17
[FIX] cannot create user due to required name.

NL66278:
Use @api.onchange to prevent error with required name when creating user.

### DIFF
--- a/partner_firstname/res_user.py
+++ b/partner_firstname/res_user.py
@@ -1,4 +1,5 @@
 # -*- coding: utf-8 -*-
+"""Extend res.users to be compatible with split name in res.partner."""
 ##############################################################################
 #
 #    Author: Nicolas Bessi. Copyright Camptocamp SA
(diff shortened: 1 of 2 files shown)


NL66278:
Solves: https://github.com/OCA/partner-contact/issues/112

partner_firstname/res_user.py
```diff
@@ -17,14 +18,19 @@
...
+class ResUsers(models.Model):
+    """Extend res.users to be compatible with split name in res.partner."""
     _inherit = 'res.users'
 
+    @api.onchange('firstname', 'lastname')
+    def change_name(self):
+        names = [name for name in [self.firstname, self.lastname] if name]
+        self.name = ' '.join(names)
+
     def copy_data(self, cr, uid, _id, default=None, context=None):
```
pedrobaeza, line 34: You can take the occasion to change this to new API.

yvaucher:
:+1:

NL66278:
@pedrobaeza Hello Pedro, I tried to make the changes as small as possible. The copy_data method itself in openerp/models.py has still the same interface as before.

yvaucher:
I agree with @NL66278 to keep this diff small. API 8.0 on copy_data is out of context for a fix.

pedrobaeza:
OK then :+1:

PR #115, opened 2015-04-30
Do no longer set the 'ref' field readonly

guewen:
Because it is set to readonly conditionally on the view.
Setting the field to readonly prevents users to import it with csv files.


pedrobaeza:
:+1:

fclementic2c:
functional test OK
:+1:

yvaucher:
:+1:

PR #118, opened 2015-05-18
Add module partner_gender.

yajo:
This is a very simple module that adds gender to partners. The field is hidden on companies.


partner_gender/README.rst
```diff
@@ -0,0 +1,59 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+    :alt: License: AGPL-3
+
+Partner Gender
+==============
+
+This module was written to extend the functionality of `base_contact`_ to
```
pedrobaeza: You should call it _base_contact_gender_ then.

pedrobaeza:
The icon represents those that doesn't consider male nor female (something with the sexual identity), so they blend both signs. You should use an icon with each traditional gender sign.

partner_gender/README.rst
bwrsandman: Agreed, we should be upfront about that dependency. They are begining to be more of a family of modules.

pedrobaeza: Anyways, I see that the field is added to res.partner, so there's no problem to inherit from other view and keep the name, and those that don't like _base_contact_ (like me), can enjoy this module.

yajo: I think the whole point is that _base_contact_ is wrong. It makes many things that are totally independent.

I inherit the module just for [the view sheet of personal data](https://github.com/OCA/partner-contact/pull/118/files#diff-93b96a5c10e7af0bbc086441331b2017R27), but I agree that I don't need that dependency for anything else.

IMHO _base_contact_ should be just that: a base for all modules that modify the contact form (BTW, it should be _base_partner_ for more general usage). It should define the new views and nothing more. Then 3 modules out of there should appear:
- _base_contact_type_
- _base_contact_birthdate_
- _base_contact_nationality_

And, of course, _base_contact_gender_, which is what we are really talking about.

pedrobaeza: _base_contact_ is intended for another thing: having a notion of contact, that can be attached to multiple companies, but this concept has been misfigured through versions, and with the new v7 partner model, it remains in a grey land which overlaps some expected functionalities from new model (contacts exists already as res.partner with a parent defined).

So all should be _base_partner__, and better partner__, because there are plans to dettach partner from base module.

yajo:
All fixed (except the dependency problem, which is a different issue IMHO).

pedrobaeza:
Well, I prefer the partner_gender approach, so I'm not going to block this, but don't approve it.

One last thing. If the icon has been fetched from Internet, you should put in credits the URL where you got it.

yajo:
> Well, I prefer the partner_gender approach, so I'm not going to block this, but don't approve it.

I prefer it too. I can do this:
- Create a new base (_partner_personal_data_base_, for example) that only creates a placeholder view with that sheet, since it's the only piece in common for all other modules.
- Rebase _base_contact_ on that.
- Strip birthdate and nationality from _base_contact_ and make them separate modules.
- Rebase this one on _partner_personal_data_base_.

Of course, it will break things, so I need approval before starting such work.

Another solution is moving the gender outside the personal data sheet, but seems to me that it's the right place for this field to be.

> One last thing. If the icon has been fetched from Internet, you should put in credits the URL where you got it.

It's from FontAwesome, which is included in Odoo, so I thought it was not necessary, but I will put it.

pedrobaeza:
But I don't understand why you need such refactoring. _partner_lastname_ also add a field for only persons, and make it without that dependency. Can't you do the same?

yajo:
It's different. In that module we are talking about the name, which is the main field of any partner. I don't know if you tested yourself the module, but what it does is that when unchecking the box _is it a company?_ it _replaces_ the name field with a firstname and lastname.

So, there I take one existing field and replace it with two new.

This time, I'm creating a new one, so I need a place to put it.

yajo:
BTW, I also need the birthdate functions, but not the rest, so I'd be glad to do that work. Also, the birthdate functions are a bit buggy, I could fix it then too.

pedrobaeza:
Well, for a stable v8 version maybe it's too disruptive to change all of this, but propose it to the community to see.

yajo:
> Well, for a stable v8 version maybe it's too disruptive to change all of this, but propose it to the community to see.

There is one way to make it non-disruptive: deprecation.

Replace _base_contact_ with the other modules. Then, make _base_contact_ 8.0 a dummy module that just installs the other ones, with a deprecation warning. Then, remove it from master. Finally, add _partner_gender_.

No functionality is lost, new modules get installed in next update, and no blocking for progress. What do you think?

pedrobaeza:
Yes, that can be a viable roadmap. Let's see what others say.

yajo:
Then, I'll close this PR for now and open a bug to discuss that. Thanks!

Issue #119, opened 2015-05-19
Deprecate base_contact and split it in other modules


yajo:
After discussion in https://github.com/OCA/partner-contact/pull/118, it became clear that the current implementation of _base_contact_ is not what its name suggests (a base for contact modules), but a collection of unrelated functionalities that remain from old times and block further development.

So, my suggestion is:
1. Create a new base module for contact stuff. It will just have a sheet called _Personal data_ that will display only for contact partners. It's intended to serve as a base for other modules, so it will do nothing more. Let's call it _contact_base_.
2. Split the functionalities from _base_contact_ in 3 modules:
   1. _contact_in_several_partners_: set several job positions in different companies per contact.
   2. _contact_birthdate_.
   3. _contact_nationality_.
3. Deprecate _base_contact_, making it a dummy module that depends on those 3 new modules. This way, after next update all will continue working as it did before.
4. Remove _base_contact_ whenever branch 9.0 appears.

After all that is done, new modules for contact personal data may come easier. I myself will submit some: contact gender, disability, and some others just for the l10n-spain team.

My company needs all these changes, so I'll start working on this right now unless somebody tells me they are already doing it or have a better roadmap for those modules. I don't mind having our own fork forever if nobody wants these changes, but I think this work will benefit everyone, so I'd prefer to have approval from the community and give back a little of the help we have received and appreciate so much.

pedrobaeza:
I agree, but prefer to prepend partner_ to all modules

yajo:
No problem. Do you mean for instance _partner_contact_birthdate_ or _partner_birthdate_?

pedrobaeza:
It seems to be more coherent to name them partner_contact_base, partner_contact_birthdate...

yajo:
:+1:

yvaucher:
IMO a contact being a partner I don't see the need of adding `partner_` to all module technical names

And indeed a split of functionalities would be better.

yajo:
Then please come to a consensus. For now I've started with @pedrobaeza suggestion. You can track progress if you want in https://github.com/grupoesoc/partner-contact/tree/deprecate-base_contact.

pedrobaeza:
@yvaucher, partner prefix identify the "area" you are working with.

yvaucher:
[...]

PR #120, opened 2015-05-20
Deprecate base_contact

yajo:
This fixes https://github.com/OCA/partner-contact/issues/119.

The new modules:
- partner_contact_personal_information_page: The new base.
- partner_contact_birthdate
- partner_contact_in_several_companies
- partner_contact_nationality

The old base_contact warns deprecation everywhere.

Old translations preserved where possible with credits.

After merging we can start again with https://github.com/OCA/partner-contact/pull/118.


base_contact/README.rst
```diff
@@ -34,6 +45,7 @@ Contributors
 * Sandy Carter <bwrsandman@gmail.com>
```
bwrsandman: Could you use my savoirfairelinux email? sandy.carter@savoirfairelinux.com while you're there

dreispt:
This is a case study for me. 
I'm curious: how will this behave in existing installations, when doing an upgrade? Is there risk of losing data?

yajo:
I'll do some checks and let you know.

yajo:
I did this:
- Create an empty demo database.
- Install `base_contact`.
- Allow technical features for admin.
- Create a contact.
- Put there birthdate, nationality and some job positions.
- Shutdown the server.
- `git checkout deprecate-base_contact`.
- Start the server.
- **Configuration > Update modules list.**
- **Manually upgrade module `base_contact`.**
- Checked that all data from that contact is OK (and it was).
- Change birthdate, nationality and job positions.
- Save.
- Record updated OK.

The bold lines are the necessary steps for sysadmins to follow, and nothing was broken.

hbrunn:
@Yajo great initiative, but you definitely need migration scripts for that. At the very least, you need to migrate the xmlids of the fields's ir.model.fields record to the appropriate new module (see https://github.com/OpenUpgrade/OpenUpgrade/pull/303 for discussion, I just stumpled upon this one yesterday). Otherwise, odoo will drop the columns if you deinstall the then obsolete base_contact.

yajo:
Ignore last update, it's just a rebase.

@hbrunn Thanks but I don't understand you very well... Can you explain better please? And if you have any docs (or at least examples) for that, they'll be very appreciated.

hbrunn:
@Yajo when you deinstall a module, odoo unlinks all records that have an xml id (look at the table `ir_model_data`) that were created by this module. This happens automatically on module installation. When a module adds a field, an entry in `ir_model_fields` is created and a record in `ir_model_data` pointing to this record.

For documentation, use the source: https://github.com/odoo/odoo/blob/8.0/openerp/models.py#L356

Now on module deinstallation, odoo unlinks the `ir.model.fields` entry, but this results in dropping the column: https://github.com/odoo/odoo/blob/8.0/openerp/addons/base/ir/ir_model.py#L355 - that's a good thing, because uninstalled modules don't leave stuff behind this way.

But in this case, the xmlid belongs to the module `base_contact`, so if you uninstall that, you lose the columns now living in the other modules.

To fix that, create a file `base_contact/migrations/8.0.2.0/pre-migrate.py` saying

```
def migrate(cr, version):
  cr.execute("update ir_model_data set module='the new module' where module='base_contact' and name in ('the generated xmlid name(s) of the field(s) in question')")
  #repeat for all moved fields
```

This will run if you update `base_contact` from a version lower than `8.0.2.0` to this version. (For testing purposes, you can modify `latest_version` in `ir_module_module` to trigger this)

Please don't hesitate to ask if something is still unclear

yajo:
Thanks, it's clear now. I also found https://github.com/odoo/odoo/blob/8ff7230e2d6d49e5b880fbee9589de1e8a78fb95/openerp/modules/migration.py#L38 useful.

I'm on it.

yajo:
@hbrunn Well spotted. After uninstalling, fields were removed. Last update fixes that. Thank you!

dreispt:
@hbrunn Many thanks for the advice!

yajo:
Rebased to see if that fixes the runbot builds.

yajo:
Can anyone help me fixing the runbot error please? It seems to me like it's treating the deprecation warning as an error, but it's the expected behavior.

bwrsandman:
Runbot is sending fails to github when it has a warning status

base_contact/__init__.py
```diff
@@ -19,4 +20,7 @@
 #
 ##############################################################################
 
-from . import base_contact
+import logging
+
+
+logging.getLogger(__name__).warn("This module is DEPRECATED. See README.")
```
yajo: Warnings from runbot come from here, but it's a normal deprecation warning. Nothing to fix.

bwrsandman: That this will flash every time a ~~server~~ worker is launched even if the module isn't installed.

gurneyalex:
@yajo: can you replace your logging statement with the following code:

```
from warnings import warn
warn('base_contact: this module is deprecated. See README', DeprecationWarning)
```

base_contact/__init__.py
yajo: I'm not sure if that can be avoided...

yajo:
@gurneyalex Doing that in my tests logs nothing. Are you sure you want me to push that?

Another solution is to downgrade the message to info, but IMHO it should be a warning.

yajo:
I opted for transforming the warn in info to fix the checks.

gurneyalex:
@Yajo yes DeprecationWarnings are ignored by default (https://docs.python.org/2.7/library/warnings.html?highlight=warnings#warning-categories)

If you prefer having a WARNING log level (and I agree this is better than Info), you could check the hostname of the server. If it matches runbot.*.odoo-community.org then don't log. I can make sure that the runbot has an environment variable set (e.g. OCA_RUNBOT) and you can use this rather than the fqdn for the test. And the best way of doing this is to use code such as the following:

in the **init**.py of your module:

```
import logging
import os
logger = logging.getLogger(__name__ + ".deprecated")
if os.environ.get("OCA_RUNBOT"):
    logger.setLevel(logging.ERROR)
logger.warning("Module %r is DEPRECATED. See %s/README.", 
                         __name__, 
                         os.path.dirname(__file__))
```

(caution, I did not test this, there may by typos). 

Addind '.deprecated' is important so that the loggers in the other files of the modules are not impacted.

yajo:
I removed the __name__ part because anyway the logger already logs that automatically. Let's see if this time it passes the checks.

gurneyalex:
@Yajo I did not have time to update the runbot, so it will fail.

yajo:
Can you explain yourself better, please? Sorry, I did not understand.

gurneyalex:
fixed typo in my comment.

yajo:
Then what should I do now?

yajo:
@gurneyalex Have you updated it already?

yajo:
It's been almost one month from the last comment. Any chance to fix the runbot errors? Thanks.

hbrunn:
I just restarted the builds, let's see what happens now

yajo:
Still fails for the warning. I'm trying another method to skip it.

yajo:
[The OCA_RUNBOT trick fails](https://runbot.odoo-community.org/runbot/build/3109936), any other ideas?

yajo:
After this morning's attempts I ran out of ideas, do I raise an INFO instead of a WARNING?

pedrobaeza:
You can try:

```
logger_call = logger.info if os.environ.get("OCA_RUNBOT") else logger.warning
logger_call("Module %r is DEPRECATED. See %s/README.",
            __name__,
            os.path.dirname(__file__))
```

bwrsandman:
I don't like the idea of hardcoding OCA_RUNBOT in modules, that's not very clean.

@gurneyalex had a solution at some point

pedrobaeza:
This is his solution reworked, as what he proposed is not working: https://github.com/OCA/partner-contact/pull/120#issuecomment-106201428

yajo:
@pedrobaeza [That's exactly what I was meaning](https://github.com/grupoesoc/partner-contact/commit/2e39bfd1470accd1a57f6031c4d4fc9d0aa505b9) when I said "this morning attempts". :grin:

That leads me to think that the real problem is that either `$OCA_RUNBOT` does not exist, or it evaluates to `False`, so I'm going to do one last try to discard that last option.

yajo:
There it is, let's see...

pedrobaeza:
It's still the same. Maybe @gurneyalex (who is now on holidays) doesn't put that environment variable in runbot, and that's why this problem. I propose to unlock this putting always log.info, and we can go back again to this afterwards.

yajo:
It's a solution, but if any sysadmin has a filter to not see INFO logs, they will not get noticed. We would be fixing a runbot environment to hurt the real world one, and it does not seem very sensible to me. I'll do what you OCA people say, but I'd like to hear more opinions.

gurneyalex:
@Yajo sorry, been really busy and could not work on this until now. I'll update the set up this week and ping you when it's done

gurneyalex:
:+1: build is :green_heart:

yajo:
So it was a bug in the runbot after all, no?

gurneyalex:
@Yajo not a bug. Just needed to find the time to update the runbot servers and the setup procedure etc...

pedrobaeza:
OK, finally merging

lepistone:
Hi,

sorry I arrive late in the discussion and this is now merged, but I think the warning should be moved somewhere that is executed only if the module is installed. I'm unsure where though. Maybe `auto_init` in a model?

Thanks!

dreispt:
@hbrunn Just checking:

> This will run if you update base_contact from a version lower than 8.0.2.0 to this version. (For testing purposes, you can modify latest_version in ir_module_module to trigger this)

The module version was bumped to 2.0.0, which is equivalent to 8.0.2.0.0.
This is the version the migration refers to. And the extra `.0` is OK.
Correct?

hbrunn:
Good you asked, I'm not so sure: https://github.com/OCA/OCB/blob/8.0/openerp/modules/migration.py#L96 ff

Did anyone try if the migration script runs when upgrading from base_contact after this version number change? If not, that's a serious problem when people uninstall base_contact.

@lepistone I have the same problem. `init_hook` seems the good choice to me.

hbrunn:
ping @dreispt (just to avoid you missed my previous non-attributed post)

dreispt:
@hbrunn I didn't test. I guess the relevant pieces of code are:
- https://github.com/OCA/OCB/blob/8.0/openerp/modules/migration.py#L140
- https://github.com/odoo/odoo/blob/8.0/openerp/tools/parse_version.py#L30

From the code, I'm guessing it should work...

hbrunn:
@dreispt yes I just debugged thourgh it. There are some curious transformations of the version string, and it works out in the end. So nothing to do and sorry for the noise

dreispt:
@hbrunn Thanks for your time checking that.

lepistone:
@hbrunn about the init hook: I'm not sure myself, as that would be shown only on module installation. Where would you put them, and with which log level? Thanks!

hbrunn:
@lepistone I was thinking about raising in the init hook, but that indeed doesn't do anything for existing installations. Another possibility might be setting `'installable': False` in the manifest, but I never tried our how this behaves if the module in question is already installed.

For new users, I'd declare an abstract model that gives a warning in its `_register_hook`. We might even set the module's state to 'to uninstall' there, then it will be gone with the next database update.

But on the other hand: What's the problem with keeping this module as sort of metapackage?

yajo:
> But on the other hand: What's the problem with keeping this module as sort of metapackage?

The only problem is maintaining deprecated code, but it should work as it is.

Indeed the warning raises even when it is not installed, but no idea about how to avoid that. It should be a separate issue anyway.

PR #125, opened 2015-06-01
Keep partner's lastname if it has one

steveferry:
On creation of a user, keep the partner's lastname if it has already been set to not have the email instead

### DIFF
--- a/partner_firstname/res_user.py
+++ b/partner_firstname/res_user.py
@@ -42,8 +42,7 @@ class ResUsers(orm.Model):
             res_partner = self.pool.get('res.partner')
             partner = res_partner.browse(cr, user, vals2['partner_id'],
                                          context)
-            if partner.lastname:
-                vals2['lastname'] = partner.lastname
+            vals2['lastname'] = partner.lastname
         elif 'login' in vals and 'lastname' not in vals:
             vals2['lastname'] = vals2['login']
         return super(ResUsers, self).create(cr, user, vals2, context=context)
--- a/partner_firstname/res_user.py
+++ b/partner_firstname/res_user.py
@@ -36,13 +36,14 @@ class ResUsers(orm.Model):
         """
         vals2 = vals.copy()
 
-        res_partner = self.pool.get('res.partner')
-        partner = res_partner.browse(cr, user, vals2['partner_id'], context)
-
         if 'name' in vals:
             vals2['lastname'] = vals2['name']
-        elif partner.lastname and 'lastname' not in vals:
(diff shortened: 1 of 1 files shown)


partner_firstname/res_user.py
```diff
@@ -35,8 +35,14 @@ def create(self, cr, user, vals, context=None):
         installed
         """
         vals2 = vals.copy()
+
+        res_partner = self.pool.get('res.partner')
+        partner = res_partner.browse(cr, user, vals2['partner_id'], context)
+
         if 'name' in vals:
             vals2['lastname'] = vals2['name']
+        elif partner.lastname:
```
StefanRijnhart: The line below takes into account that 'lastname' can be present in vals. Do you need to do that too?

steveferry: Yes you're right! I missed this case, thanks!

gurneyalex:
CLA ok

gurneyalex:
This change breaks the travis build ==> needs fixing

ecino:
:+1: We need this bug fix

partner_firstname/res_user.py
```diff
@@ -35,8 +35,15 @@ def create(self, cr, user, vals, context=None):
...
         """
         vals2 = vals.copy()
+
         if 'name' in vals:
             vals2['lastname'] = vals2['name']
+        elif 'lastname' not in vals and 'partner_id' in vals:
+            res_partner = self.pool.get('res.partner')
+            partner = res_partner.browse(cr, user, vals2['partner_id'],
+                                         context)
+            if partner.lastname:
```
StefanRijnhart: Can it actually be the case that the partner has no lastname? In that case, you need to refactor the code below to also assign the login to vals['lastname'] if there is a partner but it does not have a lastname. This clause is now skipped as soon as there is a partner_id.

ecino: lastname is a required field, so I think this case could not happen.

StefanRijnhart: In that case, the condition can be removed.

ecino:
Any review for this one ?

StefanRijnhart:
@steveferry when you follow up the comments from the review, please write a note saying that you did. It is difficult for me as a reviewer to keep up to date if I don't get such notifications.

Changes look good, but runbot failed and its logs have expired. I'm requesting a rebuild.

StefanRijnhart:
Runbot warnings are unrelated to this particular change. :+1:

pedrobaeza:
:+1:

Issue #129, opened 2015-06-17
[8.0] partner_relations


antespi:
```
File "/srv/openerp/instances/openerp-oca-runbot/parts/odoo-extra/runbot/static/build/2947363-8-0-f99b1d/openerp/addons/partner_relations/model/res_partner_relation_all.py", line 187, in write
    underlying_objs.type_selection_id.id
AttributeError: 'list' object has no attribute 'type_selection_id'
```

Use case:
1. Create two relations for Company-Company ('Is competitor of' and 'Representative')
2. Go to a Partner and create a relation with another partner
3. Save
4. Edit the created relation and change 'Relation Type' column
5. Save

NOTE: If you select the inverse relation of the current relation, bug doesn't appear

Is this the expected behavior? 
- If yes, then 'Relation Type' domain must be more restrictive in order to avoid user to select a bad relation type
- If not, write method must support this use case

Thanks for your support

PR #132, opened 2015-06-17
[FIX] Type Exception when writing res.partner.relation.all model

antespi:
Fix https://github.com/OCA/partner-contact/issues/129

### DIFF
--- a/partner_relations/model/res_partner_relation_all.py
+++ b/partner_relations/model/res_partner_relation_all.py
@@ -177,6 +177,7 @@ class ResPartnerRelationAll(models.AbstractModel):
     def write(self, vals):
         """divert non-problematic writes to underlying table"""
         underlying_objs = self.get_underlying_object()
+        underlying_objs = underlying_objs and underlying_objs[0]
         vals = {
             key: val
             for key, val in vals.iteritems()
@@ -209,4 +210,6 @@ class ResPartnerRelationAll(models.AbstractModel):
     @api.one
     def unlink(self):
         """divert non-problematic creates to underlying table"""
-        return self.get_underlying_object().unlink()
+        underlying_objs = self.get_underlying_object()
+        underlying_objs = underlying_objs and underlying_objs[0]
+        return underlying_objs.unlink()
--- a/partner_relations/model/res_partner_relation_all.py
+++ b/partner_relations/model/res_partner_relation_all.py
@@ -138,8 +138,7 @@ class ResPartnerRelationAll(models.AbstractModel):
         return super(ResPartnerRelationAll, self)._auto_init(
             cr, context=context)
 
-    @api.one
-    def get_underlying_object(self):
+    def _get_underlying_object(self):
         """Get the record on which this record is overlaid"""
         return self.env[self._overlays].browse(self.id / PADDING)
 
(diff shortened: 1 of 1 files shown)


partner_relations/model/res_partner_relation_all.py
```diff
@@ -209,4 +210,6 @@ def create(self, vals):
     @api.one
     def unlink(self):
         """divert non-problematic creates to underlying table"""
-        return self.get_underlying_object().unlink()
+        underlying_objs = self.get_underlying_object()
+        underlying_objs = underlying_objs and underlying_objs[0]
+        return underlying_objs.unlink()
```
pedrobaeza: Why only remove the first one?

partner_relations/model/res_partner_relation_all.py
```diff
@@ -177,6 +177,7 @@ def onchange_type_selection_id(self):
     def write(self, vals):
         """divert non-problematic writes to underlying table"""
         underlying_objs = self.get_underlying_object()
+        underlying_objs = underlying_objs and underlying_objs[0]
```
pedrobaeza: You can avoid this sentence putting at the end of the upper one `[:1]`

partner_relations/model/res_partner_relation_all.py
antespi: This is an @api.one method

partner_relations/model/res_partner_relation_all.py
antespi: But self.get_underlying_object() could return False, isn't it?

partner_relations/model/res_partner_relation_all.py
antespi: After I've got a shower and re-thought what you wrote, you,re rigth. Here we need a loop or make self.get_underlying_object() returns a recordset not a list.

partner_relations/model/res_partner_relation_all.py
antespi: Same that other coment: make self.get_underlying_object() returns a recordset not a list.

partner_relations/model/res_partner_relation_all.py
adrienpeiffer: Don't hesitate to correct me if i'm wrong but `self.get_underlying_object()` is call in an @api.one method.
Then if `get_underlying_object` use `self.id`. This method should return a list with only 1 object.

adrienpeiffer:
Why not change the decorator of `get_underlying_object()` to avoid to have a list of recordset. We can change this to `@api.model` and add id in the method declaration. What do you think ?

antespi:
@adrienpeiffer you're right. Remove api.one is better solution. I've marked `get_underlying_object()` method as private, it's used only in this model as internal method.

antespi:
Please @bwrsandman, could you check this bugfix?

bwrsandman:
Looks fine, but I am no longer doing any Odoo.

rafaelbn:
Looks fine :+1:

pedrobaeza:
The method is still unprotected from being called from a multi-recordset, raising the ensure_one in `self.id`, but I see that this method is only called from `@api.one` methods, so :+1:

adrienpeiffer:
:+1:

antespi:
@bwrsandman We'll miss you so much. Good luck wherever you go.

PR #133, opened 2015-06-19
Added portuguese translation files po

danimaribeiro:
I've requested a new idiom on transifex:
https://www.transifex.com/projects/p/OCA-partner-contact-8-0/
But I'm not sure if I should add the .po files here first to translate later, so in any case I'm sending this PR to add those files.

### DIFF
--- a/base_contact/i18n/pt_BR.po
+++ b/base_contact/i18n/pt_BR.po
@@ -1,0 +1,186 @@
+# Translation of OpenERP Server.
+# This file contains the translation of the following modules:
+# 	* base_contact
+#
+msgid ""
+msgstr ""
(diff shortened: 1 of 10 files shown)


danimaribeiro:
Brazilian Portuguese translated on transifex.
https://www.transifex.com/projects/p/OCA-partner-contact-8-0/language/pt_BR/

gurneyalex:
@pedrobaeza not sure how transifex works for this. Do you have an idea what is needed here?

danimaribeiro:
I think we can close this, I translated in Transifex, and the translations are ok.

pedrobaeza:
Exactly, this is not needed, as an script push back the translations from Transifex periodically.

PR #134, opened 2015-06-20
8.0 add module partner_external_maps (module google_map in better !)

alexis-via:
See the module description to get more information about this module.


youring:
:thumbsup: Suggestion: would be better to fetch the address_format set in Sales--Configuration--Address Book--Localization--Countries when doing _address_as_string method. Since country, state, city, address sequences may vary on diffrent countries.

partner_external_maps/README.rst
```diff
@@ -0,0 +1,54 @@
+Partner External Maps
+=====================
+
+In the old days of Odoo/OpenERP, back in version 6.1, there was an official *google_map* module ; this module added a *Map* button on the partner form view and, when the user clicked on that button, it would open a new tab on its web browser and go to Google Map with a search on the address of the partner.
+
+This module aims at restoring this feature with several improvements:
+* each user can select the map website he wants to use in its preferences
```
eLBati, line 8: Please add newline around '\* '

partner_external_maps/partner_external_maps.py
```diff
@@ -0,0 +1,197 @@
...
+            ('route_address_url', '!=', False),
+            ('route_lat_lon_url', '!=', False)],
+        help="Map provided used when you click on the car icon on the partner "
+        "form to display an itinerary.")
+    context_route_start_partner_id = fields.Many2one(
+        'res.partner', string='Start Address for Route Map')
+
+    @api.model
+    def _default_map_settings(self):
+        '''Method called from post-install script
```
pedrobaeza: For docstring, use triple double quota instead of single one

pedrobaeza: Although you need the post-init script for the existing users, it still would be useful to provide a default value for new created users since the installation of the module.

partner_external_maps/partner_external_maps.py
```diff
@@ -0,0 +1,197 @@
...
+                })
+
+
+class ResPartner(models.Model):
+    _inherit = 'res.partner'
+
+    @api.model
+    def _address_as_string(self):
+        addr = u''
+        if self.street:
```
pedrobaeza, line 116: Isn't better to add all existing fields to a list and build the string with the `' '.join()` expression?

partner_external_maps/partner_view.xml
```diff
@@ -0,0 +1,31 @@
...
+<data>
+
+<!--  Inherit partner view to add 'Map' button -->
+<record id="view_partner_form" model="ir.ui.view">
+    <field name="name">map.button.res.partner.form</field>
+    <field name="model">res.partner</field>
+    <field name="inherit_id" ref="base.view_partner_form"/>
+    <field name="arch" type="xml">
+        <xpath expr="//div[@name='buttons']" position="inside">
+            <button class="oe_inline oe_stat_button"
```
pedrobaeza, line 18: Isn't better to provide these buttons next to the street information or at least inside the same tab instead of being mixed with the bunch of buttons that is already present on the partner in the smart-buttons area?

alexis-via:
@pedrobaeza I took into account your remarks
About the buttons, I think it's a good choice to group them all at the same place ; that's how the form view of the partner form has been designed by Odoo S.A. Of course, we can all have different opinions about this...

pedrobaeza:
Thanks :+1:

cdubuit:
Merci :+1:

pedrobaeza:
Trying on runbot, I'm getting `Error: Unknown field street in domain [["street","=",false],["city","=",false]]` accessing the form view. This should happend when the field is not included in the view. Do you know the case where street field is not included in the view. Is there any module that replaces it?

About @youring suggestion, what do you think?

alexis-via:
@pedrobaeza it's the module "partner_street_number" that removes the street field from the view. It's not a problem to only keep the city = False in the domain, so the module is now compatible with "partner_street_number"

alexis-via:
@pedrobaeza after my last commit, we still have an error msg in runbot: "Error: NameError: name 'street' is not defined". But the issue is not in my module.

I think it is a very bad practise to remove fields from view as they did it in partner_street_number, because it breaks all modules that use the street field in domain or in view inheritance. So I make a PR to fix it:
https://github.com/OCA/partner-contact/pull/135

pedrobaeza:
#135 is merged, so you can rebase, Alexis.

alexis-via:
I tested runbot after my rebase, and it works fine: no more error msg when you access the form view of a partner !

I think we can merge now.

pedrobaeza:
Indeed it's working! Thanks

:+1:

pedrobaeza:
Following the singular conventions, I'm going to rename the module to _partner_external_map_

pedrobaeza:
Well, thinking twice, we can stay as is as the meaning is not kept. On v9 we can find another name.

PR #135, opened 2015-06-28
8.0 FIX partner_street_number: do not remove street field as it breaks all modules that use it in views

alexis-via:
Make the street field invisible instead


pedrobaeza:
Thanks for the patch. I'll fast track when Travis and runbot finish

guewen:
:+1:

StefanRijnhart:
Too late to approve, but thanks @alexis-via!

PR #138, opened 2015-07-02
Update "display name" fields when updating first / last names.

houzefa-abba:
Hey,

The "partner_firstname" addon keeps the "name" field of "res.partner" records synchronized with first / last names; but it does not do so for the "display_name" field.

Consequently, places that make use of the "display_name" field show wrong values (eg partner list / kanban views).


antespi:
@houzefa-abba this issue has been fixed in https://github.com/OCA/partner-contact/pull/104

houzefa-abba:
Hey,

Indeed this issue is gone after #104. =)

PR #139, opened 2015-07-06
[8.0] partner_firstname: Recursive @onchange misbehavior

yajo:
@pedrobaeza This is the bug that I discovered from https://github.com/OCA/partner-contact/pull/104.

The problem is that I don't know how to fix it, but at least I'm sending the tests to get some help.

Current failure:
- User creates a new partner.
- User unchecks `is_company`.
- User sets `firstname` to "John".
  - This triggers `_compute_name()`.
    - This updates `name`.
      - This triggers `_onchange_name()`.
        - This inverts name.
          - This sets `lastname` to "John" and `firstname` to False.

The result is that the user writes in `firstname`, and that ends up in `lastname` after triggering all the onchanges.

**1st idea** tried: Remove `_onchange_name()` function, and let `_inverse_name_after_cleaning_whitespace()` be called only on save, but then:
- User creates a new partner.
- User checks `is_company`.
- User writes in `name`.
- User saves.
  - Odoo sends the correct `name` value, but also sends `lastname = firstname = False` because hidden fields are sent always too.
    - This triggers `_compute_name()`.
      - This sets `name = False`
        - This makes `_check_name()` fail.

**2nd idea** was to disable `_onchange_name()` when called from `_compute_name()`, so the code was:

```
@api.one
@api.depends("firstname", "lastname")
def _compute_name(self):
    """Write the 'name' field according to splitted data."""
    self.with_context(skip_invert=True).name = u" ".join(
        filter(None, (self.lastname, self.firstname)))

[...]

@api.one
@api.onchange("name")
def _onchange_name(self):
    """Ensure :attr:`~.name` is inverted in the UI."""
    if not self.env.context.get("skip_invert"):
        self._inverse_name_after_cleaning_whitespace()
```

But seems like the _onchange magic_ does not work when switching the context. `_compute_name()` never changes anything. For some reason if you don't set new values directly to `self`, no record is updated.

In fact, that does not even let you install the module, since `_compute_name()` is called from `_install_partner_firstname()`.

**3rd idea** was same as 2nd but returning the modified record, but I ended up with same result as for the 2nd:

```
@api.one
@api.depends("firstname", "lastname")
def _compute_name(self):
    """Write the 'name' field according to splitted data."""
    other = self.with_context(skip_invert=True)
    other.name = u" ".join(
        filter(None, (self.lastname, self.firstname)))
    return other
```

So **4th idea** was to leave `_compute_name()` uncahnged and add `context="{'skip_invert': True}"` to firstname and lastname `<field/>` in views, but [seems like the `context` attribute is only used in relational fields](https://www.odoo.com/documentation/8.0/reference/views.html#semantic-components). That context was ignored.

So, right now I'm on a dead end. I'd appreciate any suggestions.

### DIFF
--- a/partner_firstname/tests/test_name.py
+++ b/partner_firstname/tests/test_name.py
@@ -27,8 +27,7 @@
 
 """Test naming logic.
 
-Please have in mind that some tests will fail if the ``mail`` module is
-installed.
+To have more accurate results, remove the ``mail`` module before testing.
 """
 
 from .base import BaseCase
@@ -60,6 +59,7 @@ class PartnerCompanyCase(BaseCase):
         self.original.is_company = True
 
     def test_copy(self):
+        """Copy the partner and compare the result."""
         super(PartnerCompanyCase, self).test_copy()
         self.expect(self.name, False, self.name)
 
@@ -72,6 +72,19 @@ class PartnerCompanyCase(BaseCase):
 
 class UserCase(PartnerContactCase):
     def create_original(self):
-        self.original = self.env["res.users"].create({
-            "name": u"%s %s" % (self.lastname, self.firstname),
-            "login": "firstnametest@example.com"})
+        name = u"%s %s" % (self.lastname, self.firstname)
+
+        # Cannot create users if ``mail`` is installed
+        if self.mail_installed():
+            self.original = self.env.ref("base.user_demo")
+            self.original.name = name
+        else:
+            self.original = self.env["res.users"].create({
+                "name": name,
+                "login": "firstnametest@example.com"})
+
+    def test_copy(self):
+        """Copy the partner and compare the result."""
+        # Skip if ``mail`` is installed
+        if not self.mail_installed():
+            super(UserCase, self).test_copy()
--- a/partner_firstname/tests/test_name.py
+++ b/partner_firstname/tests/test_name.py
@@ -25,6 +25,12 @@
 # along with this program; if not, write to the Free Software
 # Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA  02111-1307, USA.
 
+"""Test naming logic.
+
+Please have in mind that some tests will fail if the ``mail`` module is
+installed.
+"""
+
 from .base import BaseCase
 
 
--- a/partner_firstname/tests/test_onchange.py
+++ b/partner_firstname/tests/test_onchange.py
@@ -1,23 +1,21 @@
 # -*- coding: utf-8 -*-
 """These tests try to mimic the behavior of the UI form.
 
-The form operates in onchange mode, and has only some limitations that must be
-met.
+The form operates in onchange mode, with its limitations.
 """
 
-from openerp.tests.common import TransactionCase
+from .base import OnChangeCase
 
 
-class PartnerCompanyCase(TransactionCase):
+class PartnerCompanyCase(OnChangeCase):
+    is_company = True
+
     def test_create_from_form(self):
         """A user creates a company from the form."""
         name = u"Sôme company"
         with self.env.do_in_onchange():
             # User presses ``new``
-            partner = self.env["res.partner"].create({})
-
-            # User ensures it is a company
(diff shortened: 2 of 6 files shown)


andreparames:
Why is `_compute_name()` setting the `name` even when `first_name` and `last_name` are both `False`? Seems like the first version could work if an `if` to detect this case was added, no?

yajo:
I already tried that too, but when you enter a compute method, the computed field is set to False before entering. You must give it a value.

And BTW, if firstname & lastname are both False, then name must be False too.

yajo:
I finally found the solution! Thanks @andreparames, you gave me the idea!

BTW I found a UI problem when sometimes lastname was set to `u""` instead of `False`. It was not being required when it should. Tested & fixed.

pedrobaeza:
There are still errors on tests

partner_firstname/tests/test_onchange.py
```diff
@@ -0,0 +1,116 @@
...
+from openerp.tests.common import TransactionCase
+
+
+class PartnerCompanyCase(TransactionCase):
+    def test_create_from_form(self):
+        """A user creates a company from the form."""
+        name = u"Sôme company"
+        with self.env.do_in_onchange():
+            # User presses ``new``
+            partner = self.env["res.partner"].create({})
```
yajo: This (and others equal) is the failing line, but I cannot reproduce the error in my tests, with an empty database. When a user presses "New", an empty record is created in an onchange environment, and that's exactly what I do here. Sholud I call a different method?

yajo:
Tests fixed, thanks!

antespi:
Hi @Yajo, I've just propuse another way (https://github.com/OCA/partner-contact/issues/143) to fix this bug, please check it out and express your opinion

antespi:
Hi @Yajo, I've propuse a new test suite here https://github.com/grupoesoc/partner-contact/pull/2, please review it and merge if you're agreed with it

antespi:
:+1: All current tests pass and it's working in production environment

@Yajo and me are agree to go step by step and wait for merge this PR before refactoring test suite. By now, this PR fix onchange bug, when merged I'll propuse a new one with my improvements and we can discuss on it without affecting this one.
@gurneyalex,  @pedrobaeza, thanks for you reviewing and merging if all it's correct as soon as possible

yajo:
Anything else for merging?

partner_firstname/tests/test_name.py
```diff
@@ -25,6 +25,12 @@
 # along with this program; if not, write to the Free Software
 # Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA  02111-1307, USA.
 
+"""Test naming logic.
+
+Please have in mind that some tests will fail if the ``mail`` module is
```
pedrobaeza: Can you check this case from the test and avoid the failure?

yajo: I don't know how.

pedrobaeza: Querying ir_module_module to see if module with name='mail' has state='installed'

yajo: And what should I do if it is? Don't run the tests?

pedrobaeza: Yeah, I think so. It's not the best option, but if you don't know how to run the test with the other module installed, it can serve

yajo: OK let me try.

yajo:
New tests will not fail with `mail` installed. Sometimes it means to skip those tests, but whatever.

yajo:
All seems fixed. Merge?

(BTW, linking this to https://github.com/OCA/l10n-spain/issues/174).

gurneyalex:
:+1:

rafaelbn:
This module has a bug. After installing this module you cannot delete any res.partner.
Please add this test to module (deleting a partner).
Ref: https://github.com/OCA/partner-contact/issues/154

PR #140, opened 2015-07-06
[IMP][8.0] partner_firstname order selectable

antespi:
This PR improve partner_firstname allowing admin to select how names are ordered to compute `name` and `display_name` fields

There are two possibilities:
- Firstname Lastname
- Lastname Firstname (default, for backward compatibility)

This setting is company dependant

### DIFF
--- a/partner_firstname/models.py
+++ b/partner_firstname/models.py
@@ -91,8 +91,8 @@ class ResPartner(models.Model):
                     first = parts[0]
                     last = u" ".join(parts[1:])
                 else:
-                    first = parts[-1]
-                    last = u" ".join(parts[:-1])
+                    last = parts[0]
+                    first = u" ".join(parts[1:])
                 parts = [last, first]
             else:
                 while len(parts) < 2:
--- a/partner_firstname/models.py
+++ b/partner_firstname/models.py
@@ -46,7 +46,7 @@ class ResPartner(models.Model):
         names = (self.lastname, self.firstname)
         if self.company_id.names_order == 'first_last':
             names = (self.firstname, self.lastname)
-        self.name = u" ".join((p for p in names if p))
+        self.name = u" ".join(filter(None, names))
 
     @api.one
     def _inverse_name_after_cleaning_whitespace(self):
@@ -89,10 +89,10 @@ class ResPartner(models.Model):
             if len(parts) > 1:
                 if self.company_id.names_order == 'first_last':
                     first = parts[:1][0]
-                    last = ' '.join(parts[1:])
+                    last = u" ".join(parts[1:])
                 else:
                     first = parts[-1:][0]
-                    last = ' '.join(parts[:-1])
+                    last = u" ".join(parts[:-1])
                 parts = [last, first]
             else:
                 while len(parts) < 2:
--- a/partner_firstname/models.py
+++ b/partner_firstname/models.py
@@ -4,6 +4,7 @@
 #    Copyright (C)
 #       2014:       Agile Business Group (<http://www.agilebg.com>)
 #       2015:       Grupo ESOC <www.grupoesoc.es>
+#       2015:       Antonio Espinosa <antonioea@antiun.com>
 #
 #    This program is free software: you can redistribute it and/or modify
 #    it under the terms of the GNU Affero General Public License as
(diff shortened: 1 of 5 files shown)


antespi:
@Yajo, clone from https://github.com/grupoesoc/partner-contact/pull/1

partner_firstname/models.py
```diff
@@ -80,13 +83,20 @@ def _inverse_name(self):
...
+                    first = parts[0]
+                    last = u" ".join(parts[1:])
+                else:
+                    first = parts[-1]
+                    last = u" ".join(parts[:-1])
+                parts = [last, first]
+            else:
+                while len(parts) < 2:
+                    parts.append(False)
         self.lastname, self.firstname = parts
```
yajo, line 100: You should use `self.update()` here, since there's a possibility that constraint fails (if firstname="something" and lastname=False).

yajo:
And don't forget the tests :wink:

partner_firstname/models.py
antespi, line 100: @Yajo I'm not following you, what is self.update()?

antespi:
@Yajo You're right, I have to finish some urgent tasks and then I'll read testing document and implement more test.

partner_firstname/models.py
yajo, line 100: It's an undocumented feature. It's the same as [`write()`](https://www.odoo.com/documentation/8.0/reference/orm.html#openerp.models.Model.write), but works in temporary enviroments (onchange). Read https://github.com/odoo/odoo/issues/7417 and you'll understand.

antespi, line 100: @Yajo, So, this is what do you propuse?

Replace this

```
self.lastname, self.firstname = parts
```

with this?

```
self.update({'lastname': parts[0], 'firstname': parts[1]})
```

antespi, line 100: @pedrobaeza could you help us in this point?
AFAIK, inside in onchange method if self.xxx is assigned, Odoo updates only cache, do not perform any write operation to database, just in UI. Write operations will do when user press `Save` button, isn't it?
I didn't know `update()` models feature

antespi, line 100: @Yajo, @pedrobaeza Actually, reading the code from models.py:

```
    def update(self, values):
        """ Update record `self[0]` with `values`. """
        for name, value in values.iteritems():
            self[name] = value
```

This is the same that assing self.xxx, isn't it?

antespi:
@Yajo, here you have tests, and bugfixes to pass them ;)
I'm not an Odoo Test virgin anymore

partner_firstname/models.py
yajo, line 100: AFAIK update does the same, but all in one call, so all fields get their values before the constraints check. But indeed the code you show makes me wonder if it's true. Anyway with the way you have it, it will fail for sure, so it's worth trying IMHO.

Just make the test and see if it fails with each solution.

yajo:
@antespi See how it was not that hard? Seems fine, but I think this should be merged after https://github.com/OCA/partner-contact/pull/139, since the tests there will probably make you see some additional bugs.

antespi:
@Yajo, you're rigth, waiting to merge https://github.com/OCA/partner-contact/pull/139 first, then i can rebase and adapt again this new feature.
I'll be watching ;)

partner_firstname/models.py
yajo, line 100: I just wanted to point out that after testing in https://github.com/OCA/partner-contact/pull/141/, `update()` fails too.

See the workaround I did in that PR: https://github.com/OCA/partner-contact/pull/153/files#diff-707f7a50bf3ebd6c95af51564f66788cR64

antespi:
I close this PR in order to add this feature after https://github.com/OCA/partner-contact/issues/159 were merged.

rafaelbn:
@antespi ,

```
I close this PR in order to add this feature after #159 were merged.
```

yajo:
If you reopen this, keep in mind:
- Rebase/merge.
- There is now `partner_second_lastname`, and its algorithm depends on `partner_firstname`'s.

Also, naming order should be by user preference IMHO, not by company preference. Otherwise setting names for customers for different companies with different settings would confuse the user.

yajo:
And companies can have users from different countries that are comfortable with different order for last names.

rafaelbn:
Please @antespi rebase, we must continue this work. Take care on last comments of @Yajo

rafaelbn:
Closing as Discussed here https://github.com/OCA/partner-contact/issues/210

PR #142, opened 2015-07-09
Remove migration script

ghost:
This script will not work because when loading the module res.better.zip
other models related to it with m2o or m2m might not be already loaded to the database.
The script can still be found on Openupgrade but should be used
in a seperate module once all modules are loaded.


pedrobaeza:
:+1:

gurneyalex:
:+1:

PR #149, opened 2015-07-31
Add module partner_contact_affected_by_terrorism.

yajo:
Aiming to fix https://github.com/OCA/l10n-spain/issues/174, this module lets the user indicate if any contact has been affected by terrorism.


yvaucher:
This might be a bit specific no?

In which context of company or organisation do you need it?

yajo:
@yvaucher Please read https://github.com/OCA/l10n-spain/issues/174. It's a common requirement for some Spanish legal things, but I'm posting it here because it could be helpful for other people too.

yvaucher:
@Yajo Couldn't it be done with partner categories ?

If not you could add more information in README.rst

No idea if other countries needs it.

yajo:
It's hard to rely hard-dependency modules in just categories. We thought it at first, but partners would end with too much categories, and it's not easy to use for end users.

In Spain, this field is needed to acquire an European subsidy for training. That leads me to think that the same could be needed in other countries in Europe.

rafaelbn:
This field is very specific. In which cases (different of European subsidy) could apply?
In my opinion this field must be included in the "Training module" unless we find indications of general use in the worldwide.

yajo:
Closing. See https://github.com/OCA/l10n-spain/issues/174#issuecomment-127557716.

PR #155, opened 2015-08-12
Partner activity base

vrenaville:
This is the first part of partner activity:
it refer to the discussion on #144 , I will propose additional module for each 
module part.
this one just print partner informations and contacts

PR #144, opened 2015-07-21
Partner activity report

vrenaville:
#  This module give a report with recent activity on the partner
- Display adresses and contact of the partner
- Display crm open for this partner
- Display invoices related to this partner
- Display sale order related to this customer
- Display delivery lines related to this customer

PR #156, opened 2015-08-12
[ADD] activity on crm for partner

vrenaville:
It's add activity on CRM for partner, depend on PR #155

### DIFF
--- a/partner_activity_report_crm/README.rst
+++ b/partner_activity_report_crm/README.rst
@@ -1,0 +1,45 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+    :alt: License: AGPL-3
+
+Partner Activity Details for CRM
+================================
+
+This module give a report with recent activity in CRM for this partner
+----------------------------------------------------------------------
+
+* Display crm open for this partner
+
+
+Bug Tracker
+===========
+
+Bugs are tracked on `GitHub Issues <https://github.com/OCA/partner-contact/issues>`_.
+In case of trouble, please check there if your issue has already been reported.
+If you spotted it first, help us smashing it by providing a detailed and welcomed feedback
+`here <https://github.com/OCA/partner-contact/issues/new?body=module:%20partner_activity_report%0Aversion:%208.0%0A%0A**Steps%20to%20reproduce**%0A-%20...%0A%0A**Current%20behavior**%0A%0A**Expected%20behavior**>`_.
+
+
+Credits
+=======
+
+Contributors
+------------
+
+* Vincent Renaville <vincent.renaville@camptocamp.com>
+* Damien Crier <damien.crier@camptocamp.com>
+
+Maintainer
+----------
+
+.. image:: https://odoo-community.org/logo.png
+   :alt: Odoo Community Association
+   :target: https://odoo-community.org
+
+This module is maintained by the OCA.
+
+OCA, or the Odoo Community Association, is a nonprofit organization whose
+mission is to support the collaborative development of Odoo features and
+promote its widespread use.
+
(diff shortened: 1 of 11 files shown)


partner_activity_report_crm/README.rst
```diff
@@ -0,0 +1,45 @@
...
+* Display crm open for this partner
+
+
+Bug Tracker
+===========
+
+Bugs are tracked on `GitHub Issues <https://github.com/OCA/partner-contact/issues>`_.
+In case of trouble, please check there if your issue has already been reported.
+If you spotted it first, help us smashing it by providing a detailed and welcomed feedback
+`here <https://github.com/OCA/partner-contact/issues/new?body=module:%20partner_activity_report%0Aversion:%208.0%0A%0A**Steps%20to%20reproduce**%0A-%20...%0A%0A**Current%20behavior**%0A%0A**Expected%20behavior**>`_.
```
damdam-s: partner_activity_report_crm

pedrobaeza:
Maybe this one should be on OCA/crm repo?

rafaelbn:
Which is the criteria in OCA to decide this? This module is a report of a Partner/contact but it has dependencies with CRM.

pedrobaeza:
The criteria is the very subjected one of "The best repo that fits to the purpose of the module". As some repos can overlap their boundaries, sometimes it's not very clear where to put. For example: _website_event_photo_, a module that you know. It's a module about events, but it corresponds to the website part, so it can be hosted on event or website projects. In this case, as we have a specialized repo for events, I would bet that a better hosting is event project, because the website project is also very crowded with other website related things. In this case, I would say the contrary, because in this case, this repo is the one very crowded, and if you think it, almost everything is partner-related, so I would bet for moving to the crm part, that isolate a bit that part of the ecosystem.

gurneyalex:
@pedrobaeza I'd rather have all the modules linked to #155 live in the same repository as it makes maintenance of the code much easier.

pedrobaeza:
Well, I still prefer the other option, but I won't block a PR here, because as I said, it's a matter of taste.

rafaelbn:
OK, this means that all modules related with https://github.com/OCA/partner-contact/pull/144 purpose will be here in this report (all together)

And all will have name: partner_activity_report_\* (crm: lead, oportinities, calls; sales: sales orders; ...)

¿OK @gurneyalex @pedrobaeza ?

cc @vrenaville

pedrobaeza:
I must insist with OCA/crm question, because I found a case where people navigating through OCA repo expect to find all modules related to CRM on the link that he/she has on the website about this area, that leads to https://github.com/OCA/crm. Having there the _partner_activity_crm_ module, he/she can read the description and decides that the module is of them interest. Installation section then tell them to install the dependency with partner-contact repo.

What do you think about this reasoning?

rafaelbn:
@gurneyalex how much complicated makes maintenance having modules in different projects?

I agree with @pedrobaeza. It will be more intuitive to find.

Moreover, we planned to make a module for reporting partners about the events the have participate, I thought event's project would be the best project.

"The best repo that fits to the purpose of the module": Is the purpose of this module to inform Partners about their CRM activity? This means OCA/crm fits better.

gurneyalex:
let's move then.

gurneyalex:
@vrenaville can you move this module to OCA/crm?

rafaelbn:
Hi @vrenaville do you want to continue with this PR? Could you please move it to OCA/crm. Let me know please

rafaelbn:
WIP as depends on https://github.com/OCA/partner-contact/pull/155

NL66278:
Closed as related PR closed and no activity for a long time.

PR #157, opened 2015-08-13
Add module base_partner_helper from original bzr repo

Viggor:
original authors: @sebastienbeau @bealdav 
base_partner_help: Module for additional methods for manage partner. For now, there is a method to deal with addresses (get_split_address).


Viggor:
Already merged under a different name.

PR #159, opened 2015-08-17
[REF][partner_firstname] Method refactoring to use algorithms in other model

adrienpeiffer:
This refactoring will allow to use this intelligence in hr_employee_firstname

https://github.com/OCA/hr/pull/128

### DIFF
--- a/partner_firstname/models.py
+++ b/partner_firstname/models.py
@@ -40,7 +40,9 @@ class ResPartner(models.Model):
 
     @api.model
     def _get_computed_name(self, lastname, firstname):
-        """Write the 'name' field according to splitted data."""
+        """Compute the 'name' field according to splitted data.
+        You can override this method to change the order of lastname and
+        firstname the computed name"""
         return u" ".join((p for p in (lastname, firstname) if p))
 
     @api.one
@@ -77,6 +79,8 @@ class ResPartner(models.Model):
         - Otherwise, make a guess.
 
         This method can be easily overriden by other submodules.
+        You can also override this method to change the order of name's
+        attributes
 
         When this method is called, :attr:`~.name` already has unified and
         trimmed whitespace.
--- a/partner_firstname/models.py
+++ b/partner_firstname/models.py
@@ -38,12 +38,16 @@ class ResPartner(models.Model):
         required=False,
         store=True)
 
+    @api.model
+    def _get_computed_name(self, lastname, firstname):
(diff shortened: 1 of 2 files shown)


partner_firstname/models.py
```diff
@@ -38,12 +38,16 @@ class ResPartner(models.Model):
         required=False,
         store=True)
 
+    @api.model
+    def _get_computed_name(self, lastname, firstname):
+        """Write the 'name' field according to splitted data."""
+        return u" ".join((p for p in (lastname, firstname) if p))
```
pedrobaeza, line 46: Didn't we talk about putting a comma in the middle to correctly write the form "Lastname, Firstname"?

lmignon:
:+1: (Code review) No opinion on comment of @pedrobaeza but I can imagine that some customers prefer the comma in the middle and others not.

partner_firstname/models.py
sbidoul, line 46: @pedrobaeza I think we can't change the behaviour of this module in this respect. 

`_get_computed_name` and `_get_inverse_name` are intented to be overridden to implement other schemes such as "name, firstname".

@adrienpeiffer I suggest you add a word in the README and the docstrings of `_get_computed_name` and `_get_inverse_name` to explain this.

pedrobaeza, line 46: I know that they can be overriden, but we are talking about the default behaviour, that all agree that now is not the best choice and even annoying. A change in these methods with a migration script will handle it.

sbidoul, line 46: I think we need more feedback before changing the default behaviour. Now this PR is only here to improve https://github.com/OCA/hr/pull/128. 

Would you agree to leave this PR as is and have another PR or issue to change the default behaviour and gather feedback on that specific point.

pedrobaeza, line 46: OK, I agree

pedrobaeza:
:+1:

sbidoul:
:+1: when the doc is updated to explain which methods to override to customize the name+firstname concatenation.

adrienpeiffer:
@sbidoul documentation is updated. Thanks !

antespi:
:+1: I think your solution is more versatile and allows a cleaner inherit for fitting customer needs. I'll re-adapt this PR: https://github.com/OCA/partner-contact/pull/140 after merged in order to allow admin to select most common name shemes directly in company form. Thanks

Issue #160, opened 2015-08-17
Bad compare onchange_type_selection_id


ghost:
On line 169 of res_partner_relation_all.py
is_company = self.type_selection_id.partner_category_other == 'c'

It gives apples/oranges python error...

Not sure what this is supposed to do, it is getting a category but the 'c', 'p' portions of this module are for the is_company flag.  I commented and removed the filter and it now works, but it would be nice of setting the relationship type filtered the right partner dropdown.

rafaelbn:
@idncom you must open issue with the bugtracker template:

> module: l10n_es_pos_simplified_invoice
> version: 8.0
> 
> **Steps to reproduce**
> - ...
> 
> **Current behavior**
> 
> **Expected behavior**

Also take care on tittle

ghost:
Thank you, I actually forked, corrected and submitted a pull request.  We
just submitted our agreement so we can contribute as we find issues.

On Wed, Aug 19, 2015 at 3:44 AM, Rafael Blasco notifications@github.com
wrote:

> @idncom https://github.com/idncom you must open issue with the
> bugtracker template:
> 
> module: l10n_es_pos_simplified_invoice
> version: 8.0
> 
> _Steps to reproduce_
> - ...
> 
> _Current behavior_
> 
> _Expected behavior_
> 
> Also take care on tittle
> 
> —
> Reply to this email directly or view it on GitHub
> https://github.com/OCA/partner-contact/issues/160#issuecomment-132495411
> .

## 

Best Regards,
Mark Meadows
www.apptuate.com
636.678.9924

PR #161, opened 2015-08-18
[FIX] Other partner on change filter

ghost:
The code was incorrect for is_company filter, I also added a filter for category.


pedrobaeza:
Fixes #160

hbrunn:
sorry @idncom, I just proposed #164 without checking if someone else fixed the problem already. It does the same as your PR, but also filters the possible relations by the preselected partner's contact type and labels, so it's more comprehensive.

Thanks for your efforts in any case!

ping @pedrobaeza

PR #164, opened 2015-08-20
[FIX] respect restrictions on possible relation

### LINKED ISSUE #160: Bad compare onchange_type_selection_id

hbrunn:
types and other partner
fixes #160


pedrobaeza:
Thanks for the patch :+1:

yvaucher:
:+1:

PR #165, opened 2015-08-20
[FIX] _get_inverse_name in case of space in name

bealdav:
## issue

in case of installation of the module with existing data.

if name == `' my_name'` then firstname is `u' '` and lastname is `'my_name'`

but value of self.lastname is lost in def _check_name() and EmptyNamesError is raised
## workaround

This PR
Avoid side effects to firstname and lastname in _check_name.

EDIT(@yvaucher) add backquote for readability

### DIFF
--- a/partner_firstname/models.py
+++ b/partner_firstname/models.py
@@ -90,7 +90,7 @@ class ResPartner(models.Model):
             parts = [name or False, False]
         # Guess name splitting
         else:
-            parts = name.split(" ", 1)
+            parts = name.strip().split(" ", 1)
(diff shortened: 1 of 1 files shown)


pedrobaeza:
:+1:, but please make a test for this use case

antespi:
@bealdav, @pedrobaeza , I'm refactoring test suite in order to test this case and other ones in a more comprensible way. So, could you give some time to finish them and propuse in a PR? Thanks.

yvaucher:
:+1:

PR #174, opened 2015-09-21
[8.0] base_contact: Show warning without install it

moylop260:
I have next warning in my instance without install the module: `base_contact`

``` bash
2015-09-21 18:52:38,338 174 WARNING openerp_template openerp.addons.base_contact.deprecated: 
This module is DEPRECATED. See /root/partner-contact/base_contact/README.rst.
```

IMHO a warning message of a module that is not installed is a bad idea.

### DIFF
--- a/base_contact/__init__.py
+++ b/base_contact/__init__.py
@@ -19,14 +19,3 @@
 #    along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #
 ##############################################################################
-
-import logging
-import os
-
-
-_logger = logging.getLogger(__name__ + ".deprecated")
-
-# Skip warnings on runbots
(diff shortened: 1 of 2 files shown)


moylop260:
Why don't use a `"installable": False` in manifest file?

charbeljc:
Hi @moylop260, thanks for your hard qa work ! :+1:

moylop260:
You are welcome

yvaucher:
@moylop260 :-1:  I don't think this will work. Here you simply want to remove the icon to get ride of the message. We do want this message on instances having it installed.

charbeljc:
Hum, you are right @yvaucher, @moylop260 could you revert this part of your PR ? thx !

yajo:
Please see #250

PR #177, opened 2015-10-01
[8.0][partner_contact_vat] Module to have and check VAT field for contacts

yajo:
Related to https://github.com/OCA/l10n-spain/issues/174.

This module allows to have a VAT for each contact, separate from the company's.

It does not interfere with any other logic with VAT


partner_contact_vat/README.rst
```diff
@@ -0,0 +1,67 @@
...
+
+Bug Tracker
+===========
+
+Bugs are tracked on `GitHub Issues <https://github.com/OCA/
+partner-contact/issues>`_. In case of trouble, please check there if your issue
+has already been reported. If you spotted it first, help us smashing it by
+providing a detailed and welcomed feedback `here <https://github.com/OCA/
+partner-contact/issues/new?body=module:%20
+partner_contact_vat%0Aversion:%208.0.3.0.0%0A%0A**Steps%20to%20reproduce**%0A-%20...%0A%0A**Current%20behavior**%0A%0A**Expected%20behavior**>`_.
```
pedrobaeza: Put 8.0

partner_contact_vat/README.rst
```diff
@@ -0,0 +1,67 @@
...
+=======================
+VAT in contact partners
+=======================
+
+This module was written to extend the functionality of contacts management to
+support setting a VAT in contact partners (those partners that are not not
+companies) and allow you to have separate VAT codes for contact and for their
+company.
+
+The company's VAT will still be used for invoiced and other legal documents.
```
dreispt: IMHO it would be helpful to expand on why we would want this.

rafaelbn: @Yajo please expand information on why we would want this as @dreispt said. When we will want to have the contact's VAT of a customer? Thanks

yajo: Odoo by default only allows VAT in parent companies. When a contact belongs to a company and you go to _accounting_ page, it shows _Accounting is managed in the parent company_.

If you manage any kind of personal information of your clients (prevention of occupational hazards, training, event assistants, etc.) you probably need their VAT.

Indeed the company VAT will still be used for invoices and such, but for other things you need the person's. For that reason, instead of overwriting the VAT field, I just added a new one that will be hidden for companies.

dreispt: OK. You should add that text to the README.

yajo:
README updated.

rafaelbn:
Just a question, It's not need to validate it? Any number is ok?

yajo:
Validation is performed by `partner_contact_vat_check`, which is on hold until this is merged.

pedrobaeza:
Why 2 modules for this little task? The validation should be performed also in this PR.

yajo:
Odoo separates validation of VAT in `base_vat` module, so I followed their schema. What if you do not want validation?

pedrobaeza:
But that it's because base module contains a lot of things, so isolation is logical, but not in this case. You shouldn't allow to enter a vat that can be invalid, and later let that vat to get into the partner.

yajo:
We are arguing about future code that has not been reviewed, let's keep present comments to present code please. If in the future PR seems so useless to have 2 modules, I'll just merge it into this one, but that's a problem for the future us.

pedrobaeza:
No, it's not  future problem. It's a feature needed to validate this PR, so please include it.

yajo:
Done.

yajo:
Both modules have been simplified and merged. Now only 1 module to have VAT and check it.

yajo:
Runbot fails. Who can we ping? Seems like it lacks the OCA_RUNBOT env var.

It's also affected by a problem that is fixed in #233.

yajo:
Travis fails until https://github.com/OCA/partner-contact/pull/245 is merged.

rafaelbn:
Well I think we can review this again and get it merge. Rebuilding

yajo:
Closing in favor of partner_identification.

PR #180, opened 2015-10-07
[FIX][8.0] base_partner_merge: correct check for functional fields

alhashash:
Functional fields are not correctly skipped. As a result, any functional field of type reference will be overwritten with value `res.partner,XX` regardless its current value.

This is a disaster for functional fields that implement `fnct_inv`. For example, trying to merge partners will corrupt the target action for ALL menus!

### DIFF
--- a/base_partner_merge/base_partner_merge.py
+++ b/base_partner_merge/base_partner_merge.py
@@ -274,9 +274,8 @@ class MergePartnerAutomatic(orm.TransientModel):
             if record.model == 'ir.property':
                 continue
 
-            field_type = proxy_model._columns.get(record.name).__class__._type
-
-            if field_type == 'function':
+            column = proxy_model._columns[record.name]
+            if isinstance(column, fields.function):
                 continue
 
             for partner in src_partners:


pedrobaeza:
What about new fields with compute?

alhashash:
@pedrobaeza, fields with compute but no store will not have column in model._columns, so, they will be skipped before the check. 

It is still not perfect. If the computed field has `store`, The module call `write` on it. 

I think the module should always use direct SQL to update the data. Also, there are many other issues like not handling SQL unique constrains and many2many tables are handled via a dirty hack for two column tables.

Anyway, I may work on this module later but I wanted to push this fix now to protect other users from having a disaster like I had today.

pedrobaeza:
OK for me then :+1: 

I want to make for some time a generic base_merger with hooks for specific models, but it's in my wishlist.

alhashash:
It is on my wishlist too. I'll try to work on it in the near future.

guewen:
Could you see https://github.com/OCA/partner-contact/pull/183 which is a slightly different fix?

alhashash:
@guewen, Fix #183 will not work as fields defined using the old style `fields.function` will have `compute` value of `None`. For example, the field `action` in `ir.ui.menu` is functional field but will not be skipped in the update function

Fix #180 (this one) will not work too for computed new style Reference fields as the corresponding column type will be `fields.reference` not `fields.function`.

We can apply both checks but the module will still be broken anyway. Stored computed fields still need to be updated. We should only use direct SQL for search and update. The ORM has too many side effects to watch for.

alhashash:
I'm closing this PR  in favour of #183 which included this fix.

PR #185, opened 2015-10-14
[9.0][partner_relations] add unique constraints on `name` and `name_inverse`

charbeljc:


pedrobaeza:
You should change installable=False flag on manifest and try the module before proposing changes to it. If this is only for keeping both versions in sync, there's no need of it. It can be done via a rebase strategy. See https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-9.0

charbeljc:
@pedrobaeza thanks for the pointer on migration document. I close this PR.

PR #146, opened 2015-07-29
add unique constraints on `name` and `name_inverse`

charbeljc:

PR #186, opened 2015-10-14
[8.0][partner_relations] add unique constraints on `name` and `name_inverse`

charbeljc:


pedrobaeza:
I see correct to have a constraint on name, but not on inverse_name, because multiple relations can have the same inverse.

charbeljc:
Hum, right @pedrobaeza I did overlooked this, thx !

charbeljc:
Hi again @pedrobaeza, could you elaborate on your remark, as I can't thing of any concrete use case and I would start adding some unit tests. Thanks in advance.

Also, just saw that I did already submitted this patch some time ago here: #146. Apologies for this, I will close the old one.

pedrobaeza:
For example, relation son and grandson, inverse relation: ancestor, because you don't want to specify father and grandfather.

rafaelbn:
Hi @gurneyalex @pedrobaeza @antespi you give 3 :+1: to this PR https://github.com/OCA/partner-contact/pull/146 and this one is the continue of it so, do you think we can make here the last review and merge it?

@charbeljc do you want to go ahead with this PR?

Thanks to all

pedrobaeza:
Reviewing deeper, I found an issue that is not solved.

NL66278:
@pedrobaeza I was looking into this and I think it is a good idea to make name and inverse_name unique. Even stronger: name should also not appear as inverse_name and the other way around.

In your counterexample it should really be 'son or grandson' and inverse 'father or grandfather'. Unless you find a way to be your own father. Or it should be ancestor with inverse 'descendant'.

What do you think?

antespi:
:+1:, as I did in the other PR

rafaelbn:
@pedrobaeza which issue is not solved? https://github.com/OCA/partner-contact/pull/186#issuecomment-230072629

pedrobaeza:
I'm talking about https://github.com/OCA/partner-contact/pull/186#issuecomment-148455764

pedrobaeza:
Is this going to be continued?

NL66278:
@pedrobaeza It is not my PR, but I think this should not be continued. The problem is you might want to use the same labels for a relationship in different contexts, which therefore would not clash. For instance, say you have a 'board of directors' category for organisations and a 'director' category for persons. To establish the relation you define a relation on the organisation side: 'has member', on the person side 'is member of' . But the same labels 'has member'  and 'is member of' you might want to reuse for an ad hoc committee which has committee members. Because both types of relation require different categories, there is no actual name clash in practice.

pedrobaeza:
OK, closing then

PR #187, opened 2015-10-14
[9.0][PORT] port base location

yvaucher:


pedrobaeza:
Maybe you can take the occasion to make some tests for this module.

charbeljc:
@pedrobaeza I would tend to merge as is, and focus on adding coverage reports first, so :+1:

pedrobaeza:
The coverage reports are already there... via coveralls.io

charbeljc:
hum, can't see it on this page ?

pedrobaeza:
It can be due to the high-traffic that is happening because the 9.0 branch migration, but it was working fine.

yvaucher:
@pedrobaeza Here we go all onchanges tested

pedrobaeza:
Thank you very much! Waiting for Travis result, :+1:

yvaucher:
Travis is :green_heart:

pedrobaeza:
Perfect, I merge then

PR #191, opened 2015-10-29
Upgrade paersonal_information_page module.

richard-willdooit:


.gitignore
```diff
@@ -45,6 +45,11 @@ coverage.xml
 # Pycharm
```
elicoidal: Please remove this file

README.md
```diff
@@ -20,6 +20,7 @@ Available addons
 addon | version | summary
```
elicoidal, line 20: Please format the README file following https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md

richard-willdooit, line 20: I can't see anything in there about formatting README files?

Also, given that this is a repository wide file, not module specific, I am a bit loathe to making changes beyond what is relevant to what I have changed.

elicoidal, line 20: There is one README for the repo but you should have one for the module 
as well.

richard-willdooit, line 20: There is already one in the module, as well.

elicoidal, line 20: This is the one that should be normalized following OCA template

pedrobaeza, line 20: You don't even need to change this one, because it's automatically updated with an script.

Issue #154, opened 2015-08-11
[8.0] partner_relations - Error when deleting a res.partner


rafaelbn:
Tested in OCA's runbot
https://runbot.odoo-community.org/runbot/repo/github-com-oca-partner-contact-134
- Create a partner
- Don't select "is a company"
- Filled "firstname" and "lastname"
- Save partner
- Delete the partner we have created
- You get this error:

> AccessError
> No value found for res.partner(75,).birthdate_date

I have tested in Odoo's runbot and you can delete any res.partner if it hasn't any other relation with other objects.

Maybe this bug it not from this module partner_contact_birthdate

@pedrobaeza I think that any module in this project must at least pass test of creating, modifying and deleting a res.partner, if not shouldn't be merged.

cc @sysadminmatmoz @Yajo

pedrobaeza:
Tests are not mandatory for now in OCA to not discourage the contributions, and the review process usually detect these things.

I also think that this module is not the guilty, but something is happening.

yajo:
Is `base_contact` installed? If so, try going to modules list, searching for `base_contact` and pressing _Update_.

rafaelbn:
@Yajo base_contact is installed. After update still error.

I have uninstall lots of modules. Test & Try...

After uninstalling partner_firstname the problems disapears...

So ther is a bug in partner_firstname module https://github.com/OCA/partner-contact/pull/139

eLBati:
@rafaelbn just tested on
http://3112595-8-0-359f6c.runbot1.odoo-community.org/
using the **base** database, installed `partner_firstname`, `partner_contact_birthdate` and `base_contact`: the error does not happen.

Please try to reproduce it on a new database, pointing out the modules to install. Traceback would be useful too.

Thanks

yajo:
As a rule of thumb, I always press _Update_ in the `base` module after pulling new changes to any modules (core or OCA, any). That makes Odoo perform any updates needed and usually makes everything work smoothly. You could try that.

This bug seems to me more related to https://github.com/OCA/partner-contact/pull/120 than to #139. You can sniff all the code in `partner_firstname` and will find nothing related to birthdate.

And also you should try what @eLBati says.

rafaelbn:
Please @Yajo add a test for deleting res.partner to partner_firstname, this could help now and in the future. Eventhogh I'm going now to test runbot again.

@eLBati howto traceback?

rafaelbn:
@Yajo @eLBati 
[...]

PR #170, opened 2015-09-07
Test bug #154.

yajo:
I still do not understand the cause of this issue. Normal behavior is that the modified line raises a `MissingError` and it works fine. However installing `partner_relations` makes it to raise an `AccessError` which does not get handled and prevents unlinking.

This just checks the existence of the partner before computing it, and skips if false.

This adds also a test case. Merge it to fix #154.

PR #200, opened 2015-11-11
FIX when partner_contact_birthdate and partner_relations are installa…

eLBati:
…ed together, get

AccessError: ('AccessError', 'No value found for res.partner(43,).birthdate_date')
See https://github.com/OCA/partner-contact/issues/154

### DIFF
--- a/partner_relations/model/res_partner.py
+++ b/partner_relations/model/res_partner.py
@@ -298,12 +298,11 @@ class ResPartner(models.Model):
             cr, uid, args + date_args + active_args, offset=offset,
             limit=limit, order=order, context=context, count=count)
 
-    def read(
-            self, cr, uid, ids, fields=None, context=None,
-            load='_classic_read'):
-        return super(ResPartner, self).read(
-            cr, uid, ids, fields=fields,
-            context=self._update_context(context, ids), load=load)
+    @api.multi
+    def read(self, fields=None, load='_classic_read'):
+        self.env.context = self._update_context(
+            self.env.context, [r.id for r in self])
+        return super(ResPartner, self).read(fields=fields, load=load)
 
     def write(self, cr, uid, ids, vals, context=None):
         return super(ResPartner, self).write(
--- a/partner_relations/model/res_partner.py
+++ b/partner_relations/model/res_partner.py
@@ -78,7 +78,8 @@ class ResPartner(models.Model):
         if context is None:
             context = {}
         relation_obj = self.pool.get('res.partner.relation')
-        context2 = self._update_context(context, ids)
+        context2 = self.with_partner_relations_context(
+            cr, uid, ids, context=context).env.context
         for value in field_value:
             if value[0] == 0:
                 relation_obj.create(cr, uid, value[2], context=context2)
@@ -300,20 +301,18 @@ class ResPartner(models.Model):
 
     @api.multi
     def read(self, fields=None, load='_classic_read'):
-        self.env.context = self._update_context(
-            self.env.context, [r.id for r in self])
-        return super(ResPartner, self).read(fields=fields, load=load)
+        return super(ResPartner, self.with_partner_relations_context())\
+            .read(fields=fields, load=load)
 
-    def write(self, cr, uid, ids, vals, context=None):
-        return super(ResPartner, self).write(
-            cr, uid, ids, vals, context=self._update_context(context, ids))
+    @api.multi
+    def write(self, vals):
+        return super(ResPartner, self.with_partner_relations_context())\
+            .write(vals)
 
-    def _update_context(self, context, ids):
-        if context is None:
-            context = {}
-        ids = ids if isinstance(ids, list) else [ids] if ids else []
-        result = context.copy()
-        result.setdefault('active_id', ids[0] if ids else None)
-        result.setdefault('active_ids', ids)
-        result.setdefault('active_model', self._name)
-        return result
+    @api.multi
+    def with_partner_relations_context(self):
+        context = dict(self.env.context)
+        context.setdefault('active_id', self.ids[0] if self.ids else None)
+        context.setdefault('active_ids', self.ids)
+        context.setdefault('active_model', self._name)
+        return self.with_context(context)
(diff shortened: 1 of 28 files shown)


moylop260:
:+1: 
Could you fix?
`partner_relations/model/res_partner.py:303:80: E501 line too long (87 > 79 characters)`

If you want maybe could fix too:

``` bash
partner_relations/__openerp__.py:21: [C8102(manifest-required-key), ] Missing required key "license" in manifest file
partner_relations/model/res_partner_relation_type_selection.py:1: [C8201(no-utf8-coding-comment), ] No UTF-8 coding comment found: Use `# coding: utf-8` or `# -*- coding: utf-8 -*-`
partner_relations/tests/test_partner_relations.py:1: [C8201(no-utf8-coding-comment), ] No UTF-8 coding comment found: Use `# coding: utf-8` or `# -*- coding: utf-8 -*-`
```

Thank you!

rafaelbn:
Thank to you! @eLBati :-D

Tested functionally :+1: fix the problem

rafaelbn:
cc @Yajo @antespi

eLBati:
@moylop260 oops, thanks. Fixed.
How could I fix `Uses of a deprecated module 'openerp.osv'`?

Yajo:
replace by openerp.orm

yajo:
replace by openerp.models and openerp.fields

yajo:
Please could you merge into this PR this another: #170? Then run in unittest for `partner_relations,partner_firstname,partner_contact_birthdate` twice and confirm. Maybe that can be added to travis somehow, I don't know. Read https://github.com/OCA/partner-contact/pull/170#issuecomment-147684039 to know why.

BTW after all the headache, a test in this module (the guilty one) should be added IMHO to avoid further problems of this kind.

Thanks!

partner_relations/model/res_partner.py
```diff
@@ -298,12 +298,11 @@ def search(self, cr, uid, args, offset=0, limit=None, order=None,
...
-            self, cr, uid, ids, fields=None, context=None,
-            load='_classic_read'):
-        return super(ResPartner, self).read(
-            cr, uid, ids, fields=fields,
-            context=self._update_context(context, ids), load=load)
+    @api.multi
+    def read(self, fields=None, load='_classic_read'):
+        self.env.context = self._update_context(
+            self.env.context, [r.id for r in self])
+        return super(ResPartner, self).read(fields=fields, load=load)
```
hbrunn: don't assign to env.context - this messes up caching. Use
`return super(ResPartner, self.with_context(self._update_context(self.env.context, self.ids)).read(fields=fields, load=load)`

eLBati:
@Yajo I tried merging https://github.com/OCA/partner-contact/pull/170 into this, but it gives conflicts as it probably needs rebase

partner_relations/model/res_partner.py
eLBati: @hbrunn I tried it but did not solve

hbrunn: what happens then?

eLBati: Using `with_context` I get the same error:

```
File "/home/elbati/workspace/odoo/demo8/parts/odoo/openerp/fields.py", line 55, in get
    raise self.exception
AccessError: ('AccessError', 'No value found for res.partner(17,).birthdate_date')
```

hbrunn: how exactly does your code look then?

eLBati: The failing code:

```
    @api.multi
    def read(self, fields=None, load='_classic_read'):
        return super(ResPartner, self.with_context(self._update_context(
            self.env.context, self.ids))).read(fields=fields, load=load)
```

hbrunn: That's an indicator then that this is not the (only) thing which is wrong with the code. Chances are you just hide the problem. What you effectively do with your assignment is to change context in a global way. You should go hunting for the reason this goes wrong in the first place.

Put a breakpoint in https://github.com/OCA/OCB/blob/9.0/openerp/fields.py#L36 and look at the backtrace, that should give you clues

eLBati: I already did. I found that the difference between when `partner_relations` is installed and when it is not is, in first case, `FailedValue` object is instanciated with `AccessError` exception. While, in second case, it is instanciated with `MissingError` exception.
The correct behaviour is to have a `MissingError` exception, because it is later correctly intercepted by ORM. Instead, `AccessError` raises the error pointed out by https://github.com/OCA/partner-contact/issues/154
Thus I noticed it was something related to `read` method and tried rewriting it with new API.
I don't have additional information at the moment.

@hbrunn maybe you can go further with debugging?

yajo:
@eLBati Try now please.

eLBati:
@Yajo done, thanks

partner_relations/model/res_partner.py
eLBati: @hbrunn BTW what problem do you see with writing `self.env.context` directly?

hbrunn: @eLBati have a look at this gist: https://gist.github.com/hbrunn/174be7844b409666a281
There you will see how values in your context don't match values in the cache. This is asking for trouble.

yajo:
Then I'm closing that PR and branch for it's included here.

Did you run the tests twice as says https://github.com/OCA/partner-contact/pull/200#issuecomment-156037770? Do they pass?

I'll give my thumb up if those 2 answers are true, because runbot test works.

hbrunn:
@eLBati I made a PR https://github.com/eLBati/partner-contact/pull/1 that actually fixes the issue instead of masking the problem. Please merge it, then we can go on with this PR

eLBati:
Thanks @hbrunn , merged.

@Yajo I ran 

```
bin/start_odoo -d demo_8 --stop-after-init --log-level info  --test-enable --init partner_relations,partner_firstname,partner_contact_birthdate
```

twice without errors

yajo:
Then :+1: for me.

pedrobaeza:
:+1:

PR #211, opened 2015-12-03
[7.0] [FIX] Fix issue in partner_relations.

NL66278:
There was a problem with the code to determine wether the right hand partner in a
relation was the active partner.

### DIFF
--- a/partner_relations/__init__.py
+++ b/partner_relations/__init__.py
@@ -1,21 +1,4 @@
 # -*- coding: utf-8 -*-
-##############################################################################
-#
-#    OpenERP, Open Source Management Solution
-#    This module copyright (C) 2013 Therp BV (<http://therp.nl>).
-#
-#    This program is free software: you can redistribute it and/or modify
-#    it under the terms of the GNU Affero General Public License as
-#    published by the Free Software Foundation, either version 3 of the
-#    License, or (at your option) any later version.
-#
-#    This program is distributed in the hope that it will be useful,
-#    but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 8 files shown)


hbrunn:
@NL66278 it's quite hard to see what the relavant changes are because of all the formatting stuff you changed in the same commit. Please do that in different commits.

For the change itself: Could you explain why moving the same code out of the class fixes the issue?

hbrunn:
Isn't the problem rather how `get_values` is called: https://github.com/Therp/partner-contact/commit/e7a5657458a028bc44cb36a6e47eecc7437e185a#diff-2aea9944c966acbd4b8672e5716546dcL103 - we pass a browse record where a class instance is expected. I'd rather change this (either pass `self` instead of `i` or remove the parameter completely in order to use self from the main function's scope)

NL66278:
The problem was that a browse record was passed, where an instance of model was expected. But then this was never used, leading to a no self use warning in pylint. Therefore the obvious solution to me was moving the method out of the class altogether and just calling it from the module.

rafaelbn:
Hi, is this PR needed? Maybe customer is already in v8...

pedrobaeza:
Is this also still relevant?

NL66278:
This is still used in production at customer side, so yes, still relevant.

pedrobaeza:
The same pylint issues here, so you can clean them in one of them, and when merged (I can fast-track it as exception as this one is for an old version), and then rebase the other.

pedrobaeza (approved):
OK, I'm merging this as you have made the hard work and I'll fix the remaining pylint issue from the other module. Please rebase the other PR branch in about 5 minutes.

PR #212, opened 2015-12-07
[ADD] partner_sector

JavierIniesta:
This module adds the concept of sector to partners.

PR #218, opened 2015-12-14
8.0 partner industry

jt-xx:
Adds an industry sector object and a field 'Industry sector' after the tags on the partner form view.


gurneyalex:
see discussion on #212

PR #226, opened 2015-12-18
[8.0][partner_contact_job_position] Add new module to categorize job positions in contacts

yajo:
Add a field for almost the same as the `function` field of `res.partner`, but allowing categorizations in tree form; without replacing `function`, so it remains usable as a kind of detailed/uncategorized job position.

@rafaelbn @JavierIniesta.


elicoidal:
LGTM

yajo:
Runbot is broken with https://github.com/odoo/odoo/issues/10159, not my fault.

rafaelbn:
Tested in local :+1: 

Please review @JavierIniesta , thanks!

JavierIniesta:
Tested in runbot :+1:

JMODaERP:
I´ve tested in runbot :+1:

PR #237, opened 2016-01-10
9.0 port partner_external_maps

alexis-via:
Contributed by Wim Audenaert.


alexis-via:
Travis is red because "Transifex password not recognized- exiting early." ; all the other travis tests are green.

pedrobaeza:
Please rebase

alexis-via:
Rebase done. The file REAME.md is really a pain, because it will always conflict everytime a new module is ported.

pedrobaeza:
Not if you don't change it. And you mustn't. This is done automatically by an script, so you should abstent to update manually.

README.md
```diff
@@ -41,7 +42,11 @@ addon | version | summary
 [partner_auto_salesman](partner_auto_salesman/) | 8.0.1.0.0 (unported) | Partner auto salesman
 [partner_contact_address_detailed](partner_contact_address_detailed/) | 8.0.1.0.0 (unported) | All address data in summarized contact form
 [partner_contact_birthdate](partner_contact_birthdate/) | 8.0.1.0.0 (unported) | Contact's birthdate
+<<<<<<< 3398e80890c1e14aa3131baad1a5bf6f9ed6a96b
```
pedrobaeza: This has not been correctly rebase. Please remove these changes. As stated, they are done automatically after the merge by an script.

alexis-via:
Could someone merge this please ?

partner_external_maps/__openerp__.py
```diff
@@ -23,7 +23,7 @@
 
 {
     'name': 'Partner External Maps',
-    'version': '8.0.0.1.0',
+    'version': '9.0.0.1.0',
```
pedrobaeza: 9.0.1.0.0

pedrobaeza:
Some things that I consider before merging:
- Rename to a more polite singular name _partner_map_.
- Don't add and then remove changes in README.md. This will lead to conflicts when migrating the rest of the modules to v9 with the procedure in https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-9.0

elicoidal:
Can you please update the copyright headers to the simple version in all py files following OCA templates
https://github.com/OCA/maintainer-tools/blob/master/template/module/__openerp__.py#L3

alexis-via:
I think changing the name of the module is a bad idea in most cases (for me, the only good reason to change the name of the module would be if the destination/purpose of the module is radically changed). As the module is kept unchanged, I think we should keep the old name.

pedrobaeza:
We don't have to extend a firstly badly named module across versions, so we always take the migration as an occasion to change it. Please reconsider your position.

alexis-via:
Well, I don't think that the name is that bad... And if the purpose is only to remove the "s" at the end because some OCA board member had the brillant idea to invent a new "rule" that module names should not have an "s" at the end, then give him some "real" work (porting module to v9 for example) so that he can make himself helpful to the community. In the OCA, we need more guys doing the real work and less people inventing theorical "rules" about every details of modules.

pedrobaeza:
Hey, the change is not only about the `s` (I'm proposing `partner_map`), but this is not only a brilliant of an OCA board member (that's to say, me), but a general convention rule to avoid misspelling problems on every place where the module have to be named (XML-IDs, scripts, JS references...). You can find it here in the contributing conventions: https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#modules (the first one in this section).

If the problem is the job of renaming, then I can do it.

alexis-via:
In fact, I am in favor of naming conventions, including for module names, but I think that these conventions should apply on new modules only. Changing the name of modules is a pain for users, in particular for non-expert users, for no real benefit.

But OK, if you really want to change the name of this module, you can do it.

pedrobaeza:
Name change and other improvements in https://github.com/akretion/partner-contact/pull/2

pedrobaeza:
Sorry about the lint errors. May I do another PR for fixing them?

```
partner_external_map/models/map_website.py:6:1: F401 '_' imported but unused
partner_external_map/models/map_website.py:6:1: F401 'api' imported but unused
partner_external_map/models/res_partner.py:6:1: F401 'fields' imported but unused
partner_external_map/models/res_users.py:6:1: F401 '_' imported but unused
```

alexis-via:
i fixed them

pedrobaeza:
Thanks :+1:

alexis-via:
Ready to merge :+1:

lmignon:
:+1: (Code review + functional)

PR #193, opened 2015-10-29
Update Partner Contact Personal Information module to 9.0

richard-willdooit:
partner_contact_personal_information updated to version 9.0

PR #243, opened 2016-01-18
Port of partner_contact_birthdate

leemannd:
There was an error with multiple tabs "personal information".
xmids have been changed.
And duplicate views are supressed.


pedrobaeza:
The  solution is not to make one module depending on the other, but put the same XML-ID on all modules declaring the same view (the one including the new tab). This XML-ID should be of a common existing module, for example "base.base_view". And then, declare another view that depends on this one adding the field in the tab.

partner_contact_nationality/views/res_partner.xml
```diff
@@ -2,33 +2,14 @@
 <odoo>
 <data>
 
-<!-- Declared the same in every module that may need it -->
-<record id="base.personal_contact_information" model="ir.ui.view">
-    <field name="name">Personal information page for contacts form</field>
+<record id="view_personal_information_gender" model="ir.ui.view">
+    <field name="name">Partner gender</field>
```
yvaucher: Wrong name here, it seems you copy/pasted the one from partner_contact_gender

yvaucher:
Reading description of https://github.com/OCA/partner-contact/tree/9.0/partner_contact_personal_information_page

It seems this module isn't intended anymore to be an inheritable module.

Thus, you should be able to remove the dependencies on it and declare in each module the extension
base.personal_contact_information . Having the same XML ID will make sure only one is valid at the end. And you should have only 1 tab.

This change took place here https://github.com/OCA/partner-contact/pull/193

leemannd:
@pedrobaeza Ok. Making it better.

guewen:
> This XML-ID should be of a common existing module, for example "base.base_view". And then, declare another view that depends on this one adding the field in the tab.

I disagree. IMO, we should never create records hijacking the namespace of another addon. Odoo won't be able anymore to clean the records created by the addon. Example: you have 2 addons that create a view `base.partner_personal_page`, even if you uninstall both, the view remains in the database. It sounds cleaner to me to have an intermediate addon for the creation of the view.

leemannd:
@guewen It is correct. The view remains in the database. For the moment, in wich manner should I make the modifications?

yvaucher:
@pedrobaeza Have you seen @guewen comment ?
What is the benefit of the change you asked over the previous version using module dependencies?

leemannd:
@pedrobaeza @yvaucher @guewen 
I have corrected the PR to fit the actual vision of partner_contact_information.
So now this is a port in version 9.0 of partner_contact_birthdate

partner_contact_birthdate/__openerp__.py
```diff
@@ -18,15 +18,15 @@
 
 {
```
yvaucher, line 5: Short header to use here

leemannd, line 5: Updated in all the files

yvaucher:
@leemannd Thanks :+1: it comply with the current state of other addons, whereas I'm not convinced it's the best way to do it.

EDIT: added reserves about the removal of the base module. Though this PR is ok for the current state.

guewen:
> I disagree. IMO, we should never create records hijacking the namespace of another addon. Odoo won't be able anymore to clean the records created by the addon. Example: you have 2 addons that create a view base.partner_personal_page, even if you uninstall both, the view remains in the database. It sounds cleaner to me to have an intermediate addon for the creation of the view.

Another effect: when you upgrade `base`, it won't find the view `base.partner_personal_page` in there so it will delete it (it will be created again when the dependent addon is installed but that's still undesirable).

pedrobaeza:
Hi, @guewen, you're totally right. I used this trick, but I didn't realize about the side effect, so the benefits are less than the cons. Go ahead with the other approach.

leemannd:
@pedrobaeza @guewen @YannickB This PR is doing two things:
1. Port of partner_contact_birthdate & making
2. The modules using the tab "personal informations" (nationality, gender, birthdate, "in_several_companies") depends now on the "partner_contact_personal_information_page"

yvaucher:
@leemannd Thanks :+1:

guewen:
Thanks @leemannd 
:+1:

pedrobaeza:
Why Travis is failing?

yvaucher:
Hum a change is missing on 

https://github.com/OCA/partner-contact/blob/9.0/partner_contact_personal_information_page/views/res_partner.xml#L6

As we need to revert it to ensure it is not on `base` module.

guewen:
@leemannd did you see @yvaucher 's comment?

JosDeGraeve:
A rebase fixes the failing build, see: https://github.com/leemannd/partner-contact/pull/4

LGTM: :+1:

lmignon:
:+1:  LGTM (Code review + Functional test) Not a blocking issue but it would be nice to move the override of res.partner from `models.py` to `models/res_partner.py` to be conform to the OCA standards.

rafaelbn:
Hi @leemannd . Thanks a lot fot this PR. Please, as this module is for v9 it's very important to achive the simple change @lmignon commented in https://github.com/OCA/partner-contact/pull/243#issuecomment-222430403. Could you make it? I will merge after this. Thanks again

partner_contact_birthdate/__init__.py
```diff
@@ -1,19 +1,4 @@
...
-#
-# This program is distributed in the hope that it will be useful,
-# but WITHOUT ANY WARRANTY; without even the implied warranty of
-# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
-# GNU Affero General Public License for more details.
-#
-# You should have received a copy of the GNU Affero General Public License
-# along with this program.  If not, see <http://www.gnu.org/licenses/>.
-
+# © <YEAR(S)> <AUTHOR(S)>
```
rafaelbn: You should maintain original authors
cc @Yajo

partner_contact_birthdate/__openerp__.py
```diff
@@ -1,24 +1,10 @@
...
-# (at your option) any later version.
-#
-# This program is distributed in the hope that it will be useful,
-# but WITHOUT ANY WARRANTY; without even the implied warranty of
-# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
-# GNU Affero General Public License for more details.
-#
-# You should have received a copy of the GNU Affero General Public License
-# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+# © <YEAR(S)> <AUTHOR(S)>
```
rafaelbn: You should maintain original authors

partner_contact_birthdate/__openerp__.py
```diff
@@ -1,24 +1,10 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
 {
     "name": "Contact's birthdate",
-    "version": "8.0.1.0.0",
+    "version": "9.0.1.0.0",
     "author": "Odoo Community Association (OCA)",
```
rafaelbn: Here is missing Original author @Yajo

partner_contact_birthdate/__openerp__.py
yajo: Short header is 2 lines: copyright & license. Keep both please.

leemannd:
@rafaelbn @Yajo @lmignon Hello, I'm going to make the changes at the end of the day or tomorrow. So it could be merged.
Thank you for your comments

dreispt:
Changes seem good, but TravisCI build is red.

partner_contact_birthdate/__openerp__.py
leemannd: Didn't remove it, but I'm adding it.

partner_contact_birthdate/__openerp__.py
```diff
@@ -1,25 +1,12 @@
...
-# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+# Copyright 2014-2015  Grupo ESOC <www.grupoesoc.es>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
 {
     "name": "Contact's birthdate",
-    "version": "8.0.1.0.0",
-    "author": "Odoo Community Association (OCA)",
+    "version": "9.0.1.0.0",
+    "author": "Jairo Llopis",
```
yajo: The comma here goes inside the string

yajo:
Thanks for taking care of this @leemannd :wink:

dreispt:
@leemannd Still needs fixing - see TravisCI log.

leemannd:
@yvaucher @dreispt @Yajo Pep8 corrected and freshly rebased.

gurneyalex:
:+1:

PR #244, opened 2016-01-20
[8.0][FIX][partner_relations] Split read method overwrite

antespi:
Split read method in two methods:
- api.v7
- api.v8

This is needed in order to respect calling chain for API v7 methods

### DIFF
--- a/partner_relations/model/res_partner.py
+++ b/partner_relations/model/res_partner.py
@@ -299,7 +299,13 @@ class ResPartner(models.Model):
             cr, uid, args + date_args + active_args, offset=offset,
             limit=limit, order=order, context=context, count=count)
 
-    @api.multi
+    @api.v7
+    def read(self, cr, user, ids, fields=None, context=None,
+             load='_classic_read'):
+        return super(ResPartner, self).read(
+            cr, user, ids, fields=fields, context=context, load=load)
+
+    @api.v8
     def read(self, fields=None, load='_classic_read'):
         return super(ResPartner, self.with_partner_relations_context())\
             .read(fields=fields, load=load)


pedrobaeza:
Without this, some calls can give an error, so :+1:

rafaelbn:
Fasttrack :+1:

hbrunn:
@rafaelbn @antespi this will introduce subtle errors as the v7 version doesn't adapt the context. You need a v7 version of `with_partner_relations_context` (https://github.com/Antiun/partner-contact/blob/e74bc190d10af9b442c8d21920e8db75174805ce/partner_relations/model/res_partner.py#L319) and pass this context to `super`

So create a function that just returns the context keys, call super in v7 with `context=dict(context, **self.partner_relations_context(context))` and have `with_partner_relations_context` use this function too.

And please don't fasttrack stuff that if you're not absolutely sure what you're doing, and in the current case, not at all. The rules are quite clear on this: https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#please-respect-a-few-basic-rules

hbrunn:
@antespi please also elaborate which problem this fixes

pedrobaeza:
@rafaelbn, you have merged so fast without the approval of the original author.

@hbrunn, there's no need of what you say, because v7 method call the @api.v8 method, which is overwritten by your method, so the context is changed. This is due to an error in this line: https://github.com/odoo/odoo/blob/8.0/addons/project/project.py#L95, that we have traced back up to this method, because an incompatibility v7/v8 API.

antespi:
@hbrunn, IMHO It is not need to add context, like @pedrobaeza says. I though the same but base v7 is calling v8 method, so no need to add context twice.

hbrunn:
Ah, I see: https://github.com/OCA/OCB/blob/8.0/openerp/models.py#L3143, thanks for making this clear. I hereby partially retract my rant.

The error in upstream is, as so often, that a keyword argument is passed positionally, this screws up the version mangling. But given we don't have influence on this, I agree we have to jump through this hoop here.

rafaelbn:
Absolutely. I'm sorry, furthermore been partner_relations module which has give us such a big work with several bugs.
It will not happen again, for sure.

hbrunn:
no worries, mistakes happen, as my previous post proves ;-)

PR #246, opened 2016-01-25
Add ondelete option to parent_id, runbot fails to lack of attribute.

feketemihai:


feketemihai:
@OCA/crm-sales-marketing-maintainers Allow travis and runbot greenify.

pedrobaeza:
I prefer to keep children department with null as parent. Why this new behaviour?

feketemihai:
@pedrobaeza Should i change it then to 'restrict' ...

pedrobaeza:
I would let it as is. The default is "set null", which seems perfectly fine for me.

feketemihai:
@pedrobaeza Seems to be a new check in creating parent on the server side, please see https://runbot.odoo-community.org/runbot/build/3140208 errors...

feketemihai:
https://github.com/odoo/odoo/blob/8.0/openerp/models.py#L2757

pedrobaeza:
Uhm, I didn't know about that one. Then I see better 'restrict' for avoiding accidental lose of data.

StefanRijnhart:
It's because the parent store is otherwise not properly updated upon a deletion. :+1:

pedrobaeza:
:+1:

yajo:
:+1:

PR #248, opened 2016-02-15
[REF] Fix new checks of latest version of flake8

moylop260:


pedrobaeza:
:+1: when Travis goes green.

max3903:
:+1:

moylop260:
Runbot with warnings unrelated

dreispt:
Where do these rules come from? AFAIK we have no rule in place about the indentation of closing braces/brackets.

yvaucher:
:-1: About this rule on closing brackets. Both are valid.

dreispt:
I agree with @yvaucher , thus my previous question.
IMO Java-like closing brackets are not pythonic (they break the general python indentation blocks rule of practice).
Maybe not everyone shares my opinion, but forcing it on everyone is not nice, to say the least.

nhomar:
:+1: There is a pylint rule related to the change, it is not java like it is a python rule.

yvaucher:
@nhomar those are not rules but  conventions. In that case the convention states that both are valid:

From https://www.python.org/dev/peps/pep-0008/

```
 The closing brace/bracket/parenthesis on multi-line constructs may either line up under the first non-whitespace character of the last line of list, as in:

my_list = [
    1, 2, 3,
    4, 5, 6,
    ]
result = some_function_that_takes_arguments(
    'a', 'b', 'c',
    'd', 'e', 'f',
    )

or it may be lined up under the first character of the line that starts the multi-line construct, as in:

my_list = [
    1, 2, 3,
    4, 5, 6,
]
result = some_function_that_takes_arguments(
    'a', 'b', 'c',
    'd', 'e', 'f',
)
```

lmignon:
I agree with @yvaucher :-1: 
PLZ no new check! Each time you enforce a new rule you discourage new contribution. Moreover this new cosmetic rule is a questionable rule.

qtheuret:
I also agree with @yvaucher − PEP 8 defines both cases, so flake8 doesn't raise warnings/erros on this.
Which version of flake8 do you use ?

moylop260:
Sorry, I saw the travis red by flake8 and I ran the flake8 locally without our configuration file (my bad) to fix them

screen shot 2016-05-06 at 9 05 22 am

moylop260:
I closed the issue because now the travis build of 8.0 is green

yvaucher:
@moylop260 Thanks

PR #249, opened 2016-02-18
[9.0] partner_firstname - Fix user form with firstname and lastname asking for a mandatory name

yvaucher:


yajo:
It would be nice if you can add a test for this.

partner_firstname/models/res_user.py
```diff
@@ -28,3 +28,10 @@ def default_get(self, fields_list):
                 result[field] = inverted.get(field)
 
         return result
+
+    @api.one
```
leemannd: deprecated

pedrobaeza:
Please check pylint error

yvaucher:
@pedrobaeza fixed (don't over do, travis is already there to say so :smile: )

@Yajo I just added a test on the onchange of user

pedrobaeza:
:+1: 

@yvaucher, it was just in case you were not aware of it.

leemannd:
:+1:

PR #254, opened 2016-03-12
[IMP] base_location_geonames_import: Make tests more efficient

pedrobaeza:
Instead of importing full ZIP, we limit the test import to 10 entries, making very fast the test execution.


pedrobaeza:
cc @alexis-via

alexis-via:
An easier solution would have been to switch from Italy to Monaco in the YAML test. That way, no need to patch the original code of the module with some parameter that are only used for import, and we are independent from the order of the cities inside the ZIP file. For your information, Monaco has 29 entries in its zip file, so it's pretty quick. And you will also speed-up the test because the size of the file to download it much smaller (1,7 KB for MC vs 245 KB for IT).

pedrobaeza:
Well, that's a partial solution, because for example in Spanish localization I test also other things limiting the number of imported records: https://github.com/OCA/l10n-spain/pull/330/files#diff-0dd3b3350c7271e5e9f7fad2f79a44afR18

Anyway, thanks for the tip. I'll change it to save some bits for the Earth!

alexis-via:
:+1:

pedrobaeza:
@alexis-via, I have switched to Monaco, converted tests to Python, and include a test for checking deleted entries.

pedrobaeza:
@rafaelbn @sergio-teruel @carlos-incaser, please check this one.

pedrobaeza:
As this is blocking Spanish Localization tests, I merge this one.

PR #270, opened 2016-05-05
Add Local Adminitrative Units support

dreispt:


pedrobaeza:
There's already a module called base_location_nuts that makes this. Isn't it?

dreispt:
You mean reusing the NUTS table to go all the way to Level 5?
Eurostat states that NUTS 4 and 5 are deprecated, bu I can see the technical benefits for that.

As a side note, I noticed now that he 9.0 NUTS implementation on Parters in different from 8.0.
It would be nice to have a migration script and some advice on the README about this.

dreispt:
There is a con about using the NUTS module: it changes the Partner form adding a lot of fields that are not needed here. Thoughts?

pedrobaeza:
OK, I see that there are different levels of information, so it's good to have it as a separate module.

dreispt:
Tests fail because of `base_location_geonames` :confused:

pedrobaeza:
Another question. Why do you need such strict classification and not using base_location?

About base_location_geonames_import, please see https://github.com/OCA/partner-contact/pull/268.

dreispt:
I considered using it. Technically it could do the job.
But it feels like mixing apples and oranges:
The `base_location` is about a zip code table. 
This is about a municipality code table.

Allowing to select a municipality code where a zip code is expected doesn't sound good.

pedrobaeza:
base_location also allows it leaving the zip field empty (it's not required).

dreispt:
Yes, but why would a zipless municipality <s>should</s> show as an option alongside a proper zip code?

pedrobaeza:
I don't get you, Daniel. You can use the field for auto-completing the rest of the fields, or as a classification if you want.

dreispt:
Zip is an address element.
Municipality code is a reporting attribute,and is not address related.
I don't see why it should be selectable in places I expect zip codes.

pedrobaeza:
You are giving that field the address meaning, not the module itself. As I have said, zip code is optional.

dreispt:
If the municipality code is selectable from the address, then it gains address meaning, and I do not wish that.
If I use the same table for zip and municipality, when I type a city name in the address, I will get both option to select. And I want the address to show only zip codes, not municipality codes, unrelated to address.

How could I add municipalities to the zip code table without affecting addresses?

pedrobaeza:
OK, I refrain from this. I won't block the PR, but I also don't share your point of view.

dreispt:
@pedrobaeza I still would like to see your point of view, but I'm not there yet.
AFAICS your suggestion is to place municipalities in the same table as zip code.
What would be the the rationale?

pedrobaeza:
The purpose is to save one data model, as it seems that you are not going to use zip codes in the localization.

dreispt:
I can also use zip codes.
It's an unrelated decision.
To reuse a model, the best candidate is the NUTS one, but I would need it without any impact on Addresses.
It was because of these two concerns that I ended up deciding to create a third model to Municipalities, reusable for other countries LAUs.

My particular use case is to add municipality codes to employees, not addresses/partners in general.

pedrobaeza:
OK, I understood. You're right about the different concepts.

Go ahead then.

dreispt:
I need two OKs :wink: 
PS: I'm using this for https://github.com/OCA/l10n-portugal/pull/14

base_location_lau/README.rst
```diff
@@ -0,0 +1,89 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+    :alt: License: AGPL-3
```
pedrobaeza, line 3: Put target link

base_location_lau/README.rst
```diff
@@ -0,0 +1,89 @@
...
+============
+
+No data is provided by this module.
+Country specific data should be provided by localization modules.
+
+
+Configuration
+=============
+
+The Local Administrative Units table setup can be done under
```
pedrobaeza: s/done/manage

base_location_lau/README.rst
```diff
@@ -0,0 +1,89 @@
...
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/134/8.0
+
+Bug Tracker
+===========
+
+Bugs are tracked on `GitHub Issues <https://github.com/OCA/partner-contact/issues>`_.
+In case of trouble, please check there if your issue has already been reported.
+If you spotted it first, help us smashing it by providing a detailed and welcomed feedback
+`here <https://github.com/OCA/partner-contact/issues/new?body=module:%20base_location_nuts%0Aversion:%208.0%0A%0A**Steps%20to%20reproduce**%0A-%20...%0A%0A**Current%20behavior**%0A%0A**Expected%20behavior**>`_.
```
pedrobaeza: Change the module name

base_location_lau/__openerp__.py
```diff
@@ -0,0 +1,37 @@
+# -*- coding: utf-8 -*-
+##############################################################################
```
pedrobaeza: Use short headers like latest guidelines: https://github.com/OCA/maintainer-tools/blob/master/template/module/__init__.py

pedrobaeza:
Please change the icon to distinguish base_location_nuts from this one.

dreispt:
I kept it because it was NUTS related, but I'll put the OCA one then.

pedrobaeza:
No, if you want, you can simply add the word LAU, and thus they are different. It's just a question of not having the exact icon 2 times.

base_location_lau/models/res_partner_lau.py
```diff
@@ -0,0 +1,24 @@
...
+    level = fields.Integer(required=True, index=True)
+    code = fields.Char(required=True)
+    name = fields.Char(required=True, translate=True)
+    country_id = fields.Many2one('res.country', 'Country', required=True)
+    state_id = fields.Many2one('res.country.state', 'State')
+    # Parent hierarchy
+    parent_id = fields.Many2one('res.partner.lau', ondelete='restrict')
+    child_ids = fields.One2many('res.partner.lau', 'parent_id', 'Children')
+    parent_left = fields.Integer('Parent Left', select=True)
+    parent_right = fields.Integer('Parent Right', select=True)
```
andreparames: I think `select` is deprecated in the new API.

dreispt:
I must confess that image editing is not my thing. I'll just place the OCA icon then.

base_location_lau/models/res_partner_lau.py
pedrobaeza: Yeah, it's index=True.

andreparames:
You can always use a mix of the two :smile: 

oca_nuts

Besides what's already been noted, the PR looks good to me!

dreispt:
Fixed.

dreispt:
@andreparames Thanks for the icon, I used it in my last commit.

pedrobaeza:
:+1:

dreispt:
Travis looks good now.
@andreparames can I have your +1?

base_location_lau/models/res_partner_lau.py
```diff
@@ -0,0 +1,24 @@
...
+
+    level = fields.Integer(required=True, index=True)
+    code = fields.Char(required=True)
+    name = fields.Char(required=True, translate=True)
+    country_id = fields.Many2one('res.country', 'Country', required=True)
+    state_id = fields.Many2one('res.country.state', 'State')
+    # Parent hierarchy
+    parent_id = fields.Many2one('res.partner.lau', ondelete='restrict')
+    child_ids = fields.One2many('res.partner.lau', 'parent_id', 'Children')
+    parent_left = fields.Integer('Parent Left', index=True)
```
elicoidal, line 23: out of curiosity: how is this information updated? Is it automatic?
For the rest :+1:

andreparames:
:+1:

andhit-r:
@dreispt no lau1_id and lau2_id on partner form?

dreispt:
@andhit-r Correct. No obvious place to put the - that will depend on why you need LAUs.
So I decided to leave that out of the module.

andhit-r:
@dreispt thanks for the explanation. One more question though. Why don't you leave out creation lau1_id and lau2_id on other modules?

dreispt:
@andhit-r Sorry, I didn't understand.

andreparames:
@andhit-r well, having them in this module ensures the field names will be standard, instead of having different modules choosing incompatible names. Do you see any disadvantage in having them here?

andhit-r:
@dreispt @andreparames the creation of 2 LAUs field on this module makes me wonder, why only 2 level provided by this module? **the LAUs model itself give us the ability to defined more than 2 level**. Euro countries might only need 2 LAUs, but in Indonesia for example (https://en.wikipedia.org/wiki/Subdivisions_of_Indonesia), it need 3 level LAU below state. So IMO:
- Leave this module only to create LAUs model
- New module base_location_lau_level1 for creation lau1_id field on res.partner
- New module base_location_lau_level2 for creation lau2_id field on res.partner
- an so on

I know additional module will be one module one field addition, but with this approach further development for additional LAUs field on res.partner will be standardized. Or maybe you have idea how to accommodate more than 2 level LAU on res.partner?

dreispt:
I see that as a localization specific extension, adding the extra fields to Partners.

Conceptually, the NUTS levels are major regions below the Country level. LAU are low level administrative regions, such as municipalities or districts.

You should consider using the `base_location_nuts` instead of this: it has the same multi-level code table, in 8.0 it adds 4 fields to the Partner, but in 9.0 that was simplified to just 2 (region and substate).

andhit-r:
@dreispt i think LAU concept is what i need, not NUTS. I also need low level administrative regions. But okay, i will workaround my 3rd level LAUs need. 

So :+1: from me

dreispt:
Can we merge this?

dreispt:
ping!

PR #273, opened 2016-05-12
Fix error : when creating a new relation between res.partner (type : …

Creamaster:
…Person), the view filter only relation between company, without checking the partner type, making it impossible to select relation typed for two Person


rafaelbn:
Please @antespi @pedrobaeza take a look here

pedrobaeza:
@hbrunn can be a better candidate to check it

pedrobaeza:
Closing as this hasn't been reported by anyone more and no activity.

Issue #196, opened 2015-10-30
base_partner_merge duplicated with Odoo crm module


guewen:
I noticed that the `base_partner_merge` module is duplicated with the official `crm` module which includes this wizard. They share the same model name, views, etc. so it produces weird effects when both are installed. The code seems roughly the same (it was a backport originally) but I didn't checked if we have patches on this module to push upstream (at least #182 and #183 I guess).

Should we remove this module?

pedrobaeza:
Yeah, it seems to me also weird when I saw a few weeks ago some patches for this. I have in mind to make a _base_merger_ general module, but not on the near roadmap.

rafaelbn:
Yes, it's duplicated.

hbrunn:
@guewen @pedrobaeza @rafaelbn 
I see merit in this module because is gives you the ability to merge partners without getting all the cruft of crm.

I think the way to go is making it clear in the readme that this module doesn't do anything else than the crm module, and shouldn't be installed if the latter is installed. I just merged a PR that at least adds a proper readme amongst others.

Then we can copy base_partner_merge\* from crm - which we should, because it looks like there are some subtle improvements.

Or we could try to be smart and do

```
from openerp.addons.crm import base_partner_merge
```

in `__init__.py`, this always should give us the current code. Might be a problem if some environments only have the code that is actually installed, but in a standard environment, this should work. Then we could have some abstract model that pulls the relevant data files from the `crm` module in `_register_hook`. This way, it would look like it's its own module, but all code comes from `crm`. What do you think?

pedrobaeza:
I see the import as the smartest option, and I don't think there should be a problem, because almost everyone has available all the core addons in their deployments.

rafaelbn:
Well this proposal has sense :+1: .

For me (what I wish) is that base_partner_merge would improve functionality of CRM... see https://github.com/odoo/odoo/issues/10534
I Odoo is impossible to deduplicate comparing partners with `is a company` = true by VAT (for example)

In my experience this year with many migrations from other CRM, ERP, Account of any database to Odoo, is that if Odoo and OCA expect a big growing is because many companies will migrate from other software and this tools will be critical to have a qualified data base of customers.

yvaucher:
See https://github.com/OCA/partner-contact/pull/274 in which I implemented what @hbrunn proposed.

pedrobaeza:
Closing as the PR is merged.

PR #274, opened 2016-05-13
[9.0][PORT] base_partner_merge

yvaucher:
Migration of module base_partner_merge.

It proposes a simple way allows to use the crm deduplicate wizard without having to install CRM module.

See the discution here: https://github.com/OCA/partner-contact/issues/196#issuecomment-169781002

Thanks @hbrunn for the idea.

I set `_module = 'base_partner_merge'` to ensure the model are loaded.

I will let you do other PRs to extend or modify the behaviour of the wizard if needed to alter the way the official wizard works.


base_partner_merge/__openerp__.py
```diff
@@ -1,18 +1,16 @@
...
-    'author': "OpenERP S.A.,Odoo Community Association (OCA)",
+    'name': 'Deduplicate Contacts (No CRM)',
+    'summary': 'Partner merge wizard without dependency on CRM',
+    'author': "Camptocamp,Odoo Community Association (OCA)",
     'category': 'Generic Modules/Base',
-    'version': '8.0.0.1.0',
-    'description': """
-backport module, to be removed when we switch to saas2 on the private servers
-""",
+    'version': '9.0.0.1.0',
```
pedrobaeza: 9.0.1.0.0

base_partner_merge/base_partner_merge.py
```diff
@@ -1,895 +1,11 @@
-# -*- coding: utf-8 -*-
+from openerp.addons.crm.base_partner_merge import *
```
pedrobaeza: Put a try here if there are custom deployments without the crm module.

pedrobaeza: Can you also be more specific about what to import instead of using `*`?

yvaucher: @pedrobaeza I want the whole file to simulate it is there.

pedrobaeza: OK then

pedrobaeza:
Please bump the version and check Travis.

yvaucher:
Fixed

hbrunn:
cool that this works! I pondered if it's a good idea to load the view from the crm addon to in a function called from xml files, but that's a bit tricky and we wouldn't be able to add extra features, so I think this is fine the way it is :+1:

lmignon:
:+1:  Works like a charm. Thanh you @pedrobaeza

yvaucher:
@lmignon wrong PR? Or didn't got your tee/coffee yet? :smile:

yvaucher:
Thanks for the merge @pedrobaeza

lmignon:
oupssss not the wrong PR my thankfulness is for you @yvaucher

PR #276, opened 2016-05-18
partner_firstname: Write in one step

kkarolis:
Both fields should be written in one go. An inline write like this
causes validations to be called two times and fail at times where
lastname becomes False and firstname is written only in following write.

Use Case:
When the intention is to change the order of firstname and lastname, the
user is guided to override the _get_inverse_name.  One of the simpler
approaches is to switch the fields of the result returned by this
function. However, this will fail for companies.

At the moment lastname and firstname attributes are written, lastname
will be written first (if the order is changed it will be False) leaving
the record temporarily with both lastname and firstname False which
raises an exception in _check_name function.

### DIFF
--- a/partner_firstname/models.py
+++ b/partner_firstname/models.py
@@ -155,7 +155,12 @@ class ResPartner(models.Model):
     def _inverse_name(self):
         """Try to revert the effect of :meth:`._compute_name`."""
         parts = self._get_inverse_name(self.name, self.is_company)
-        self.lastname, self.firstname = parts["lastname"], parts["firstname"]
+        values = {
+            "lastname": parts["lastname"],
+            "firstname": parts["firstname"]
+        }
+        self.write(values)
+        self.update(values)
 
     @api.one
     @api.constrains("firstname", "lastname")


pedrobaeza:
There's an error in Travis. Please check.

kkarolis:
Was an issue because normal attribute assignment behaves differently compared to write in case of onchange methods. Added explicit write and update methods because the _inverse_name is used both in onchange and in "normal" methods.

partner_firstname/models.py
```diff
@@ -155,7 +155,12 @@ def _get_inverse_name(self, name, is_company=False):
     def _inverse_name(self):
         """Try to revert the effect of :meth:`._compute_name`."""
         parts = self._get_inverse_name(self.name, self.is_company)
-        self.lastname, self.firstname = parts["lastname"], parts["firstname"]
+        values = {
+            "lastname": parts["lastname"],
+            "firstname": parts["firstname"]
+        }
+        self.write(values)
+        self.update(values)
```
dreispt, line 163: I'm sorry if my question is dumb. What is `self.update()`?

partner_firstname/models.py
```diff
@@ -155,7 +155,12 @@ def _get_inverse_name(self, name, is_company=False):
     def _inverse_name(self):
         """Try to revert the effect of :meth:`._compute_name`."""
         parts = self._get_inverse_name(self.name, self.is_company)
-        self.lastname, self.firstname = parts["lastname"], parts["firstname"]
+        values = {
+            "lastname": parts["lastname"],
+            "firstname": parts["firstname"]
+        }
+        self.write(values)
```
dreispt, line 162: Nitpick: I would make the above two instructions into a single `write` instruction.

partner_firstname/models.py
hbrunn, line 163: a shorthand for multiple assignments: https://github.com/OCA/OCB/blob/8.0/openerp/models.py#L5465

partner_firstname/models.py
hbrunn, line 162: don't `write` the values, the update suffices

kkarolis, line 162: @hbrunn , I would argue that not in all cases. Isin't update supposed to work with in-memory records? (solely). This of course works fine in the interface (because it updates the viewed model) which is later written on save with updated values.

But the problem is when you want to switch the firstname and lastname places. Imagine the following situation:

Clean database,  except for some partners with the following names:
1. Candyshop [is_company=True]
2. Bobby [is_company=False]
3. John Smith [is_company=False]

Now the user installs module partner_firstname into his db, the following computation takes place:
1. lastname: Candyshop, firstname: false, name: Candyshop
2. lastname: Bobby, firstname: false, name: Bobby
3. lastname: John, firstname: Smith, name: John Smith

Now the 2,3 are incorrect. I want to correct lastname and firstname for some records. One way to achieve this (and as suggested in the _get_inverse_name docstring) is to override that function and change the ordering there. To reuse the logic, I would get the result dictionary, {"lastname":..} and switch the places of the values. E.g. something like:

``` python
@api.model
 def _get_inverse_name(self, name, is_company=False):
    res = super(ResPartner, self)._get_inverse_name(
        name, is_company=is_company)
    return {'firstname': res['lastname'], 'lastname': res['firstname']}
```

In the new module, I would also call the `_inverse_name` method to recompute the existing fields.

Now the problem in such a case would occur for entries 1 and 2. Because at the moment lastname write is made, the record does not have firstname yet (it was just returned from the function), and the constraints added will be violated (also failing to install the module).

Now I do realize, that there are ways to overcome this with post_install or migrations. Sadly, in some cases this requires some hand holding and the update is not totally automatic.I figured it would have been nicer to just write that place in one go and solve it all together. Unfortunately, as it later turned out this solution is incorrect.

I hope I made valid arguments that there is a problem in some cases.  Maybe anybody has experience how to do this differently. At the very least, an improvement on the IMO deceiving docstring would be nice?

kkarolis:
Do not merge. This change causes recursion issues because of subtle differences between direct `write` invocation and attribute assignment `self.attrb = ...`

partner_firstname/models.py
hbrunn, line 162: That's all very nice, but misses the point entirely. If you `write` in an onchange handler, the user can't undo this easily by simply cancelling the current edit action. An onchange handler (and compute functions btw) never should have side effects, because you'll _always_ run into situations where they fall on your feet.

As you write yourself, this belongs into a *install hook and/or migration scripts.

kkarolis, line 162: @hbrunn , I am not arguing. `write` shouldn't be there, but `update` doesn't fix the problem stated either.

Just more mental load to have when using this module.

PR #279, opened 2016-05-31
[ADD] init hook to set gender based on titles

hbrunn:


partner_contact_gender/tests/test_partner_contact_gender.py
```diff
@@ -0,0 +1,10 @@
+# -*- coding: utf-8 -*-
+# © 2016 Therp BV <http://therp.nl>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from openerp.tests.common import TransactionCase
+
+
+class TestPartnerContactGender(TransactionCase):
+    def test_partner_contact_gender(self):
+        from ..hooks import post_init_hook
```
pedrobaeza, line 16: You should put in the setUp a test partner with any of the titles, but without gender, call the hook, and then assert the gender to actually test something.

partner_contact_gender/__init__.py
```diff
@@ -17,3 +17,4 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 from . import models
+from hooks import post_init_hook
```
pedrobaeza: `from .hooks import post_init_hook`

hbrunn:
@pedrobaeza thanks, done

pedrobaeza:
:+1: 

Thanks

gurneyalex:
:+1:

PR #282, opened 2016-06-20
[9.0][ADD] New module partner_firstname_auth_ldap

lmignon:


partner_firstname_auth_ldap/models/res_company_ldap.py
```diff
@@ -0,0 +1,30 @@
...
+        :param tuple ldap_entry: single LDAP result (dn, attrs)
+        :return: parameters for a new resource of model res_users
+        :rtype: dict
+        """
+        values = super(ResCompanyLdap, self).map_ldap_attributes(
+            conf, login, ldap_entry)
+        values.update({
+            'lastname': ldap_entry[1]['cn'][0],
+            'firstname': ldap_entry[1]['givenName'][0],
+            'email': ldap_entry[1]['mail'][0],
```
hbrunn, line 28: isn't this already taken care of by https://github.com/OCA/server-tools/blob/8.0/users_ldap_mail/users_ldap_model.py ?
And did you consider making the values configurable as the mail module does it?

lmignon, line 28: @hbrunn Thank you for the review. I'm not aware of `users_ldap_mail`. I'll take the same approach in this module to stay consistent...

hbrunn, line 28: you might also have a look at https://github.com/OCA/server-tools/tree/8.0/users_ldap_push
What I planned back then was to split this up into two modules, one taking care of the mapping from ORM fields to ldap attributes, and the second to do the actual push. But unfortunately, time constrains didn't permit this.

lmignon, line 28: Indeed `users_ldap_push` seems to provide a good way to define the mapping between the LDAP and the ORM. I agree with you, the module should be split into two modules but as for you, I also have time constrains.

hbrunn, line 28: we'll add that sooner or later. For now, having the attributes configurable as in users_ldap_mail is perfectly fine I think.

partner_firstname_auth_ldap/__init__.py
```diff
@@ -0,0 +1 @@
+from . import models
```
NL66278, line 1: Missing coding and licence

partner_firstname_auth_ldap/__openerp__.py
```diff
@@ -0,0 +1,17 @@
+# -*- coding: utf-8 -*-
+# Copyright 2016 ACSONE SA/NV
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+{
+    'name': 'Partner Firstname Auth Ldap',
+    'summary': """
```
NL66278, line 7: Summary should fit one line (do not use triple quotes)

partner_firstname_auth_ldap/__openerp__.py
```diff
@@ -0,0 +1,17 @@
...
+        Fill firstname and lastname for partners imported from ldap""",
+    'version': '9.0.1.0.0',
+    'license': 'AGPL-3',
+    'author': 'ACSONE SA/NV,Odoo Community Association (OCA)',
+    'website': 'https://acsone.eu/',
```
NL66278, line 12: ```https://github.com/oca/partner-contact```

partner_firstname_auth_ldap/models/__init__.py
```diff
@@ -0,0 +1 @@
+from . import res_company_ldap
```
NL66278, line 1: coding and licence

NL66278:
Need to add ldap to requirements.txt to solve travis errors.

NL66278 (changes requested):
See other comments. Need to fix travis errors.

NL66278:
@lmignon Can this be closed, or would you still like to take up the comments?

NL66278:
Closing for lack of action.

PR #288, opened 2016-07-18
[FIX][partner_academic_title] Avoid to display academic titles on companies.

### LINKED ISSUE #217: [8.0] partner_academic_title - add academic tittle to Company but comanies cannot have academic tittle
module: partner_academic_title
version: 8.0

**Steps to reproduce**
- Go To a Company see field "academic tittle"

**Current behavior**
- You can see "academic titlle" in companies

**Expected behavior**
- "Academic tittle" is only for non-companies (contacts)

adrienpeiffer:
closes #217.

### DIFF
--- a/partner_academic_title/views/res_partner_view.xml
+++ b/partner_academic_title/views/res_partner_view.xml
@@ -7,8 +7,8 @@
             <field name="inherit_id" ref="base.view_partner_form" />
             <field name="arch" type="xml">
                 <xpath expr="//field[@name='title']" position="after">
-                    <field name="academic_title_ids" class="oe_edit_only" widget="many2many_tags" placeholder="Academic Titles"/>
-                    <field name="academic_title_display" class="oe_read_only" />
+                    <field name="academic_title_ids" class="oe_edit_only" widget="many2many_tags" placeholder="Academic Titles" attrs="{'invisible': [('is_company', '=', True)]}"/>
+                    <field name="academic_title_display" class="oe_read_only" attrs="{'invisible': [('is_company', '=', True)]}" />
                 </xpath>
             </field>
         </record>


guewen:
:+1:

rafaelbn:
Hi @adrienpeiffer thanks! :smile: 
Runbot is building but travis has failed. Could you review errors please?

partner_academic_title/views/res_partner_view.xml
```diff
@@ -7,8 +7,8 @@
             <field name="inherit_id" ref="base.view_partner_form" />
             <field name="arch" type="xml">
                 <xpath expr="//field[@name='title']" position="after">
-                    <field name="academic_title_ids" class="oe_edit_only" widget="many2many_tags" placeholder="Academic Titles"/>
-                    <field name="academic_title_display" class="oe_read_only" />
+                    <field name="academic_title_ids" class="oe_edit_only" widget="many2many_tags" placeholder="Academic Titles" attrs="{'invisible': [('is_company', '=', True)]}"/>
```
pedrobaeza, line 10: It should be:

```
attrs="{'invisible': ['|', ('is_company', '=', True), ('parent_id', '=', False)]}
```

adrienpeiffer, line 10: why ?

pedrobaeza, line 10: Because both conditions are the standard for determining if a partner is a "commercial partner".

adrienpeiffer, line 10: If `parent_id` is False academic titles should be visible.

guewen, line 10: @pedrobaeza a company can't have an academic title so it makes sense to hide the field for them, but a partner without `parent_id` might be an individual with an academic title

pedrobaeza, line 10: A partner without parent is also a company. @StefanRijnhart has said to me this other times. I'm on travel right now so I can't elaborate more, but I'll do it when arriving.

pedrobaeza, line 10: The problem is that with quick_create, you're creating partners without parent, but also without is_company flag checked, so a lot of times, you will see bad configured partners with this option enabled.

adrienpeiffer, line 10: I'm not sure to understand what you say. For me this little correction is correct. Any other opinion ?

pedrobaeza, line 10: OK, go ahead then. It's only a precision that maybe doesn't worth

pedrobaeza:
:+1:

lmignon:
:+1:

elicoidal:
:+1:

StefanRijnhart:
@pedrobaeza I must have meant: partners that are not a company but have no parent id are _commercial_ partners in the specific Odoo sense of the word. But they are still persons. Sorry if I caused any confusion.

PR #289, opened 2016-07-21
[FIX] Avoid writes if no changes in name, firstname, lastname ...

antespi:
Also call super write with sudo only if vals is empty. This is necessary for avoid access checking here https://github.com/odoo/odoo/blob/8.0/openerp/addons/base/res/res_partner.py#L569

### DIFF
--- a/partner_firstname/models/res_partner.py
+++ b/partner_firstname/models/res_partner.py
@@ -1,23 +1,9 @@
 # -*- coding: utf-8 -*-
-
-#    Author: Nicolas Bessi. Copyright Camptocamp SA
-#    Copyright (C)
-#       2014:       Agile Business Group (<http://www.agilebg.com>)
-#       2015:       Grupo ESOC <www.grupoesoc.es>
-#       2015:       Antiun Ingenieria S.L. - Antonio Espinosa
-#
-#    This program is free software: you can redistribute it and/or modify
-#    it under the terms of the GNU Affero General Public License as
-#    published by the Free Software Foundation, either version 3 of the
-#    License, or (at your option) any later version.
-#
-#    This program is distributed in the hope that it will be useful,
-#    but WITHOUT ANY WARRANTY; without even the implied warranty of
-#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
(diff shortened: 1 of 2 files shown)


yvaucher:
:+1:

rafaelbn:
Please review runbot and travis @antespi

antespi:
Travis errors are not related
@yvaucher Thanks for your review :smile:

partner_firstname/models/res_partner.py
```diff
@@ -236,3 +230,11 @@ def _install_partner_firstname(self):
         # Force calculations there
         records._inverse_name()
         _logger.info("%d partners updated installing module.", len(records))
+
+    @api.multi
+    def write(self, vals):
+        name = vals.get('name')
+        if name and all(name == partner.name for partner in self):
+            vals.pop('name', None)
+        this = self.sudo() if not vals else self
```
pedrobaeza, line 244: Document why you make this here

pedrobaeza:
:+1:

rafaelbn:
Thanks :+1:

partner_firstname/models/res_partner.py
```diff
@@ -236,3 +230,16 @@ def _install_partner_firstname(self):
...
+    def write(self, vals):
+        name = vals.get('name')
+        if name and all(name == partner.name for partner in self):
+            vals.pop('name', None)
+        # If vals is empty (only write name field and with the same value)
+        # Avoid access checking here
+        # https://github.com/odoo/odoo/blob/
+        #   8b83119fad7ccae9f091f12b6ac89c2c31e4bac3/openerp/addons/base/res/
+        #   res_partner.py#L569
+        this = self.sudo() if not vals else self
```
yajo, line 244: @pedrobaeza I was with @antespi working on this and I know this is a workaround needed to avoid an access error when autosubscribing somebody. Odoo behaves weirdly here.

Anyway, I think we should write a test that tries to set the same name a partner has into itself, but written with a unprivileged user, and check no exceptions are raised.

This bug was very hard to locate, and I think this simple test is worth it to avoid future replications.

What do you think?

pedrobaeza, line 244: OK, it worths to be tested

PR #291, opened 2016-07-25
[8.0][FIX] Travis using mock

antespi:
Using mock python module (https://docs.python.org/dev/library/unittest.mock.html) backported to v2.7, we can mock `check_vies` library method in order to not call it and return a sample object.

This avoid test errors in future due to VIES service downtimes or VIES database changes.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,8 +22,10 @@ env:
   matrix:
   - LINT_CHECK="1"
   - TRANSIFEX="1"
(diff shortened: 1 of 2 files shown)


antespi:
Well, the problem here is the conflict between `partner_create_by_vat` and `partner_changeset`

antespi:
:smile: It works!!
I've separated `partner_changeset` from other addons test

yajo:
:+1:

yvaucher:
:+1:

carlosdauden:
:+1:

PR #292, opened 2016-08-03
[9.0] [ADD] partner_financial_risk

carlosdauden:
Adds a new page in partner to manage its _Financial Risk_.
If any limit is exceed the partner gets forbidden to confirm sale orders.

@Tecnativa @pedrobaeza


partner_financial_risk/models/res_partner.py
```diff
@@ -0,0 +1,120 @@
...
+from dateutil.relativedelta import relativedelta
+from openerp import api, fields, models
+
+
+class ResPartner(models.Model):
+    _inherit = 'res.partner'
+
+    eval_risk_sale_order = fields.Boolean(
+        string="Eval Sale Orders", help="Compute in total risk")
+    max_risk_sale_order = fields.Monetary(
```
yajo: `fields.Monetary` exists?

partner_financial_risk/README.rst
```diff
@@ -0,0 +1,67 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+    :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+    :alt: License: AGPL-3
+
+======================
+Partner Financial Risk
+======================
+
+Adds a new page in partner to manage the partner's *Financial Risk*
+If any limit is exceed the the partner is blocked to confirm sale orders.
```
yajo: Better: Adds a new page in partner to manage its _Financial Risk_. If any limit is exceed the the partner gets forbidden to confirm sale orders.

partner_financial_risk/models/res_partner.py
```diff
@@ -0,0 +1,120 @@
...
+
+from datetime import datetime
+from dateutil.relativedelta import relativedelta
+from openerp import api, fields, models
+
+
+class ResPartner(models.Model):
+    _inherit = 'res.partner'
+
+    eval_risk_sale_order = fields.Boolean(
```
yajo: About the naming, I'd make 2 suggestions: 
1. Keep all fields grouped by prefix. For example:
   - Put all under `risk_` prefix.
   - Group by model: `risk_sale_order_*`.
2. These `eval` fields... It is a bit misleading. How about `allowed` instead?

partner_financial_risk/models/res_partner.py
carlosdauden: It isn't documented, but :https://github.com/odoo/odoo/blob/e37ba6183269b0d075833efbb9bf26e99404247e/openerp/fields.py#L1131-L1130

yajo: Cool, thanks! Shouldn't you be using currency_field then?

partner_financial_risk/models/res_partner.py
```diff
@@ -0,0 +1,120 @@
...
+        compute='_compute_risk_sale_order', store=True,
+        string='Sale Order Not Invoiced')
+
+    risk_invoice_draft_include = fields.Boolean(
+        string='Include Draft Invoices', help='Compute in total risk')
+    risk_invoice_draft_limit = fields.Monetary(
+        string='Limit In Draft Invoices', help='Set 0 if it not lock')
+    risk_invoice_draft = fields.Monetary(
+        compute='_compute_risk_invoice', store=True,
+        string='Not Validated Invoice')
```
yajo: Much better now. I know there are a lot of fields, but IMHO this module would benefit a lot from having `help` attributes that tell you what each field is used for.

carlosdauden:
Travis fails with **base_location_geonames_import** module

carlosdauden:
LINT_CHECK fails in base_partner_merge.

Unrelated travis fail

partner_financial_risk/README.rst
elicoidal: Partner's sales order cannot be confirmed whenever the limit is exceeded.

partner_financial_risk/README.rst
```diff
@@ -0,0 +1,67 @@
...
+
+Configuration
+=============
+
+To configure this module, you need to:
+
+#. Go to *Invoicing > Configuration > Settings > Invoicing & Payments >
```
elicoidal: split the line

partner_financial_risk/README.rst
```diff
@@ -0,0 +1,67 @@
...
+   Financial Risk* and set *Maturity Margin*.
```
elicoidal: Please explain the usage of Maturity Margin field

partner_financial_risk/models/res_partner.py
```diff
@@ -0,0 +1,120 @@
...
+from datetime import datetime
+from dateutil.relativedelta import relativedelta
+from openerp import api, fields, models
+
+
+class ResPartner(models.Model):
+    _inherit = 'res.partner'
+
+    eval_risk_sale_order = fields.Boolean(
+        string="Eval Sale Orders", help="Compute in total risk")
```
elicoidal: the help message is not that clear: "Full risk computation"?

partner_financial_risk/models/res_partner.py
```diff
@@ -0,0 +1,120 @@
...
+    max_risk_sale_order = fields.Monetary(
+        string='Max In Sale Orders', help="Set 0 if it not lock")
```
elicoidal: Set 0 if it is not locked

partner_financial_risk/README.rst
elicoidal: Can you elaborate the rationale behind your calculation of the risk?
Is it based on some documentation/best practices?
Detailing the criteria and / or documenting with URL/guide is necessary for this kind of module
Checking the code is not that clear and gives me the feeling that this is specific for one customer/accountant PoV.
I am not saying it is wrong but we need modules that are generic or related to best practices and we need to document them if necessary.

rafaelbn: Hi @elicoidal , the main purpose of module it that the user (CFO or similar) could just set an amount of money that customer can use before paying it. But decision about this amount is analyzed outside Odoo. For Example a company which sale computer: CFO will decide that Customer1 will have 6.000 € of risk  just because this customer usually buy around that amount in computer every month. So if Customer1 wants to buy 12.000€ one month system will block this operation and CFO o similar should approve this. All operations bellow 6.000 will be approved automatically.
Conclusion: in this case Risk is just a field.

rafaelbn:
Tested in http://3151659-292-678711.runbot2.odoo-community.org/web

This module must depends in payment mode as Risk depends on how Customer Pay. Paypal, Credit card, transfer doesn't generate any risk,

1. Delete tab "Financial Risk" in the Partner. It should be a section inside Accounting tab. Note that this information should be showed only for partner which are customers and not suppliers.
   Furthermore if not we have to take in consideration security rules. If risk in the accounting tab then security group is from accounting which is correct.
2. It don't have any sens to set a Financial Risk per sale.orders and for invoices. A Partner (customer) have a only one and single risk depending payment (payment credit which means that you sell products and customer doesn't pay them). If customer pay like bank transfer or credit card in a moment it doesn't have any sense to limit selling :smile: 
   So Just one Limit is OK
3. In this moment module doesn't take in consideration quotations. But it should because if you send a quotation with payment mode is transfer is OK (no risk to operation) but if you send a quoation of risk with 30 day payment, this is a risk so you cannot create a quotation with that kind of payment.
4. As commented in screenshot we must correct that you cannot create an invoice in draft if it it exceeds the limit with the payment mode.

rafaelbn:
Ping @pedrobaeza

partner_financial_risk/README.rst
elicoidal: @rafaelbn thanks for the explanation. My question was broader than that: there are multiple fields and logic in the code that is not crystal clear and that might require some more explanation.
Not a real blocking point but then the module is never really used if you cannot really understand it (which was my case this morning)

yajo: I agrree with @elicoidal. If a module is not well explained, it cannot be well evaluated and used.

rafaelbn:
@carlosdauden travis fails, could you check it please?

carlosdauden:
@rafaelbn I explain travis errors a few comments before

pedrobaeza:
Better call the second module _partner_sale_risk_ and the third one _partner_stock_risk_

partner_financial_risk/README.rst
```diff
@@ -0,0 +1,69 @@
...
+    :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+    :alt: License: AGPL-3
+
+======================
+Partner Financial Risk
+======================
+
+Adds a new page in partner to manage its *Financial Risk*.
+
+If any limit is exceed the partner gets forbidden to confirm sale orders.
```
pedrobaeza: s/sale orders/the corresponding sales order or customer invoice (this is needed to be implemented if not already).

ezerouali:
Hi Carlos.
Seems like "Financial Risk" tab has some rendering issue when in edit mode, at least in Chrome for Mac (see attached). Have analysed involved css and the div (id=notebook_page_73) handling the data has a correct width but the nested table has a wrong and much larger one. Hope that helps.
screen shot 2016-08-09 at 11 47 48

carlosdauden:
@makwin Now looks fine

carlosdauden:
Now, this PR is to review

rafaelbn:
functional test :+1:

partner_financial_risk/models/account_invoice.py
```diff
@@ -0,0 +1,37 @@
...
+
+from openerp import api, models, _
+
+
+class AccountInvoice(models.Model):
+    _inherit = 'account.invoice'
+
+    @api.multi
+    def invoice_open(self):
+        if not self.env.context.get('bypass_risk', False):
```
pedrobaeza: For avoiding this cyclomatic complexity, better to repeat the return:

```
if not self.env.context.get('bypass_risk', False):
    return self.signal_workflow('invoice_open')
for invoice in self:
...
return self.signal_workflow('invoice_open')
```

partner_financial_risk/models/res_partner.py
```diff
@@ -0,0 +1,150 @@
...
+             'risk_invoice_open_include'),
+            ('risk_invoice_unpaid', 'risk_invoice_unpaid_limit',
+             'risk_invoice_unpaid_include'),
+            ('risk_account_amount', 'risk_account_amount_limit',
+             'risk_account_amount_include'),
+        ]
+
+    @api.model
+    def _get_depends_compute_risk_exception(self):
+        # TODO: Improve code without lose performance
```
pedrobaeza:
```
# TODO: Improve code without performance loss
```

partner_financial_risk/models/res_partner.py
```diff
@@ -0,0 +1,150 @@
...
+    risk_invoice_open_limit = fields.Monetary(
+        string='Limit In Open Invoices', help='Set 0 if it is not locked')
+    risk_invoice_open = fields.Monetary(
+        compute='_compute_risk_invoice', store=True,
+        string='Total Open Invoices',
+        help='Residual amount of invoices in Open state and the date due is '
+             'not exceeded, considering Due Margin set in account '
+             'settings')
+    risk_invoice_unpaid_include = fields.Boolean(
+        string='Include Due Invoices', help='Full risk computation')
```
pedrobaeza: Change label to talk about 'unpaid', not 'due invoices', which is not the same.

partner_financial_risk/models/res_company.py
```diff
@@ -0,0 +1,13 @@
...
+# © 2016 Carlos Dauden <carlos.dauden@tecnativa.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from openerp import fields, models
+
+
+class ResCompany(models.Model):
+    _inherit = 'res.company'
+
+    invoice_due_margin = fields.Integer(
```
pedrobaeza: This should be called `invoice_unpaid_margin` for keeping homogeneity

partner_financial_risk/models/res_partner.py
```diff
@@ -0,0 +1,150 @@
...
+        # TODO: Improve code without lose performance
+        tuple_list = self._risk_field_list()
+        res = [x[0] for x in tuple_list]
+        res.extend([x[1] for x in tuple_list])
+        res.extend([x[2] for x in tuple_list])
+        res.append('credit_limit')
+        return res
+
+    @api.model
+    def process_due_invoice(self):
```
pedrobaeza: Call this method `process_unpaid_invoices` for consistency (in plural, as you're processing all the unpaid, not only one). Singular form is only used in model names and module names, not in methods.

partner_financial_risk/README.rst
```diff
@@ -0,0 +1,70 @@
...
+    :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+    :alt: License: AGPL-3
+
+======================
+Partner Financial Risk
+======================
+
+Adds a new page in partner to manage its *Financial Risk*.
+
+If any limit is exceed the partner gets forbidden to confirm customer invoice.
```
pedrobaeza: `If any limit is exceeded, you won't be able to confirm any of its invoices unless you are authorized.`

partner_financial_risk/README.rst
```diff
@@ -0,0 +1,70 @@
...
+
+Configuration
+=============
+
+To configure this module, you need to:
+
+#. Go to *Invoicing > Configuration > Settings > Invoicing & Payments*
+#. Go to *Financial Risk* and set *Due Margin* (This field adds days of margin
```
pedrobaeza: `In the *Financial Risk* section, fill *Unpaid Margin* for setting the number of days to last after the due date to consider an invoice as unpaid.`

pedrobaeza: Which are the authorized groups? Please explain here.

partner_financial_risk/README.rst
```diff
@@ -0,0 +1,70 @@
...
+   before that a invoice is consider as unpaid)
+
+Usage
+=====
+
+To use this module, you need to:
+
+#. Go to *Customers > Financial Risk*
+#. Set limits and choose options to compute in credit limit.
```
pedrobaeza, line 31: Explain which are the possible limits and their use.

partner_financial_risk/README.rst
```diff
@@ -0,0 +1,70 @@
...
+Adds a new page in partner to manage its *Financial Risk*.
+
+If any limit is exceed the partner gets forbidden to confirm customer invoice.
+
+Configuration
+=============
+
+To configure this module, you need to:
+
+#. Go to *Invoicing > Configuration > Settings > Invoicing & Payments*
```
pedrobaeza: Invoicing > Invoicing/Accounting (it can have both names, depending on the installed modules).

partner_financial_risk/README.rst
```diff
@@ -0,0 +1,70 @@
...
+#. Go to *Financial Risk* and set *Due Margin* (This field adds days of margin
+   before that a invoice is consider as unpaid)
+
+Usage
+=====
+
+To use this module, you need to:
+
+#. Go to *Customers > Financial Risk*
```
pedrobaeza:
```
#. Go to *Invoicing/Accounting > Sales > Customers*.
#. Select an existing customer or create a new one.
#. Open the *Financial Risk* tab.
```

partner_financial_risk/README.rst
```diff
@@ -0,0 +1,70 @@
...
+#. Set limits and choose options to compute in credit limit.
+#. Go to *Invoicing > Sales > Customer invoices* and create new customer
```
pedrobaeza: `#. Test the restriction trying to create an invoice for the partner for an amount higher of the limit you have set.`

pedrobaeza (changes requested):

pedrobaeza:
Please include also Spanish translation.

JordiBForgeFlow:
@carlosdauden @pedrobaeza Did you consider this? https://github.com/OCA/account-financial-tools/blob/8.0/account_credit_control. Why did you choose not to reuse? 

Naming a module 'financial risk' is too broad.  What you are managing here is a customer credit risk.

Customer credit control is the usual name given to this process.

pedrobaeza:
@jbeficent This is more than a credit control. See the possible options. I know that module, but it doesn't make the same and it's not the same approach.

JordiBForgeFlow:
@pedrobaeza OK. Is interesting. I worked in something similar long ago here: https://github.com/OCA/sale-financial/pull/9.

pedrobaeza (approved):

PR #299, opened 2016-08-12
[FIX] Fix test according to updated geonames list

rousseldenis:
@pedrobaeza Can you merge ?


pedrobaeza:
All green, so merging!

PR #304, opened 2016-08-31
9.0 partner relations v8api

NL66278:
Here is the promised PR to migrate partner_relations (not installable) to partner_multi_relation, based on the 8.0 API.

Beside this there are many improvements, like working domains so only appropiate values can be selected etc.


hbrunn:
:+1:

NL66278:
@rafaelbn DId you review this one? Would be nice if this PR could be accepted. Then we can move the module also up to 10.

lfreeke:
:+1: (test)

pedrobaeza:
OK, I think we can merge this one as there has been already reviewed and let any possible problem to further PRs.

PR #307, opened 2016-09-07
FIX birthdate date install

ecino:
Fix a bug where a typo would prevent to set all birthdate dates at installation of module partner_contact_birthdate


ecino:
For whatever reason it seems the install function is never called. Can someone help me on that? I'm not familiar with .yml files.

pedrobaeza:
I don't see any post_init_hook or similar, so how it's called that install function?

ecino:
My guess is that the .yml should do it :
https://github.com/CompassionCH/partner-contact/blob/b7d7bcf82fdd202ba18be5122e3b51849e02c142/partner_contact_birthdate/data/res_partner.yml#L19

But the yml file is not listed in the openerp manifest:
https://github.com/CompassionCH/partner-contact/blob/8.0/partner_contact_birthdate/__openerp__.py#L29

ecino:
Now the FIX is working.

pedrobaeza:
It's better to use a post_init_hook, or this will be executed on each update.

Garamotte:
I agree with @pedrobaeza, this should be done in a post init hook.

ecino:
I made the change.

partner_contact_birthdate/hooks.py
```diff
@@ -0,0 +1,27 @@
+# -*- coding: utf-8 -*-
+###############################################################################
```
pedrobaeza: Put short header.

partner_contact_birthdate/hooks.py
```diff
@@ -0,0 +1,27 @@
...
+#
+###############################################################################
+
+from openerp import SUPERUSER_ID
+from openerp.api import Environment
+
+
+def post_init_hook(cr, pool):
+    env = Environment(cr, SUPERUSER_ID, {})
+    env['res.partner']._birthdate_install()
```
pedrobaeza: Put here the code instead of calling a new function:

```
env['res.partner'].search(
    [('birthdate', "!=", False)])._birthdate_inverse()
```

pedrobaeza (changes requested):

ecino:
Done

pedrobaeza (approved):

Garamotte (approved):

invitu:
I did a PR for those 3 modules because they are small and depend on partner_contact_personal_information_page

requested fixes are commited

PR #207, opened 2015-11-24
[ADD] res.partner: Porting partner_street_number module to Odoo 9.0

DarkoNikolovski:

PR #314, opened 2016-10-14
[MIG][10.0] Porting partner_street_number to V10

astirpe:
Based on the job https://github.com/OCA/partner-contact/pull/207 made by @DarkoNikolovski.


partner_street_number/__init__.py
```diff
@@ -33,11 +33,11 @@ def post_init_hook(cr, registry):
...
-        'SELECT id, street FROM res_partner '
-        'WHERE street IS NOT NULL and street_name IS NULL'
-        )
-    partner_obj = registry['res.partner']
-    for partner in cr.fetchall():
-        partner_obj.write(
-            cr, SUPERUSER_ID, partner[0], {'street': partner[1]})
+
+    env = api.Environment(cr, SUPERUSER_ID, {})
+    partners = env['res.partner'].search(
```
hbrunn: add `.with_context(active_test=False)` to get all partners here.
Search can be formatted more beautifully as

``` python
myobject.search([
    (prop1),
    (prop2),
])
```

making a change in a proposition a one line change

partner_street_number/__init__.py
```diff
@@ -33,11 +33,11 @@ def post_init_hook(cr, registry):
...
+        [('street', '!=', False),
+         ('street_name', '=', False)])
+
+    for partner in partners:
+        partner.sudo().write({'street': partner.street})
```
hbrunn: your environment is already created for root, so no need to sudo here a second time

partner_street_number/static/src/css/field.css
```diff
@@ -0,0 +1,11 @@
+.o_address_street_name {
+	   float: left !important;
+       margin-right: 2% !important;
+       width: 75% !important;
+   }
+
+.o_address_street_number {
+       margin-right: 0% !important;
+       width: 25% !important;
```
hbrunn: is is possible to avoid the `!important`s by adding a longer selector? The important qualifier tends to mess up css in the sense that it overrides everything

hbrunn (changes requested):

partner_street_number/__manifest__.py
```diff
@@ -19,19 +19,21 @@
...
-    "summary": "Introduces separate fields for street name and street number.",
-    "version": "8.0.0.1.0",
-    "author": "Therp BV,Odoo Community Association (OCA)",
-    "website": "https://github.com/oca/partner-contact",
-    "category": 'Tools',
-    "depends": [
-        'base'
+    'name': 'Street name and number',
+    'summary': 'Introduces separate fields for street name and street number.',
+    'version': '10.0.0.1.0',
```
pedrobaeza: 10.0.1.0.0

partner_street_number/i18n/en.po
```diff
@@ -0,0 +1,53 @@
+# Translation of Odoo Server.
```
pedrobaeza: Remove this file

partner_street_number/i18n/partner_street_number.pot
```diff
@@ -4,7 +4,7 @@
 #
```
pedrobaeza: Remove this file

partner_street_number/static/src/css/field.css
```diff
@@ -0,0 +1,19 @@
+.o_form_view .o_address_format > .o_form_input.o_address_street_name.o_form_field {
```
pedrobaeza: Maybe you can make a less?

pedrobaeza (approved):
Well, the less is the same of the CSS. It doesn't make use of any of the available variables that can be changed in pack, but for now I'm not going to push further.

Please squash on merging.

rafaelbn (approved):
Tested in runbot

pedrobaeza:
You should now that Odoo is working on a similar approach with the recent announce of including Mexican localization in v10 EE: https://github.com/odoo/odoo/pull/13835.

partner_street_number/models/res_partner.py
```diff
@@ -41,26 +42,26 @@ def _write_street(self):
...
-        street_number = False
-        if self.street:
-            match = re.search(r'(.+)\s+(\d.*)', self.street.strip())
-            if match and len(match.group(2)) < 6:
-                street_name = match.group(1)
-                street_number = match.group(2)
         for partner in self:
-            self.street_name = street_name
-            self.street_number = street_number
+            street_name = partner.street and partner.street.strip() or False
```
OpenCode: Can you use inline if:

`partner.street.strip() if partner.street else False`

?

partner_street_number/views/res_partner.xml
```diff
@@ -7,46 +7,63 @@
...
+                               />
+                </xpath>
+                <xpath expr="//div[@class='o_address_format']//field[@name='street']"
+                       position="attributes">
+                    <attribute name="invisible">1</attribute>
+                </xpath>
+            </field>
+        </record>
+
     </data>
```
OpenCode: <openerp><data> can be replaced with <odoo>

rafaelbn:
Hi @hbrunn , here is missing you final review, thanks!

partner_street_number/__manifest__.py
dreispt: why?

pedrobaeza:
We should wait for merging this module until https://github.com/odoo/odoo/pull/13835 is resolved, because this overlaps the same functionality in core.

hbrunn (approved):

hbrunn:
I'd propose to merge as it is, and adapt to core changes when core is actually changed

pedrobaeza:
@hbrunn I prefer not, or will have to deal later with migration scripts and maybe to have a dummy module all version around.

astirpe:
@pedrobaeza what about all the modules that depend on this?

hbrunn:
I don't see a problem with needing migration scripts, and replacing all address-related modules from this repo by dummies once `base_address_structured` landed seems the right thing to do to me. We have no clue how long it will take, so if somebody moves forward before, why block it?

pedrobaeza:
Please just wait one week to see if there are moves upstream. Then we decide.

hbrunn:
one week, deal. Revisiting October 28th ;-)

astirpe:
and...?

pedrobaeza:
hehe, OK, let's go with this one, as there's no answer from Odoo so far.

pedrobaeza:
Anyway, maybe you can consider the same things as I say in the other PR: https://github.com/odoo/odoo/pull/13835#discussion_r84131784

hbrunn:
I say: Merge now, and add @pedrobaeza's suggestions which sound perfectly valid to me as occasion allows

partner_street_number/hooks.py
```diff
@@ -0,0 +1,41 @@
+# -*- coding: utf-8 -*-
+##############################################################################
```
pedrobaeza: Please short header.

astirpe:
Indeed the comments of @pedrobaeza are interesting. We could get inspiration from https://github.com/openvenues/libpostal and its Python bindings to make a generic solution. Thank you!

I will create a new PR for that. Or maybe someone else would like to work at it?

hbrunn:
@pedrobaeza can we merge this?

pedrobaeza:
OK, let's go!

Issue #308, opened 2016-09-28
[partner_address_street3] Suggestion to add street3 field in address_format using post_init_hook and uninstall_hook


atchuthan:
@gurneyalex 

The field is only available in res.partner form and is not shown in name_get of show_address at Sale order nor in its report.

So, I am planning to create post_init_hook for this module, to add street3 in the address format of each country via SQL query.

**post_init_hook query**
UPDATE res_country
SET    address_format = replace(address_format, '%(street2)s', E'%(street2)s\n%(street3)s')

**uninstall_hook query**
UPDATE res_country
SET    address_format = replace(address_format, E'\n%(street3)s', '')

Let us know your opinion

gurneyalex:
sounds like a good idea to me.

atchuthan:
thanks.. will try to make the PR in few days

atchuthan:
created PR https://github.com/OCA/partner-contact/pull/316 related to this issue

AKSHAYAGOVIND:
Hi,

I am using the Odoo 12.0 enterprise version. I  imported the partner_address_street3 module. But i am getting the below error.

-------------------------------------------------------------------
ValueError: Field `street3` does not exist

Error context:
View `res.partner.form`
[view_id: 113, xml_id: base.view_partner_form, model: res.partner, parent_id: n/a]
-----------------------------------------------------------------------
Please help me how to resolve this issue. I am getting the same error when I try to import the partner_contact_Gender module as well.  I have attached my error report.

![image](https://user-images.githubusercontent.com/43378529/55417589-9d781b80-5536-11e9-9000-3649f24bc55d.png)

PR #316, opened 2016-10-18
[9.0][FIX] added post and uninstall hook for partner_address_street3

SodexisTeam:
Ref: https://github.com/OCA/partner-contact/issues/308

cc: @gurneyalex


pedrobaeza (approved):
Smart hooks!

rafaelbn (approved):
Runbot works...

alexis-via:
GRRRR, I'm seeing this PR just after finishing my PR to port to v10 https://github.com/OCA/partner-contact/pull/318 This implementation of post_install is a bit better than mine, so I'll update my code to put your implementation

alexis-via (changes requested):
I tested the change... but it doesn't work ! The hook is not called upon installation of the module, so the field format_address is not updated on countries.

alexis-via:
@atchuthan You can see an example that really works in my PR for v10 https://github.com/OCA/partner-contact/pull/318
You forgot the 2 new keys in **openerp**.py to activate your hooks.

atchuthan:
thanks @alexis-via

partner_address_street3/hooks.py
```diff
@@ -0,0 +1,28 @@
...
+    """
+    cr.execute(query)
+
+
+def uninstall_hook(cr, registry):
+    query = """
+        UPDATE res_country
+        SET address_format = replace(
+        address_format,
+        E'%(street3)s\n',
```
yajo, line 27: I think the `\n` part here should be optional. If any `%(street3)s` is found without a trailing `\n` it should get removed too.

Also please add docstrings to these hooks to make it clearer that you are adding/dropping street3 to the standard display format for addresses automatically.

partner_address_street3/hooks.py
```diff
@@ -0,0 +1,28 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+
+def post_init_hook(cr, registry):
+    query = """
+        UPDATE res_country
+        SET address_format = replace(
+        address_format,
+        E'%(street2)s\n',
+        E'%(street2)s\n%(street3)s\n'
```
yajo, line 14: Please add a test that checks this actually worked. :blush:

yajo (changes requested):

alexis-via (approved):

partner_address_street3/hooks.py
atchuthan: if I make it drop %(street3) pattern only, then address format will have an extra line after the module is uninstalled, to resolve this I need to have 2 replace query for patterns E'%(street3)s\n' and E'%(street3)s'.

Let me know if you have any solution to this.

partner_address_street3/hooks.py
atchuthan: added it in https://github.com/OCA/partner-contact/pull/316/files#diff-f9fdb975d77b348b5bab55bc71ebf13bR14

partner_address_street3/hooks.py
yajo: I'm not sure if that `E` stuff is a regexp, but if so, `E'%(street3)s\n?'` should do the trick. Otherwise just add a second lap to the replacement as you said. :wink:

atchuthan: okay

yajo: Remember to add `# pragma: no-cover` to the uninstall hook please. No way to test that AFAIK.

gurneyalex (approved):

partner_address_street3/hooks.py
pedrobaeza: In fact, you can test them manipulating previously with SQL.

yajo: Ah good point, you mean by directly calling this method, right? It's a good idea. I guess the required SQL stuff should be available already after installing the module (that is when tests run).

pedrobaeza: Yeah, the steps are:
- Change something via SQL to leave it as "original state".
- Call this method.
- Assert that the final result is the expected after the hook.

partner_address_street3/tests/test_street_3.py
```diff
@@ -33,3 +35,11 @@ def test_partner(self):
         # test synchro of street3 on write
         homer.write({'street3': 'in OCA we trust'})
         self.assertEquals(bart.street3, 'in OCA we trust')
+
+    def test_hooks(self):
+        # Test post_init_hook & uninstall_hook
+        us_country = self.env.ref('base.us')
+        self.assertTrue('%(street3)s' in us_country.address_format)
+        uninstall_hook()
```
yajo: Travis is :x: because you lack args here: `(self.cr, self.registry)`

yajo (changes requested):

partner_address_street3/tests/test_street_3.py
```diff
@@ -1,88 +1,45 @@
...
-        country_model = self.registry('res.country')
-        country_id = country_model.create(
-            self.cr,
-            self.uid,
-            {
-                'name': 'Donut Land',
-                'code': 'DNL',
-            }
-        )
+        # Test address_format has been updated on existing countries
```
yajo: Use a correct docstring for the method. This way, the docstring appears in the logs too.

partner_address_street3/tests/test_street_3.py
```diff
@@ -1,88 +1,45 @@
...
+            'type': 'contact',
+        })
+        self.assertEquals(bart.street3, 'Tho')
+
+        # test synchro of street3 on write
+        homer.write({'street3': 'in OCA we trust'})
+        self.assertEquals(bart.street3, 'in OCA we trust')
+
+    def test_hooks(self):
+        # Test post_init_hook & uninstall_hook
```
yajo: Use a correct docstring for the method. This way, the docstring appears in the logs too.

BTW you are only testing uninstall_hook here

atchuthan: the module's post_init_hook() is executed before the test files are executed and the test for 'street3' in the address_format does the job(no need to explicitly call the method)
https://github.com/OCA/partner-contact/pull/316/files#diff-f9fdb975d77b348b5bab55bc71ebf13bR42

yajo (changes requested):
Just one more little change to go :+1: :blush:

partner_address_street3/tests/test_street_3.py
```diff
@@ -1,88 +1,45 @@
...
+            'type': 'contact',
+        })
+        self.assertEquals(bart.street3, 'Tho')
+
+        # test synchro of street3 on write
+        homer.write({'street3': 'in OCA we trust'})
+        self.assertEquals(bart.street3, 'in OCA we trust')
+
+    def test_hooks(self):
+        """"Test post_init_hook & uninstall_hook"""
```
yajo: I agree with you in https://github.com/OCA/partner-contact/pull/316#discussion_r85279191

What I mean is that this test only tests the uninstall hook. Actually the post_init_hook is tested in every test because all of them would fail otherwise, but you don't specify it. This docstring should be "Test uninstall_hook", but if you simply name the method as `test_uninstall_hook`, then simply drop the docstring (it would be redundant) and it'll be OK :wink:

yajo (changes requested):
Sorry I forgot to tell you what to change in last review! here it goes

BTW, did you know that `post_init_hook` gets called only when installing an addon, **not** when updating it?

I think we need a migration script here too.

partner_address_street3/tests/test_street_3.py
atchuthan: renamed method & docstring to test uninstall_hook

yajo (approved):

yajo:
There's still the problem of broken updates, not sure if we should block this for that reason...

pedrobaeza:
Too late...

PR #317, opened 2016-10-19
[MIG] 10.0 mig base location

OpenCode:
Migration of base_location to 10.0

### DIFF
--- a/base_location/__manifest__.py
+++ b/base_location/__manifest__.py
@@ -1,43 +1,19 @@
 # -*- coding: utf-8 -*-
-##############################################################################
-#
-#    Author: Nicolas Bessi. Copyright Camptocamp SA
-#    Contributor: Pedro Manuel Baeza <pedro.baeza@serviciosbaeza.com>
-#                 Ignacio Ibeas <ignacio@acysos.com>
-#                 Alejandro Santana <alejandrosantana@anubia.es>
-#
-#    This program is free software: you can redistribute it and/or modify
-#    it under the terms of the GNU Affero General Public License as
-#    published by the Free Software Foundation, either version 3 of the
-#    License, or (at your option) any later version.
-#
-#    This program is distributed in the hope that it will be useful,
-#    but WITHOUT ANY WARRANTY; without even the implied warranty of
-#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
-#    GNU Affero General Public License for more details.
-#
-#    You should have received a copy of the GNU Affero General Public License
-#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
-#
-##############################################################################
+# Copyright 2016 Nicolas Bessi, Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
 {
     'name': 'Location management (aka Better ZIP)',
-    'version': '9.0.1.0.0',
-    'depends': ['base'],
+    'version': '10.0.1.0.0',
+    'depends': [
+        'base',
+    ],
     'author': "Camptocamp,"
               "ACYSOS S.L.,"
               "Alejandro Santana,"
               "Serv. Tecnol. Avanzados - Pedro M. Baeza,"
               "Odoo Community Association (OCA)",
     'license': "AGPL-3",
-    'contributors': [
-        'Nicolas Bessi <nicolas.bessi@camptocamp.com>',
-        'Pedro M. Baeza <pedro.baeza@serviciosbaeza.com>',
-        'Ignacio Ibeas (Acysos S.L.)',
-        'Alejandro Santana <alejandrosantana@anubia.es>',
-        'Sandy Carter <sandy.carter@savoirfairelinux.com>',
-        'Yannick Vaucher <yannick.vaucher@camptocamp.com>',
-    ],
     'summary': '''Enhanced zip/npa management system''',
     'website': 'http://www.camptocamp.com',
     'data': ['views/better_zip_view.xml',
@@ -49,6 +25,6 @@
     'demo': [
         'demo/better_zip.xml',
     ],
-    'installable': False,
+    'installable': True,
     'auto_install': False,
 }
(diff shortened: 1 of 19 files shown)


pedrobaeza (approved):

base_location/__manifest__.py
```diff
@@ -37,6 +39,7 @@
         'Alejandro Santana <alejandrosantana@anubia.es>',
         'Sandy Carter <sandy.carter@savoirfairelinux.com>',
         'Yannick Vaucher <yannick.vaucher@camptocamp.com>',
+        'Francesco Apruzzese <f.apruzzese@apuliasoftware.it>',
```
andrea4ever: Please move contributor on README

andrea4ever (approved):

base_location/__manifest__.py
OpenCode: Add contributor in README

eLBati (approved):

pedrobaeza:
We should wait until https://github.com/odoo/odoo/pull/13835 is resolved, as the module in that PR overlaps most of the features of this module.

alexis-via:
We need to decide RAPIDLY what we do about the dependency on "sales_team":
1) should we add a dependency on sales_team just to be able to add a menu entry in Sales > Configuration > Address Book ?
2) or should we stay with a dependency on "base" only and put the menu entry elsewhere, for example under base.menu_custom (i.e. "Settings > Technical > xxx").
3) any other bright idea ?
We face this problem in many module, so we need to find the right solution now.

alexis-via (approved):
Code review, no test. Approve with my remark about the dependency on sales_team ; I would prefer to avoid it !

pedrobaeza:
One suggestion is to copy the XML menu definition with the same XML-IDs here, so that we don't overlap when sale_team in installed.

alexis-via:
@OpenCode Did you migrate base_location_geonames_import too ? I don't see any PR that migrate this module, but I see your PR to migrate l10n_it_base_location_geonames_import... Otherwise, I'll do a PR to port base_location_geonames_import today or tomorrow.

OpenCode:
@alexis-via I need sale_team to use Sales > Configuration > Address Book menu. I don't like this dependence too, my I need it.
I will port base_location_geonames_import after this pull request but if you can do it for me it's ok :)

alexis-via:
@OpenCode Yes, but @pedrobaeza suggests a solution to work around this problem, even if I'm not sure that it really works. @pedrobaeza did you already test the solution that you suggest ?

alexis-via:
@pedrobaeza Could you make a PR against this PR to show us exactly the solution you propose ? I tried to implement it, but it's very difficult in real life:
- it's not just 1 menu entry, but all the "sales" hierarchy: sales_team.menu_base_partner > sales_team.menu_sale_config > sales_team.menu_config_address_book > sales_team.menu_localisation
- these menu entries use groups such as group_sale_salesman or group_sale_manager, which are also part of the sales_team module
  So I fear that it's more complex than what you think when you try to implement it.

alexis-via:
@OpenCode I ported base_location_geonames_import in this PR https://github.com/OCA/partner-contact/pull/320

OpenCode:
@alexis-via Thanks :)

@alexis-via @pedrobaeza I don't like to use this workaround

pedrobaeza:
OK, the workaround implies too much records to replicate.

alexis-via:
@pedrobaeza OK, which option do we take ? I proposed 3 options in an earlier comment... This impacts a lot of OCA modules, so we'd better choose the right option now.

pedrobaeza:
As base module doesn't have anything except Settings menu, we should locate there the corresponding menu if we don't want to depend in any other module, but maybe people with no access to Settings should be able to modify zip entries. A new top menu is no better solution, as is going to be underused. The less bad is then the Settings one I think.

OpenCode:
I propose to keep sale_team as dependence. We can ask to Odoo (using a PR) to move the menu in the next version.

eLBati:
Depending on `sales_team` sounds strange to me too, however `Countries` and other localization menus are added by `sales_team`, so, if we follow how odoo 10 is designed, adding `Cities` menu under `sales_team.menu_localisation` should be fine.
As a user, I would look for `Cities` in the neighbourhood of `Countries`, so, having to go to `Settings` would be strange

alexis-via:
@eLBati We all know that there is not perfect solution:
- if we prefer a well-designed configuration menu, we choose option 1
- if we prefer well-designed dependencies, we choose option 2
  I tend to prefer option 2, but if OCA opts for option 1, it's not a big deal for me.

If someone has a bright idea for an option 3, it's time to post it here now !

andrea4ever:
I agree with @alexis-via to use option 1

pedrobaeza:
@alexis-via, any progress in this?

alexis-via:
@pedrobaeza What do you expect me to do ? I'm not the author of this PR, and it don't think any decision has been made between the 2 options (depend on sales_team or not). But it would be good to take a decision and move on.

I said in my previous comment that I prefered option 2. @andrea4ever said he preferd option 1. Here we are...

lasley (commented):
I'm for option 2, moving to Settings => Technical or somewhere. This menu doesn't need to be seen by a normal user, and adding a depends just for a technical menu seems a bit overboard IMO

rafaelbn:
Hi,

@andrea4ever said in https://github.com/OCA/partner-contact/pull/317#issuecomment-256022714 that "agree with Alexis option 1"
@alexis-via said in https://github.com/OCA/partner-contact/pull/317#issuecomment-264030902 that "I prefered option 2"

Something doesn't fit 😉 

My opinion is option 2 -> Settings

pedrobaeza:
@OpenCode, can you do then the honors to put the menu under Settings?

OpenCode:
Ok

ok ok ok

alexis-via (commented):
To avoid this warning "Field res.company.better_zip_id: parameter 'select' is no longer supported; use 'index' instead." you should remove select=1 in company.py (or convert it to index=True, but I don't think it's useful here)

rafaelbn:
Thanks @OpenCode for this migration, waiting for last changes 😄

eLBati:
@pedrobaeza @OpenCode @alexis-via @rafaelbn @lasley , moved menu under settings

alexis-via:
OK, let's merge then :+1:

pedrobaeza:
Please, let me squash

PR #319, opened 2016-10-22
[10.0] Port partner_external_map

alexis-via:
Remove Yahoo Maps
Update URLs for MapQuest

### DIFF
--- a/partner_external_map/README.rst
+++ b/partner_external_map/README.rst
@@ -26,7 +26,6 @@ This module supports several map websites:
 * `Bing Maps <https://www.bing.com/maps/>`
 * `Here Maps <https://www.here.com/>`
 * `MapQuest <http://www.mapquest.com/>`
-* `Yahoo! Maps <https://maps.yahoo.com/>`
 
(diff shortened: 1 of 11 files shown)


pedrobaeza (approved):
Tried on runbot. Some tests would be welcome.

alexis-via:
This module has unittests... you are the author of them ! The tests work fine, that's why I didn't change them (apart from the "import" line that I updated).

pedrobaeza:
Yeah, but the coverage drops, so they are not enough! Anyway, it's enough for now.

alexis-via:
I changed didn't add nor remove any line of python code, so I don't see why the coverage drops.

pedrobaeza:
Looking at coveralls, the problem is that the test doesn't include street, street2 and other fields, and that lines remain untested. Not your fault, but as the coverage up to now was so high, it means that drop.

lasley (approved):
Thanks @alexis-via 

I agree that more tests would be awesome, but non-blocking for me due to the previous code already being untested in production

PR #322, opened 2016-11-01
[ADD] partner_default_sale_discount

pedrobaeza:
Default sales discount per partner
==================================

This module allows to define at partner level a default discount to be applied
on sales orders as the default one. As the field is a company dependent one,
you can configure a different value for the partner for each of your companies.

Configuration
=============

Enabling the use of discounts in Odoo:

* Go to *Sales > Configuration > Settings*
* On the group "Quotations & Sales", select the option "Allow discounts on
  sales order lines" for the "Discount" section.

Setting a default sales discount:

* Then, go to *Sales > Customers*.
* Create or modify a customer.
* Go to "Sales & Purchases" page.
* In the "Sale" section, there's a field called "Default sales discount (%)"
  where you can fill the default discount percentage.
* This value can only be filled at company partners, not for contacts.

Usage
=====

To use this module, you need to:

* Go to *Sales > Quotations*.
* Create a new quotation.
* Select a partner with a default discount applied.
* Create an order line, and you will see this discount as default in the line.

Known issues / Roadmap
======================

* This module doesn't work with *product_visible_discount*, as it overwrites
  the discount field with its own value.

### DIFF
--- a/partner_default_sale_discount/README.rst
+++ b/partner_default_sale_discount/README.rst
@@ -1,0 +1,85 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+==================================
+Default sales discount per partner
+==================================
+
+This module allows to define at partner level a default discount to be applied
+on sales orders as the default one. As the field is a company dependent one,
+you can configure a different value for the partner for each of your companies.
+
+Configuration
+=============
+
+Enabling the use of discounts in Odoo:
+
+#. Go to *Sales > Configuration > Settings*
+#. On the group "Quotations & Sales", select the option "Allow discounts on
+   sales order lines" for the "Discount" section.
+
+Setting a default sales discount:
+
+#. Then, go to *Sales > Customers*.
+#. Create or modify a customer.
+#. Go to "Sales & Purchases" page.
+#. In the "Sale" section, there's a field called "Default sales discount (%)"
+   where you can fill the default discount percentage.
+#. This value can only be filled at company partners, not for contacts.
+
+Usage
+=====
+
+To use this module, you need to:
+
+#. Go to *Sales > Quotations*.
+#. Create a new quotation.
+#. Select a partner with a default discount applied.
+#. Create an order line, and you will see this discount as default in the line.
+
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/134/9.0
+
+Known issues / Roadmap
+======================
+
+* This module doesn't work with *product_visible_discount*, as it overwrites
+  the discount field with its own value.
+
+Bug Tracker
+===========
+
+Bugs are tracked on `GitHub Issues
+<https://github.com/OCA/partner-contact/issues>`_. In case of trouble, please
+check there if your issue has already been reported. If you spotted it first,
+help us smashing it by providing a detailed and welcomed feedback.
+
+Credits
+=======
+
+Images
+------
+
+* Odoo Community Association: `Icon <https://github.com/OCA/maintainer-tools/blob/master/template/module/static/description/icon.svg>`_.
+
+Contributors
+------------
+
+* Pedro M. Baeza <pedro.baeza@tecnativa.com>
+
+Maintainer
+----------
+
+.. image:: https://odoo-community.org/logo.png
+   :alt: Odoo Community Association
+   :target: https://odoo-community.org
+
+This module is maintained by the OCA.
+
+OCA, or the Odoo Community Association, is a nonprofit organization whose
+mission is to support the collaborative development of Odoo features and
+promote its widespread use.
+
+To contribute to this module, please visit https://odoo-community.org.
--- a/partner_default_sale_discount/i18n/es.po
+++ b/partner_default_sale_discount/i18n/es.po
@@ -1,0 +1,34 @@
+# Translation of Odoo Server.
+# This file contains the translation of the following modules:
+# 	* partner_default_sale_discount
+#
+msgid ""
+msgstr ""
+"Project-Id-Version: Odoo Server 9.0c\n"
+"Report-Msgid-Bugs-To: \n"
+"POT-Creation-Date: 2016-11-01 16:13+0000\n"
+"PO-Revision-Date: 2016-11-01 17:13+0100\n"
+"Last-Translator: <>\n"
+"Language-Team: \n"
+"MIME-Version: 1.0\n"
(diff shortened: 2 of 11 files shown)


pedrobaeza:
cc @Tecnativa

partner_default_sale_discount/i18n/es.po
```diff
@@ -0,0 +1,34 @@
...
+
+#. module: partner_default_sale_discount
+#: model:ir.model,name:partner_default_sale_discount.model_res_partner
+msgid "Partner"
+msgstr "Empresa"
+
+#. module: partner_default_sale_discount
+#: model:ir.model,name:partner_default_sale_discount.model_sale_order
+msgid "Sales Order"
+msgstr "Pedido de venta"
```
sergio-teruel, line 34: s/Pedido/Pedidos

partner_default_sale_discount/models/sale_order.py
```diff
@@ -0,0 +1,43 @@
...
+        if xml_order_line and xml_discount:
+            # This should be handled in "string" mode
+            context = xml_order_line[0].get('context', '{}')
+            index = context.find('{')
+            context = (
+                context[:index + 1] +
+                "'default_discount': default_sale_discount, " +
+                context[index + 1:]
+            )
+            xml_order_line[0].set('context', context)
```
sergio-teruel, line 39: Why do you not use a safe_eval for update the context?

partner_default_sale_discount/models/sale_order.py
```diff
@@ -0,0 +1,43 @@
...
+
+
+class SaleOrder(models.Model):
+    _inherit = 'sale.order'
+
+    default_sale_discount = fields.Float(
+        related="partner_id.commercial_partner_id.default_sale_discount",
+        string="Default sales discount (%)",
+    )
+
```
sergio-teruel, line 16: Although the field is invisible in view, I think put it in readonly mode

sergio-teruel (changes requested):

partner_default_sale_discount/i18n/es.po
pedrobaeza: It's singular, not plural: Sales **Order**

partner_default_sale_discount/models/sale_order.py
pedrobaeza: No need for that, but I'll do it if another change is needed.

partner_default_sale_discount/models/sale_order.py
pedrobaeza: Because context can contain JS keys that can't be interpreted with eval. Take for example this injected context as example. We have the variable `default_sale_discount`, that should be interpreted in "execution time" by the JS client, so if there are another variables like this one in the previous context, `safe_eval` will fail.

partner_default_sale_discount/i18n/es.po
sergio-teruel: Oks..

sergio-teruel (approved):
Oks for me.. :+1:

partner_default_sale_discount/README.rst
```diff
@@ -0,0 +1,85 @@
...
+======================
+
+* This module doesn't work with *product_visible_discount*, as it overwrites
+  the discount field with its own value.
+
+Bug Tracker
+===========
+
+Bugs are tracked on `GitHub Issues
+<https://github.com/OCA/partner-contact/issues>`_. In case of trouble, please
```
yajo, line 55: Shouldn't this one go to sale-workflow?

partner_default_sale_discount/models/sale_order.py
yajo: It's OK, although you could simply use `context.replace("{", "{'default_discount': default_sale_discount, ", 1)`

partner_default_sale_discount/models/sale_order.py
```diff
@@ -0,0 +1,43 @@
...
+    @api.model
+    def fields_view_get(self, view_id=None, view_type='form', toolbar=False,
+                        submenu=False):
+        """Inject the default in the context of the line this way for
+        making it inheritable.
+        """
```
yajo, line 22: Isn't it easier to add the context directly to the model field?

partner_default_sale_discount/README.rst
pedrobaeza: As the information is attached to each partner, I decided to put it here.

partner_default_sale_discount/models/sale_order.py
pedrobaeza: No, because other module that changes the context will invalidate this module.

partner_default_sale_discount/models/sale_order.py
pedrobaeza: Good idea, @Yajo, changing it...

partner_default_sale_discount/models/sale_order.py
yajo: OK, then please add a comment to explain that, so next people reviewing this don't refactor it.

partner_default_sale_discount/models/sale_order.py
yajo: Please also add a comment solving @sergio-teruel's doubt.

yajo (changes requested):

partner_default_sale_discount/models/sale_order.py
```diff
@@ -31,12 +31,8 @@ def fields_view_get(self, view_id=None, view_type='form', toolbar=False,
         xml_discount = eview.xpath("//field[@name='default_sale_discount']")
         if xml_order_line and xml_discount:
             # This should be handled in "string" mode
```
yajo: Sorry, last review was on outdated changes; in any case please explain _why_ here.

yajo (changes requested):

partner_default_sale_discount/models/sale_order.py
pedrobaeza: The comment is already there.

partner_default_sale_discount/models/sale_order.py
pedrobaeza: The existing comment is enough for me.

yajo: If it were enough, I wouldn't have asked before :wink: 

If you don't explain why it should be handled as a string, then a future code reviewer can believe this is overengineered. Simply say `This should be handled in "string" mode to avoid conflicts in case more than 1 addon modify the field's or view's context` and all will be clear.

pedrobaeza: That's not the reason, but there you have it for better comprehension.

yajo (approved):

rafaelbn (approved):
Tested in runbot

pedrobaeza:
Please merge squashing now that there are 2 approvals

Issue #321, opened 2016-10-29
Licensing of some basic modules


blaggacao:
I guess there is very interesting and valuable, and above all reusable work done in this repository, however it seems that for a licensing conflict we cannot use it for localisation modules. First it needs to be acknowledge, that we can only publish under LGPL in order not to automatically produce kind of a license conflict with our more ample private developments. But then, on the other hand we would like to incorporate some modules of this repository (and possibly others) in order to honour the idea of reusability and shared effort (to avoid "reinventing the wheel"). 
Therefore, I would like to ask very humbly, if there could be any chance of relicensing some key common modules under a LGPL license? I can understand why this can be a bad idea, but unfortunately this automatically means, that no LGPL licensed module can use them, which inhibits wider adoption (among honourable and license-respecting people, which cannot adopt AGPL for whatever multiple reasons, like company policy, externally licensed mixins, etc.)

hbrunn:
for any of my commits, it's a non-negotiable no

blaggacao:
I see that the following modules could be of special interest:
- `base_location` - as it is almost inevitable that more advanced modules would depend on some kind of data produced under this facilities. (read=city based conditionals)
- `base_location_geonames_import` - the previous argument goes unchanged here
- `partner_identification` - this is a great base for amplification and adaption and make it a core component of private logic all around any possible imaginable ID.

yajo:
As long as any of you want to sue bad uses of your AGPL-licensed addons at any point in the future, it's rational to keep that license.

If you just license it in AGPL but either use it as a dependency of an LGPL/private addon (illegal) or don't want to take any action on anybody doing that, then just license it under LGPL and forget about legal issues. It's open source too, after all.

OCA's standing point is more the 2nd point, so we should either change that position or relicense to LGPL.

Of course, you must respect every author before doing that on any addon. You could open a relicense PR for each of those and ping all original authors and wait for their input, but I'm pretty sure many will not agree...

In fact, the most pragmatic approach would be you licensing all under AGPL and continuing under that license.

I'm sure I just started a flamewar, but I don't want to, so I'm closing this. Choose your option and do it:
1. Relicense yours to AGPL.
2. Open relicense PRs and ping authors.

blaggacao:
@Yajo can you clarify:

> OCA's standing point is more the 2nd point, so we should either change that position or relicense to LGPL.
> From the previous paragraphs I understand:

[...]

PR #323, opened 2016-11-02
[LICENSE] relicensing to LGPL

blaggacao:
Hello, I kindly request evaluating to agree in relicensing base_location to LGPL. It is a very basic module, and one of the OCA objectives is to foster code reusability across OCA repos. This module is so basic, that is is very easy to become infected down the line. On the other hand it is a very useful module for localisation repositories. Being still AGPL, it cannot be reused by LGPL localisations.

Thanks for your consideration! It's not too many so I guess, eventually this is feasible, until then, some people, including me are forced to reimplement the logic or similar independently under LGPL.

- [ ] @alejandrosantana 
- [ ] @tafaRU 
- [ ] @gurneyalex 
- [ ] @alexis-via 
- [ ] @bwrsandman 
- [ ] franco.tampieri
- [ ] @hhgabelgaard 
- [x] @eLBati 
- [ ] @nbessi 
- [ ] @pedrobaeza 
- [ ] @jjscarafia 
- [ ] @sbidoul 
- [ ] @yvaucher 

Based on previous discussion/issue: https://github.com/OCA/partner-contact/issues/321

### DIFF
--- a/base_location/README.rst
+++ b/base_location/README.rst
@@ -1,6 +1,6 @@
-.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
-   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
-   :alt: License: AGPL-3
+.. image:: https://img.shields.io/badge/licence-LGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/lgpl-3.0-standalone.html
+   :alt: License: LGPL-3
 
 =======================
 Enhanced ZIP management
--- a/base_location/__init__.py
+++ b/base_location/__init__.py
@@ -1,24 +1,5 @@
 # -*- coding: utf-8 -*-
-##############################################################################
-#
-#    Author: Nicolas Bessi. Copyright Camptocamp SA
-#    Contributor: Pedro Manuel Baeza <pedro.baeza@serviciosbaeza.com>
-#                 Ignacio Ibeas <ignacio@acysos.com>
-#                 Alejandro Santana <alejandrosantana@anubia.es>
-#
-#    This program is free software: you can redistribute it and/or modify
-#    it under the terms of the GNU Affero General Public License as
-#    published by the Free Software Foundation, either version 3 of the
-#    License, or (at your option) any later version.
-#
-#    This program is distributed in the hope that it will be useful,
-#    but WITHOUT ANY WARRANTY; without even the implied warranty of
-#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
-#    GNU Affero General Public License for more details.
-#
-#    You should have received a copy of the GNU Affero General Public License
-#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
-#
-##############################################################################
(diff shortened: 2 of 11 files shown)


rafaelbn:
Hi @blaggacao , you should write to contributors@odoo-community.org

eLBati:
Hi @blaggacao , to avoid future conflicts, These changes should be proposed to https://github.com/OCA/partner-contact/pull/317

About moving to LGPL, I am OK with that

yvaucher:
I personally disagree for this relicensing without more input. What kind of module would require to have a proprietary license on top on this? If you want to sell a module you can still do it in AGPL or if it is something specific you are not forced to reverse it to the community, just to the users who uses it.

Furthermore, base location implements a very basic feature for countries that should really be available for everyone in each country, I don't want to open the door to have dozen of partner doing the exact same work for the same country and sell it on apps.odoo.com while they could collaborate.

For me LGPL makes more sense for website modules as you probably want to not having to give the code to all visitors. But for backend usage the user must be able to ask for the code source thus it should stay AGPL.

yvaucher (changes requested):
I disagree

yajo:
> For me LGPL makes more sense for website modules as you probably want to not having to give the code to all visitors. But for backend usage the user must be able to ask for the code source thus it should stay AGPL.

Just think that you cannot create a LGPL addon as long as an AGPL one is in its inheritance chain, that's why @blaggacao asks for base addon on LGPL.

Not pressuring you, just clarifying a little bit. If you still don't agree, just say it and we'll close this.

blaggacao:
It's ok, I don't want to question motives. Odds have been limited since the beginning, that all would agree. I guess for this specific module, and strategically speaking, inclusion into upstream 11.0 could be a possible solution. On the other hand, technically, creating a siblings module partner_location_lgpl would be warranted. But let's just not take this too far. I guess I have tried what's possible to be able to reuse the code - according to OCA's objectives. Now, I would simply not use it. 👋 

EDIT
Well reading again I guess I misunderstood "without more input". Well, @yvaucher if you feel what @Yajo said, convinces you, let me know.

yajo:
If I were you, I'd simply publish my work under AGPL for OCA and be happy with that...

blaggacao:
Yes, I understand that many people ask themselves, why is this not an option. In short: I want to take licensing seriously given our strategic options and threads. And I think that's ok.

Furthermore, it's just as respectable to go with LGPL as it is to not to do so.
Or in your words: I unfortunately can't be "happy" with AGPL. But that's ok. I'm not feeling bad about it.

eLBati:
> Just think that you cannot create a LGPL addon as long as an AGPL one is in its inheritance chain

Actually, you can, as LGPL 3 is GPL-compatible.
See
https://www.gnu.org/licenses/gpl-faq.en.html#IfLibraryIsGPL
and
https://www.gnu.org/licenses/license-list.en.html#GPLCompatibleLicenses

yajo:
Well, true you can, but under those circumstances, combination becomes fully AGPL automatically: https://www.gnu.org/licenses/gpl-faq.html#AllCompatibility

PR #324, opened 2016-11-02
[10.0] [MIG] partner_identification

jjscarafia:


rafaelbn:
Hello @jjscarafia , please could you check travis? Thanks!

partner_identification/__manifest__.py
```diff
@@ -22,10 +22,10 @@
         'security/ir.model.access.csv',
     ],
     'author': 'ChriCar Beteiligungs- und Beratungs- GmbH, '
-              'Antiun Ingeniería S.L.',
+              'Antiun Ingeniería S.L.,'
```
rafaelbn: Please change to the new name "Tecnativa"

rafaelbn (commented):
Still pending functional test

jjscarafia:
@rafaelbn done! thanks

rafaelbn (approved):
Tested in runbot

partner_identification/__manifest__.py
```diff
@@ -11,19 +11,22 @@
 {
     'name': 'Partner Identification Numbers',
     'category': 'Customer Relationship Management',
-    'version': '9.0.1.0.0',
+    'version': '10.0.1.0.0',
+    'depends': [
+        'sales_team',
```
yajo, line 16: Can't you put the settings view in another place under _Sales > Configuration_ that does not require this dependency?

Not a blocker, though, but would be nice.

yajo (approved):

tarteo:
Hi @jjscarafia please update the "Try me on Runbot" button to 10.0 and use <odoo> instead of <openerp><data>.

jjscarafia:
@tarteo done
@Yajo as far as I can see the hole sales menu is created by sales_team module.

yajo:
Anyway I just saw that [`sale` depends on `sales_team` and not the other way around](https://github.com/odoo/odoo/blob/10.0/addons/sale/__manifest__.py#L44), so not a big problem.

PR #325, opened 2016-11-04
[[MIG] partner_create_by_vat: Migrated to 10.0.

feketemihai:
The module is directly migrated from 8.0 to 10.0, and it was merged in 8 after the 9 branch creation.


feketemihai:
@pedrobaeza Should i update travis and add requirements just like in 8.0 branch?

https://github.com/OCA/partner-contact/commit/3e7a21bd5eb3fd9a48f27892a71b5f185c4e19b2

pedrobaeza:
Yeah, I think so

partner_create_by_vat/models/res_partner.py
```diff
@@ -0,0 +1,64 @@
...
+        if country:
+            res['country_id'] = country[0].id
+        return res
+
+    @api.onchange('vat')
+    def vat_change(self):
+        res = {'value': {}}
+        # Update fields with the values available in the upper method
+        # Skip required name error
+        with self.env.do_in_onchange():
```
yajo, line 54: This is redundant? because this is already an onchange method...

yajo (approved):
Code review OK, with little non-blocking question.

partner_create_by_vat/models/res_partner.py
feketemihai: @Yajo I've already checked and no vat onchange method defined in base_vat, it was changed to api.model the check for vat and validation.

feketemihai:
@pedrobaeza Can i have your review, please...

pedrobaeza (approved):

partner_create_by_vat/README.rst
```diff
@@ -0,0 +1,70 @@
...
+#. install the module
+
+Usage
+=====
+
+On the partner's form view you will have a button in the header, called
+"Get Vies Data", available only on companies (is_company field set to True).
+Clicking the button will fetch data, when available, from the VIES Webservice, for most of
+the EU countries.
+
```
rafaelbn, line 32: @feketemihai could you please show my an example? I'm not able to use it. I've tried with ESA09000928 but it looks module doesn't works, it only takes Country field

rafaelbn (changes requested):
I don't know how to use it or it doesn't work

feketemihai:
@Yajo Hi there, sorry for the late reply, the module is working, but you see, not all the countries companies are listed with name and adress on Vies, you can check if the module is working with HU11045201, for Spain companies seems that the datas are only the country and if it's valid or not, http://ec.europa.eu/taxation_customs/vies/vatRequest.html, if it's valid it completes the country and Vat Legal Statement field, because there is no other data to be fetched, it is the same for Romania, but there we have over the localisation on OCA, not ported yet since it depends on this module other local sources that fetch data for them, see https://github.com/OCA/l10n-romania/tree/8.0/l10n_ro_partner_create_by_vat

feketemihai:
So to work, you check the is_company field, complete the vat field and push the button....

feketemihai:
Examples: BE0889734181, ATU19949705, DE217522543(only country and valid)..

yajo:
@rafaelbn Please check again like @feketemihai said.

rafaelbn (approved):

rafaelbn:
OK, tested. it works. Thanks

PR #326, opened 2016-11-04
Remove images from readme, update icon.

feketemihai:


pedrobaeza:
Why removing images? Maybe better to put images zooming in the button part?

feketemihai:
@pedrobaeza It seems that when you set images in manifest, apps.odoo.com is replacing the icon with the image...that's why...

pedrobaeza:
Well, I'm seeing both:

seleccion_001

feketemihai:
Yes...if you are on it, but in the search...
https://www.odoo.com/apps/modules/browse?search=create_by_vat

pedrobaeza:
The trick then can be to put as first image the icon...

feketemihai:
Ok...i will do it like this...Thanks. I will update the 10.0 the same.

feketemihai:
Error is in partner_changeset:

2016-11-04 09:40:48,726 6506 ERROR openerp_test openerp.addons.partner_changeset.tests.test_changeset_field_type:

yajo:
Travis' :x: is related?

feketemihai:
@Yajo The error is in partner_changeset, just like my upper comment...

yajo (approved):

rafaelbn (approved):

PR #328, opened 2016-11-07
9.0 [MIG] partner_contact_job_position migration

cubells:


partner_contact_job_position/README.rst
```diff
@@ -0,0 +1,71 @@
...
+check there if your issue has already been reported. If you spotted it first,
+help us smashing it by providing a detailed and welcomed feedback.
+
+Credits
+=======
+
+Images
+------
+
+* `Antiun Ingeniería S.L. <http://www.antiun.com>`_: Icon.
```
pedrobaeza: This icon link is not correct.

partner_contact_job_position/__openerp__.py
```diff
@@ -0,0 +1,30 @@
...
+# © 2016 Tecnativa S.L. - Jairo Llopis
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+{
+    "name": "Partner job position",
+    "summary": "Categorize job positions for contacts",
+    "version": "9.0.1.0.0",
+    'category': 'Customer Relationship Management',
+    "website": "http://www.antiun.com",
+    'author': 'Serv. Tecnolog. Avanzados - Pedro M. Baeza, '
```
pedrobaeza: Switch to `Tecnativa` both website and author.

partner_contact_job_position/models/res_partner.py
```diff
@@ -0,0 +1,39 @@
...
+
+class ResPartnerJobPosition(models.Model):
+    _name = "res.partner.job_position"
+    _order = "parent_left"
+    _parent_order = "name"
+    _parent_store = True
+    _description = "Job position"
+
+    name = fields.Char(required=True, translate=True)
+    parent_id = fields.Many2one(
```
pedrobaeza, line 27: Check recursion

pedrobaeza (changes requested):
Choose another icon different, because this one is used in partner_sector.

Squash "OCA transbot" and the "en.po" commits  together.

pedrobaeza:
cc @Tecnativa

partner_contact_job_position/models/res_partner.py
pedrobaeza: And add a test for that

pedrobaeza: Well, if you use something like https://github.com/odoo/odoo/blob/af865c2e9ea1dc1890ef33add9a6e4bc657068e0/addons/hr/hr.py#L264, there's no need for the test, as this method is already tested.

rafaelbn (approved):
Tested in runbot

yajo (approved):

pedrobaeza:
@cubells, please make the changes I requested.

pedrobaeza (approved):

PR #330, opened 2016-11-10
[9.0][partner_contact_in_several_companies] Several companies 90

richard-willdooit:


petrus-v (changes requested):
Can you give us more insight about your use case?

You should use the new api anyway according OCA guilde lines

richard-willdooit:
@petrus-v In version 9, there is a bug in the api inheritance with read, and it has to do with the fact that it can be passed an id or a list of ids.  Essentially, tne new api will always return a recordset and back translate it to a list result when downgrading.  However, some old api calls expect a list (when ids were passed) and some old api calls expect a single result.

Read is particularly vulnerable as it is called in both ways in different spots, and especially when the client has to resolve action ids, it frequently does so for a single id.

Hence, it is far simpler to just always use old api for "read" inheritance, otherwise you will need to create a double wrapper to ensure it always provides the correct result.

rafaelbn:
Hi @richard-willdooit , thanks for this contributions. We are reviewing all PR, please could make a review and fix travis errors? Thanks!

richard-willdooit:
@rafaelbn Up to date :)

rafaelbn:
Hi @richard-willdooit thanks! have you signed OCA's CLA?

rafaelbn:
@petrus-v could you please review again? thanks!

richard-willdooit:
@rafaelbn Yes - the signed document was linked to my old username.  It has been updated in the meantime.

rafaelbn:
OK, thanks! @pedrobaeza is CLA OK?

In the other hand, we need @petrus-v review again

In my side 👍

pedrobaeza:
CLA OK

antoniocanovas:
Testing in #ocacodesprint ...
I don't understand this topics:
- When I add a new position on a contact there is no field to assign this second company.
- I was waiting to see this contact in both companies, when I click on each one.

richard-willdooit:
@antoniocanovas It does not present all the companies for the contact, but rather allows the contact to be represented in multiple companies.

rafaelbn (changes requested):
After testing long time... IMHO this module is not well described or it is not intuitive, I don't understand how to use it and I cannot reproduce the functionality for what it was done.

The module description says:

> Partner Contact in Several Companies
> This module extends the contact management functionality. It allows one contact to have several job positions in different companies.

As @richard-willdooit said in https://github.com/OCA/partner-contact/pull/330#issuecomment-305036619
> @antoniocanovas It does not present all the companies for the contact, but rather allows the contact to be represented in multiple companies.

How to?



Please reconsider a contribution writing down a detailed "Usage" section, then other users could try or test it.

Thanks

richard-willdooit:
@rafaelbn Can I just hightlight that this PR is not to merge the module.  The module is already in the repository.  This PR is not about functionality.  It is fixing a bug between old and new api when some calls to action expect a dictionary and others expect a list of dictionaries depending if the ids are a list or integer.

While I am in agreement that it is "not well described or it is not intuitive", it does work.  It was available in 8 and our client needed it to be available in 9, so I upgraded it.  Perhaps it can be re-thought for 10 and redone for 10.

yajo (approved):

rafaelbn (approved):
AS commented in https://github.com/OCA/partner-contact/pull/330#issuecomment-305359521 this PR is a FIX and improved TEST are maken.

PR #315, opened 2016-10-16
[MIG] Migration partner_contact_birthdate partner_contact_gender part…

cvinh:
…ner_contact_personal_information_page to v10

PR #331, opened 2016-11-12
[MIG][10.0] partner_contact_birthdate

lasley:
WIP - not quite sure if this works yet, some of the naming conflicted with vertical-medical cols so my local may have been dirty. I don't think the original birthdate exists in partner anymore.

* Change openerp to odoo
* Update versions
* Remove api.one


lasley:
Oops my bad I didn't see #315 - I'll review instead

Issue #335, opened 2016-11-16
Nutrition information


max3903:
# partner_contact_height

Height: float
Height UOM: many2one with uom in the Distance category

# partner_contact_weight

Weight: float
Weight UOM: many2one with uom in the Weight category

# partner_contact_nutrition

Add a "Nutrition" tab on the partner form view. Only visible if the partner is an individual.

Calorie: float
Calorie UOM: many2one with uom
Fat: float
Fat UOM: many2one with uom
Protein: float
Protein UOM: many2one with uom
Carbohydrate: float
Carbohydrate UOM: many2one with uom

# partner_contact_nutrition_exclusion

Exclusions: many2many with product.product

# partner_contact_nutrition_allergen

Allergens: many2many with product.product

# partner_contact_nutrition_goal

Nutrition goal: selection between 'Lose weight', 'Maintain weight', and 'Gain weight'

# partner_contact_nutrition_activity_level

Activity level: selection between Sedentary, Moderately Active, Very Active, Extremely Active

# partner_contact_nutrition_diet

Diet: many2one with nutrition_diet. 
Nutrition_diet is only an object with a name field.  It can be extended later to have a list of inclusions or exclusions.
The module would provide the following data: Lifestyle, Carnivore, Pescatarian, Vegetarian.

pedrobaeza:
You already have partner_gender for the first one

max3903:
@pedrobaeza Good catch. Thanks! I just removed it from this issue.

lasley:
Maybe instead of a Medical tab, just add into the `Personal Information` tab that the gender and birthdate go under:

![image](https://cloud.githubusercontent.com/assets/7397578/20365352/7906f960-abfb-11e6-9610-272d75422711.png)

We just finished completely removing the Medical tab from vertical medical 😆 . It would conflict in v8 and 9 though.

max3903:
@lasley Done. Thanks!

lasley:
Question on the allergies. Given the context, it seems like maybe we want the term `allergen` instead?

Looking at the nutritional goals, it would be interesting if the `weight` was tracked over time instead of simple float. Maybe I'm getting too far ahead though.

max3903:
@lasley My scope does not include the history, but it would be easy for another module to make this field get the last value from the weight history. Same with height if you want to track it for children.

NL66278:
[...]

PR #340, opened 2016-11-18
[ADD] partner_contact_nutrition_allergen

max3903:
#335 

Depends on #337


partner_contact_nutrition_allergen/models/res_partner.py
```diff
@@ -0,0 +1,18 @@
...
+from openerp import models, fields
+
+
+class ResPartner(models.Model):
+    _inherit = 'res.partner'
+
+    nutrition_allergen_ids = fields.Many2many('product.product',
+                                              'res_partner_allergen_rel',
+                                              'partner_id',
+                                              'product_id',
```
lasley: Any reason for explicitly defining these? I could be wrong, but I think the guidelines say it's best to leave this automatic unless there's an issue.

lasley (approved):
One comment, non-blocking

susport (approved):
👍

PR #342, opened 2016-11-18
[ADD] partner_contact_nutrition_activity_level

max3903:
#335 

Depends on #337


lasley (approved):
LGTM thanks @max3903

susport (commented):
This is 👍  But I would like the activity level to be a Many2one field to a model example ActivityLevel with this default data already loaded. Thank you.

max3903:
@susport Can we keep that for version 2 of the module? I have run out of time.

susport:
@max3903 yes no problem, it was only a suggestion that can be taken in v2.

susport (approved):
👍

max3903:
#346

PR #343, opened 2016-11-18
[10.0] port base_vat_sanitized

alexis-via:


astirpe (approved):

base_vat_sanitized/models/res_partner.py
```diff
@@ -0,0 +1,25 @@
...
+# @author Alexis de Lattre <alexis.delattre@akretion.com>
+
+from odoo import models, fields, api
+import re
+
+
+class ResPartner(models.Model):
+    _inherit = 'res.partner'
+
+    sanitized_vat = fields.Char(
```
yajo, line 13: I think it would make more sense to sanitize the main VAT field instead of having a hidden sanitized one.

yajo (changes requested):
Please also squash all translation-related commits.

base_vat_sanitized/models/res_partner.py
alexis-via: Some users may want to keep the "spacing" of the VAT number. The approach is exactly the same as the field "sanitized_acc_number" of the "base" module.

yajo: OK then just the squashing please.

pedrobaeza (approved):
Can you please squash the same way as I have done on v9?

alexis-via:
Commits squashed.

lmignon (approved):

yajo (approved):

yajo:
Shouldn't this new field be exposed somewhere to the user?

tarteo (approved):

tarteo:
@Yajo It's meant to be a technical module used to make it easier for other modules to search for partners on vat number. I don't know if it's handy for the user to see this field.

pedrobaeza:
I think we can merge then.

PR #345, opened 2016-11-22
[FIX] domain

max3903:
#335

### DIFF
--- a/partner_contact_height/models/res_partner.py
+++ b/partner_contact_height/models/res_partner.py
@@ -11,6 +11,5 @@ class ResPartner(models.Model):
 
     height = fields.Float("Height")
     height_uom = fields.Many2one("product.uom", "Height UoM",
-                                 domain="[('category_id', '=', "
-                                        "self.env.ref('product.\
-                                        uom_categ_length').id)]")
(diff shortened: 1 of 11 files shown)


partner_contact_height/models/res_partner.py
```diff
@@ -11,6 +11,5 @@ class ResPartner(models.Model):
 
     height = fields.Float("Height")
     height_uom = fields.Many2one("product.uom", "Height UoM",
-                                 domain="[('category_id', '=', "
-                                        "self.env.ref('product.\
-                                        uom_categ_length').id)]")
+                                 domain=lambda self: [('category_id', '=', self.env.ref('product.uom_categ_length').id)] # noqa
```
lasley: Why disabling QA?

lasley:
```
    height_uom = fields.Many2one(
        "product.uom", 
        "Height UoM",
        domain=lambda s: [
            ('category_id', '=', s.env.ref('product.uom_categ_length').id),
        ],
    )
```

susport (commented):
👍  But please correct travis error.

PR #355, opened 2016-12-05
[PORT] Change installable to True and modify parent of Localization menu

open-net-sarl:
The main goal is to migrate l10n_ch_zip and l10n_ch_states but before we need to migrate this module.


pedrobaeza:
There's already a previous PR for migrating this module: https://github.com/OCA/partner-contact/pull/317

PR #358, opened 2016-12-15
[9.0] [MIG] base_partner_sequence module

cubells:
- Updated README
- Updated views and sequence

cc @Tecnativa


rafaelbn (approved):
Runbot is Red, not sure why... because I 've tested in runbot and functionally is OK, thanks!

base_partner_sequence/__openerp__.py
```diff
@@ -1,54 +1,22 @@
...
-    "author": "Tiny/initOS GmbH & Co. KG,Odoo Community Association (OCA)",
+    "version": "9.0.0.1.0",
+    "author": "Tiny/initOS GmbH & Co. KG,"
+              "ACSONE SA/NV,"
+              "Tecnativa, "
+              "Odoo Community Association (OCA)",
     "category": "Generic Modules/Base",
-    "website": "http://www.initos.com",
-    "depends": ["base"],
+    "website": "http://www.tecnativa.com",
```
pedrobaeza: Don't change URL

base_partner_sequence/models/partner.py
```diff
@@ -0,0 +1,72 @@
...
+        """
+        Checks whether a sequence value should be assigned to a partner's 'ref'
+
+        :param parnter_id: id of the partner object
+        :param vals: known field values of the partner object
+        :return: true iff a sequence value should be assigned to the\
+                      partner's 'ref'
+        """
+        if not vals and not partner_id:
+            raise Exception('Either field values or an id must be provided.')
```
pedrobaeza: Use Odoo exception

base_partner_sequence/views/partner_view.xml
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<odoo>
+
+    <record id="view_partner_form" model="ir.ui.view">
+        <field name="name">Make partner reference readonly when non-commercial</field>
+        <field name="model">res.partner</field>
+        <field name="inherit_id" ref="base.view_partner_form"/>
+        <field name="arch" type="xml">
+            <field name="ref" position="attributes">
+                <attribute name="attrs">{
```
pedrobaeza: Put this in only line for not messing up architecture view in Odoo

pedrobaeza (changes requested):
Please add a test for creating a contact to see which reference has (it has to be the same as the parent, and the sequence mustn't increase).

pedrobaeza:
There's also a strange repetition of commits that you should put together

base_partner_sequence/models/partner.py
```diff
@@ -0,0 +1,72 @@
...
+    @api.multi
+    def write(self, vals):
+        for partner in self:
+            ref = vals.get('ref') if 'ref' in vals else partner.ref
+            if not ref and self._needsRef(partner.id, vals):
+                vals['ref'] = self._get_next_ref(partner, vals)
+            super(ResPartner, partner).write(vals)
+        return True
+
+    @api.one
```
pedrobaeza: Don't use @api.one here.

base_partner_sequence/models/partner.py
```diff
@@ -0,0 +1,74 @@
...
+    _inherit = 'res.partner'
+
+    @api.model
+    def _needsRef(self, partner_id=None, vals=None):
+        """
+        Checks whether a sequence value should be assigned to a partner's 'ref'
+
+        :param parnter_id: id of the partner object
+        :param vals: known field values of the partner object
+        :return: true iff a sequence value should be assigned to the\
```
gdgellatly: typo - partner
typo - non blocking
typo - unnecessary trailing slash

base_partner_sequence/__openerp__.py
```diff
@@ -1,54 +1,22 @@
...
-    "version": "8.0.1.1.0",
-    "author": "Tiny/initOS GmbH & Co. KG,Odoo Community Association (OCA)",
+    "version": "9.0.0.1.0",
+    "author": "Tiny/initOS GmbH & Co. KG,"
+              "ACSONE SA/NV,"
+              "Tecnativa, "
+              "Odoo Community Association (OCA)",
     "category": "Generic Modules/Base",
     "website": "http://www.initos.com",
-    "depends": ["base"],
```
gdgellatly, line 29: I know ordinarily there is an implicit dependency on base, but in this case, because the module is directly referencing base xml ids I think it would be better to explicitly state.  I've only done a static review and not tested, so don't know whether this would ever actually cause a problem on a fresh database.

gdgellatly (approved):
My comments are non-blocking.  I don't know the answer to the base dependency question. I imagine the code is identical for 10.0, are we going to forward port when this gets approved?

base_partner_sequence/__openerp__.py
gdgellatly: actually, having tested in runbot, it seems that on a fresh install it will pick up base xml ids so ignore this comment if you choose.

pedrobaeza (approved):
I have amended last Graeme's comments and this is ready to be merged when CIs finish

cubells:
Tanks @pedrobaeza

PR #362, opened 2016-12-20
[FIX] partner_firstname: fix for address 'type' 

mmuch:
[IMP] partner_firstname: name, firstname and lastname are only required if address field 'type' is of type contact.

In many cases addresses don't necessarily have a first- or last-name. 
For example in the case of delivery addresses, required fields are not needed.

### DIFF
--- a/partner_firstname/views/res_partner.xml
+++ b/partner_firstname/views/res_partner.xml
@@ -19,10 +19,12 @@
                 <group attrs="{'invisible': [('is_company', '=', True)]}">
                     <field name="lastname" attrs=
                         "{'required': [('firstname', '=', False),
-                                       ('is_company', '=', False)]}"/>
+                                       ('is_company', '=', False),
+                                       ('type', '=', 'contact')]}"/>
                     <field name="firstname" attrs=
                         "{'required': [('lastname', '=', False),
-                                       ('is_company', '=', False)]}"/>
+                                       ('is_company', '=', False),
+                                       ('type', '=', 'contact')]}"/>
                 </group>
             </xpath>
         </data>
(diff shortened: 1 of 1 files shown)


Armitxes:
A much needed improvement :+1:

rafaelbn:
Hi, thanks for this contribution. The only required field is 'lastname' not 'firstname'. It's true that module show both as required (purple) but you onyl have to field one of them. Functionally I cannot reproduce the BUG or problem that you are having. Could you please give me the steps to reproduce it in order to understand this patch please? Thanks again!

Armitxes:
Hello @rafaelbn,

the issue here is when you create a new address related to a company you got to select between "contact", "invoice address", "shipping address" and "other address". As you mentioned, the field last name is an always required field which makes sense for "contact" but not for any of the other options as you often will address the parent company itself (and take it's name only once, i.e. for the warehouses of your own company) and not a third-party / external contact.

Hope I could clarify the issue a bit.

rafaelbn:
I understand, but the IMP is not correct (introduce a BUG), because in Odoo, `name` is required and after this PR you allow to create contacts with empty `name`

This is a usability issue maybe. But this PR is not correct.

Armitxes:
> because in Odoo, name is required and after this PR you allow to create contacts with empty name

That's not quite correct. Actually we're only bringing back the Odoo standard which had its sense. Without this module "name" isn't a mandatory field except for the type 'contact' (see the modificator on the screenshot).

It's the module "partner_firstname" that currently breaks the odoo standard by making the field "name" always required (via lastname) so this this would rather be a [FIX] for my understanding.

bildschirmfoto vom 2016-12-21 11 24 56

EDIT:
I can understand that you don't want a blank contact box in the parent but for that issue another solution has to be found if the address & name isn't filled. It's also quite questionable why u'd add an address without providing an address so you may rather just make the address fields required (type 'contact' excepted) ;)

On top "partner_firstname" with it's mandatory "lastname" breaks another feature where Odoo sets the (translated) type automatically behind the parent name in listviews if no name is provided. This also would be fixed [FIX] with this PR.

@mmuch maybe we can change this to a [FIX], can you add your thoughts?

rafaelbn (approved):
OK, tested again 👍 . Thanks for explanations. Please @Yajo @pedrobaeza take a look here.

yajo (approved):
This comes from the v8 days, where name was always required. Good improvement, thanks 👍

pedrobaeza (approved):

PR #369, opened 2017-01-11
[ADD] partner_password_reset: Create module

tedsalmon:
* Create module that allows you to send the password reset emails from the Partner view to any associated `res.users`

cc @lasley


rousseldenis (commented):
I don't know if I like this because action server is not based on groups, so this action is accessible to everyone... Is this wanted ?

lasley:
@rousseldenis - Actually yeah this was intended. Generating a password reset email actually doesn't require admin permissions. Locking it behind those permissions makes it so that when a customer requests a password reset, an admin has to do it, or the customer has to be walked through the process of doing it themselves. While self-help is obviously ideal, most customers in pharmacy (old people) absolutely refuse to do so.

@tedsalmon That said, I think this should be modularized a bit. While this code works fine when you're making a feature for a customer on the fly, this is totally unrefined right now. I don't think a server action is a good solution for a permanent module.

IMO we should inject the button using proper xpath & use real model methods to perform this.

On the note of security too, we should definitely only show this to employees. Portal users should not be able to generate a reset this way. 

We also need to make sure to not expose the password reset link on the partner page, because seeing that should definitely require admin privileges. I haven't looked whether this is being done or not currently, just a note.

tedsalmon:
Hey @lasley  / @rousseldenis,

I've converted this into a proper action inside a module. Let me know if there's anything else required.

Thanks!

partner_password_reset/models/res_partner.py
```diff
@@ -0,0 +1,14 @@
...
+from openerp import api, models
+
+
+class ResPartner(models.Model):
+    _inherit = 'res.partner'
+
+    @api.multi
+    def action_reset_user_password(self):
+        for rec in self:
+            rec.user_ids.action_reset_password()
```
lasley: `user_ids` is not a partner attr

I'm curious how your test passes?

partner_password_reset/models/res_partner.py
```diff
@@ -0,0 +1,14 @@
...
+
+from openerp import api, models
+
+
+class ResPartner(models.Model):
+    _inherit = 'res.partner'
+
+    @api.multi
+    def action_reset_user_password(self):
+        for rec in self:
```
lasley: self.mapped('user_id').action_reset_password()

lasley (changes requested):

partner_password_reset/models/res_partner.py
tedsalmon: Yes it is, please see:

https://github.com/odoo/odoo/blob/9.0/openerp/addons/base/res/res_partner.py#L246

`user_id` is definitely not the droid I'm looking for:
https://github.com/odoo/odoo/blob/9.0/openerp/addons/base/res/res_partner.py#L198

😆

tedsalmon:
Thanks for the review! Comments have been responded or attended to :)

lasley:
I think maybe we still need some ACLs though, as mentioned in https://github.com/OCA/partner-contact/pull/369#issuecomment-274876333

> On the note of security too, we should definitely only show this to employees.

rousseldenis:
@lasley IMO, having base.group_user on button is sufficient. Agree ?

lasley:
@rousseldenis Totally agree, `base.group_user` is perfect for the intent here!

tedsalmon:
Cool, that's the group already assigned to the button :)!

lasley (approved):
Oh snap my bad, I dunno how i missed that! 

LGTM, thanks

rousseldenis (approved):
:+1:

hparfr:
Good work.
+1 for the feature available for backend users.

But the button is always visible, even if there is no user linked to the partner.

tedsalmon:
@hparfr,

Interesting, I'll take a look at the domain on the button and adjust as necessary.

Thanks!

lasley:
@tedsalmon while you're there can you please attend to @elicoidal comments in #383, then cross port all the things?

tedsalmon:
@hparfr,

That was a doozy! It should now be fixed though, so please take a look when you have a chance.

@lasley,

Comments in #383 have been attended to in this PR. I'll cross port now!

Thanks!

hparfr (approved):

tedsalmon:
Per https://github.com/OCA/partner-contact/pull/383#issuecomment-299045172, I have turned the functionality added in this module into a wizard. I'll squash commits once these changes are approved.

Thanks!

partner_password_reset/wizard/res_partner_password_reset_wizard.py
```diff
@@ -0,0 +1,43 @@
...
+                        submenu=False):
+        """ Override to check that there are associated users when called """
+        res = super(ResPartnerPasswordResetWizard, self).fields_view_get(
+            view_id=view_id, view_type=view_type, toolbar=toolbar,
+            submenu=False)
+        if not self._default_user_ids():
+            raise exceptions.Warning(
+                _('The selected partners have no associated portal users')
+            )
+        return res
```
lasley, line 37: This untested line makes me sad panda

partner_password_reset/wizard/res_partner_password_reset_wizard.py
```diff
@@ -0,0 +1,43 @@
...
+
+    @api.multi
+    def action_submit(self):
+        """ Reset the user passwords on submission """
+        for rec in self:
+            rec.user_ids.action_reset_password()
```
lasley: `self.mapped('user_ids').action_reset_password()` instead of record loop.

Also there's a panda here too

lasley (changes requested):

tedsalmon:
@lasley,

Thanks for the review - comments attended to.

lasley (approved):

chienandalu (approved):

PR #373, opened 2017-01-21
[IMP] partner_identification: Add context override

lasley:
* Allow for context override of validations using ``__id_no_validate__``

This was causing me some issues with an importer, where all data may not be good but is still required.

### DIFF
--- a/partner_identification/README.rst
+++ b/partner_identification/README.rst
@@ -36,7 +36,7 @@ Code:
   Code, abbreviation or acronym of this ID type. For example, 'driver_license'
 Python validation code:
   Optional python code called to validate ID numbers of this ID type. This functionality can be
-  overridden by setting ``__id_no_validate__`` to ``True`` in the context, such as:
+  overridden by setting ``id_no_validate`` to ``True`` in the context, such as:
 
   .. code-block:: python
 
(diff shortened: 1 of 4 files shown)


rousseldenis (approved):
:+1:  Code review

pedrobaeza:
You can put as context variable only `id_no_validate` for being shorter and clearer to pass the value I think.

lmignon:
@lasley I also prefer `id_no_validate` as key.   Apart from this remark your proposal makes sense.

lasley:
I prefer that naming as well. Done!

pedrobaeza (approved):

partner_identification/README.rst
```diff
@@ -35,8 +35,15 @@ Name:
 Code:
   Code, abbreviation or acronym of this ID type. For example, 'driver_license'
 Python validation code:
-  Optional python code called to validate ID numbers of this ID type.
+  Optional python code called to validate ID numbers of this ID type. This functionality can be
+  overridden by setting ``__id_no_validate__`` to ``True`` in the context, such as:
```
lmignon: @lasley  s/__ id_no_validate __/id_no_validate

lmignon (changes requested):

lasley:
Thanks for the catch @lmignon, ReadMe updated

lmignon (approved):
Thank you @lasley

PR #377, opened 2017-01-26
[ADD][10.0] Add partner_contact_configuration module

rousseldenis:


adrienpeiffer (approved):
:+1:

lmignon:
@rousseldenis Good Idea. Once merged it will be possible to remove the annoying dependency on sale_team in partner_indentification and ....

lmignon:
:+1:

lmignon (approved):

rousseldenis:
@lmignon That was the purpose, thanks!

partner_contact_configuration/README.rst
```diff
@@ -0,0 +1,57 @@
...
+
+
+Usage
+=====
+
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/134/10.0
+
+.. repo_id is available in https://github.com/OCA/maintainer-tools/blob/master/tools/repos_with_ids.txt
```
pedrobaeza: Remove this

pedrobaeza:
But depending on module _contacts_ is similar than depending on _sales_team_. The latter has also dependency on mail, but it's a dependency that you are going to install for sure, and I would even prefer it because I have only one app on the top menu: Sales, instead of Sales and Contacts. Any reason that I don't get?

rousseldenis:
@pedrobaeza In some cases, you don't want to have Sales menu. Moreover, your users don't have necessarily sales rights (that is mandatory to have menus in that case).
You want just to manage contacts ... so, this is the purpose of this module

pedrobaeza:
OK, got it. It would be good to have this explanation on the README too.

lmignon:
This one seems ready to merge...

pedrobaeza:
Here you have!

PR #387, opened 2017-02-13
[10.0][IMP][partner_external_map] Refactored as an AbstractModel so it can be used as a mixin.

george-mengelberg:
Reason: I wanted to use partner_external_map on other models with address related fields.

Note: This is my very first submission to any OSS project...so please let me know how I can do a better job in the future.

### DIFF
--- a/partner_external_map/models/res_partner.py
+++ b/partner_external_map/models/res_partner.py
@@ -116,5 +116,3 @@ class ExternalMap(models.AbstractModel):
 class ResPartner(models.Model):
     _name = 'res.partner'
     _inherit = ['res.partner', 'external.map']
-
-
--- a/partner_external_map/models/res_partner.py
+++ b/partner_external_map/models/res_partner.py
@@ -116,3 +116,4 @@ class ExternalMap(models.AbstractModel):
 class ResPartner(models.Model):
     _name = 'res.partner'
     _inherit = ['res.partner', 'external.map']
+
--- a/partner_external_map/models/res_partner.py
+++ b/partner_external_map/models/res_partner.py
@@ -11,8 +11,8 @@ import logging
 logger = logging.getLogger(__name__)
 
 
-class ResPartner(models.Model):
-    _inherit = 'res.partner'
+class ExternalMap(models.AbstractModel):
+    _name = 'external.map'
 
     @api.multi
     def _address_as_string(self):
@@ -26,6 +26,8 @@ class ResPartner(models.Model):
             addr.append(self.city)
         if self.state_id:
             addr.append(self.state_id.name)
+        if self.zip:
+            addr.append(self.zip)
         if self.country_id:
             addr.append(self.country_id.name)
         if not addr:
@@ -109,3 +111,8 @@ class ResPartner(models.Model):
             'url': url,
             'target': 'new',
         }
+
+
+class ResPartner(models.Model):
+    _name = 'res.partner'
+    _inherit = ['res.partner', 'external.map']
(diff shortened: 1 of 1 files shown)


george-mengelberg:
Hmmm...I am not sure how to fix the issues w/ codecov.

gurneyalex (approved):

partner_external_map/models/res_partner.py
```diff
@@ -26,6 +26,8 @@ def _address_as_string(self):
             addr.append(self.city)
         if self.state_id:
             addr.append(self.state_id.name)
```
lasley, line 28: These three conditionals are untested, which is why Codecov is griping. 

The solution here would be to
* update the tests to include the three attributes when creating [the test record](https://github.com/george-mengelberg/partner-contact/blob/5ccc17b06c4fc455142c0db4811e200f64ac2076/partner_external_map/tests/test_partner_external_map.py#L22). 
* update the [expected URL](https://github.com/george-mengelberg/partner-contact/blob/5ccc17b06c4fc455142c0db4811e200f64ac2076/partner_external_map/tests/test_partner_external_map.py#L47) to look how it would with the attributes

I won't block though, because the lines you are emulating are also untested. I would love it if you tested your part though, and bonus points if you add the street and street2 attributes at the same time to cover the other lacking coverage here that you didn't cause.

Tests are insanely important IMO, so well worth the effort.

lasley (approved):
Congrats on your first OSS contribution @george-mengelberg & thanks for the feature!

I commented inline regarding the CodeCov issue, which is due to adding untested code.

Tests are insanely important for the maintainability of code, so I recommend you add them & give a shout if you have any questions.

I won't block though, especially considering you're a new contributor, so if you'd prefer to just merge without the tests let me know & I will add them by submitting a PR to your branch before this is merged.

partner_external_map/models/res_partner.py
george-mengelberg: @lasley I will do it the correct way and update the tests and the expected URL.

pedrobaeza: Any news of this?

rafaelbn:
hi @george-mengelberg  could you please review/update last comments? We will do las review for merge. thanks!

rafaelbn:
Well @gurneyalex and @lasley approved it long time ago.. I merge

pedrobaeza:
There's still an unattended comment

ghost (approved):
Tested on Runbot

partner_external_map/models/res_partner.py
```diff
@@ -26,6 +26,8 @@ def _address_as_string(self):
...
+        if self.zip:
+            addr.append(self.zip)
```
ghost, line 30: I think this small addition is not related with the commit purpose. Maybe redescribe the commit.

pedrobaeza:
Please check the comment

partner_external_map/models/res_partner.py
lasley: IMO this should be squashed on merge, so the individual commit comments are fairly irrelevant.

NL66278:
@george-mengelberg Are you going to take up the last comments? I think the addition of the zip field to the address should not be in this PR.

george-mengelberg:
I am not actively developing on Odoo anymore so it would be best to remove
me as a POC and allow someone else to take over if they desire.

NL66278:
@george-mengelberg Sorry to hear that you are not working with Odoo anymore. Anyway success in whatever you are doing now, and thanks for your contribution!

PR #388, opened 2017-02-15
[8.0] fix partner merge

luc-demeyer:
partner merge  fails on migrated databases.
If a field no longer exists in  the python code but still in ir_model_fields, than the partner merge fails.

### DIFF
--- a/base_partner_merge/base_partner_merge.py
+++ b/base_partner_merge/base_partner_merge.py
@@ -295,7 +295,8 @@ class MergePartnerAutomatic(TransientModel):
             legacy = proxy_model._columns.get(record.name)
             field_spec = proxy_model._fields.get(record.name)
 
-            if isinstance(legacy, fields.function) or field_spec.compute:
+            if not legacy or isinstance(legacy, fields.function) \
+                    or field_spec.compute:
                 continue
 
             for partner in src_partners:


pedrobaeza:
But the problem seems to be more on the migration side. Purge that fields with database_cleanup or improve the migration script removing offending entries.

luc-demeyer:
@pedrobaeza 
I checked both databases upgraded with openupgrade and odoo migration service.
In both cases we have legacy fields remaining in the database (concrete case if 'copy_object' field on ir_actions_server that is renamed to 'ref_object'.
We can fix the openupgrade but we cannot make a PR on the Odoo migration scripts to fix this, hence I believe it's better to merge this PR.

pankk:
The change is simple and doesn't seem to break anything, so -if I understand correctly- this seems preferable to letting everyone that uses the Odoo migration service or OpenUpgrade run database_cleanup to remove those specific fields. I can see the benefit of having the database as clean as possible though, so perhaps there is a better unexplored alternative?

rafaelbn:
Hi @luc-demeyer please could you review travis errors? We are in #ocacodesprint we will review al PRs! thanks

luc-demeyer:
The Travis errors are in other modules (e.g. 'partner_identification') hence that should not be a reason to block the merge of this PR.

rafaelbn (approved):
OK

cubells (approved):
:+1: 

LGTM

PR #392, opened 2017-02-27
[FIX] partner_firstname: fix for address 'type'

mmuch:
[FIX] partner_firstname: name, firstname and lastname are only required if address field 'type' is of type contact.

In many cases addresses don't necessarily have a first- or last-name.
For example in the case of delivery addresses, required fields are not needed.
(A Warehouse doesn’t have a first- and/or lastname )

As https://github.com/OCA/partner-contact/pull/362 but for 10.0 branch.


pedrobaeza (approved):

pedrobaeza:
As this has been approved in previous branch, I can fast-track here.

PR #395, opened 2017-03-07
[10.0][MIG] Partner Contact Weight

woodbrettm:
Hi there,

This migrates partner_contact_weight to 10.0.

Thanks,
Brett


lasley (approved):
LGTM thanks @bmw95. Please squash the consecutive Transifexes into each other though please.

tarteo (approved):
LGTM. Thanks!

woodbrettm:
@lasley squashed

rafaelbn (approved):
Tested in runbot

dreispt:
I feel that this should also have added the height.
In fact, for health purposes, you height is important factor to consider when evaluating weight.

woodbrettm:
@dreispt what about something like a partner_contact_size module for that? You could have height, foot size, leg length, waist size, etc.

dreispt:
For some applications height weight and maybe IMC computation are enough.
Personally I' have use them as basic screening factors for applicants for security guard jobs.

lasley:
@dreispt - While I do agree that height is necessity, there is already [`partner_contact_height`](https://github.com/OCA/partner-contact/tree/9.0/partner_contact_height) for this. Seems better to leave as two independent modules IMO, even though I can't really think of an instance where one would be relevant without the other.

dreispt:
> even though I can't really think of an instance where one would be relevant without the other.

exactly my opinion 😉

lasley:
@max3903 - Any thoughts on merging these two modules (weight & height) in v11?

max3903:
@lasley Sure, why not.

PR #400, opened 2017-03-20
[FIX] Fix usage in Readme.md

hurrinico:
I have fixed readme.md file modifying path of the wizard button with the right one

### DIFF
--- a/base_location_geonames_import/README.rst
+++ b/base_location_geonames_import/README.rst
@@ -26,7 +26,7 @@ system parameter.
 Usage
 =====
 
-Go to *Sales > Configuration > Address book > Localization > Import from Geonames*,
+Go to *Settings > Technical > Cities/Locations Management > Import from Geonames*,
 and click on it to open a wizard.
 
 When you start the wizard, it will ask you to select a country. Then, for the


base_location_geonames_import/README.rst
```diff
@@ -26,7 +26,7 @@ system parameter.
 Usage
 =====
 
-Go to *Sales > Configuration > Address book > Localization > Import from Geonames*,
+Go to *Configurations > Technical > Cities/Locations Management > Import from Geonames*,
```
pedrobaeza: s/Configurations/Configuration

tarteo: Settings > Technical > Cities/Locations Management > Import from Geonames
Settings*

tarteo: Settings > Technical > Cities/Locations Management > Import from Geonames
Settings*

tarteo (changes requested):

base_location_geonames_import/README.rst
pedrobaeza: You are totally right! I was only correcting the put name without s, but it's not the English name.

hurrinico:
@pedrobaeza @tarteo Fixed

tafaRU (approved):

pedrobaeza (approved):

PR #271, opened 2016-05-06
[REF] requirements.txt: Add pip install from .travis.yml

moylop260:
Now MQT support `requirements.txt` https://github.com/OCA/maintainer-quality-tools/pull/311

PR #403, opened 2017-04-06
[9.0][REF] Install python requirements with req file and not travis

jjscarafia:


pedrobaeza:
I remember that it was put in that place outside requirements.txt because it was needed before, but @moylop260 can tell

moylop260 (approved):
This looks like as a cherry-pick from https://github.com/OCA/partner-contact/pull/271 so 👍

pedrobaeza (approved):
OK then

nicomacr (approved):

PR #405, opened 2017-04-15
[10.0][IMP] base_location(geonames): Add Lat/Long

lasley:
This PR contains two modules `base_location` and `base_location_geonames_importer`, but they are minor changes in both and entirely related:

[IMP] base_location: Add lat & long to `better.zip`
* Add latitude and longitude columns to `better.zip`

[IMP] base_location_geonames_import: Add lat/long
* Add support for latitude & longitude to genomes importer

I specifically did not use geoengine so as to not require PostGIS. We can create a glue module pretty easily I think, so we can have the best of both worlds (see https://github.com/OCA/geospatial/issues/136).

I also did not add the lat/long into any views, because I don't really feel they will be useful for humans. Please let me know if you feel otherwise.

### DIFF
--- a/base_location/models/better_zip.py
+++ b/base_location/models/better_zip.py
@@ -20,6 +20,8 @@ class BetterZip(models.Model):
     city = fields.Char('City', required=True)
     state_id = fields.Many2one('res.country.state', 'State')
     country_id = fields.Many2one('res.country', 'Country')
+    latitude = fields.Float()
+    longitude = fields.Float()
 
     @api.one
     @api.depends(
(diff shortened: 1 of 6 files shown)


lasley:
Assuming tests pass, can a PSC please mark as 10.0 milestone & Ready for Review please and thanks!

rvalyi:
:+1:  seems both useful and very simple

base_location/models/better_zip.py
```diff
@@ -20,6 +20,8 @@ class BetterZip(models.Model):
     city = fields.Char('City', required=True)
     state_id = fields.Many2one('res.country.state', 'State')
     country_id = fields.Many2one('res.country', 'Country')
+    latitude = fields.Float()
+    longitude = fields.Float()
```
lasley, line 25: Note that there is also an `accuracy` in GeoNames that I don't personally care about but would have no problems implementing if someone does.
```
accuracy          : accuracy of lat/lng from 1=estimated to 6=centroid
```

elicoidal (approved):

lasley:
Good for merge?

rafaelbn:
AS @rvalyi approve it also I merge https://github.com/OCA/partner-contact/pull/405#issuecomment-294266956

PR #406, opened 2017-04-18
[WIP][BACKPORT][8.0] Backport of base_partner_merge from 9.0, and add cron features

thomaspaulb:
This has become a project of its own, probably have to resubmit this properly after we get it working. Besides backporting the module as it is in 9.0, we also try to add a few features such as specifying a search domain for the records to deduplicate, initiating deduplication from a cron job, and deduplication on the "ref" field.


thomaspaulb:
@hbrunn Continuing the discussion from here: https://github.com/sunflowerit/partner-contact/pull/1

thomaspaulb:
Something I can't wrap my head around just yet, is how to solve the fact that when `base_partner_merge` is already installed, and you want to install `crm` also, the following is raised:

`Cannot create a consistent method resolution order (MRO) for bases ResPartner, res.partner`

For now am working around it by deinstalling `base_partner_merge` and then installing `crm` before reinstalling `base_partner_merge` again, but later should come back to this.

thomaspaulb:
I achieved to delete only the "double" models, but now have run into another issue:

```
  File "/home/antiflu/Code/buildout/vwnoost/parts/odoo/openerp/modules/registry.py", line 102, in __getitem__
    return self.models[model_name]
KeyError: 'base.partner.merge.automatic.wizard'
```

I think this is because `crm` is loaded after `base_partner_merge`, so the wizard model we need is just not there yet to inherit from. But we also cannot add `crm` to the dependencies to force this one to be loaded after crm. What can we do?

hbrunn:
what's the full traceback? This looks like you either delete too many models, or some code tries to access it too early

hbrunn:
concerning https://github.com/sunflowerit/partner-contact/pull/1#issuecomment-302431568: this is the other way around which I didn't consider. I think we can solve this by intercepting https://github.com/OCA/OCB/blob/8.0/openerp/addons/base/module/module.py#L411 and do something very similar as in the register hook

cubells:
@thomaspaulb  please review travis errors

hbrunn:
I hijacked this in #441

PR #415, opened 2017-04-27
[8.0] fix partner_relations addit'l fields

luc-demeyer:
The use of the _additional_view_fields private attribute for adding new fields fails in the following scenario:

- odoo 8.0 running without workers
- db1 with 'partner_relations' module installed
- db1 has also a 'custom_partner_relations' module adding fields as documented in the docstring (via the autoinit)

As soon as we create a second database db2 on this odoo server and we try to install the partner_relations module (NOT the 'custom_partner_relations' module, the module install fails since  the _additional_view_fields added in the 'custom_partner_relations' module are added to the postgres view creation of db2, but since those fields are not defined the module install fails.

By changing from a private attribute to a method (cf. diff infra) this issue is solved.

### DIFF
--- a/partner_relations/__openerp__.py
+++ b/partner_relations/__openerp__.py
@@ -3,7 +3,7 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 {
     "name": "Partner relations",
-    "version": "8.0.1.1.1",
+    "version": "8.0.1.1.2",
     "author": "Therp BV,Camptocamp,Odoo Community Association (OCA)",
     "complexity": "normal",
     "category": "Customer Relationship Management",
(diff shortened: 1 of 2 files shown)


rafaelbn:
Hi @luc-demeyer , could you review travis errors plesase? Thanks!

luc-demeyer:
It's unclear to me what's wrong with travis but I don't think there is a relation with the patch I made (which I tested and deployed at customer site in the meanwhile).
IMHO we should not block fixes in case of travis 'failures' which are unrelated to the fix.
In parallel, we should of course have a look to why travis is red and fix the travis config or other modules causing travis to turn red.

cubells (approved):
:+1: 
LGTM

cubells:
Travis is ok.

rafaelbn (approved):
I agree checked: https://travis-ci.org/OCA/partner-contact/jobs/226429632 so 👍

PR #420, opened 2017-05-21
[8.0][IMP][partner_contact_birthdate] feat(res.partner): add field birthdate_month

agb80:
Add field birtdate_month that allow create a new filter to display only res.partners which birthdate

is on current month.

### DIFF
--- a/partner_contact_birthdate/models.py
+++ b/partner_contact_birthdate/models.py
@@ -16,6 +16,8 @@
 # You should have received a copy of the GNU Affero General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+from datetime import datetime
+
 from openerp import _, api, fields, models
 import logging
 
@@ -29,6 +31,10 @@ class Partner(models.Model):
 
     # New birthdate field in date format
     birthdate_date = fields.Date("Birthdate")
+    birthdate_month = fields.Char(
+        'Birthdate month', compute='_birthdate_compute', store=True,
+        help='Helper field to get current month birthdate partners',
+    )
 
     # Make the old Char field to reflect the new Date field
     birthdate = fields.Char(
@@ -41,6 +47,10 @@ class Partner(models.Model):
     def _birthdate_compute(self):
         """Store a string of the new date in the old field."""
         self.birthdate = self.birthdate_date
+        if self.birthdate_date:
+            self.birthdate_month = datetime.strptime(
+                self.birthdate_date, '%Y-%m-%d',
+            ).strftime('%m')
 
     @api.one
     def _birthdate_inverse(self):
--- a/partner_contact_birthdate/views/res_partner.xml
+++ b/partner_contact_birthdate/views/res_partner.xml
@@ -15,5 +15,17 @@
   </field>
 </record>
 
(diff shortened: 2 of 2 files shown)


antoniocanovas:
Hi @rafaelbn I see two entries in advanced search called "Birthday" on runbot.
The filter for "birthday this month" works fine.

antoniocanovas:
@rafaelbn , I go on testing...
The advanced filter for "Birthday month" works with numbers.
It would be fine to can search by the name of the month too, or advise to use only numbers.

aming27:
I test the functionability and works fine.

chienandalu:
Is this PR too specific in its purpose?

agb80:
@chienandalu the purpose for this PR is being able to get a list for customers whose birthday is this month for being able to send them promotions. 

@antoniocanovas the two birthdate fields are already present on partner_bitthday module before this PR, didn't change it because I think will be out of scope and will need an independent PR if needed.

AaronHForgeFlow (commented):
It works fine. However, In my opinion it'd be more useful if the system tell us in advance when is the birthday. With this approach it would be too late for the contacts whose birthday falls on the first day of the month.

rafaelbn:
Hi @agb80 , sorry but in this case this module makes something very specific. This is not a general improvement for OCA. Furthermore you don't need to add this field yo get this information, you just have yo add and option with a "Group by" date in the field birthday to get the same information. Or just you can add that filter of "This month birthdays" using same field birthday without adding a new fields... It's just a formula... I'm closing.

rafaelbn:
Hi  @agb80 , after a discussion here in the Code Sprint we have reconsider you PR.

Please add a group by also, it would be useful to group by month too.

Thanks

partner_contact_birthdate/models.py
```diff
@@ -41,6 +47,10 @@ class Partner(models.Model):
     def _birthdate_compute(self):
         """Store a string of the new date in the old field."""
         self.birthdate = self.birthdate_date
+        if self.birthdate_date:
+            self.birthdate_month = datetime.strptime(
```
lasley, line 51: `fields.Datetime.from_string`

partner_contact_birthdate/views/res_partner.xml
```diff
@@ -15,5 +15,17 @@
...
 </record>
 
+<record id="res_partner_view_search" model="ir.ui.view">
+    <field name="name">res.partner.view.search</field>
+    <field name="model">res.partner</field>
+    <field name="inherit_id" ref="base.view_res_partner_filter"/>
+    <field name="arch" type="xml">
+        <filter string="Suppliers" position="after">
+            <separator/>
+            <filter string="This month birthdates" name="this_month_birthdates" domain="[('birthdate_month','=', context_today().strftime('%%m'))]"/>
```
lasley, line 25: s/This month birthdates/Birthdays This Month

lasley (changes requested):

NL66278:
This seems very specific to me and not a general improvement. Anyway, would you take up the change request and resolve the conflicts?

NL66278:
Branch contains conflicts and question is not taken up.

PR #391, opened 2017-02-25
[BACKPORT][8.0] partner_sale_risk

andhit-r:

PR #449, opened 2017-07-21
[FIX][8.0] partner_financial_risk: credit_limit field is not correctly updated.

Daniel-CA:
The credit_limit field already exist in the partner form view of the account module, should be replaced to avoid incorrect behavior.

When merged #391 could be merged too.

### DIFF
--- a/partner_financial_risk/views/res_partner_view.xml
+++ b/partner_financial_risk/views/res_partner_view.xml
@@ -10,6 +10,7 @@
     <field name="priority" eval="20"/>
     <field name="groups_id" eval="[(4, ref('base.group_sale_salesman')), (4, ref('account.group_account_manager'))]"/>
     <field name="arch" type="xml">
+        <field name="credit_limit" position="replace"/>
         <page name="sales_purchases" position="after">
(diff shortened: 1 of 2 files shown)


partner_financial_risk/views/res_partner_view.xml
```diff
@@ -56,5 +56,13 @@
         </page>
     </field>
 </record>
+    <record id="view_partner_property_form" model="ir.ui.view">
```
pedrobaeza: Don't create other view, but put it on the same.

Daniel-CA: Hi @pedrobaeza, you mean leave the credit_limit field in the account view as it is or I should add the new attributes of the field?

pedrobaeza: No, I mean to include this replace in the previous view, before redeclaring it.

Daniel-CA: oh, ok!

Daniel-CA:
Done.

pedrobaeza (approved):

yajo (approved):
Quite obvious.

andhit-r (approved):

Issue #439, opened 2017-06-23
[9.0] Uncaught Error: Unknow field parent_id in domain [["type","=","contact"],["parent_id","!=",false]]


VisiionSoluciones:
Hi,
I have this error when I try to create a new contact inside a partner.

I think the problem may be in the following file:

base_location/views/partner_view.xml

in attrs values

StefanRijnhart:
Solved in https://github.com/OCA/partner-contact/commit/3496517550ee1b

PR #450, opened 2017-07-24
[FIX] Error displayed when create a contact in existing partner

VisiionSoluciones:
[FIX] Error displayed when create a contact in existing partner


StefanRijnhart:
@VisiionSoluciones what is the error, and how can I reproduce? I tried clicking around on the 9.0 runbot but I was able to create contacts on existing partners without a problem.

StefanRijnhart:
I found the related issue: https://github.com/OCA/partner-contact/issues/439. The code conflict between the PR branch and the main branch pointed me to https://github.com/OCA/partner-contact/commit/3496517550ee1b, which has solved the problem.

PR #460, opened 2017-08-25
[10.0][IMP] partner_firstname: Add indices

lasley:
* Add index to first and last name in order to allow for quicker searches - fixing an incredibly slow update time with large dbs


lmignon (approved):

yajo (approved):
I guess that when you say "update time" you refer to "view update time", not "record update time", because remember that indexes speed up selects but slow down slightly inserts and updates (I mean SQL queries).

In any case I still think this is 👍

pedrobaeza (changes requested):
Change version number

lasley:
> I guess that when you say "update time" you refer to "view update time", not "record update time", because remember that indexes speed up selects but slow down slightly inserts and updates (I mean SQL queries).

Sorry I should have been more clear - when you update this module on a large database, it takes forever & that seems to stem from the lack of indices on the first/last name - there is a search in the installation method.

Now that I'm looking further, this installation method should be noupdate. I'll add a new commit bumping the version and changing the [datafile](https://github.com/LasLabs/partner-contact/blob/f0606006b5b652063a6a99956fc5dee54ea1e6f0/partner_firstname/data/res_partner.yml) to xml with a noupdate

partner_firstname/__manifest__.py
```diff
@@ -7,7 +7,7 @@
 {
     'name': 'Partner first name and last name',
     'summary': "Split first name and last name for non company partners",
-    'version': '10.0.2.0.0',
+    'version': '10.0.2.0.1',
```
pedrobaeza: Version should be 10.0.2.1.0

pedrobaeza (changes requested):

partner_firstname/__manifest__.py
lasley: Hmmmm I would consider this more of a fix, but no prob (amended)

pedrobaeza: But it requires an update, isn't it?

lmignon:
@lasley If the function must be run only at install IMO it would be better to call the method from a `post_init_hook` https://github.com/OCA/maintainer-tools/blob/master/template/module/__openerp__.py#L15 and remove the call from the xml file.

lmignon (changes requested):
@lasley If the function must be run only at install IMO it would be better to call the method from a post_init_hook https://github.com/OCA/maintainer-tools/blob/master/template/module/__openerp__.py#L15 and remove the call from the xml file.

yajo:
Hmm yes, in such case I agree with @lmignon

lasley:
Good call, thanks @lmignon!

yajo:
Thanks @lasley; @lmignon and @pedrobaeza, could you update your reviews please? This seems ready to merge.

lmignon (approved):
@lasley Thank you!

njeudy (approved):
Code check and test

pedrobaeza:
Please squash at your taste

lasley:
Squashed! I left two independent commits in this case because IMO both add value as a change log (indices and hook-based install)

PR #464, opened 2017-08-30
[ADD] new module for partner business id

jarmokortetjarvi:
Business id is a necessity for b2b sales in some countries.
This field has a slightly different function than fields "company_registry" or "vat"


pedrobaeza:
I think you should use https://github.com/OCA/partner-contact/tree/9.0/partner_identification instead.

jarmokortetjarvi:
Closing this for the moment. The discussion about the module/field in https://github.com/OCA/partner-contact/pull/465

PR #465, opened 2017-08-30
[ADD] new module for partner business id

jarmokortetjarvi:
Business id is a necessity for b2b sales in some countries.
This field has a slightly different function than fields "company_registry" or "vat"


pedrobaeza:
I think you should use https://github.com/OCA/partner-contact/tree/10.0/partner_identification instead.

jarmokortetjarvi:
We actually checked into that module when we were having a debate about this issue within the Finnish Odoo community earlier this year.

While it wouldn't be impossible to use the partner_identification module, it wouldn't be very practical.

At least in Finland and Sweden the business id is basically as important as the VAT-number, and the VAT-number is actually derived from business id. All companies don't even have a VAT-number, if they aren't VAT registered, so we can't reliably get the business id from the VAT-number

Using partner_identification module for storing business ids would mean writing some custom constraints and checks for the records of certain category  saved (a company can't have but one business id, and it has to match a certain format).

Several Finnish companies have created their own business id module, which is a bit problematic when trying to make generic modules that require that field.

The decision to try to get the module to partner-contact instead of l10n-finland is the fact that the same problem exists in Sweden too.

pedrobaeza:
One question: the business ID is the same for both Finland and Sweden? That's to say: a Finnish supplier that is registered in a Swedish Odoo company, have the same business ID that if you register that supplier in a Finnish Odoo company?

If not, my advice would be:

* Create a module per localization depending on partner_identification.
* Each module adds as data an identification type called "Finnish business ID"/"Swedish business ID".
* Add a computed filed at res.partner level that checks in the one2many the line with that type and make it accessible at partner level.

This can also apply although the number is the same for both countries.

mlaitinen:
The business ID is for domestic use only and the format differs between Finland and Sweden.

I understand your point of view, but depending on `partner_identification` feels pretty complex and encourages people to continue making non-shared and non-standard `business_id` implementations.

Maybe we are over-generalizing this? Man I wish that business ID would be as important in Belgium as it is in Finland and Sweden. We'd already have a `res.partner.business_id` field in Odoo core.

pedrobaeza:
Yeah, but then the recommendation is to have one module per localization with only that field and call it "Finnish business ID" and "Swedish business ID"...

pedrobaeza:
Because having only one module for both IDs will make that a multi-company Odoo with both Finnish and Swedish companies and shared partners won't be able to put correctly that data.

mlaitinen:
In my eyes this module is multi-company compatible. If a multi-company Odoo instance has one company for Sweden and one for Finland, they only need one `business_id` field, even when there are shared partners. A Finnish company can only have a Finnish business ID and a Swedish company can only have a Swedish business ID.

pedrobaeza:
But not being company-dependent field will make that the number will be shared also. Any way, I see this module very specific to be put in this crowded repository. Let's see what others think.

mlaitinen:
There's no harm involved in sharing the business ID across different companies. It behaves just like `vat` field and `vat` isn't company-dependent either.

jarmokortetjarvi:
The business id functions pretty much exactly the same way in domestic business as the VAT code does in business done within EU. Not being company-dependent is more pro than con.

We have actually been using a module like this for years in multi-company setups.

I initially thought about changing the logic into using `partner_identification` as base, but it would make the module a lot more complicated, as we would have to add
- A new `res.partner.id_category` for each country using the business id
- A computed field for `business_id`
- Getter and setter to handle the the basic usage of a/m field to utilize `partner_identification`

Later for validation (not in current PR)
- Business id validation that is dependent on `res.partner.id_category` and partner `country_id`
- (Constraint to disallow having multiple companies with same business id)

The current PR will just add the field for business id, but it will also need a country-dependent validation
https://github.com/Tawasta/partner-contact/pull/1/commits/84118d8c123ccbbd7c6c9fb3274f88afecaf6c61#diff-832e566ba716868d85201853cda2e086

pedrobaeza:
partner_identification already includes a mechanism for embedding validation code, so I think is the way to go. Please read again https://github.com/OCA/partner-contact/pull/465#issuecomment-326009579

jarmokortetjarvi:
Understood. We'll re-visit using `partner_identification` as a base for this functionality

jarmokortetjarvi:
Re-done this for `l10n-finland` using `partner_identification` as base. Closing this PR

PR #472, opened 2017-09-12
[MIG] partner_contact_lang: Migration to 10.0

cubells:
- Update README
- Update tests

cc @Tecnativa


partner_contact_lang/models/res_partner.py
```diff
@@ -21,21 +22,12 @@ def write(self, vals):
...
-    def onchange_parent_id(self, parent_id):
+    @api.onchange('parent_id')
+    def onchange_parent_id(self):
         """Change language if the parent company changes and there's no
         language defined yet"""
-        res = super(ResPartner, self).onchange_parent_id(parent_id)
-        if parent_id and self.parent_id.id != parent_id and not self.lang:
-            parent = self.browse(parent_id)
+        res = super(ResPartner, self).onchange_parent_id()
+        if self.parent_id and self.parent_id.id != self.id and not self.lang:
```
yajo: You can do `if self != self.parent_id and not self.lang` I think 🤔

yajo (changes requested):

chienandalu (approved):
:+1: Tested on runbot

partner_contact_lang/models/res_partner.py
cubells: I can't. It must be checked before if parent_id exists or not.

yajo (approved):

tafaRU (approved):

pedrobaeza (approved):

PR #473, opened 2017-09-13
[9.0][IMP] partner_financial_risk: Add track visibility to credit limit

carlosdauden:
@Tecnativa

### DIFF
--- a/partner_financial_risk/__openerp__.py
+++ b/partner_financial_risk/__openerp__.py
@@ -5,7 +5,7 @@
 {
     'name': 'Partner Financial Risk',
     'summary': 'Manage partner risk',
-    'version': '9.0.1.0.1',
+    'version': '9.0.1.0.2',
     'category': 'Sales Management',
     'license': 'AGPL-3',
     'author': 'Tecnativa, Odoo Community Association (OCA)',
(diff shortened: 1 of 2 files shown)


pedrobaeza:
Change module version

carlosdauden:
@pedrobaeza I was in it :smile:

partner_financial_risk/__openerp__.py
```diff
@@ -5,7 +5,7 @@
 {
     'name': 'Partner Financial Risk',
     'summary': 'Manage partner risk',
-    'version': '9.0.1.0.1',
+    'version': '9.0.1.0.2',
```
yajo, line 8: I think it should be `9.0.1.1.0`, since it requires a module update.

partner_financial_risk/models/res_partner.py
```diff
@@ -56,6 +56,7 @@ class ResPartner(models.Model):
         help='It Indicate if partner risk exceeded')
     credit_policy = fields.Char()
     risk_allow_edit = fields.Boolean(compute='_compute_risk_allow_edit')
+    credit_limit = fields.Float(track_visibility='onchange')
```
yajo, line 59: 🤔 Do you add a field and nothing more? Don't you add it to a view or something like that?

pedrobaeza: This field already exists before.

partner_financial_risk/__openerp__.py
pedrobaeza: Not really. That attribute gets refreshed with a service restart.

pedrobaeza (approved):

yajo (approved):

chienandalu (approved):

PR #475, opened 2017-09-15
[ADD][partner_phonecall_schedule] Know partner's best phonecall time

yajo:
Moving https://github.com/OCA/crm/pull/172

@Tecnativa


partner_phonecall_schedule/README.rst
```diff
@@ -0,0 +1,89 @@
...
+   :alt: License: AGPL-3
+
+===========================
+Partner phonecalls schedule
+===========================
+
+If you know the best moment to call your partners, use this addon to keep
+track of it and be able to filter partners that you can comfortably call now.
+
+Installation
```
pedrobaeza: Remove this

chienandalu: I think you need to drop this

chienandalu (commented):
Tested on runbot:

The only thing I couldn't achieve is to aggregate schedules. For example:

- Schedule 1: Monday from 10.00 to 15.00
- Schedule 2: Tuesday from 14.00 to 18.00

Only the the first one is considered. Shouldn't both be combined?

partner_phonecall_schedule/tests/test_schedule.py
```diff
@@ -0,0 +1,112 @@
...
+
+from mock import patch
+from datetime import datetime, timedelta
+from openerp.tests.common import TransactionCase
+
+PATH = ("openerp.addons.partner_phonecall_schedule.models"
+        ".res_partner.datetime")
+
+
+class CanICallCase(TransactionCase):
```
pedrobaeza: Use SavePointCase for avoiding succesive calls to `setUp`

partner_phonecall_schedule/tests/test_schedule.py
```diff
@@ -0,0 +1,112 @@
...
+        self.Partner = self.env["res.partner"]
+        self.dude = self.Partner.create({
+            "name": "Dude",
+        })
+        self.dude.phonecall_calendar_ids = self.some_mornings
+        # This is a friday morning
+        self.datetime = datetime(2017, 9, 15, 10, 53, 30)
+
+    @patch(PATH)
+    def tearDown(self, mocked_dt):
```
pedrobaeza: It's strange to do this on tearDown. Why not calling a method?

partner_phonecall_schedule/views/res_partner_view.xml
```diff
@@ -0,0 +1,38 @@
...
+
+<odoo>
+
+    <record id="view_partner_form" model="ir.ui.view">
+        <field name="name">Partner phonecall schedule</field>
+        <field name="model">res.partner</field>
+        <field name="inherit_id" ref="base.view_partner_form"/>
+        <field name="arch" type="xml">
+            <xpath expr="//notebook">
+                <page string="Phone calls" name="phonecalls">
```
pedrobaeza: Do you think this deserves a new tab? I would put it in a many2many_tags widget inside "Sales & Purchases".

pedrobaeza (changes requested):

partner_phonecall_schedule/views/res_partner_view.xml
chienandalu: I agree that the schedules would fit best in a widget. Anyway, the aggregated phonecall schedule needs plenty of space to show all the available ones:
image

yajo: You nailed it. Besides there's a UI problem: to manage available `resource.calendar` records, you need to have admin permissions, and `many2many_tabs` wouldn't allow the user to click & edit a calendar record as the list widget does.

So with this widget, they can edit it if anything went wrong, and still there's no need to move the resource menus or give extra privileges to users.

pedrobaeza: OK, so if this is required, drop a comment in the XML file for explaining it to avoid the same future comments

yajo:
I think all should be fixed now.

chienandalu (approved):
It works perfect now :+1:

yajo:
It should go 100% ☔️ now

partner_phonecall_schedule/models/res_partner.py
```diff
@@ -65,7 +66,13 @@ def _phonecall_available_domain(self):
             tz = timezone(self.env.context["tz"])
         except KeyError:
             tz = None
-        now = datetime.now(tz=tz)
+        now = self.env.context.get("now", datetime.now(tz=tz))
+        try:
+            now = datetime.strptime(now, DEFAULT_SERVER_DATETIME_FORMAT)
```
pedrobaeza: Please use `fields.Datetime.from_string` here.

partner_phonecall_schedule/models/res_partner.py
```diff
@@ -65,7 +66,13 @@ def _phonecall_available_domain(self):
             tz = timezone(self.env.context["tz"])
         except KeyError:
             tz = None
-        now = datetime.now(tz=tz)
+        now = self.env.context.get("now", datetime.now(tz=tz))
```
pedrobaeza: Use `fields.Datetime.now()`

partner_phonecall_schedule/tests/test_schedule.py
```diff
@@ -5,16 +5,17 @@
 from datetime import datetime, timedelta
 from mock import patch
 from openerp.tests.common import SavepointCase
+from openerp.tools import DEFAULT_SERVER_DATETIME_FORMAT
```
pedrobaeza: The same about this

partner_phonecall_schedule/models/res_partner.py
yajo: Nope, I need a datetime object here.

pedrobaeza: `fields.Datetime.from_string(fields.Datetime.now())` and you don't have to deal with tz.

yajo: But that's exactly the same as `datetime.now()`, and the problem with tz is that hour fields are floats, and thus not auto-localized, so I need to localize them manually here.

pedrobaeza: OK then

partner_phonecall_schedule/tests/test_schedule.py
pedrobaeza: This one is still valid though

pedrobaeza (approved):

PR #480, opened 2017-09-26
[FIX] partner_phonecall_schedule: Allow partner managers to use addon

yajo:
Avoids users receiving message:
image

@Tecnativa

### DIFF
--- a/partner_phonecall_schedule/__openerp__.py
+++ b/partner_phonecall_schedule/__openerp__.py
@@ -4,7 +4,7 @@
 {
     "name": "Partner phonecalls schedule",
     "summary": "Track the time and days your partners expect phone calls",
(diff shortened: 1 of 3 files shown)


pedrobaeza (changes requested):
groups should be put at view level, not page level.

yajo:
Why? The purpose is to avoid the error message if you cannot edit records for any reason.

pedrobaeza:
For avoiding to give extra permission to everyone that access res.partner form view. Although the page is invisible, the fields are loaded. Same as https://github.com/OCA/account-financial-tools/pull/537

yajo:
OK I get it, it's fixed now.

carlosdauden (approved):

pedrobaeza (approved):

PR #486, opened 2017-10-06
Add module partner_multi_relation_parent_only

gfcapalbo:
Adding module for having  Parent contact Hierarchy Mapping in relations  (2-way)


partner_multi_relation_parent/README.rst
```diff
@@ -0,0 +1,61 @@
...
+    :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+    :alt: License: AGPL-3
+
+=============================================
+Parent contact Hierarchy Mapping in relations
+=============================================
+
+This module maps automatically the relations between parent partners and their
+children as relations. It has an init hook that will create such relations for
+existing partners and their children partner. If a child partner changes it's
```
hbrunn: its

partner_multi_relation_parent/README.rst
```diff
@@ -0,0 +1,61 @@
...
+create the relations "has contact" and "is contact of" for partners and their
+contacts. This will allow to search using this key, and therefore have an
+updated search option for partners and their contacts.
+
+
+
+Known issues / Roadmap
+======================
+
+* hide/forbade the delition of the "Is contact of" and "Has Contact" installed
```
hbrunn: delEtion

partner_multi_relation_parent/hooks.py
```diff
@@ -0,0 +1,12 @@
...
+# © 2017 Therp BV <http://therp.nl>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from odoo import api, SUPERUSER_ID
+
+
+def post_init_hook(cr, registry):
+    env = api.Environment(cr, SUPERUSER_ID, {})
+    partner_model = env['res.partner']
+    # get all fields with a parent
+    partners = partner_model.search(['!', ('parent_id', 'in', [False])])
```
hbrunn: add `active_test=False`

hbrunn: what's wrong with `[('parent_id', '!=', False)]`?

partner_multi_relation_parent/models/res_partner.py
```diff
@@ -0,0 +1,58 @@
+# -*- coding: utf-8 -*-
+# © 2017 Therp BV <http://therp.nl>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from openerp import api, models
+
+
+class ResPartner(models.Model):
+    _inherit = 'res.partner'
+
+    # called without parameters from the init hook
```
hbrunn: where?

partner_multi_relation_parent/models/res_partner.py
```diff
@@ -0,0 +1,58 @@
...
+    def find_current_relation(self, left, type_relation, right):
+        par_rel_mod = self.env['res.partner.relation']
+        return par_rel_mod.search([
```
hbrunn: there's a high chance that this is already in the cache, so better use `self.relation_all_ids` and restrict to what you search for. I also don't see the point of this function given you use it only once and there's no obvious reason why other modules should want to override exactly this function

partner_multi_relation_parent/models/res_partner.py
```diff
@@ -0,0 +1,58 @@
...
+        res = super(ResPartner, self).create(vals=vals)
+        if "parent_id" in vals:
+            res.update_relations(None, vals['parent_id'])
+        return res
+
+    @api.multi
+    def write(self, vals):
+        for this in self:
+            if this.env.context.get('no_relation_update'):
+                res = super(ResPartner, this).write(vals=vals)
```
hbrunn: do this check out of the loop. Otherwise, you degrade all bulk updates on parts to a loop

hbrunn: also check if you need the loop at all out of the loop, and only go there when necessary

partner_multi_relation_parent/models/res_partner.py
```diff
@@ -0,0 +1,58 @@
...
+                return res
+            if "parent_id" in vals and vals.get('parent_id'):
```
hbrunn: so you actually only care about the result of `vals.get('parent_id')`

partner_multi_relation_parent/models/res_partner_relation.py
```diff
@@ -0,0 +1,113 @@
...
+            # check that whatever relation will come out of this write does
+            # not exist already , but check only if type_id doesn't change, if
+            # it does we don't care about uniqueness
+            if vals.get('type_id', this.type_id) == type_relation:
+                relation = this.relation_exists(
+                    vals.get('left_partner_id', this.left_partner_id),
+                    type_relation,
+                    vals.get('right_partner_id', this.right_partner_id)
+                )
+                if relation:
```
hbrunn, line 45: this is logaclly wrong. You need to check that the left partner has no other relations of this type, not only no relations with the current partner

partner_multi_relation_parent/models/res_partner_relation.py
```diff
@@ -0,0 +1,113 @@
...
+    def create(self, vals):
+        type_relation = self.env.ref(
+            'partner_multi_relation_parent.parent_relation_type'
+        ).id
+        current = self.relation_exists(
+            vals['left_partner_id'],
+            type_relation,
+            vals['right_partner_id']
+        )
+        if current:
```
hbrunn: see above

hbrunn (changes requested):

partner_multi_relation_parent/models/res_partner_relation.py
gfcapalbo: checking all possible right partners in the function

NL66278:
I think it is wrong to make invoice, shipping or other addresses automatically into contacts. This should only work for contact addresses.

partner_multi_relation_parent/models/res_partner.py
```diff
@@ -0,0 +1,78 @@
...
+    def write(self, vals):
+        for this in self:
+            if this.env.context.get('no_relation_update'):
+                continue
+            if vals.get('parent_id'):
+                this.update_relations(this.parent_id.id, vals['parent_id'])
+        res = super(ResPartner, self).write(vals=vals)
+        return res
+
+    # if you unlink a partner you must unlink all relations in which it is
```
NL66278: I think there is no need for this as left_partner_id and right_partner_id have already defined a cascading delete

NL66278:
@gfcapalbo Please close this in favour of https://github.com/OCA/partner-contact/pull/493. Would be nice if you could test 493 as well.

PR #488, opened 2017-10-09
[11.0] [mig] base_location: Migration to 11.0

etobella:
Migration to 11.0.

No major changes


base_location/demo/better_zip.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version = "1.0" encoding="utf-8"?>
+<openerp>
```
rgarnau: change to "odoo" and remove "data"

base_location/models/better_zip.py
```diff
@@ -0,0 +1,47 @@
...
+    name = fields.Char('ZIP')
+    code = fields.Char('City Code', size=64,
+                       help="The official code for the city")
+    city = fields.Char('City', required=True)
+    state_id = fields.Many2one('res.country.state', 'State')
+    country_id = fields.Many2one('res.country', 'Country')
+    latitude = fields.Float()
+    longitude = fields.Float()
+
+    @api.one
```
rgarnau: deprecated

base_location/models/company.py
```diff
@@ -0,0 +1,24 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import models, fields, api
+
+
+class ResCompany(models.Model):
+
+    _inherit = 'res.company'
+
+    @api.onchange('better_zip_id')
```
rgarnau: this goes after the fields

rgarnau (commented):
LGTM :+1:  Just minor comments inline

base_location/README.rst
```diff
@@ -0,0 +1,66 @@
...
+=====
+
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/134/10.0
+
+Bug Tracker
+===========
+
+Bugs are tracked on `GitHub Issues <https://github.com/OCA/
```
JordiBForgeFlow: Adjust this section, to follow the same structure as in https://github.com/OCA/maintainer-tools/blob/master/template/module/README.rst

base_location/README.rst
```diff
@@ -0,0 +1,66 @@
...
+It enables zip, city, state and country auto-completion on partners and companies.
+
+Also allows different search filters.
+
+Usage
+=====
+
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/134/10.0
```
JordiBForgeFlow: 11.0

base_location/README.rst
```diff
@@ -0,0 +1,66 @@
...
+
+Icon
+----
+* http://icon-park.com/icon/location-map-pin-orange3/
+
+
+Maintainer
+----------
+
+.. image:: http://odoo-community.org/logo.png
```
JordiBForgeFlow: use https in all references in the readme

JordiBForgeFlow (approved):
:+1: Functional and technical review. Minor comments.

pedrobaeza:
Odoo now contains a city_id field. Maybe we should adapt this module to the new core data model.

base_location/models/company.py
```diff
@@ -6,19 +6,19 @@
...
+        'res.better.zip',
+        string='Location',
+        help='Use the city name or the zip code to search the location',
+    )
+
     @api.onchange('better_zip_id')
     def on_change_city(self):
         if self.better_zip_id:
             self.zip = self.better_zip_id.name
+            self.city_id = self.better_zip_id.city_id
```
JordiBForgeFlow: res.company does not have a field 'city_id'

base_location/views/partner_view.xml
```diff
@@ -1,24 +1,49 @@
...
+    </record>
 
     <record id="view_partner_form" model="ir.ui.view">
         <field name="name">res.partner.zip_id.2</field>
         <field name="model">res.partner</field>
-        <field name="inherit_id" ref="base.view_partner_form" />
+        <field name="inherit_id" ref="base.view_partner_form"/>
         <field name="arch" type="xml">
             <field name="city" position="before">
                 <field name="zip_id"
```
JordiBForgeFlow: The field should not be editable for contact persons, as it happens will other fields in this circumstance.
image


Just add `attrs="{'readonly': [('type', '=', 'contact'),('parent_id', '!=', False)]}"`

JordiBForgeFlow (changes requested):
Minor changes after adding city_id

base_location/models/better_zip.py
```diff
@@ -13,24 +13,22 @@ class BetterZip(models.Model):
     _order = "name asc"
     _rec_name = "display_name"
 
-    display_name = fields.Char('Name', compute='_get_display_name', store=True)
+    display_name = fields.Char(
+        'Name', compute='_compute_display_name', store=True
```
pedrobaeza: command at the end for reducing diff on later changes

base_location/demo/better_zip.xml
```diff
@@ -1,12 +1,8 @@
 <?xml version = "1.0" encoding="utf-8"?>
-<openerp>
-  <data>
-
+<odoo>
     <record id="demo_brussels" model="res.better.zip">
       <field name="name">1000</field>
       <field name="city">Brussels</field>
       <field name="country_id" ref="base.be"/>
```
pedrobaeza: Isn't demo data in base for completing city_id field?

base_location/models/better_zip.py
```diff
@@ -13,24 +13,22 @@ class BetterZip(models.Model):
...
+    )
     name = fields.Char('ZIP')
     code = fields.Char('City Code', size=64,
                        help="The official code for the city")
     city = fields.Char('City', required=True)
+    city_id = fields.Many2one('res.city', 'City')
     state_id = fields.Many2one('res.country.state', 'State')
     country_id = fields.Many2one('res.country', 'Country')
+    enforce_cities = fields.Boolean(related='country_id.enforce_cities')
```
pedrobaeza: `readonly=True` for avoiding undesired writes on res.country

base_location/views/better_zip_view.xml
```diff
@@ -7,11 +7,23 @@
...
                 <group col="4">
-                <field name="name"/>
-                <field name="code"/>
-                <field name="city"/>
-                <field name="state_id"/>
-                <field name="country_id"/>
+                    <group>
+                        <field name="name"/>
+                        <field name="city_id"
+                               attrs="{
```
pedrobaeza: Don't split this attribute through several lines or the visualization inside UI will be ugly

base_location/views/better_zip_view.xml
```diff
@@ -7,11 +7,23 @@
...
+                                'invisible': [('enforce_cities', '=', False)],
+                                'required': [('enforce_cities', '=', True)]}"
+                               domain="[('country_id', '=', country_id)]"/>
+                        <field name="city"
+                               attrs="{'invisible': [('enforce_cities', '=', True)]}"/>
```
pedrobaeza: If you hide this field, then hide also state field that is also auto-assigned from city

base_location/views/better_zip_view.xml
```diff
@@ -7,11 +7,23 @@
...
+                        <field name="country_id"/>
+                        <field name="enforce_cities" invisible="1"/>
+                    </group>
+                    <group>
+                        <field name="code"/>
+                        <field name="state_id"
+                               domain="[('country_id', '=', country_id)]"/>
```
pedrobaeza: This should be dynamically added on country_id onchange, as this way, no state is shown if no country is selected.

base_location/views/better_zip_view.xml
```diff
@@ -7,11 +7,23 @@
...
-                <field name="city"/>
-                <field name="state_id"/>
-                <field name="country_id"/>
+                    <group>
+                        <field name="name"/>
+                        <field name="city_id"
+                               attrs="{
+                                'invisible': [('enforce_cities', '=', False)],
+                                'required': [('enforce_cities', '=', True)]}"
+                               domain="[('country_id', '=', country_id)]"/>
```
pedrobaeza: This should be dynamically added on country_id onchange, as this way, no city is shown if no country is selected.

base_location/views/better_zip_view.xml
```diff
@@ -59,16 +73,16 @@
     </record>
 
     <menuitem
-        name="Cities/Locations Management"
-        id="locations_root_menu"
-        parent="base.menu_custom"
-        />
+            name="Cities/Locations Management"
```
pedrobaeza: Don't change this only for stylistic reasons

base_location/views/better_zip_view.xml
```diff
@@ -59,16 +73,16 @@
...
+            parent="base.menu_custom"
+    />
 
     <menuitem
-        name="Cities/Locations"
-        id="locations_menu"
-        parent="locations_root_menu"
-        action="action_zip_tree"
-        />
+            name="Cities/Locations"
```
pedrobaeza: Don't change this only for stylistic reasons

base_location/models/partner.py
```diff
@@ -13,6 +13,7 @@ class ResPartner(models.Model):
     def onchange_zip_id(self):
         if self.zip_id:
             self.zip = self.zip_id.name
+            self.partner_id.city_id = self.zip_id.city_id
```
pedrobaeza: `self.city_id = self.zip_id.city_id`

pedrobaeza (changes requested):

base_location/demo/better_zip.xml
JordiBForgeFlow: No, the module 'base_address_city' contains no demo data.

pedrobaeza:
Please check CIs status

base_location/views/res_country_view.xml
```diff
@@ -0,0 +1,34 @@
...
+            <xpath expr="//div[hasclass('oe_button_box')]" position="inside">
+                <button name="%(action_zip_tree)d"
+                    class="oe_stat_button"
+                    icon="fa-globe"
+                    type="action"
+                    context="{'default_country_id': active_id, 'search_default_country_id': active_id}"
+                    string="Locations">
+                </button>
+            </xpath>
+            <xpath expr="//field[@name='phone_code']" position="after">
```
HviorForgeFlow: Already defined on module base_address_city.

base_location/views/better_zip_view.xml
```diff
@@ -0,0 +1,85 @@
...
+    <record model="ir.ui.view" id="better_zip_form">
+        <field name="name">res.better.zip.form</field>
+        <field name="model">res.better.zip</field>
+        <field name="arch" type="xml">
+            <form string="City">
+                <group col="4">
+                    <group>
+                        <field name="name"/>
+                        <field name="city_id"
+                               attrs="{'invisible': [('enforce_cities', '=', False)],'required': [('enforce_cities', '=', True)]}"
```
HviorForgeFlow: Close tag

base_location/tests/test_completion.py
```diff
@@ -30,20 +40,57 @@ def test_onchange_company_city_completion(self):
...
+        self.env['res.country'].browse(self.ref('base.es')).write({
+            'enforce_cities': True
+        })
         self.company = self.env.ref('base.main_company')
         self.better_zip1 = self.env['res.better.zip'].create({
             'name': 1000,
             'city': 'Lausanne',
             'state_id': state_vd.id,
-            'country_id': self.ref('base.ch'),
+            'country_id': self.ref('base.ch')
```
pedrobaeza: Why do you remove final commas for raising the diff lines and making in later changes to have 2 lines diff?

base_location/tests/test_completion.py
```diff
@@ -30,20 +40,57 @@ def test_onchange_company_city_completion(self):
...
         })
         self.partner1 = self.env['res.partner'].create({
-            'name': 'Camptocamp',
+            'name': 'Camptocamp'
```
pedrobaeza: Please end all lines that are elements of a dictionary with the trailing comma as a good habit (you have added a lot with it).

base_location/models/partner.py
```diff
@@ -7,12 +7,16 @@
...
     def onchange_zip_id(self):
         if self.zip_id:
             self.zip = self.zip_id.name
+            self.city_id = self.zip_id.city_id
             self.city = self.zip_id.city
-            self.state_id = self.zip_id.state_id
+            if self.country_id.enforce_cities:
+                self.state_id = self.city_id.state_id
+            else:
+                self.state_id = self.better_zip_id.state_id
```
pedrobaeza: s/better_zip_id/zip_id

base_location/models/company.py
```diff
@@ -6,19 +6,22 @@
 
 
 class ResCompany(models.Model):
-
     _inherit = 'res.company'
 
+    better_zip_id = fields.Many2one(
```
pedrobaeza: This shouldn't be needed as company_id is inherited by delegation of res.partner. And please, don't call it `better_zip_id`.

base_location/models/company.py
pedrobaeza: You have to rename it to `self.zip_id.city_id`

base_location/models/partner.py
```diff
@@ -7,12 +7,16 @@
 
 class ResPartner(models.Model):
     _inherit = 'res.partner'
-    zip_id = fields.Many2one('res.better.zip', 'City/Location')
+    zip_id = fields.Many2one('res.better.zip', 'Location')
```
pedrobaeza: Call it `ZIP location`

pedrobaeza (changes requested):

rgarnau (approved):

base_location/models/better_zip.py
```diff
@@ -11,37 +12,108 @@ class BetterZip(models.Model):
...
+        for rec in self:
+            name = []
+            if rec.name:
+                name.append(rec.name)
+            name.append(rec.city)
+            if rec.state_id:
+                name.append(rec.state_id.name)
+            if rec.country_id:
+                name.append(rec.country_id.name)
+            result.append((rec.id, ", ".join(name)))
```
yajo: Shouldn't you be using a `res.lang`-specific template if available? Such as `%(city)s, %(state)s (%(country)s)` for example

base_location/models/better_zip.py
```diff
@@ -11,37 +12,108 @@ class BetterZip(models.Model):
     _name = "res.better.zip"
     _description = __doc__
     _order = "name asc"
-    _rec_name = "display_name"
 
-    display_name = fields.Char('Name', compute='_get_display_name', store=True)
+    def domain_city_id(self):
+        if self.country_id:
+            return [('country_id', '=', self.country_id.id)]
+        return ""
```
yajo: Shouldn't this return `[]`?

base_location/models/better_zip.py
```diff
@@ -11,37 +12,108 @@ class BetterZip(models.Model):
...
+
+    def domain_state_id(self):
+        if self.country_id:
+            return [('country_id', '=', self.country_id.id)]
+        return ""
```
yajo: Shouldn't this return `[]`?

base_location/models/partner.py
```diff
@@ -2,17 +2,79 @@
...
+            if self.country_id.enforce_cities:
+                self.city_id = self.zip_id.city_id
             self.zip = self.zip_id.name
-            self.city = self.zip_id.city
             self.state_id = self.zip_id.state_id
-            self.country_id = self.zip_id.country_id
+            self.city = self.zip_id.city
+
+    @api.constrains('zip_id', 'country_id', 'city_id', 'state_id')
+    def constrains_zip(self):
```
yajo: Call it `_check_zip`

base_location/models/company.py
```diff
@@ -6,19 +6,45 @@
...
-            self.country_id = self.better_zip_id.country_id
+    city_id = fields.Many2one(
+        'res.city',
+        compute='_compute_address',
+        inverse='_inverse_city_id',
+        string="City"
+    )
 
-    better_zip_id = fields.Many2one(
+    zip_id = fields.Many2one(
```
yajo: Add `oldname="better_zip_id"`

yajo (changes requested):

JordiBForgeFlow (approved):

JordiBForgeFlow:
@Yajo @pedrobaeza Can you do a last review? Seems that this one is ready to merge.

base_location/README.rst
```diff
@@ -12,24 +12,48 @@ It enables zip, city, state and country auto-completion on partners and companie
...
 
+Configuration
+=============
+
+* Users in group 'Contact Creation' can maintain the Locations.
+
 Usage
 =====
 
+#. Activate the developer mode in *Settings*.
```
pedrobaeza: This should be in Configuration, and in the Usage section, explain that going to a partner, and filling the field "Autocomplete...", they will get the rest of the fields...

base_location/models/company.py
```diff
@@ -6,19 +6,46 @@
...
         'res.better.zip',
-        string='Location',
+        string='ZIP Location',
+        compute='_compute_address',
```
pedrobaeza: This should be a simply related field to `partner_id.zip_id`. Don't complicate things...

base_location/models/company.py
```diff
@@ -6,19 +6,46 @@
...
     _inherit = 'res.company'
 
-    @api.onchange('better_zip_id')
-    def on_change_city(self):
-        if self.better_zip_id:
-            self.zip = self.better_zip_id.name
-            self.city = self.better_zip_id.city
-            self.state_id = self.better_zip_id.state_id
-            self.country_id = self.better_zip_id.country_id
+    city_id = fields.Many2one(
```
pedrobaeza: Why adding this field here? It's out of this module's scope.

base_location/models/company.py
```diff
@@ -6,19 +6,46 @@
...
+    def _inverse_city_id(self):
+        for company in self:
+            company.partner_id.city_id = company.city_id
+
+    def _inverse_zip_id(self):
+        for company in self:
+            company.partner_id.zip_id = company.zip_id
+
+    @api.onchange('zip_id')
+    def on_change_city(self):
```
pedrobaeza: `_onchange_zip_id`

base_location/models/company.py
```diff
@@ -6,19 +6,46 @@
...
     )
+
+    def _get_company_address_fields(self, partner):
+        res = super(ResCompany, self)._get_company_address_fields(partner)
+        res['city_id'] = partner.city_id
+        res['zip_id'] = partner.zip_id
+        return res
+
+    def _inverse_city_id(self):
+        for company in self:
```
pedrobaeza: Remove this method

base_location/models/company.py
```diff
@@ -6,19 +6,46 @@
...
+            company.partner_id.city_id = company.city_id
+
+    def _inverse_zip_id(self):
```
pedrobaeza: Remove this method

base_location/models/company.py
```diff
@@ -6,19 +6,46 @@
...
         'res.better.zip',
-        string='Location',
+        string='ZIP Location',
+        compute='_compute_address',
+        inverse='_inverse_zip_id',
+        oldname="better_zip_id",
         help='Use the city name or the zip code to search the location',
     )
+
+    def _get_company_address_fields(self, partner):
```
pedrobaeza: Explain what this method does.

base_location/models/partner.py
```diff
@@ -2,17 +2,79 @@
...
+        if self.city_id:
+            return {
+                'domain': {
+                    'zip_id': [('city_id', '=', self.city_id.id)]
+                },
+            }
+        return {'domain': {'zip_id': []}}
+
+    @api.onchange('state_id')
+    def _onchange_state_id(self):
```
pedrobaeza, line 29: Is this really needed? Nobody searches zips only filtering the state: or you type the zip code or you type the city name

base_location/models/partner.py
```diff
@@ -2,17 +2,79 @@
...
+            return {
+                'domain': {
+                    'zip_id': [('state_id', '=', self.state_id.id)]
+                },
+            }
+        return {'domain': {'zip_id': []}}
+
+    @api.onchange('country_id')
+    def _onchange_country_id(self):
+        if self.zip_id and self.zip_id.country_id != self.country_id:
```
pedrobaeza: The same comment about its need. KISS principle.

base_location/models/partner.py
```diff
@@ -2,17 +2,79 @@
...
+        if self.country_id and self.country_id.enforce_cities:
+            res['domain']['zip_id'] = [('country_id', '=', self.country_id.id)]
+        return res
 
     @api.onchange('zip_id')
-    def onchange_zip_id(self):
+    def _onchange_zip_id(self):
         if self.zip_id:
+            self.country_id = self.zip_id.country_id
+            self.state_id = False
```
pedrobaeza: Why putting this to False?

base_location/models/partner.py
```diff
@@ -2,17 +2,79 @@
...
+                self.city_id = self.zip_id.city_id
             self.zip = self.zip_id.name
-            self.city = self.zip_id.city
             self.state_id = self.zip_id.state_id
-            self.country_id = self.zip_id.country_id
+            self.city = self.zip_id.city
+
+    @api.constrains('zip_id', 'country_id', 'city_id', 'state_id')
+    def _check_zip(self):
+        for rec in self:
```
pedrobaeza: Put `for rec in self.filtered('zip_id'):` and you remove the need of putting it on the subsequent `if`

base_location/tests/test_base_location.py
```diff
@@ -0,0 +1,282 @@
...
+            'zip_id': better_zip1.id,
+            'state_id': better_zip1.state_id.id,
+            'country_id': better_zip1.country_id.id,
+            'city_id': better_zip1.city_id.id,
+            'city': better_zip1.city,
+            'zip': better_zip1.name,
+        })
+        company = self.env['res.company'].new({
+            'name': partner.name,
+            'partner_id': partner.id
```
pedrobaeza: This is not Odoo flow: you have to create the company, and the partner is created from it.

base_location/models/better_zip.py
```diff
@@ -11,35 +12,102 @@ class BetterZip(models.Model):
     _name = "res.better.zip"
     _description = __doc__
     _order = "name asc"
-    _rec_name = "display_name"
 
-    display_name = fields.Char('Name', compute='_get_display_name', store=True)
+    def domain_city_id(self):
```
pedrobaeza: Why splitting these methods here if they are used only one time?

base_location/models/better_zip.py
```diff
@@ -11,35 +12,102 @@ class BetterZip(models.Model):
...
+            if rec.state_id:
+                name.append('%(name)s' % {'name': rec.state_id.name})
+            if rec.country_id:
+                name.append('%(name)s' % {'name': rec.country_id.name})
+            result.append((rec.id, ", ".join(name)))
+        return result
+
+    @api.onchange('country_id')
+    def _onchange_country_id(self):
         if self.country_id:
```
pedrobaeza: No need of this initial `if`, as `self.state_id.country_id != self.country_id` is going to be handled correctly in case of country_id = False.

base_location/models/better_zip.py
```diff
@@ -11,35 +12,102 @@ class BetterZip(models.Model):
...
+            if self.state_id.country_id != self.country_id:
+                self.state_id = False
+            if self.city_id.country_id != self.country_id:
+                self.city_id = False
+        else:
+            self.state_id = False
+            self.city_id = False
+        return {
+            'domain': {
+                'state_id': self.domain_city_id(),
```
pedrobaeza: Handle here directly the domains (and look that the methods names are very fuzzy.

base_location/models/better_zip.py
```diff
@@ -11,35 +12,102 @@ class BetterZip(models.Model):
...
+        }
+
+    @api.onchange('city_id')
+    def _onchange_city_id(self):
+        if self.city_id:
+            self.city = self.city_id.name
+            self.country_id = self.city_id.country_id
+            if not self.country_id.enforce_cities:
+                self.state_id = self.city_id.state_id
+                self._onchange_state_id()
```
pedrobaeza: AFAIK, this is called automatically when you change the value of state_id.

pedrobaeza (dismissed):

base_location/models/company.py
etobella: We are following the same configuration used in [odoo company](https://github.com/odoo/odoo/blob/11.0/odoo/addons/base/res/res_company.py#L59)

yajo: If I'm not wrong, `res.company` extends `res.partner`, so `zip_id` should be available out of the box, isn't it?

yajo (changes requested):

base_location/models/company.py
```diff
@@ -0,0 +1,51 @@
...
+        'res.city',
+        compute='_compute_address',
+        inverse='_inverse_city_id',
+        string="City"
+    )
+
+    zip_id = fields.Many2one(
+        'res.better.zip',
+        string='ZIP Location',
+        compute='_compute_address',
```
yajo: Not sure what happens with GH, but I re-comment to get it displayed.

AFAIK `res.company` extends `res.partner`, so `zip_id` should be there out of the box, right?

pedrobaeza: Yeah, definitively I don't understand why the rest of the stuff would be needed having this easy approach of simply adding the field to the view, as in previous versions.

etobella: That is not true, as you can check on [odoo company definition](https://github.com/odoo/odoo/blob/11.0/odoo/addons/base/res/res_company.py#L11).

yajo: Well, indeed I was wrong: `res.company` does not extend `res.partner`. In any case, wouldn't it be enough to add the field as related instead of repeating the compute methods?

JordiBForgeFlow: @Yajo But odoo chose to duplicate the address fields into res.company. And we should respect that IMHO.

etobella:
@pedrobaeza @Yajo is it necessary to change how it works? Or should we keep similar to the original address fields of `res.company`?

base_location/models/company.py
yajo: Yes, but [look at how it does it: basically reading and writing the field from the linked partner](https://github.com/odoo/odoo/blob/11.0/odoo/addons/base/res/res_company.py#L101-L130). That's basically a more complex way of using a related field, isn't it?

JordiBForgeFlow: @Yajo I propose that we stick to the way odoo works now. And maybe we can propose to odoo to change their approach upstream, and so when they fix in future versions we also fix.

JordiBForgeFlow:
@Yajo @pedrobaeza Can we merge this one? I think that hitting 96.74% coverage is not too bad, right? :)

base_location/models/company.py
pedrobaeza: I don't agree. If there are no reasons for doing that way, we don't have to do the same as Odoo. Let's keep the code simple. The only possible reason I see if it's there's any issue when you first create a company. Please check that case. If not, KISS principle.

yajo (approved):
I don't really like overengineering things, and just doing that because Odoo does that doesn't seem a good decision IMHO. However, I won't block anymore on that matter.

base_location/models/company.py
JordiBForgeFlow: @pedrobaeza I have been testing and it is proving a mess to mix the two methods. When you make the fields related, the constrains that we added in res.partner, https://github.com/OCA/partner-contact/pull/488/files#diff-2f72fbcaf854c88b3e175783b321cd16R64, will fail. because when you change the city_id in the company, you are effectively changing it in the partner. The constrains on the partner are evaluated before the inverse methods update the other fields (city, etc..).

And we need constrains in the partner to ensure consistency.

So, as a conclusion, address fields are very related to each other. Either you make them all related to the partner in company, or you don't for all of them. But mixing methods produces inconsistencies.

pedrobaeza: Thanks for the explanation, Jordi. OK, that's a good reason for keeping things this way. That was the explanation I was looking for keeping things this way.

pedrobaeza:
For me, 2 remaining things before merging:

- Put the explanation on the code about why is done this way instead with a related (for avoiding possible future questions). Jordi's explanation is very good.
- Squash useless commits from this migration.

base_location/models/company.py
yajo: Interesting. Maybe you could add a comment or known issue on that matter, to avoid future iterations on the same subject unless odoo changes its way...

base_location/models/company.py
```diff
@@ -0,0 +1,66 @@
...
+        inverse='_inverse_zip_id',
+        oldname="better_zip_id",
+        help='Use the city name or the zip code to search the location',
+    )
+    '''
+        In order to keep the same logic used in odoo, fields must be computed
```
JordiBForgeFlow: @pedrobaeza Comments here. I think that we can merge now.

base_location/models/better_zip.py
pedrobaeza: This comment still remains

base_location/models/company.py
pedrobaeza: This comment still remains

base_location/models/company.py
pedrobaeza: This is not comment, but docstring, and can't be in the middle of the code. You should make here a real comment:
```
# In order to keep the same logic used in odoo, fields must be computed
# and inversed, not related. This way, we can ensure that it works
# ...
```

base_location/models/company.py
pedrobaeza: This comment still remains.

base_location/models/partner.py
pedrobaeza: This comment still remains

base_location/models/partner.py
pedrobaeza: This comment still remains

base_location/views/better_zip_view.xml
pedrobaeza: This comment still remains

base_location/models/company.py
etobella: City_id is not added on `base_address_city` module. So I had to add it in order to maintain its logic.

base_location/models/company.py
etobella: This method computes address fields from the partner. Used in `_compute_address` as shown in [company_code](https://github.com/odoo/odoo/blob/11.0/odoo/addons/base/res/res_company.py#L106)

base_location/models/partner.py
etobella: It has been modified the domain filtering, but not the changes on fields (in order to avoid discrepancies)

base_location/models/better_zip.py
```diff
@@ -0,0 +1,101 @@
...
+            }
+        }
+
+    @api.onchange('city_id')
+    def _onchange_city_id(self):
+        if self.city_id:
+            self.city = self.city_id.name
+            self.country_id = self.city_id.country_id
+            if not self.country_id.enforce_cities:
+                self.state_id = self.city_id.state_id
```
jesusVMayor: What is the sense of this line? will never be executed

JordiBForgeFlow: Why not? If the country is set to enforce the use of cities, it will.

jesusVMayor: Because is "if NOT self.country_id.enforce_cities" and if the country is not set to enforce cities the field city_id is invisible and it can not be changed. So IMHO the if is unnecesary.

yvaucher:
I don't want to be that guy but aren't we missing all the commits we had when the module was named `better_zip` ?

yvaucher:
https://github.com/OCA/partner-contact/commits/10.0/better_zip

etobella:
@yvaucher That is was related to the migration code because the module name was changed. I have changed it in order to recover them

etobella:
@pedrobaeza I think that all comments have been modified now or an answer was provided.

yvaucher:
@etobella nice work, thank you.

jesusVMayor (approved):
LGTM

JordiBForgeFlow:
@pedrobaeza All comments have now been addressed. Can we merge?

PR #492, opened 2017-10-11
[FIX] don't touch registry's test mode

hbrunn:
@lasley depending on when this test's class is loaded, it will break because the registry is already in test mode. What's your rationale to do this in a class method?

### DIFF
--- a/partner_identification/tests/test_res_partner.py
+++ b/partner_identification/tests/test_res_partner.py
@@ -7,6 +7,7 @@ from odoo.tests import common
 from odoo.exceptions import ValidationError
 
 
+# pylint: disable=consider-merging-classes-inherited
 class ResPartner(models.Model):
     _inherit = 'res.partner'
 
@@ -23,10 +24,9 @@ class ResPartner(models.Model):
     )
 
 
-class TestResPartner(common.SavepointCase):
+class TestResPartner(common.TransactionCase):
 
-    @classmethod
-    def _init_test_model(cls, model_cls):
+    def _init_test_model(self, model_cls):
         """ Build a model from model_cls in order to test abstract models.
         Note that this does not actually create a table in the database, so
         there may be some unidentified edge cases.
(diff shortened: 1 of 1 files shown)


lasley:
@hbrunn - This strategy has been an ongoing experiment, the details of which can be seen in https://github.com/OCA/maintainer-tools/pull/242. 

The rationale behind the classmethod was honestly because we copied it from code linked in https://github.com/OCA/maintainer-tools/pull/242#issuecomment-262457147. A lot of the stuff has bitten us in a lot of different spots though, so I absolutely welcome any suggestions if you have them.

I really feel like we had a strategy at the instance level, but something failed in it. I wish I could remember what.. @obulkin am I making things up?

lasley (approved):

hbrunn:
what failed before on the instance level is the savepoint base class, that's why I changed this to transaction case. Somewhere in the model setup code either a real rollback or a real commit is triggered which destroys existing savepoints. As I couldn't find another reason than performance to use this, I figured that's a safe change.

obulkin (changes requested):
@hbrunn is on point as far as base class selection. ``SavepointCase`` is not an option because it conflicts with the savepoints used by the registry test mode, but ``TransactionCase`` and ``SingleTransactionCase`` should both work. The only reason for choosing the latter is that it can speed up testing considerably when there are a lot of cases and no need for isolation between them.

That being said, I don't see why the ``SingleTransactionCase`` approach would be more likely to run into a registry that's already in test mode than the ``TransactionCase`` one. They both use the same logic to get a reference to the current registry or generate a new one. I've yet to experience this issue in practice, but it seems possible in theory, so perhaps we need to add some extra code to handle it?

Lastly, there are some problems with the tests as they are written here:

- Simply dropping the use of test mode is not an option if the tester model needs to be reflected in the DB since the logic for this in ``_auto_init`` includes a commit, making the changes permanent without a test cursor.
- If test mode is being used, there needs to be a cleanup step where it's reverted (e.g. https://github.com/OCA/maintainer-tools/pull/242/files#diff-df5f8c3e82e9b80bcac67e5c00488fafR49)

hbrunn:
seems to be fixed by https://github.com/OCA/partner-contact/commit/bc93e7bbc3e0f059b228970f0a05e57f0efba310#diff-57c0be4b932b0c2f4c5898b79b4a26ac

PR #493, opened 2017-10-12
[10.0] partner multi relation parent turbo

NL66278:
Easy view of partner addresses as relations


NL66278:
Remaining error is in unrelated module partner_create_by_vat, due to vies service not being available.

NL66278:
Not going to develop this further for the moment.

PR #494, opened 2017-10-13
[11.0][mig][base_location_geonames_import]

JordiBForgeFlow:
Migration to 11.0. 
Relevant changes:
* The module now creates res.city if the country requires it (flag 'enforce_cities' is set for that country)

Depends on 

- [x] base_location #488 

I discussed with @pedrobaeza The possibility to import only cities. But after reviewing, it can be very confusing to do so in this module, and in any case we could discuss about creating a separate `base_address_city_geonames_import` that does not depend on base_location, but only on `base_address_city.`


rgarnau (approved):
LGTM :+1:

base_location_geonames_import/__init__.py
```diff
@@ -0,0 +1,3 @@
+# -*- coding: utf-8 -*-
```
etobella, line 1: Coding is no longer necessary, as it can be checked in the [migration document to 11.0](https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-11.0)

etobella (approved):
LGTM :+1: 
Just a minor comment

base_location_geonames_import/__manifest__.py
```diff
@@ -1,12 +1,15 @@
 # -*- coding: utf-8 -*-
-# © 2014-2016 Akretion (Alexis de Lattre <alexis.delattre@akretion.com>)
-# © 2014 Lorenzo Battistini <lorenzo.battistini@agilebg.com>
-# © 2016 Pedro M. Baeza <pedro.baeza@serviciosbaeza.com>
-# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+# Copyright 2014-2016 Akretion (Alexis de Lattre
+#                     <alexis.delattre@akretion.com>)
+# Copyright 2014 Lorenzo Battistini <lorenzo.battistini@agilebg.com>
+# Copyright 2016 Pedro M. Baeza <pedro.baeza@tecnativa.com>
+# Copyright 2017 Efient Business and IT Consulting Services, S.L.
```
MiquelRForgeFlow: Eficent

base_location_geonames_import/__init__.py
mreficent: But here it still appears:
https://github.com/OCA/maintainer-tools/blob/master/template/module/models/__init__.py

MiquelRForgeFlow (commented):
LGTM

base_location_geonames_import/__init__.py
MiquelRForgeFlow: But here it appears:
https://github.com/OCA/maintainer-tools/blob/master/template/module/models/__init__.py

pedrobaeza: Because for Python 2.7 is still needed...

MiquelRForgeFlow: Oh, yes!! 😂

jesusVMayor (approved):
LGTM

pedrobaeza:
Any possible improvement in speed?

AaronHForgeFlow (approved):
It works correctly :+1: Couldn't find a way to make it faster.

yajo:
Can you rebase now that #488 is merged? (BTW I don't understand how bots were ✔️ )

JordiBForgeFlow:
@pedrobaeza @Yajo Ready to be merged.

pedrobaeza:
There are some duplicated commits at the beginning. please squash them.

JordiBForgeFlow:
@pedrobaeza I am not sure to understand which ones you mean. I found a few Transifex commits that were one after each other and I squashed them.

pedrobaeza:
Wow, it seems that right now everything has blown away with lots of commits that doesn't belong to this...

yajo:
Yeah, it seems something weird happened in the branch, please check that 😕

JordiBForgeFlow:
This is now fixed, and rebased. @pedrobaeza Not sure what commits you wanted to squash.

JordiBForgeFlow:
@pedrobaeza Will you do the honors to merge?

pedrobaeza:
Yes, of course.

pedrobaeza:
I see that your name is not correct yet (Efient instead of Eficent). Do you want to fix it?

JordiBForgeFlow:
@pedrobaeza fixed!

PR #498, opened 2017-10-19
[10.0] partner multi relation tabs migration

NL66278:


NL66278:
This PR depend on/includes https://github.com/OCA/partner-contact/pull/497

(Update 2018-04-13: PR 497 has now been merged)

partner_multi_relation/models/res_partner_relation_all.py
```diff
@@ -1,18 +1,100 @@
...
+    res_model: model._name of the underlying table,
+    res_id: base key in the underlying table,
+    this_partner_id: must refer to a partner
+    other_partner_id: must refer to a related partner
+    type_id: refers to res.partner.relation.type,
+    date_start: start date of relation if relevant or NULL,
+    date_end: end date of relation if relevant or NULL,
+    %(is_inverse)s as is_inverse: used to determine type_selection_id,
+    """
+    global _last_key_offset
```
hbrunn: this will fail for multithreaded servers. We need to think of another mechanism, I'd suggest splitting up the sql statement into function similar to https://github.com/OCA/OCB/blob/10.0/addons/account/report/account_invoice_report.py, then overrides can add what they need without destroying multithreading

hbrunn (changes requested):
test coverage needs to be increased

partner_multi_relation/models/res_partner_relation_all.py
NL66278: I don't think it can fail as the variable is only used during module loading to gather all the parts of the statement to create the view. I would be amazed if the loading of one module could be interrupted by the loading of anpther module.

NL66278:
@hbrunn Tests have been added, so I think your point is satisfied now.

hbrunn (approved):

hbrunn:
thanks!

NL66278:
@ddufresne Would you be so kind as to have a look at this PR? I also intent to migrate it to 11.0.

ddufresne:
@NL66278 this is quite a lot of code. Before I go deep into review, I would prefer to see the big picture.

Could you extend your README with screenshots and explanations on how to use your module step by step. Then, it's going to be easier for me to review.

NL66278:
@ddufresne I added a usage section with screenshots (taking as an example your code for partner_multi_relation in 12.0). However the images will only be rendered properly from the README.rst after the merge, because of the way the url's have been setup. But of course they are part of the code so you should be able to see them there.

lfreeke (approved):
Functional test :+1:

PR #504, opened 2017-10-21
[11.0][IMP] Update Gitignore to new template

lasley:


.gitignore
```diff
@@ -1,11 +1,3 @@
-# buildout
```
pedrobaeza, line 1: Isn't buildout still being used?

lasley: I noticed this as well, but this gitignore is a straight copy from our [template](https://github.com/OCA/maintainer-quality-tools/blob/master/sample_files/.gitignore), which does not contain the buildout stuff. I assumed it was intentional.

yajo (approved):

pedrobaeza (approved):

PR #506, opened 2017-10-23
[10.0][FIX] Prevent flake8 error for non-critical bare except.

NL66278:
I am getting Travis errors on flake8 because of a bare except in the base_location_nuts module. The bare except in this particulare case should not be any problem. Added just a #noqa comment to ignore it.


pedrobaeza:
Can't we be more specific about the exception? Maybe it's `ValueError`

Yajo:
Usually it's a TypeError:

```

yajo:
Oops, butterfingers... Yep, I agree with @pedrobaeza

pedrobaeza (approved):
Fast-tracking this

PR #514, opened 2017-11-02
[9.0][FIX] partner_contact_job_position: error loading form view.

LoisRForgeFlow:
**Current behavior:**
Error raises when trying to open form view of a children partners from a company partner.
image

**Desired behavior:**
No error raises.

### DIFF
--- a/partner_contact_job_position/__openerp__.py
+++ b/partner_contact_job_position/__openerp__.py
@@ -6,7 +6,7 @@
 {
     "name": "Partner job position",
     "summary": "Categorize job positions for contacts",
-    "version": "9.0.1.0.0",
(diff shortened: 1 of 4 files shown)


yajo (approved):
Not sure why merging these 2 patches in the same PR, but in any case I like both

LoisRForgeFlow:
@yajo the second one was a minor thing I wouldn't likely open a PR for, but since I open this one and I was working with both modules I took the occasion to also fix the little string issue.

pedrobaeza:
Please change module version number in both (each one in its commit).

LoisRForgeFlow:
@pedrobaeza Done!

pedrobaeza (approved):

PR #515, opened 2017-11-02
[IMP] mute logger in test

hbrunn:
otherwise, we keep a warning in the log

### DIFF
--- a/partner_contact_birthdate/__openerp__.py
+++ b/partner_contact_birthdate/__openerp__.py
@@ -29,4 +29,5 @@
         "views/res_partner.xml",
     ],
     "post_init_hook": "post_init_hook",
+    "license": "AGPL-3",
 }
--- a/partner_contact_birthdate/hooks.py
+++ b/partner_contact_birthdate/hooks.py
@@ -7,4 +7,4 @@ from openerp.api import Environment
 def post_init_hook(cr, pool):
     env = Environment(cr, SUPERUSER_ID, {})
     env['res.partner'].search(
-        [('birthdate', "!=", False)])._birthdate_inverse()
+        [('birthdate', "!=", False)])._inverse_birthdate()
--- a/partner_contact_birthdate/models.py
+++ b/partner_contact_birthdate/models.py
@@ -32,22 +32,24 @@ class Partner(models.Model):
 
     # Make the old Char field to reflect the new Date field
     birthdate = fields.Char(
-        compute="_birthdate_compute",
-        inverse="_birthdate_inverse",
+        compute="_compute_birthdate",
+        inverse="_inverse_birthdate",
         store=True)
 
-    @api.one
+    @api.multi
     @api.depends("birthdate_date")
-    def _birthdate_compute(self):
+    def _compute_birthdate(self):
         """Store a string of the new date in the old field."""
-        self.birthdate = self.birthdate_date
+        for this in self:
+            this.birthdate = this.birthdate_date
 
-    @api.one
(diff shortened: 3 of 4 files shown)


lasley (approved):

pedrobaeza:
Which is the warning?

hbrunn:
this: https://github.com/OCA/partner-contact/blob/8.0/partner_contact_birthdate/models.py#L51
and yes, i fucked up here. that's what i meant before with making it simple for the reviewer. this PR is a very good example of how you should not do it

hbrunn:
this PR is enigmatic enough to have you asking, the title could tell more. and there's no reason advocated at all why it's a good thing that's proposed

hbrunn:
we should acknowledge that we're to some degree an old boy's network and we need to fix this

lasley:
Yeah honestly my 👍  was just based on assuming you know what you're doing + verifying that there was in fact a warning on the build that is no longer here. That's probably a bad thing in hindsight

pedrobaeza:
On v10 it has been not silently ignored the problem: https://github.com/OCA/partner-contact/blob/10.0/partner_contact_birthdate/models/res_partner.py#L33

Maybe it's better to remove that try artifact

hbrunn:
I was hesitating to do that because it's a kind of API change (thing break when assigning incorrect dates vs it's skipped silently). I don't mind which way, if @lasley also agree with this change, I'll be happy to do so

lasley:
I think I agree with @hbrunn on this one - adding an exception where something once wasn't could have cascading effects on systems that are working fine in production.

I agree from a conceptual standpoint that we shouldn't be covering the error up (v10 is the correct way), but that's the way it was released in this version so I think we need to keep with it.

pedrobaeza:
OK, but Travis still needs to be fixed...

lasley:
Hrm I didn't think new lints were supposed to fail on old branches. That's a bit annoying

pedrobaeza:
OK, if this is because the new flake8 version, we froze it, so restarting Travis should be enough.

lasley:
Restarted; we'll see!

NL66278:
@hbrunn If you put licence key in manifest as well, you should get rid of the lint error in travis

hbrunn:
thanks for the reminder, also fixed a few other lints

pedrobaeza:
Travis is still red

NL66278:
But new errors have not nothing to with this PR: 
```openerp.addons.partner_identification.tests.test_partner_identification: ` Dummy CRUD test
2018-04-11 07:14:51,228 22525 ERROR openerp_test openerp.sql_db: bad query: INSERT INTO "res_partner_id_number" ("id", "active", "partner_id", "name", "create_uid", "write_uid", "create_date", "write_date") VALUES(nextval('res_partner_id_number_id_seq'), true, 6, '1234', 1, 1, (now() at time zone 'UTC'), (now() at time zone 'UTC')) RETURNING id
File "/home/travis/odoo-8.0/openerp/sql_db.py", line 247, in execute
res = self._obj.execute(query, params)
IntegrityError: null value in column "category_id" violates not-null constraint
DETAIL:  Failing row contains (4, null, null, null, 2018-04-11 07:14:51.19689, 1234, 1, null, null, 1, 2018-04-11 07:14:51.19689, t, null, null, 6, null).```
```

NL66278 (approved):
:+1: LGTM and working in practice

StefanRijnhart (approved):

PR #521, opened 2017-11-22
[add] partner_contact_role

simahawk:
Excerpt from README:

> This module extends the functionality of partners to add roles.
> Each partner can have a job position / function but can be assigned
> to several roles / responsibilities for his/her duty.

Note: it includes already the `setup/partner_contact_role` folder for setuptools since I'm already pip installing it. Since this is usually autogenerated I can remove it if needed.


pedrobaeza:
I think this conflicts with partner_contact_job_position, isn't it?

TDu (approved):
LGTM
About the conflict with partner_contact_job_position, here we have a many2many allowing multiple role per partner which does not seem to be the case on partner_contact_job_position.

simahawk:
@pedrobaeza IMO nope. job_position works on one single position and - as I get it - it addresses another use case which is the one explained in the readme: "Each partner can have a job position / function but can be assigned to different roles". Note that "roles" is a generic term that can cover "duty" or "responsibilities" etc.

I thought about extending job_position but it would be over-complicated at the end so I decided to give users the freedom to decide if they really need a further customization on this subject.
They can work together tho.

pedrobaeza:
OK, thanks for the explanation.

partner_contact_role/README.rst
```diff
@@ -0,0 +1,71 @@
...
+
+* Go to any partner's form.
+* Use the new *Roles* field.
+
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/134/10.0
+
+.. repo_id is available in https://github.com/OCA/maintainer-tools/blob/master/tools/repos_with_ids.txt
+.. branch is "8.0" for example
```
yajo: 🔥

partner_contact_role/__manifest__.py
```diff
@@ -0,0 +1,22 @@
...
+    'depends': [
+        'sales_team',
+    ],
+    'data': [
+        'security/ir.model.access.csv',
+        'views/res_partner_role.xml',
+        'views/res_partner.xml',
+    ],
+    'demo': [
+    ],
```
yajo: 🔥

partner_contact_role/security/ir.model.access.csv
```diff
@@ -0,0 +1,5 @@
+id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
+access_res_partner_role_public,res_partner_role group_public,model_res_partner_role,base.group_public,1,0,0,0
+access_res_partner_role_portal,res_partner_role group_portal,model_res_partner_role,base.group_portal,1,0,0,0
```
yajo: I think you can just omit the group if your plan is giving everybody read access. That way you can remove these 2 lines...

partner_contact_role/security/ir.model.access.csv
```diff
@@ -0,0 +1,5 @@
...
+access_res_partner_role_group_partner_manager,res_partner_role group_partner_manager,model_res_partner_role,base.group_partner_manager,1,1,1,1
+access_res_partner_role_group_user,res_partner_role group_user,model_res_partner_role,base.group_user,1,0,0,0
```
yajo: ... and just change this one.

It's just a suggestion.

partner_contact_role/views/res_partner_role.xml
```diff
@@ -0,0 +1,36 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Copyright 2017 Camptocamp
+     License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl). -->
+
+<odoo>
+
+    <record model="ir.ui.view" id="res_partner_role_form_view">
```
yajo: Missing tree and search views.

partner_contact_role/views/res_partner.xml
```diff
@@ -0,0 +1,18 @@
...
+
+<odoo>
+
+    <record model="ir.ui.view" id="view_partner_form_role">
+        <field name="name">Partner form with roles</field>
+        <field name="model">res.partner</field>
+        <field name="inherit_id" ref="base.view_partner_form"/>
+        <field name="arch" type="xml">
+            <field name="function" position="after">
+                <field name="role_ids" widget="many2many_tags" attrs="{'invisible': [('is_company','=', True)]}" />
```
yajo: According to the invisible domain, this one seems a good candidate as submodule for `partner_contact_personal_info_page` or whatever it is called.

yajo (approved):
A couple of things to remove, and another ones to consider.

grindtildeath (approved):
Out of what's already stated by @Yajo LGTM :+1:

partner_contact_role/views/res_partner_role.xml
simahawk: search view is not needed IMO, as we have one single field `name` and search is working by default w/ it

partner_contact_role/views/res_partner.xml
simahawk: I don't think so, in the sense that this should be a prominent information right below "function"

simahawk:
Attended comments BUT personal page one :wink:

simahawk:
since the diff is small I've already squashed and rebased on 10.0.

PR #522, opened 2017-11-26
Migrate base_continent v10.0

awa-si:


base_continent/models/models.py
```diff
@@ -0,0 +1,59 @@
...
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+
+from odoo import models, fields
+
+
+class Continent(models.Model):
+    _name = 'res.continent'
```
acsonefho, line 29: The file's name should match with the name of your model. Same for the class name.
In this case, the file should be named `res_continent.py` (instead of `models.py`) and your class should be named `ResContinent` (instead of `Continent`).

base_continent/models/models.py
```diff
@@ -0,0 +1,59 @@
...
+        help="The full name of the continent.", required=True)
+
+    code = fields.Char(string='Continent Code', size=2, required=True)
+
+    countries = fields.One2many(comodel_name='res.country',
+        inverse_name='continent_id', string='Countries', readonly=True)
+
+
+class Country(models.Model):
+    _inherit = 'res.country'
```
acsonefho, line 43: Also, you should have 1 model per python file. Please split this file (don't forget to update your `__init__.py` file)

base_continent/README.rst
```diff
@@ -0,0 +1,81 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl
+   :alt: License: AGPL-3
+
+==============
+base_continent
+==============
+
+This module extends the functionality of 'base' module to support continents
+and to allow you to link each country with a continent
```
acsonefho, line 10: I have a functional question about this module:
What's the difference between this module and the Odoo standard with the `res.country.group` model (Sales/Configuration/Contact/localization/Country Group).
Into this standard model, you can create a group of country (so a continent or something else) and add many countries as needed.
It is not more simple to keep the standard instead of create this new module?

acsonefho (changes requested):
Hello,
I made few remarks and one question about this module.

emagdalenaC2i:
@pedrobaeza I think this PR should be closed because it is already merged

PR #525, opened 2017-11-29
[ADD] partner_vat_unique, check unique VAT number

ismaelcj:
Module to avoid duplicate VAT number.


partner_vat_unique/models/res_partner.py
```diff
@@ -0,0 +1,22 @@
...
+
+
+class ResPartner(models.Model):
+    _inherit = 'res.partner'
+
+    @api.constrains('vat')
+    def _check_vat_unique(self):
+        for partner in self:
+            if partner.vat:
+                duplicated_vat = self.search_count([
```
rousseldenis: @ismaelcj Maybe it could be done with a read_group before the loop as one can experiment performance problems on huge db's .

`results = self.env['res.partner'].read_group([('parent_id', '=', False), ('id', 'in', self.ids), ('vat', '!=', False)], ['vat'], ['vat'])
    for result in results:
        if result.get('vat_count') > 1:
            ...
`

feketemihai:
@ismaelcj Hi there, we have a module that is doing the same, but it's adding a pair of company_id and nrc, from Romanian localisation, please check the PR for 11.0: https://github.com/OCA/l10n-romania/pull/15/files
I think you should take in consideration at least the company_id, in a multicompany database, as long as you don't share partners between companies..

partner_vat_unique/models/res_partner.py
ismaelcj: Thanks @rousseldenis I will do the changes

ismaelcj:
Hi @feketemihai. I thought consider also the `company_id` field. But what if you share the suppliers between companies, as in my case? @rousseldenis what do you think about this?

The `nrc` is only for Romanian localisation.

rousseldenis:
@ismaelcj I think as you check through python, record rules are applied and company_id is taken into account and your code will work in case of 'specific company partners - different partners by company' or in case of shared partner through companies.

partner_vat_unique/__manifest__.py
```diff
@@ -0,0 +1,20 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017 Grant Thornton Spain - Ismael Calvo <ismael.calvo@es.gt.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+{
+    "name": "Partner VAT Unique",
+    "version": "10.0.1.0.0",
+    "category": "Customer Relationship Management",
+    "website": "https://odoo-community.org/",
```
yajo: https://github.com/OCA/partner-contact

partner_vat_unique/__manifest__.py
```diff
@@ -0,0 +1,20 @@
...
+    "author": "Grant Thornton S.L.P, Odoo Community Association (OCA)",
+    "contributors": [
```
yajo: 🔥 delete this key

partner_vat_unique/__manifest__.py
```diff
@@ -0,0 +1,20 @@
...
+        "Ismael Calvo <ismael.calvo@es.gt.com>"
+    ],
+    "license": "AGPL-3",
+    "application": True,
+    "installable": True,
+    "depends": ["base"],
+    "data": [],
+    "demo": [],
+    "qweb": []
```
yajo: 🔥 empty keys

partner_vat_unique/models/res_partner.py
```diff
@@ -0,0 +1,21 @@
...
+
+class ResPartner(models.Model):
+    _inherit = 'res.partner'
+
+    @api.constrains('vat')
+    def _check_vat_unique(self):
+        results = self.env['res.partner'].read_group([
+            ('parent_id', '=', False),
+            ('vat', 'in', map(lambda p: p.vat, filter(lambda p: p.vat, self)))
+        ], ['vat'], ['vat'])
```
yajo: Yikes, this is so hard to read that you possibly didn't notice its bugs. What is exactly the criteria you want to follow?

Bulk-checking vats would probably lead you to weird corner cases where editing 1 record in a batch that raises a false positive... Better just iterate over records and check one by one.

I guess you want to only check partners that have no parent, so first of all document that behavior, which might not be so obvious. Then start by excluding those that have a parent:

```python
for record in self:
    if record.parent_id:
        # Contacts use parent's VAT, so they are not checked
        continue
```


Next, you want to check a record doesn't exist, but you forgot to add current record's id exclusion:

```python
domain = [('parent_id', '=', False), ('vat', '=', record.vat), ('id', '!=', record.id)]
```

partner_vat_unique/models/res_partner.py
```diff
@@ -0,0 +1,21 @@
...
+        for result in results:
+            if result.get('vat_count') > 1:
+                raise ValidationError(
+                    _("The VAT must be unique per partner!"))
```
yajo: Finally, provide a specific error by saying which vat failed. Now you can know it because you are iterating. Besides, this error message is not very human-friendly. Try with "Vat %s already exists in another partner."

yajo (changes requested):

ismaelcj:
Thanks for your comments @Yajo, I just made the changes.

partner_vat_unique/README.rst
```diff
@@ -0,0 +1,49 @@
...
+============
+
+Will not check previous VAT duplicates, so it is recomended make sure there
+isn't any duplicated VAT before installation.
+
+
+Configuration
+=============
+
+Doesn't need particular configurations.
```
yajo: Then drop section.

partner_vat_unique/models/res_partner.py
```diff
@@ -0,0 +1,24 @@
...
+            if not record.parent_id:
+                results = self.env['res.partner'].read_group([
+                    ('parent_id', '=', False),
+                    ('vat', '=', record.vat),
+                    ('id', '!=', record.id)
+                ], ['vat'], ['vat'])
+            for result in results:
+                raise ValidationError(_(
+                    "The VAT %s already exists in another "
+                    "partner." % record.vat))
```
yajo: Wrong usage of `_()`

partner_vat_unique/models/res_partner.py
```diff
@@ -0,0 +1,24 @@
...
+from odoo.exceptions import ValidationError
+
+
+class ResPartner(models.Model):
+    _inherit = 'res.partner'
+
+    @api.constrains('vat')
+    def _check_vat_unique(self):
+        for record in self:
+            if not record.parent_id:
```
yajo: Change for:

```python
if record.parent_id:
    continue
```

This makes it easier to read. Did you notice that the `for result in results:` line below will be executed, no matter if this if is true or not?

Readability matters. 😉

partner_vat_unique/models/res_partner.py
```diff
@@ -0,0 +1,24 @@
...
+                results = self.env['res.partner'].read_group([
```
yajo: Do you really need a `read_group` here? You're just checking existence or not of results... `search()` should be enough.

partner_vat_unique/models/res_partner.py
```diff
@@ -0,0 +1,24 @@
...
+                    ('parent_id', '=', False),
+                    ('vat', '=', record.vat),
+                    ('id', '!=', record.id)
+                ], ['vat'], ['vat'])
+            for result in results:
```
yajo: This `for` makes no sense given what it does inside (and what I commented above). Replace by an `if`.

yajo (changes requested):

ismaelcj:
New changes for your comments @Yajo

yajo:
Nice, now please fix travis. It seems your addon violates some tests on `partner_firstname` and `base_vat_sanitized`. Thanks!

pedrobaeza:
Closing in favor of #580

PR #526, opened 2017-11-29
[11.0][MIG] partner_identification

simahawk:
squashed and reworded commits "meaningfully"

### DIFF
--- a/partner_identification/README.rst
+++ b/partner_identification/README.rst
@@ -7,11 +7,11 @@ Partner Identification Numbers
 ==============================
 
 This module allows to manage all sort of identification numbers
-and certificates which are assigned to a partner and vary from country
-to country.
+and certificates which are assigned to a partner (company or individual)
+and vary from country to country.
 
 * Commercial register
-* VAT ID (eventually)
+* VAT ID
 * Fiscal ID's
 * Membership numbers
 * Driver license
@@ -21,7 +21,7 @@ to country.
 Installation
 ============
 
-Install this addon from Settings > Local Modules, as usual.
+No specific installation step required
 
 
 Configuration
@@ -30,9 +30,13 @@ Configuration
 Configure all ID types you need in Sales > Configuration > Address Book > Partner ID Categories.
 For example, we create a category 'Driver License':
 
-* Name : Name of this ID type. For example, 'Driver License'
-* Code : Code, abbreviation or acronym of this ID type. For example, 'driver_license'
-* Python validation code: Optional python code called to validate ID numbers of this ID type.
+Name:
+  Name of this ID type. For example, 'Driver License'
+Code:
+  Code, abbreviation or acronym of this ID type. For example, 'driver_license'
+Python validation code:
+  Optional python code called to validate ID numbers of this ID type.
+
 
 Usage
 =====
@@ -40,12 +44,22 @@ Usage
 In partner form you will see another tab called 'ID Numbers'. You can add
 any IDs to this partner, defining:
 
-* Category    : ID type defined in configuration. For example, Driver License
-* ID Number   : The ID itself. For example, Driver License number of this person
-* Issued by   : Another partner, who issued this ID. For example, Traffic National Institution
-* Valid from  : Issued date. For example, date when person approved his driving exam, 21/10/2009
-* Valid until : Expiration date. For example, date when person needs to renew his driver license, 21/10/2019
-* Notes       : Any further information related with this ID. For example, vehicle types this person can drive
+Category:
+   ID type defined in configuration. For example, Driver License
+ID Number:
+  The ID itself. For example, Driver License number of this person
+Issued by:
+  Another partner, who issued this ID. For example, Traffic National Institution
+Place of Issuance:
+  The place where the ID has been issued. For example the country for passports and visa
+Valid from:
+  Issued date. For example, date when person approved his driving exam, 21/10/2009
+Valid until:
+  Expiration date. For example, date when person needs to renew his driver license, 21/10/2019
+Status:
+  ID status. For example new/to renew/expired
+Notes:
+  Any further information related with this ID. For example, vehicle types this person can drive
 
 .. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
    :alt: Try me on Runbot
@@ -55,8 +69,8 @@ any IDs to this partner, defining:
 Known issues / Roadmap
 ======================
 
-If you want to search a partner by ID you will use advance search form.
-You can't search by issuer, valid dates, category or notes.
+* If you want to search a partner by ID you will use advance search form.
+  You can't search by issuer, valid dates, category or notes.
 
 
 Bug Tracker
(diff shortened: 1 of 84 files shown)


partner_identification/README.rst
```diff
@@ -1,3 +1,4 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
 .. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
```
yvaucher: You have twice the same line
You have twice the same line

partner_identification/__manifest__.py
```diff
@@ -1,17 +1,18 @@
 # -*- coding: utf-8 -*-
```
yvaucher: header for encoding can be removed utf8 is default

yvaucher (approved):
Just some nitpicking otherwise LGTM

partner_identification/__manifest__.py
```diff
@@ -0,0 +1,32 @@
...
+        'views/res_partner_view.xml',
+        'security/ir.model.access.csv',
+    ],
+    'author': 'ChriCar Beteiligungs- und Beratungs- GmbH, '
+              'Tecnativa,'
+              'Camptocamp,'
+              'ACSONE SA/NV,'
+              'LasLabs,'
+              'Odoo Community Association (OCA)',
+    'website': 'https://odoo-community.org/',
```
elicoidal: better use the URL of the repo

partner_identification/models/__init__.py
```diff
@@ -0,0 +1,7 @@
+# -*- coding: utf-8 -*-
```
elicoidal: coding line is not necessary in v11 (P3)
remove copyright notice from `__init__.py` files too

elicoidal (approved):
some details.
LGTM

partner_identification/__manifest__.py
simahawk: sure

partner_identification/models/__init__.py
simahawk: MQT is not ready yet and logs warnings... I was trying to get a clear build. I'll remove them anyway ;)

pedrobaeza: But remember that warnings don't change green status.

partner_identification/tests/test_res_partner.py
```diff
@@ -2,77 +2,33 @@
...
-        search=lambda s, *a: s._search_identification(
-            'SSN', *a
-        ),
-    )
-
-
 class TestResPartner(common.SavepointCase):
 
-    @classmethod
-    def _init_test_model(cls, model_cls):
```
lmignon: @simahawk Why do you remove this way to create test models? What's the problem?

simahawk: @lmignon does not work on v11, it breaks transaction. In any case, my replacement sounded like less "black magic" and is working fine in a lot of modules ;)

lmignon: @simahawk It breaks transaction because one line is missing to disable the commit done by Odoo when a model is created. `self.env.cr.commit = mock.MagicMoxk()` or `self.env.cr.commit = lambda a: True` 
 With your code, after the tests the partner model is modified into database.

simahawk: what do you mean w/ modified? You run this only in test mode, so: db is a test one, everything is rolled back after tests (if you don't commit). Am I missing anything?

yajo:
Another one with truncated git history due to rename. Could you fix that please? I think a special case for this should be added to https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-11.0

pedrobaeza:
I'm not sure if that can be done. It's the only drawback of this method, but I think it's not the most important thing, as renamings should be exceptional cases (and more if everyone takes care of the names on the first review phase). Commit history is still in the corresponding branch.

yajo:
🤔 I think that if we care about history, we care about history, and if we dont, we don't.

I got the method, it's not that hard:

```bash
# New temp branch to get the commits until the rename
git checkout -b tmp1 origin/10.0

# Extract oldest commits
module=partner_identifiers
git filter-branch --force --subdirectory-filter $module

# Simplify history by setting all commits directly in the good addon name
module=partner_identification
git filter-branch --force --tree-filter "mkdir -pv $module; git mv -k * $module" HEAD

# New temp branch to get the commits from the rename
git checkout -b 11.0-$module origin/10.0

# Extract commits from rename
git filter-branch --force --subdirectory-filter $module
git filter-branch --force --tree-filter "mkdir -pv $module; git mv -k * $module" HEAD

# Get all commits together
git rebase tmp1
git branch -D tmp1

# Rebase on 11.0 and do manual removal of unneeded commits (translations, addon rename...)
git rebase -i 11.0

# Add your migration commit
git cherry-pick 260925815802f4e7cf1aece2ad7f7a090f695aac
```

With this system, we'll avoid having the problem on 12.0 also because all history will be in `partner_identification` folder 😊

pedrobaeza:
OK, I haven't dug enough. Let's try the procedure on some of the renamed modules to see if it works.

simahawk:
I'll give it a try

grindtildeath:
@simahawk Can you update commit history and squash your fixup commits ?

simahawk:
@Yajo I'm giving a try to your approach but there are no more commits before these:
```
$ g lg
*       4dc8550 2016-03-16 Laurent Mignon (ACSONE) (HEAD -> 11-mig-partner_identification-bis) [IMP] Rename partner_identifiers -> partner_identification
*       6153f76 2016-03-15 Laurent Mignon (ACSONE) [IMP] Add missing contributors in README.rst
*       3ab9f9c 2016-03-15 Laurent Mignon (ACSONE) [IMP] Add the possibility to specify a validation rule on the ID category
*       e4b926f 2016-03-15 Laurent Mignon (ACSONE) [IMP] Add dummy tests
*       c5f6a07 2016-03-15 Laurent Mignon (ACSONE) [IMP] Update to latests OCA rules
*       7958968 2016-03-15 Laurent Mignon (ACSONE) [MIG] Rename chricar_partner_id_number after extat from the CampToCamp repository
```
and if you look here https://github.com/OCA/partner-contact/commits/10.0?after=f7ec110676a490448baf08c9820cc35a3643c610+34&path%5B%5D=partner_identification is the same.
v10 has just this piece of history.... and as Pedro said: the old history is on some other branch.

pedrobaeza:
I think we can let it as it. This was more a curiosity than a real thing to be taken care of.

simahawk:
ok, I'll adapt fake models for tests and squash

simahawk:
done. I also replaced svg license image w/ png

grindtildeath (approved):

jjscarafia (approved):

partner_identification/models/res_partner.py
```diff
@@ -0,0 +1,170 @@
...
+
+    id_numbers = fields.One2many(
+        comodel_name='res.partner.id_number',
+        inverse_name='partner_id',
+        string="Identification Numbers",
+    )
+
+    @api.multi
+    @api.depends('id_numbers')
+    def _compute_identification(self, field_name, category_code):
```
yelizariev, line 24: Does it really make sense to put ``api.depends``? Will it work when we use it via lambda?
```

 compute=lambda s: s._compute_identification(
                    'social_security', 'SSN',
                ),
```

PR #529, opened 2017-12-09
Enhaced geonames 10.0

FrancoTampieri:
This Patch modify the country Model to add the properly filed for State Name and Code.

With this patch we can specify where is the right information to import from the downloaded country geonames, so modules like oca/l10n_italy/l10n_it_base_location_geonames_import can be removed and replaced with a configuration in a xml datafile like in this patch.

I have added a menu under the base_location.locations_root_menu for country to allow user to modify the value of column name and code for the correct import.


pedrobaeza:
Hi, Franco, please use the technique in https://github.com/OCA/l10n-spain/blob/10.0/l10n_es_toponyms/wizard/geonames_import.py#L23 instead of changing this.

FrancoTampieri:
@pedrobaeza u suggest me that to not extend the general country object, but extend the general wizard?

Regards

Franco

pedrobaeza:
Yeah, but on your localization module.

FrancoTampieri:
@pedrobaeza mmm but in that way, if I have a company that work with all the rest of the world I don't resolve completly the problem, because for example:

In italy, in spain the correct colum for name and code are 6 and 5, in France I see there was 4 and 3, so I have to track that changes derived from the other localizations.

But I like the Spain approach rather than the italian one https://github.com/OCA/l10n-italy/blob/3bba3e9e3880aff409ac31039a2774b2e0eb93db/l10n_it_base_location_geonames_import/geonames_import.py#L29

Do you have discussed that approach before?

FrancoTampieri:
@pedrobaeza I think that I can open a discussion in the OCA ML.
Another precision, my patch doesn't conflict with the existing *geonames addons of other the localisations

pedrobaeza:
We can use both approaches, no problem on that. What is a problem is the current implementation in Italian localization, that overwrites the configuration without taking into account other countries.

base_location_geonames_import/__init__.py
```diff
@@ -1,3 +1,4 @@
 # -*- coding: utf-8 -*-
 
-from . import wizard
+from . import models
+from . import wizard
```
pedrobaeza: Add an extra final line.

base_location_geonames_import/__manifest__.py
```diff
@@ -16,9 +16,11 @@
               'Tecnativa,'
               'Odoo Community Association (OCA)',
     'website': 'http://www.akretion.com',
-    'depends': ['base_location'],
+    'depends': ['base', 'base_location'],
```
pedrobaeza: No need to add base (it's implicit)

pedrobaeza: Change also module version number according https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#version-numbers

base_location_geonames_import/data/country.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+    <data noupdate="0">
```
pedrobaeza: No need of `<data>` tag

base_location_geonames_import/data/country.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+    <data noupdate="0">
+
+        <record id="base.it" model="res.country">
+            <field name="geonames_state_code_column" eval="6" />
+            <field name="geonames_state_name_column" eval="5" />
+        </record>
+
```
pedrobaeza: Include also the Spanish configuration as you know what it is :wink:

base_location_geonames_import/models/__init__.py
```diff
@@ -0,0 +1,5 @@
+# -*- coding: utf-8 -*-
+# Copyright 2016 Nicolas Bessi, Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from . import country
```
pedrobaeza: Line at the end

base_location_geonames_import/models/country.py
```diff
@@ -0,0 +1,19 @@
+# -*- coding: utf-8 -*-
+# Copyright 2016 Nicolas Bessi, Camptocamp SA
```
pedrobaeza: Put your copyright here

base_location_geonames_import/models/country.py
```diff
@@ -0,0 +1,19 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import models, fields
```
pedrobaeza: Sort alphabetically imports

base_location_geonames_import/models/country.py
```diff
@@ -0,0 +1,19 @@
...
+
+
+class ResCountryState(models.Model):
+
+    _inherit = 'res.country'
+
+    geonames_state_name_column = fields.Integer(
```
pedrobaeza: Remove `name` part as it makes the field name very long and doesn't add a special meaning

base_location_geonames_import/models/country.py
```diff
@@ -0,0 +1,19 @@
...
+        u'Geoname State Name Column',
+        default=3
```
pedrobaeza: Comma at the end.

base_location_geonames_import/models/country.py
```diff
@@ -0,0 +1,19 @@
...
+    )
+    geonames_state_code_column = fields.Integer(
+        u'Geoname State Code Column',
+        default=4
```
pedrobaeza: Comma at the end.

base_location_geonames_import/models/country.py
```diff
@@ -0,0 +1,19 @@
...
+
+from odoo import models, fields
+
+
+class ResCountryState(models.Model):
+
+    _inherit = 'res.country'
+
+    geonames_state_name_column = fields.Integer(
+        u'Geoname State Name Column',
```
pedrobaeza: No need of `u'..'`, as it's not an unicode character and with Python 3 this will be needed to be changed again

base_location_geonames_import/views/res_country.xml
```diff
@@ -0,0 +1,24 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
pedrobaeza: Name it `res_country_views.xml`

base_location_geonames_import/models/country.py
```diff
@@ -0,0 +1,19 @@
+# -*- coding: utf-8 -*-
```
pedrobaeza: Name it `res_country.py`

base_location_geonames_import/views/res_country.xml
```diff
@@ -0,0 +1,24 @@
...
+        <field name="type">form</field>
+        <field name="arch" type="xml">
+            <field name="code" position="after">
+                <field name="geonames_state_name_column" />
+                <field name="geonames_state_code_column" />
+            </field>
+        </field>
+    </record>
+
+    <menuitem
```
pedrobaeza: Why adding an extra menu? Leave the current one.

base_location_geonames_import/wizard/geonames_import.py
```diff
@@ -109,6 +109,10 @@ def _get_state(self, country_id, code, name):
     @api.model
     def select_or_create_state(
             self, row, country, code_row_index=4, name_row_index=3):
+        if country.geonames_state_code_column != 0:
```
pedrobaeza: No need of `!= 0` part.

pedrobaeza (changes requested):

base_location_geonames_import/models/country.py
FrancoTampieri: ok, what do u think of geonames_state_name and geonames_state_code, I've removed the _column that was redundant and not helpfull

base_location_geonames_import/views/res_country.xml
FrancoTampieri: Is to allow people to access every country item and "manually" modify the two field to fix the import of states for that country

base_location_geonames_import/models/country.py
pedrobaeza: I prefer `geonames_state_column`, which is more relevant.

pedrobaeza: Wait, I didn't remember about the code/name duality. OK then

base_location_geonames_import/views/res_country.xml
pedrobaeza: This menu should go away, as the maintenance should be done in their natural place, not here.

FrancoTampieri: Ok that's reasonable, I remove it

FrancoTampieri (commented):
I have fixed all request

base_location_geonames_import/models/res_country.py
```diff
@@ -0,0 +1,19 @@
...
+
+from odoo import fields, models
+
+
+class ResCountryState(models.Model):
+
+    _inherit = 'res.country'
+
+    geonames_state_name = fields.Integer(
+        u'Geoname State Name Column',
```
pedrobaeza: It still contains `u'..'`

FrancoTampieri: Ooops excuse me I remove it now :P

base_location_geonames_import/data/country.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="utf-8"?>
```
pedrobaeza: It should be better to call this `res_country_data.xml`, but I'm not blocking anymore.

pedrobaeza (approved):

FrancoTampieri:
@pedrobaeza Thank you veri much! I like the way u have teach me to improve the patch. I hope this patch will be merged soon.

Regards

Franco

base_location_geonames_import/models/res_country.py
```diff
@@ -0,0 +1,19 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class ResCountryState(models.Model):
+
+    _inherit = 'res.country'
+
+    geonames_state_name = fields.Integer(
```
OpenCode: I think the fields can be named "geonames_state_name_column" and "geonames_state_code_column". This avoids misunderstanding.

OpenCode (changes requested):

base_location_geonames_import/models/res_country.py
pedrobaeza: Well, I suggested the contrary because the long names, but now I don't have a strong opinion on this.

FrancoTampieri: I have no problem to restore the old field names

pedrobaeza: OK, what you prefer

FrancoTampieri:
@pedrobaeza @OpenCode I've renamed the two fields and the data file to be more compliant to the odoo/OCA guideline.

Tell me what do u think :)

pedrobaeza:
Thanks

OpenCode (approved):

base_location_geonames_import/models/res_country.py
```diff
@@ -0,0 +1,19 @@
...
+from odoo import fields, models
+
+
+class ResCountryState(models.Model):
+
+    _inherit = 'res.country'
+
+    geonames_state_name_column = fields.Integer(
+        'Geoname State Name Column',
+        default=3,
```
pedrobaeza: One last thing: please don't put defaults on these columns, and let the keyword argument to act

FrancoTampieri: @pedrobaeza ok I fix the default and push

pedrobaeza:
Thank you very much, merging! You can consider to forward port this to 11.0

PR #530, opened 2017-12-13
[9.0][IMP] partner_financial_risk: Improve multicompany cron

carlosdauden:
Force company to match property_account_receivable_id 

@Tecnativa

### DIFF
--- a/partner_financial_risk/models/res_partner.py
+++ b/partner_financial_risk/models/res_partner.py
@@ -2,6 +2,7 @@
 # Copyright 2016 Carlos Dauden <carlos.dauden@tecnativa.com>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
+from collections import defaultdict
 from datetime import datetime
 from dateutil.relativedelta import relativedelta
 from openerp import api, fields, models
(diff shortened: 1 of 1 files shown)


partner_financial_risk/models/res_partner.py
```diff
@@ -249,12 +250,24 @@ def process_unpaid_invoices(self):
         ConfigParameter = self.env['ir.config_parameter']
         last_check = ConfigParameter.get_param(
             'partner_financial_risk.last_check', default='2016-01-01')
-        move_lines = self.env['account.move.line'].search([
-            ('reconciled', '=', False),
-            ('account_id.internal_type', '=', 'receivable'),
-            ('date_maturity', '>=', last_check),
-            ('date_maturity', '<', max_date)])
-        move_lines.mapped('partner_id')._compute_risk_account_amount()
+        groups = self.env['account.move.line'].read_group(
```
pedrobaeza: Shouldn't this be process with `sudo()`?

carlosdauden: Cron execs this process with admin user, but I don't have any problem to add sudo

pedrobaeza: Being a public method that can be call from any place, I think it's convenient to be executed with `sudo`

carlosdauden:
Changes done

pedrobaeza (approved):
@sergio-teruel please review

yajo (approved):

pedrobaeza:
Please bump module version number directly, @Yajo

PR #462, opened 2017-08-28
[MIG] Migration of 'base_continent' module to v10

qtheuret:

PR #532, opened 2017-12-15
10.0 mig base continent

qtheuret:
Following problems described in https://github.com/OCA/partner-contact/pull/462 I made a new pull request to fix them.


StefanRijnhart:
Thanks for this new version! The migration seems fine now. I see you added a small functional extension: a code field on the continents. You also added codes to the continents in the data, but this won't be processed when people migrate from 9.0 to 10.0 because the data is noupdate. Can you add a migration script that adds the codes to continents upon migration to version 10.0.1.0.1?

StefanRijnhart (approved):
Thanks!

rvalyi (approved):

pedrobaeza:
Can you please squash together adjacent "OCA Transbot..." commits? The one removing en.po can be squashed together as well.

qtheuret:
Ok, done. I hope I did it correctly.

pedrobaeza:
I'm afraid not:

seleccion_003

and now there's even a lot of more commits...

qtheuret:
@pedrobaeza It seems good now but there is an error with Travis. Did I missed something ?

pedrobaeza:
The squash still is not good, as you have joined all the string at the beginning, but there should be 2 commits. You only need to squash adjacent commits.

About Travis, it was a temporary flick. I have restarted and for sure it will be correct now.

PR #548, opened 2018-01-11
set birthdate to user lang date_format if exist

mFlayyan:
the old code used to change DOB from char to Date field without taking into consideration the user language date_format , since when a char value is assigned to a Date field then the function of from_string will be called which uses the default date_format https://github.com/OCA/OCB/blob/10.0/odoo/fields.py#L1519

### DIFF
--- a/partner_contact_birthdate/models/res_partner.py
+++ b/partner_contact_birthdate/models/res_partner.py
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 # Copyright (C) 2014-2015  Grupo ESOC <www.grupoesoc.es>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from datetime import datetime
 from odoo import api, fields, models
 import logging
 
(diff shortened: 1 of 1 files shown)


partner_contact_birthdate/models/res_partner.py
```diff
@@ -30,4 +32,6 @@ def _birthdate_compute(self):
     @api.one
     def _birthdate_inverse(self):
         """Convert the old Char date to the new Date format."""
-        self.birthdate_date = self.birthdate
+        date_format = self.env['res.lang']._lang_get(
```
hbrunn: make this more charitable: Try the language's format if there is one, but also the internal one if this fails. For performance reasons, it might actually be better to first try to parse an internal date, and only afterwards try the language format. Then your code also doesn't fail one of the tests: https://travis-ci.org/OCA/partner-contact/jobs/327695653#L1176

hbrunn (changes requested):

partner_contact_birthdate/models/res_partner.py
hbrunn: oh, and be nice and don't pass through the value error, but make it a `UserError` explaining what's going on

hbrunn (approved):

yajo (approved):

yajo:
Check travis ❌

NL66278:
Trying if open/close solves travis error.

PR #550, opened 2018-01-16
[FIX] don't override defaults from context

hbrunn:
the current implementation makes it impossible to set default values for first and last name.

To me this function looks a bit botched anyways, I think it only makes sense to do anything here if `result.get('name')` is truthy. Can anyone remember why it's unconditional?

### DIFF
--- a/partner_firstname/__manifest__.py
+++ b/partner_firstname/__manifest__.py
@@ -7,7 +7,7 @@
 {
     'name': 'Partner first name and last name',
     'summary': "Split first name and last name for non company partners",
-    'version': '10.0.2.1.0',
+    'version': '10.0.2.1.1',
     'author': "Camptocamp, "
               "Grupo ESOC Ingeniería de Servicios, "
               "Tecnativa, "
--- a/partner_firstname/models/res_partner.py
+++ b/partner_firstname/models/res_partner.py
@@ -68,13 +68,16 @@ class ResPartner(models.Model):
         """Invert name when getting default values."""
         result = super(ResPartner, self).default_get(fields_list)
 
+        if not result.get('name'):
+            return result
+
         inverted = self._get_inverse_name(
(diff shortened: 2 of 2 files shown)


pedrobaeza:
Check Travis and change module version number, please.

hbrunn:
let's first discuss if lines 71ff shouldn't be under the scope of a

```python
if result.get('name'):
```

which in my opinion would make a lot of sense. @Yajo an opinion?

travis failure is [unrelated](https://travis-ci.org/OCA/partner-contact/jobs/329555575#L1666)

hbrunn:
...added my proposal without discussion

yajo (approved):
It makes sense.

pedrobaeza:
Problem comes from `partner_changeset`, that should mask its features when testing other modules. 

@yvaucher you were the one who migrates it. Can you please implement a mechanism to inhibit the behavior/raising of exception using something similar to for example: 

https://github.com/OCA/bank-payment/blob/748172f914cdcee1c4c560c65b240b436a35b7c2/account_payment_transfer_reconcile_batch/models/payment_order.py#L28-L32

yvaucher:
@pedrobaeza @hbrunn here is a fix for the issues coming from `partner_changeset` module:

https://github.com/OCA/partner-contact/pull/551

pedrobaeza:
@hbrunn I have merged the patch. Please rebase.

hbrunn:
done, even though I very much doubt this was neccessary

hbrunn:
...and turning on my brain I lost the doubt

yajo (approved):

pedrobaeza (approved):

pedrobaeza:
But I don't get it, why do you say it's not needed?  Now the PR is green...

Merging any way

pedrobaeza:
This should be forward-ported to 11.0

PR #552, opened 2018-01-22
[11.0][MIG] base_vat_sanitized

astirpe:


pedrobaeza:
Please squash together adjacent "OCA Transbot..." commits

astirpe:
Squashed!

pedrobaeza (approved):

feketemihai:
@astirpe Isn't it easier to update the default methods from Odoo, please see 
https://github.com/odoo/odoo/blob/11.0/addons/base_vat/models/res_partner.py#L106
i agree that this one is only for Europe group and i don't thing is working since on europe it will fetch datas from Vies, i checked for Romania and didn't upper the country code...but i think it's easier and we will not need another field for sanitized_vat...

astirpe:
@feketemihai this module is required by other modules (like `base_business_document_import`). I think that the reason why this module was created was to provide an alternative method to the one you mentioned. I guess the choice to keep it detached from the standard Odoo was made on purpose.
But I would turn your question to @alexis-via who is the author of this module and of the other modules that need this one.

njeudy (approved):
Ok code review + tests

alexis-via (approved):

astirpe:
@feketemihai is it ok for you to merge? Other pending PRs require this one.

feketemihai:
@astirpe Yes no problem, it was just a question if we can use vat field, not vat_sanitized.

PR #563, opened 2018-02-21
[10.0] partner_changeset - Fix condition on test_enable

yvaucher:
Context key `test_enable` will always be empty.

Thus previous patch had no effect.
Still this might speed the tests of other modules as we will only trigger changeset methods in thoses tests.

### DIFF
--- a/partner_changeset/models/res_partner.py
+++ b/partner_changeset/models/res_partner.py
@@ -3,6 +3,7 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
 from odoo import api, fields, models
+from odoo.tools import config
 
 
 class ResPartner(models.Model):
@@ -28,7 +29,7 @@ class ResPartner(models.Model):
     @api.multi
     def write(self, values):
         if (self.env.context.get('__no_changeset') or
-                self.env.context.get('test_enable') and
+                config['test_enable'] and
                 not self.env.context.get('test_partner_changeset')):
             return super(ResPartner, self).write(values)
         else:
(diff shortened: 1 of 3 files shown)


yvaucher:
See other example here:

https://github.com/OCA/account-invoicing/blob/10.0/sale_timesheet_invoice_description/models/sale.py#L59

yajo (approved):

pedrobaeza:
Why Travis is red?

yvaucher:
@pedrobaeza no idea the fixed condition seems right to me now.

By pass change set if:
Specificaly using __no_changeset
or
we are testing a module which is not this one.

Or maybe we are missing some definition of the context `test_partner_changeset`

acsonefho (approved):

grindtildeath (approved):
I approve although travis is red because `self.env.context.get('test_enable')` doesn't make much sense.

yajo (changes requested):
This PR breaks the branch.

NL66278:
@yvaucher Do you still intend to fix this branch?

yvaucher:
@NL66278 thx for the reminder I'll have to dive into that again. It has been a while.

NL66278:
Any update on this? Travis errors are in the tests for this module...

yvaucher:
@NL66278 finally fixed it, it was stupidly missing the use of the context key in the tests.

Should be green now.

:heavy_plus_sign:  rebase

yajo (approved):

pedrobaeza:
There has been a problem with merge bot, but it's fixed now, so you can relaunch.

PR #566, opened 2018-03-03
Update res_partner.xml

satriani-vai:
Company also need a fax number

### DIFF
--- a/partner_fax/views/res_partner.xml
+++ b/partner_fax/views/res_partner.xml
@@ -40,4 +40,15 @@
         </field>
     </record>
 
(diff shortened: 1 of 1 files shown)


partner_fax/views/res_partner.xml
```diff
@@ -40,4 +40,15 @@
         </field>
     </record>
 
+    <record id="view_company_form_fax" model="ir.ui.view">
```
njeudy, line 43: can you add a new file res_company_view.xml and add this code in it.

njeudy (changes requested):
Thanks for your contribution !

@satriani-vai can you keep just on PR and merge this code to #567

PR #567, opened 2018-03-03
Update res_partner.py

satriani-vai:
Fax field for company

### DIFF
--- a/partner_fax/models/res_partner.py
+++ b/partner_fax/models/res_partner.py
@@ -9,3 +9,9 @@ class ResPartner(models.Model):
     _inherit = 'res.partner'
 
     fax = fields.Char()
+
+class ResCompany(models.Model):
+
+    _inherit = 'res.company'
+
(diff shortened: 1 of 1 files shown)


partner_fax/models/res_partner.py
```diff
@@ -9,3 +9,9 @@ class ResPartner(models.Model):
     _inherit = 'res.partner'
 
     fax = fields.Char()
+
+class ResCompany(models.Model):
```
njeudy, line 13: You should create a res_company.py file and add this code on it
Update __init__.py according to this change.

njeudy (changes requested):
just code organisation

njeudy (changes requested):
could you please add README

partner_fax/__init__.py:1: [C7902(missing-readme), ] Missing ./README.rst file. Template here: https://github.com/OCA/maintainer-tools/blob/master/template/module/README.rst

PR #547, opened 2018-01-10
[ADD] partner_fax

OpenCode:
Add fax field on partner (again)

PR #572, opened 2018-03-03
[11.0] Company also need a fax number

satriani-vai:


bealdav:
Thanks a lot

Not sure it's required because, res.company have a m2o to res.partner (partner_id)

satriani-vai:
It means a fax number for the own company and not for the person.

bealdav:
Yes but partner in company is not a contact. Just define it as a company in partner form

NL66278:
@satriani-vai Please close your previous PR's that had the same intention as this one.

partner_fax/models/res_company.py
```diff
@@ -0,0 +1,8 @@
+from odoo import fields, models
+
+
+class ResCompany(models.Model):
+
+    _inherit = 'res.company'
+
+    fax = fields.Char()
```
JayVora-SerpentCS, line 8: I don't think this is needed. make a related field to partner_id like we have phone and email.
Save redundancy.

rafaelbn: I think this is the same opinion as @bealdav said in https://github.com/OCA/partner-contact/pull/572#issuecomment-370159649 . Isn't it? :thinking:

JayVora-SerpentCS: yes, Similar, we can add field, but related type only.

rafaelbn: @satriani-vai thanks for this contribution, I think recommendation from @bealdav and @JayVora-SerpentCS is correct. Don't you think so?

satriani-vai: Yes it is. I would keep fax in all contact types.

rafaelbn: Hello @satriani-vai if you change this to ma a related field to partner_id, then @JayVora-SerpentCS and @bealdav could give you the :+1 for mergin :smiley: Could you make the change please? Thank you!

SalahAdDin: @satriani-vai What happened bro?

We are still waiting for this.

SalahAdDin:
:+1:

chienandalu (commented):
This PR is still pending of the simplest change. Is it going to be finished?

emagdalenaC2i:
@pedrobaeza I think this PR should be closed because it is already merged in #547

SalahAdDin:
Thanks man!

PR #576, opened 2018-03-21
[FIX] config_parameter must be readed with sudo

etobella:
It was detected on oca/l10n-spain#797


pedrobaeza (changes requested):
Fast-tracking as obvious.

pedrobaeza (approved):

PR #582, opened 2018-04-25
[11.0] [IMP] base_location: Allow city autocompletion form to search name, city and ZIP

WohthaN:
Autocompletion form is now much more intuitive to use.

### DIFF
--- a/base_location/tests/test_base_location.py
+++ b/base_location/tests/test_base_location.py
@@ -231,6 +231,66 @@ class TestBaseLocation(TransactionCase):
             ).name
         )
 
+    def test_name_search___can_find_using_city_name_or_zip_or_code(self):
+        barcelona_data = {
+            'city_id': self.city_bcn.id,
+            'city': self.city_bcn.name,
+            'name': '444',
+            'code': 'BA',
+            'state_id': self.state_bcn.id,
+            'country_id': self.ref('base.es'),
+        }
+        madrid_data = {
+            'city_id': self.city_madrid.id,
+            'city': self.city_madrid.name,
+            'name': '555',
+            'code': 'MD',
+            'state_id': self.state_madrid.id,
+            'country_id': self.ref('base.es'),
+        }
+        lausanne_data = {
+            'city_id': self.city_lausanne.id,
+            'city': self.city_lausanne.name,
+            'name': '666',
+            'code': 'LA',
+            'state_id': self.state_vd.id,
+            'country_id': self.ref('base.ch'),
+        }
+
+        barcelona = self.env['res.better.zip'].create(barcelona_data)
+        madrid = self.env['res.better.zip'].create(madrid_data)
+        lausanne = self.env['res.better.zip'].create(lausanne_data)
+
+        found_recs = self.env['res.better.zip'].name_search(name='444')
+        self.assertEqual(len(found_recs), 1)
+        self.assertEqual(found_recs[0][0], barcelona.id)
+        found_recs = self.env['res.better.zip'].name_search(name='Barcelona')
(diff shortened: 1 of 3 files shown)


chienandalu (changes requested):
Thanks @WohthaN ! Could you improve tests coverage?

rafaelbn:
Hi @WohthaN thanks for this contribution!

As @chienandalu said, Do you know how to make test?

**codecov/patch — 33.33% of diff hit (target** 96.78%)

Let us know please!

WohthaN:
Hello!
I should be able to write the  tests, though I could use some hints on how to run all the same test pipelines (with coverage reports and such) as on the runbot locally while developing... Maybe there is some documentation out there I wasn't able to find?

pedrobaeza:
The coverage is only available here in CI tools, but check it here, and develop locally for completing lines without coverage.

base_location/tests/test_base_location.py
```diff
@@ -231,6 +231,66 @@ def test_display_name(self):
             ).name
         )
 
+    def test_name_search___can_find_using_city_name_or_zip_or_code(self):
+        barcelona_data = {
```
pedrobaeza, line 235: Don't create again this cities, but modify setUp to include missing fields like code

pedrobaeza (changes requested):

base_location/tests/test_base_location.py
WohthaN: I'm not creating the cities, I'm creating res.better.zip records in exactly the same way as the other tests are doing through `values_better_zip1` and `values_better_zip2`, reusing all the available data (and relative records) from the cities created in `setUp` and adding the missing pieces of information which are needed only for this test. 

If this approach is wrong/difficult to understand (which I do strongly agree with), the entire test suite for the module could be refactored to use a common set of data and records created in `setUp`, but that should be out of the scope of this PR.

pedrobaeza: OK, thanks for the explanation, and yes, the test suite is rather confusing, starting with the setUp method in last place. Any way, at least include that res.better.zip entries on the setUp and we are good to go.

WohthaN: None of the other tests is doing that either... why should the setup create res.better.zip records that other tests don't expect to exist?

WohthaN: Or you meant to move the city data entries there?

pedrobaeza: No, I mean to move that to the setUp. It's not mandatory. It's only a question of readibility. OK, let's keep it this way right now.

pedrobaeza (approved):

base_location/tests/test_base_location.py
WohthaN: Cool, thank you very much for the quick review!

pedrobaeza:
Why closing this?

WohthaN:
Sorry, I noticed i opened the request from the wrong branch on my repository, I'm trying to figure out how to change the source branch... Maybe you know how to do that?

pedrobaeza:
No, for that you have to make a new PR, but I don't see problem with keeping that branch...

WohthaN:
It makes my life a bit more miserable but ok.
I removed the unwanted commit from my WohthaN:zelo branch, but I don't see it disappearing from the MR, nor I can reopen it myself :(

pedrobaeza:
That's because you close and deleted the branch. Now there's no possible return. You have to make a new PR.

WohthaN:
Meh, sorry for this mess. I'm used to gitlab which allows to change both target and source branch of a merge request if needed... I'll open another MR and point to this one for history.

PR #584, opened 2018-04-29
Rename correct python dependency

kh-badep:
The correct python dependency is python-stdnum not stdnum: https://pypi.org/project/python-stdnum/

### DIFF
--- a/partner_create_by_vat/__manifest__.py
+++ b/partner_create_by_vat/__manifest__.py
@@ -15,7 +15,7 @@
     'application': False,
     'installable': True,
     'external_dependencies': {
-        'python': ['stdnum', 'suds'],
+        'python': ['python-stdnum', 'suds'],
     },
     'depends': ['base_vat'],
     'data': ['views/res_partner_view.xml'],


sbidoul:
As you can see in the travis log, the current dependency is correct. Indeed Odoo checks external python dependencies by trying an `import` instead of checking if the distribution is installed.

setuptools-odoo correctly maps stdnum to python-stdnum.

There is a problem with the `suds` dependency however, it should be mapped to `suds-jurko`. I improved setuptools-odoo for that in https://github.com/acsone/setuptools-odoo/commit/265e5859575ef53f08902072ba0e785b47e3a29e. 

So the only change needed here is a version bump to force the regeneration of the python distribution.

kh-badep:
The real issue I'm getting is that I cannot get it to install from pip:
`root@server3:~# pip install odoo10-addon-partner-create-by-vat
Looking in links: https://wheelhouse.odoo-community.org/oca
Collecting odoo10-addon-partner-create-by-vat
  Using cached https://files.pythonhosted.org/packages/c7/88/0d659a9a6ed7f00a054521af6ec56f2c4cb5cc33e524eabb3854511f5368/odoo10_addon_partner_create_by_vat-10.0.1.0.0-py2-none-any.whl
Collecting suds (from odoo10-addon-partner-create-by-vat)
  Using cached https://files.pythonhosted.org/packages/bc/d6/960acce47ee6f096345fe5a7d9be7708135fd1d0713571836f073efc7393/suds-0.4.tar.gz
Collecting stdnum (from odoo10-addon-partner-create-by-vat)
  Could not find a version that satisfies the requirement stdnum (from odoo10-addon-partner-create-by-vat) (from versions: )
No matching distribution found for stdnum (from odoo10-addon-partner-create-by-vat)
`

I thought that it was correlated to the python dependencies in the manifest.

NL66278:
Can this be closed?

PR #587, opened 2018-05-10
[11.0][ADD] New module partner_social_media

oihane:


pedrobaeza:
What do you want to put: nickname?, full URL? I think this should be also clickable for going to the corresponding profile/URL, and it should be extensible for other social providers. As is, I see this very basic module and incomplete module for serving as foundation. I also doubt about the target, as we have OCA/social that matches best.

oihane:
I have taken a look to the modules in OCA/social and they refer to mail, this will be information in the partner, that's why I have proposed it here.
This will be the first approach with the Social Media proposed by our client. I think this can be extended to show in website, as it is done with the ones in company.

Please, feel free to propose other social providers.

hieulucky111 (approved):

partner_social_media/__manifest__.py
```diff
@@ -0,0 +1,19 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+{
+    "name": "Partner Social Media",
+    "version": "11.0.1.0.0",
+    "category": "Customers, Vendors, Partners,...",
+    "license": "AGPL-3",
+    "author": "AvanzOSC, "
+              "Odoo Community Association (OCA)",
+    "website": "http://www.avanzosc.es",
```
MeetKD: website should be `https://github.com/OCA/partner-contact`

MeetKD (changes requested):

oihane:
@MeetKD change done

tarteo (changes requested):
What do think about using https://github.com/OCA/partner-contact/tree/11.0/partner_identification for the new fields?

funbaker:
Is this going to get merged?

I made a similar, not yet PR'ed change for 10.0 (Of course without checking PR's before) and now i would either like to backport this one or propose my own PR.

pedrobaeza:
Well, there are several unanswered suggestions nor any review here.

houssine78:
Hi @pedrobaeza,
What end should be given to this PR as [this](https://github.com/OCA/partner-contact/tree/10.0/partner_socialmedia) module has been merged on 10.0 branch ? We may need this features on 12.0 and we're wondering which module should be taken as candidate to be ported to 12.0. Obviously it should be the merged one but we never know so I prefer to ask ;)

houssine78:
then maybe this PR should be closed ?

pedrobaeza:
Well, both modules have the same problems that I have outlined, so take the path you consider taking that into account.

NL66278 (changes requested):
I think adding social media information can be useful. However I think it is a bad idea to add a limited list of fields to partner directly. New social media providers might come up. Not all partners will be interested in all possible social media providers. Media providers might disappear.

So it would be better, also from the viewpoint of database normalization, to have a table for social media providers (let us call it partner.social.media), and a many2many between this and res.partner.

NL66278:
Comments not answered, so closing.

PR #590, opened 2018-05-18
[10.0] [ADD] partner_identification install post-hook migration

victor-ditay:
Migration to OCA module “partner_identification”


victor-ditay:
@elicoidal

pedrobaeza:
From what module are you migrating?

victor-ditay:
>From what module are you migrating?

That is a good question
From an non OCA or standard module so I could not use upgrading features

Proof of concepts:
We expect to populate data on installation. I am already skip Demo DB and checking if table exist

Could you give advise on which kind of solutions should I try?

External scripts or other kind of imports?
Use a fork of this project on specific migration?

elicoidal:
Closing: not the right place

PR #592, opened 2018-05-28
[IMP] partner_affiliate - Changes to avoid some data being changed when swapping the parent from the Affiliate. 

BT-rmartin:
Created new Affiliate type and set it as default when creating a new affiliate.
Fixing functionality not to change address of existing company when assigning a parent

### DIFF
--- a/partner_affiliate/models/res_partner.py
+++ b/partner_affiliate/models/res_partner.py
@@ -23,3 +23,28 @@ class ResPartner(models.Model):
                                     string='Affiliates',
                                     domain=[('active', '=', True),
                                             ('is_company', '=', True)])
+
+    def get_original_address(self):
+        def convert(value):
+            return value.id if isinstance(value, models.BaseModel) else value
+
+        result = {'value': {key: convert(self[key])
+                            for key in self._address_fields()}}
+
+        return result
+
+    @api.onchange('parent_id')
+    def onchange_parent_id(self):
+        # Keep the original address info to set it back if its a company.
+        original_address = self.get_original_address()
+
+        new_partner = super(ResPartner, self).onchange_parent_id()
+
+        # When the affiliate is a company, we must set back its address
+        # because the super call changes its address by the new parent address.
+        # In addition, the type must be set to affiliate instead of contact.
+        if self.is_company:
+            new_partner.update(original_address)
+            new_partner['value'].update({'type': 'affiliate'})
+
+        return new_partner
--- a/partner_affiliate/models/res_partner.py
+++ b/partner_affiliate/models/res_partner.py
@@ -3,12 +3,14 @@
 # Copyright 2018 brain-tec AG - Raul Martin
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
-from odoo import fields, models
+from odoo import fields, models, api
 
 
 class ResPartner(models.Model):
     """Add relation affiliate_ids."""
     _inherit = "res.partner"
+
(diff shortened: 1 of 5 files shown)


BT-mfasola:
could this be merged? would definitely help!

pedrobaeza:
@yvaucher you were one of the original authors of this. What do you think?

partner_affiliate/tests/__init__.py
```diff
@@ -0,0 +1 @@
+from . import test_partner_affiliate
```
yvaucher, line 1: Missing an EOL here

partner_affiliate/models/res_partner.py
```diff
@@ -3,13 +3,15 @@
...
 
-from odoo import fields, models
+from odoo import fields, models, api
 
 
 class ResPartner(models.Model):
     """Add relation affiliate_ids."""
     _inherit = "res.partner"
 
+    type = fields.Selection(selection_add=[('affiliate', 'Affiliate')])
```
yvaucher, line 13: Please add a note in README.rst to explain there that there is a new type.

yvaucher (approved):
Thanks for that improvement!

Makes sense to me to have a type for affiliate as `other` is not helpful.

Plus, disabling the copy of address seems reasonable as you don't want to erase your address if you already entered it. And address would more likely be different anyway.

Just one remark, can you update the readme with a note about the type of partner?

Plus a little nitpicking on a missing EOF.

Already approving as required changes are minor.

partner_affiliate/models/res_partner.py
```diff
@@ -21,3 +23,28 @@ class ResPartner(models.Model):
...
+    def onchange_parent_id(self):
+        # Keep the original address info to set it back if its a company.
+        original_address = self.get_original_address()
+
+        new_partner = super(ResPartner, self).onchange_parent_id()
+
+        # When the affiliate is a company, we must set back its address
+        # because the super call changes its address by the new parent address.
+        # In addition, the type must be set to affiliate instead of contact.
+        if self.is_company:
```
katyukha, line 46: I think it is better to use `if new_partner and self.is_company:` condition, because `onchange_parent_id` may return `None`

katyukha:
It would be good to add following in partner's form view:
```diff
diff --git a/partner_affiliate/views/res_partner_view.xml b/partner_affiliate/views/res_partner_view.xml
index beeb9a45..3f1faa5f 100644
--- a/partner_affiliate/views/res_partner_view.xml
+++ b/partner_affiliate/views/res_partner_view.xml
@@ -10,6 +10,9 @@
             <xpath expr="/form/sheet//div[hasclass('o_row')]/field[@name='parent_id']" position="attributes">
                 <attribute name="attrs"></attribute>
             </xpath>
+            <xpath expr="/form/sheet//group/group/field[@name='vat']" position="attributes">
+                <attribute name="attrs">{'readonly': [('type', '!=', 'affiliate'), ('parent_id','!=',False)]}</attribute>
+            </xpath>
             <xpath expr='//page[@name="internal_notes"]' position="before">
                 <page string="Affiliates" attrs="{'invisible': [('is_company','=',False)]}">
                     <field name="affiliate_ids"

```

This change works good in 12.0 too.

NL66278:
@BT-rmartin Could you take up the requested changes? Not too much work I think, and then I could approve and merge your PR.

yvaucher:
@NL66278 feel free to take over that PR by duplicating the branch, it has been more than a year that it didn't changed.

NL66278:
Closed as review comments taken up in: https://github.com/OCA/partner-contact/pull/897

PR #594, opened 2018-06-08
[FIX] fix name constrains

mourad-ehm:
If we create contact from company partner, the type is 'contact'. But if we create a partner directly from customer menu this type is "false" and constraint doesn't work.
Otherwise, in my case I have a strange bug where First-name and last-name are not required (in form view) but only for 1 partner. This partner was created by an user (without any name and I get error on bank reconciliation). I can't reproduce this issue and don't know what way was used to create it. (when I change partner First-name and last name become required again !!)
image

### DIFF
--- a/partner_firstname/exceptions.py
+++ b/partner_firstname/exceptions.py
@@ -8,5 +8,7 @@ class EmptyNamesError(exceptions.ValidationError):
     def __init__(self, record, value=_("No name is set.")):
         self.record = record
         self._value = value
-        self.name = _("Error(s) with partner %d's name.") % record.id
+        self.name = _(
+            "Error(s): at least one name (name, firstname or "
+            "lastname)  is mandatory for partner (id : %d).") % record.id
         self.args = (self.name, value)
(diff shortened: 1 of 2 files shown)


mourad-ehm:
ping @bealdav , @florian-dacosta, @pedrobaeza

bealdav:
a lot of check fails but not about this PR

bealdav (approved):
A really important bug fixed. Thanks

mourad-ehm:
When I analyse the errors in the travis, I found some inconsistency in test and (and constraint).
And I understood what way was used by user to create partner without name. 
1 . From an company we add a contact with type = 'delivery' or "invoice".
2. From the menu we search for this address. and we remove parent company, then we get and address with out name.

We must have name if there is no parent_id

rousseldenis:
@mourad-ehm Any news on this ?

NL66278:
Question not answered. Unsupported version of Odoo. Therefore closing.

PR #599, opened 2018-06-18
[FIX] partner_contact_nutrition: correct lint error

cubells:
cc @Tecnativa


pedrobaeza:
Isn't there 2 errors?

cubells:
There is one error in two different modules.

pedrobaeza:
Yes, but you can include both in this PR. No need of 2 separate PRs. We won't see a green status in neither of both.

cubells:
done!

pedrobaeza:
You can now rebase the other

PR #603, opened 2018-06-29
[10.0] partner_email_check: Multiple email check

Abranes:
Check multiple emails separated by semicolon.

### DIFF
--- a/partner_email_check/models/res_partner.py
+++ b/partner_email_check/models/res_partner.py
@@ -29,8 +29,10 @@ class ResPartner(models.Model):
             self.email_check(rec.email)
 
     @api.model
-    def email_check(self, email):
-        if validate_email(email):
-            return True
-        else:
-            raise UserError(_('Invalid e-mail!'))
+    def email_check(self, emails):
+        for email in emails.split(';'):
+            if not validate_email(email):
+                raise UserError(
+                    _("%s is an invalid email") % email.strip()
+                )
+        return True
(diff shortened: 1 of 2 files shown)


pedrobaeza:
`;` is not the only possible separator. @Yajo isn't standard RFC already supports multiple emails and its separators?

fuentes73:
AFAIK Odoo is handling this job, it gets partner emails and uses email.utils [formataddr](https://docs.python.org/2/library/email.util.html#email.utils.formataddr) and  [getaddresses](https://docs.python.org/2/library/email.util.html#email.utils.getaddresses) from python in order to build a valid (RFC) To or Cc headers.

We've tested and it works, if you have multiple emails in a partner saparated by `;` Odoo sends an email to those emails.

pedrobaeza:
Yeah, but `,` also works as valid email separator. Does module in previous state fail with commas?

fuentes73:
Yes, it failed without this changes.
I think that / it's also a valid separator. Abraham can handle those characters too. What do you think?

pedrobaeza:
That's why I ask for the valid separators according RFC standard, so that we handle all of them at once.

yajo:
https://stackoverflow.com/questions/12120190/what-is-the-best-separator-to-separate-multiple-emails#12120203

fuentes73:
@Abranes please, change the code to use as separator commas instead of semicolon.

Abranes:
@fuentes010 Done! Now the only accepted separator is comma.

Abranes:
@pedrobaeza @Yajo @fuentes010 changed the separator according RFC standard.
Any other suggestions for change or is ready to merge?

pedrobaeza (approved):

fuentes73 (approved):

PR #609, opened 2018-07-13
[10.0] base_location: improve res.partner form view

alexis-via:
zip_id readonly when the partner is a contact of a parent partner like the other address fields, cf https://github.com/odoo/odoo/blob/10.0/odoo/addons/base/res/res_partner_view.xml#L169

### DIFF
--- a/base_location/__manifest__.py
+++ b/base_location/__manifest__.py
@@ -5,7 +5,7 @@
 
 {
     'name': 'Location management (aka Better ZIP)',
-    'version': '10.0.1.0.2',
+    'version': '10.0.1.0.3',
     'depends': [
         'base',
     ],
--- a/base_location/__manifest__.py
+++ b/base_location/__manifest__.py
@@ -5,7 +5,7 @@
 
 {
     'name': 'Location management (aka Better ZIP)',
-    'version': '10.0.1.0.3',
+    'version': '10.0.1.0.4',
     'depends': [
         'base',
     ],
--- a/base_location/views/partner_view.xml
+++ b/base_location/views/partner_view.xml
@@ -8,6 +8,7 @@
         <field name="arch" type="xml">
             <field name="city" position="before">
                 <field name="zip_id"
+                    attrs="{'readonly': [('type', '=', 'contact'), ('parent_id', '!=', False)]}"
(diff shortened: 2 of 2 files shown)


pedrobaeza (approved):
As always, please change module version number

alexis-via:
This is really a big big pain, creates merge conflicts, etc... we need to find a better way.

pedrobaeza:
Well, only if there are several PRs for the same module, but I agree with you it's not the ideal. We want to put this to be done by an OCA bot, but it's not direct also to be done. Step by step...

By the way, I force the change in the module version number for the pip packages generation. For other things, we can miss it.

alexis-via:
Travis error is unrelated

sbidoul:
It's not really necessary for pip package generation, because it autoincrement versions by counting commits since the last change in the manifest.

IMO bumping version is the role of maintainer not contributors.

pedrobaeza:
@sbidoul I remember you told me about the version number for generating a proper package. If not, there was something extra to do (maybe `--pre` or similar?).

sbidoul:
@pedrobaeza Currently there is actually little chance to get a "round" version number (ie without the .devN) on pypi, because between the moment you bump the manifest until it gets published, there are all sorts of additional commits (readme generator, translations). So you always want to use `pip install --pre` to get the last version. 

So yeah, bumping version is important for semver, but our merging==releasing does not mesh well with that. Asking contributors to bump is not ideal. Many addons have no regular maintainer that care about versions, so expecting maintainers to do it is not ideal either.

Not an easy topic...

pedrobaeza:
So if you say me that it's not needed at all (or in practical terms, you should use the alternate method for installing latest version with pip - you know that I'm not a pip user), then I don't ask anymore to bump the version. Is that or am I interpreting something wrong?

pedrobaeza:
@BT-astauder is pointing another reason for bumping the version number in https://github.com/OCA/account-financial-reporting/pull/440#issuecomment-405186556

alexis-via:
@pedrobaeza version number updated + rebase done

pedrobaeza:
OK, let's merge then as obvious although only 1 reviewer.

PR #605, opened 2018-07-09
[8.0][MIG] Backport of partner_coc

pankk:
Backport of the 10.0 module partner_coc.

Note that this depends on #612.

PR #436, opened 2017-06-07
[FIX] partner_identification: Infinite loop in search

lasley:
This PR reimplements the partner_identification search method to fix #434

Issue #434, opened 2017-06-01
[10.0] partner_identification: Infinite loop for _search_identification


moylop260:
The following test has a [search](https://github.com/OCA/partner-contact/blob/1bc94dff59c405c4ac6c3eecb6ff2499accfacb1/partner_identification/tests/test_res_partner.py#L123) with a infinite loop computing the domain from [odoo/osv/expression.py#L781](https://github.com/odoo/odoo/blob/79477b8d96be47cca5609ce5e4dd84a7e7f64f0a/odoo/osv/expression.py#L781)

`category_id.code='SSN'` increase for each iteration of the loop:
![screen shot 2017-06-01 at 12 11 13](https://cloud.githubusercontent.com/assets/6644187/26691178/cececcc2-46c1-11e7-9807-fe7f2b3cdc47.png)

There is a issue using the following field:
https://github.com/OCA/partner-contact/blob/1bc94dff59c405c4ac6c3eecb6ff2499accfacb1/partner_identification/tests/test_res_partner.py#L13-L23

And using the following method:
https://github.com/OCA/partner-contact/blob/1bc94dff59c405c4ac6c3eecb6ff2499accfacb1/partner_identification/models/res_partner.py#L139

More info about: https://github.com/OCA/maintainer-quality-tools/issues/448#issuecomment-305477622

PR related
https://github.com/OCA/partner-contact/pull/419

yajo:
Thanks @moylop260! Great debugging!

Hi there @lasley, do you think you can fix this, given https://github.com/OCA/partner-contact/pull/419 was yours?

pedrobaeza:
@lasley any news about this bug?

lasley:
Been in and out of town (Pycon + a vacation), so wasn't able to hit this yet. I'm starting to catch up on workload though, so I should have something here soon.

PR #612, opened 2018-07-19
[8.0][MIG] partner_identification backport

pankk:
Backport of partner_identification, based on the 10.0 version which adds additional features. These features are required for the partner_coc PR here: https://github.com/OCA/partner-contact/pull/605

### DIFF
--- a/partner_identification/__openerp__.py
+++ b/partner_identification/__openerp__.py
@@ -11,7 +11,7 @@
 {
     'name': 'Partner Identification Numbers',
     'category': 'Customer Relationship Management',
-    'version': '8.0.1.0.0',
+    'version': '8.0.1.1.1',
     'data': [
         'views/res_partner_id_category_view.xml',
         'views/res_partner_id_number_view.xml',
@@ -19,10 +19,11 @@
         'security/ir.model.access.csv',
     ],
     'author': 'ChriCar Beteiligungs- und Beratungs- GmbH, '
-              'Antiun Ingeniería S.L.',
+              'Tecnativa,'
               'Camptocamp,'
               'ACSONE SA/NV,'
-              'Odoo Community Association (OCA)'
+              'LasLabs,'
+              'Odoo Community Association (OCA)',
     'website': 'https://odoo-community.org/',
     'license': 'AGPL-3',
     'installable': True,
--- a/partner_identification/models/res_partner.py
+++ b/partner_identification/models/res_partner.py
@@ -8,12 +8,165 @@
 #        Antonio Espinosa <antonioea@antiun.com>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
-from openerp import models, fields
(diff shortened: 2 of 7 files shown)


partner_identification/__openerp__.py
```diff
@@ -11,12 +11,13 @@
...
+    "depends": ["base"],
     'data': [
+        'security/ir.model.access.csv',
         'views/res_partner_id_category_view.xml',
         'views/res_partner_id_number_view.xml',
         'views/res_partner_view.xml',
-        'security/ir.model.access.csv',
     ],
     'author': 'ChriCar Beteiligungs- und Beratungs- GmbH, '
               'Antiun Ingeniería S.L.',
```
tarteo: Can you move the comma inside the string

tarteo (changes requested):
Can you also fix this:
partner_identification/README.rst:18: [E7901(rst-syntax-error), ]  Unexpected possible title overline or transition.|Treating it as ordinary text because it's so short.

partner_identification/__openerp__.py
pankk: Thanks @tarteo, I will cherry-pick the commits from 11.0 that solve this (and some other) issues.

partner_identification/models/res_partner.py
```diff
@@ -8,12 +8,164 @@
...
+
+        Args:
+            category_code (str): Category code of the Identification type.
+            operator (str): Operator of domain.
+            value (str): Value to search for.
+
+        Returns:
+            list: Domain to search with.
+        """
+        id_numbers = self.env['res.partner.id_number'].search([
```
rousseldenis, line 166: @pankk Avoid this (I know this is on v11 - I'll try to fix it later).

It's not necessary to do a search before returning (two queries instead of one).
```

return [
    ('id_numbers.name', operator, value),
    ('id_numbers.category_id.code', '=', category_code),
]
```
It will impact performances.

rousseldenis (changes requested):
Code review

partner_identification/__openerp__.py
pedrobaeza: Can you please update `Antiun` author to the new company `Tecnativa`?

pankk: @pedrobaeza Yes, I've seen commits that push that change and will cherry-pick them. Thanks!

partner_identification/models/res_partner.py
pankk: I get #434 when doing it like this. Can we go with @pedrobaeza's suggestion in #436?
image

rousseldenis: @pankk Yes, I was not aware of that bug. Ok for me

tarteo (approved):
Lint error seems not to be related to this module.

andhit-r (approved):

pedrobaeza:
Merging as the lint error is unrelated

PR #614, opened 2018-08-02
Coc filter working

dave-prosee:
In Sales/Customer using custom filter on CoC registration number gives a dump 
because *a propagates 2 arguments (operator and value) resulting in total of 5 while the 
definition in partner-identification res.partner.py is:
_search_identification(self, category_code, operator, value)
Hence the 'coc_registration_number' is superfluous.


pankk (approved):
This looks really familiar, probably because I did this fix in #605. I seem to have forgotten to propose a fix for this version. Thank you for doing it.

pedrobaeza (approved):

PR #509, opened 2017-10-31
11.0 mig partner contact birthdate

jprobineau:

PR #617, opened 2018-08-20
[11.0][MIG] partner contact birthdate

Olivier-LAURENT:
* Complete #509 by removing legacy birthdate field because it has been removed by Odoo since V10
* Add setup


hieulucky111 (approved):

partner_contact_birthdate/views/res_partner.xml
```diff
@@ -0,0 +1,18 @@
...
+<!-- License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+     © 2015 Grupo ESOC Ingeniería de Servicios, S.L.U. -->
+
+<odoo>
+    <record id="view_personal_information_birthdate_date" model="ir.ui.view">
+        <field name="name">Birthdate Date field</field>
+        <field name="model">res.partner</field>
+        <field name="inherit_id" ref="partner_contact_personal_information_page.personal_information"/>
+        <field name="arch" type="xml">
+            <data>
```
nikul-serpentcs: @Olivier-LAURENT Why used `data` tag?

pedrobaeza:
I'm not sure about depending on partner_contact_personal_information_page and having only this field in that page. What do you think?

Olivier-LAURENT:
Hi @pedrobaeza 
I noticed that some other modules `partner_contact_*` have the same approach.
To put the field on a view an anchor is needed. Why not this one ?
Using a tab for such additional fields allows to not consume vertical room in the main area of the form.

partner_contact_birthdate/__manifest__.py
```diff
@@ -0,0 +1,20 @@
...
+# © 2017-Apertoso N.V. (<http://www.apertoso.be>)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+{
+    "name": "Contact's birthdate",
+    'version': '11.0.1.0.0',
+    "author": "Tecnativa,"
+              "Odoo Community Association (OCA)",
+    "category": "Customer Relationship Management",
+    "website": "https://www.tecnativa.com/",
```
nikul-serpentcs: @Olivier-LAURENT IMHO Please change URL as described in  https://github.com/OCA/maintainer-tools/blob/master/template/module/__manifest__.py#L10

ghost: @Olivier-LAURENT @pedrobaeza could this module merged?

pedrobaeza (changes requested):
I still prefer to add the field on the proper form (which is the contact popup form, or in the main area when you mark as individual the partner).

ghost:
If you have a lot of personal information a new tab would be nice. I would prefer the main tab too, but its pitty that this question is the matter not to merge the module (since a half year?)

OsoTranquilo:
We use this module habitually, for our developments for hotels, along with others also included within partner-contact, (partner_contact_gender, etc), and we think it is correct to do it in a tab. What is more personal information than the birthday?

I think you have to allow this merged already.

OsoTranquilo (approved):
I have reviewed it and it works perfectly.

pedrobaeza:
My comment still applies about not adding the page.

Olivier-LAURENT:
Hi @pedrobaeza , But what is therefore the goal of this additional addons 
`partner_contact_personal_information_page` ?

Approach is exactly the same as in the HR employee form. All personal data (gender, nationality, birth date, ...) are collected together on a single tab. This especially allows to set a visibility rule on all personal fields in one go.

This PR is just a migration, not a brainstorming on where to put this field on the partner form from now on. 

Thus I suggest to merge (this work is pending since 4 months !). Adventure can of course continue in another issues or PRs.

pedrobaeza:
Sorry but no, as having a single field requiring 2 modules is a total bloated thing. Migration PRs are for improving modules that are not good in previous versions for being better in new version. If you don't want to do it, I can do it. And you can't compare employees with contacts, as the layout of the forms and structure (parent/child) has nothing in common with the other.

OsoTranquilo:
In version 11 the item for 'gender', for example, is already available in the same table, is the same problem, and if the rest of modules do not enter here, it will be worse.
I understand what you say, and I think the criterion would be to unify this, in the partner form.
But, @pedrobaeza, I think we should consider changing the criteria for all the affected fields in V12 if it is the general vision, and unclogging this PR.

pedrobaeza:
I only want this field, not gender and so on, so I don't mind about the rest how they were migrated. If we can improve this one already, on v12 it will be correct also. For the rest, we can do the same in v12 if v11 is too late.

ghost:
@Olivier-LAURENT could you make @pedrobaeza happy and change it to the main page please.

This would be very, very nice, to use this module and field from oca-repo. I use it in a eventmodule that i want to submit to oca.

Olivier-LAURENT:
Hi @Axel-G ,

It seems my last push a week ago was doing the job.

ghost:
oh great i fail to see this commit. great. Dear @pedrobaeza now its ready to merge, isnt it?

pedrobaeza (approved):
I have optimized the commit history grouping together by author/logic set. Merging when CIs finish.

PR #615, opened 2018-08-05
[9.0] partner_sector: allow to use in individuals

JordiBForgeFlow:
By default only companies have sectors. To activate sectors in individuals also,
you need to activate the following setting:

* Go to *Settings > General Settings > Sector in contacts > Use sector in individuals*

Or, if you have *Sales* or *CRM* modules installed:

* Go to *Sales > Configuration > Settings > Partner Sector > Use sector in individuals*

PR #619, opened 2018-08-29
[11.0][MIG] partner_sector > partner_industry_secondary

MiquelRForgeFlow:
This migration includes:

- Changed dependency.
- Adapted to `res.partner.industry`
- Module renamed.

Added:

- [x] https://github.com/OCA/partner-contact/pull/622.


pedrobaeza:
Shouldn't this use now existing Odoo information about sectors and the existing many2one (creating only the many2many?

MiquelRForgeFlow:
In which part of Odoo are the sectors and the many2one you are referring?

pedrobaeza:
The field is `industry_id`, and the model `res.partner.industry`. You can see them for example on http://367102-11-0-683eca.runbot12.odoo.com/web?debug#view_type=list&model=res.partner.industry&menu_id=76&action=54

MiquelRForgeFlow:
Renamed the module to **partner_industry**.

pedrobaeza:
As you are renaming, maybe it's better to call it `partner_industry_secondary`. We will need migration scripts for converting sector records to industry records (maybe doing a matching over some known values that are standarized).

MiquelRForgeFlow:
However, this module not only adds secondary industries, also enables to add the industry fields in the partner views.

pedrobaeza:
> However, this module not only adds secondary industries, also enables to add the industry fields in the partner views.

Are you sure they are not added in core?

MiquelRForgeFlow:
> > However, this module not only adds secondary industries, also enables to add the industry fields in the partner views.
> 
> Are you sure they are not added in core?

Supersure. All odoo modules that use this model are base, contacts and website_customer. And neither of the three has that partner view connection. Really, I don't understand why odoo sometimes when they add new things they lay them unfinished like in this case.

LoisRForgeFlow:
Have you considered the changes done in https://github.com/OCA/partner-contact/pull/615? Sector (now industry) should be visible in individuals if configured to.

LoisRForgeFlow:
I would test it, but runbot is not accesible, this is happening in a lot of PR :disappointed: despite they are green

MiquelRForgeFlow:
> Have you considered the changes done in #615? Sector (now industry) should be visible in individuals if configured to.

No, I didn't remember. But I think that improvement should be ported first to v10.

MiquelRForgeFlow:
Oh. Found industry_id in partner view (in sale_purchase tab, in misc category). I will fix that.

partner_industry_secondary/views/res_partner_industry_view.xml
```diff
@@ -0,0 +1,32 @@
...
+            </field>
+        </field>
+    </record>
+
+    <record id="res_partner_industry_view_tree" model="ir.ui.view">
+        <field name="name">Industry</field>
+        <field name="model">res.partner.industry</field>
+        <field name="inherit_id" ref="base.res_partner_industry_view_tree"/>
+        <field name="arch" type="xml">
+            <!--<xpath expr="//tree" position="attributes">-->
```
nikul-serpentcs: @mreficent IMHO Remove dead code.

MiquelRForgeFlow:
I cherry-picked https://github.com/OCA/partner-contact/pull/622.

Please, could everyone update their review?

@nikul-serpentcs @lreficent @pedrobaeza @jbeficent

LoisRForgeFlow:
Can you check the option for individuals, I'm not able to see the sector for them even after enabling the setting

MiquelRForgeFlow:
@lreficent now enabling in the settings should work for individuals 😋

LoisRForgeFlow:
> @lreficent now enabling in the settings should work for individuals yum

No, that's not true :suspect:

MiquelRForgeFlow:
@lreficent Ok, I thought I had solved it before  😂 Now it's solved, truly, I have tested it. 👨‍⚖️

LoisRForgeFlow (approved):
Finally it works! :clap: :tada: :confetti_ball::confetti_ball:

nikul-serpentcs (approved):
LGTM :+1:

partner_industry_secondary/models/res_partner_industry.py
```diff
@@ -0,0 +1,41 @@
...
+    def name_get(self):
+        def get_names(cat):
+            """ Return the list [cat.name, cat.parent_id.name, ...] """
+            res = []
+            while cat:
+                res.append(cat.name)
+                cat = cat.parent_id
+            return res
+
+        return [(cat.id, " / ".join(reversed(get_names(cat)))) for cat in self]
```
yajo: So, you do a loop of a reversed loop of a loop... This sounds like a good candidate for a performance bottleneck. :thinking: 

I suggest:

```python
display_name = fields.Char(compute="_compute_display_name", store=True, index=True)

@api.depends("name", "parent_id")
def _compute_display_name(self):
    # Make your loops here; then:
    self.mapped("child_ids")._compute_display_name()
```

Storing the field will make the hyperloop only happen once on changes, and will allow better searching.

yajo (approved):
Just one suggestion.

partner_industry_secondary/models/res_partner_industry.py
MiquelRForgeFlow: Creating new industries is not something you do usually, just one day in your life. So, I think it's not justified the full performance approach by storing a <code>display_name</code> field. Also, maybe it would need a pre-migration script or a <code>pre_init_hook</code> to fill this field for those that have already using this module without this new field.

At least, I improved the performance by deleting the reversed loop.

JordiBForgeFlow:
@mreficent as discussed, please add migration script.

partner_industry_secondary/models/res_partner_industry.py
yajo: Thanks for the improvement!

> Creating new industries is not something you do usually [...] it's not justified

I'm not sure I understand your reasoning... :roll_eyes: 

That's exactly the reason why it's a good idea, don't you think? Right now you are looping over loops for each *displayed* record; by storing the value, you'd only loop when *creating* a record (which happens not much often) or when *writing* it (possibly even less often). *Reading* (most often) would be as fast as any other field.

> maybe it would need a pre-migration script

:thinking: Indeed, I didn't think of it before. Not much hard to do after all.

---

Well, in any case, you got my :heavy_check_mark: already. This can be improved later also. Do as you please :wink:

sebalix (approved):

partner_industry_secondary/migrations/11.0.1.0.0/pre-migration.py
```diff
@@ -0,0 +1,34 @@
+# Copyright 2018 Eficent Business and IT Consulting Services, S.L.
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+from openupgradelib import openupgrade
+
+_column_renames = {
+    'res_partner': [('sector_id', None)],
+}
+
+_field_adds = [
```
pedrobaeza: This is not needed. When you update the module, Odoo will add these fields from the models definition.

partner_industry_secondary/migrations/11.0.1.0.0/pre-migration.py
```diff
@@ -0,0 +1,34 @@
...
+    cr = env.cr
+    if openupgrade.table_exists(cr, 'res_partner_sector'):
+        openupgrade.add_fields(env, _field_adds)
+        openupgrade.rename_columns(cr, _column_renames)
+        openupgrade.rename_xmlids(cr, _xmlid_renames)
+        cr.execute(
+            """
+            ALTER TABLE res_partner_industry
+            ADD COLUMN %s integer;
+            """ % openupgrade.get_legacy_name('old_id')
```
pedrobaeza: A better name can be `old_sector_id` (combined with the legacy name)

partner_industry_secondary/migrations/11.0.1.0.0/post-migration.py
```diff
@@ -0,0 +1,64 @@
...
+from openupgradelib import openupgrade
+
+
+def fill_res_partner_industry(cr):
+    cr.execute(
+        """
+        INSERT INTO res_partner_industry (name, active, create_uid,
+            create_date, write_uid, write_date, parent_id, %s)
+        SELECT name, TRUE as active, create_uid, create_date, write_uid,
+            write_date, parent_id, id as %s
```
pedrobaeza: No need here to put the `as %s`, as field selection is done per order

pedrobaeza: Remove `parent_id` from this sentence, or you will have foreign key errors

partner_industry_secondary/migrations/11.0.1.0.0/post-migration.py
```diff
@@ -0,0 +1,64 @@
...
+    )
+
+
+def update_res_partner_industry(cr):
+    cr.execute(
+        """
+        UPDATE res_partner rp
+        SET industry_id = rpi.id
+        FROM res_partner_industry rpi
+        WHERE rp.%s IS NOT NULL AND rp.%s = rpi.%s
```
pedrobaeza, line 38: No need of the `IS NOT NULL` part

partner_industry_secondary/migrations/11.0.1.0.0/post-migration.py
```diff
@@ -0,0 +1,64 @@
...
+        )
+    )
+
+
+def fill_res_partner_industry_secondary(cr):
+    cr.execute(
+        """
+        INSERT INTO res_partner_res_partner_industry_rel (res_partner_id,
+            res_partner_industry_id)
+        SELECT rel.res_partner_id, rpi.id as res_partner_industry_id
```
pedrobaeza: No need of the `as res_partner_industry_id` part

pedrobaeza (changes requested):

partner_industry_secondary/migrations/11.0.1.0.0/post-migration.py
MiquelRForgeFlow: I don't have `parent_id` errors because they are corrected in line 22

partner_industry_secondary/migrations/11.0.1.0.0/post-migration.py
MiquelRForgeFlow: It's needed, because in "`AND rp.%s = rpi.%s`" you don't want `NULL = NULL` that returns `TRUE`

partner_industry_secondary/migrations/11.0.1.0.0/post-migration.py
MiquelRForgeFlow: But also no need to take it out

partner_industry_secondary/migrations/11.0.1.0.0/post-migration.py
MiquelRForgeFlow: > No need here to put the as %s, as field selection is done per order

But also no need to take it out

partner_industry_secondary/migrations/11.0.1.0.0/pre-migration.py
MiquelRForgeFlow: Yeah, I know, and I don't remember why I put them there.

partner_industry_secondary/migrations/11.0.1.0.0/pre-migration.py
MiquelRForgeFlow: Aesthetics  😏 But I like your suggestion.

partner_industry_secondary/migrations/11.0.1.0.0/pre-migration.py
MiquelRForgeFlow: Removed.

partner_industry_secondary/migrations/11.0.1.0.0/post-migration.py
pedrobaeza: Yeah, but you improve a lot the size of the query (that is logged), the readability, etc. I'm telling you this also as a didactic exercise for you to know it.

partner_industry_secondary/migrations/11.0.1.0.0/post-migration.py
pedrobaeza: > I don't have parent_id errors because they are corrected in line 22

You will have them in this statement, as the foreign key is forced in both queries. You didn't have previously the problem as the fields were created by hand (with the `add_fields` openupgradelib method). Any way, why leave it if you totally overwrite it later?

pedrobaeza: > But also no need to take it out

The same as in the other place: it's better for readability and so on.

partner_industry_secondary/migrations/11.0.1.0.0/post-migration.py
pedrobaeza: There will be previous values in `res_partner_industry` before this migration? If so, then you are right, we should keep it. Another tip: you can use keyword replacements instead of '%s' for not needing to repeat the same argument.

partner_industry_secondary/migrations/11.0.1.0.0/post-migration.py
MiquelRForgeFlow: But this one is not logged :S

And the readability is subjective. I think with the "as xx" is more understandable.

pedrobaeza: I don't agree, but I'm not going to force you on this, but please remove it in the other that you need an extra argument for formatting it.

partner_industry_secondary/migrations/11.0.1.0.0/post-migration.py
MiquelRForgeFlow: > You didn't have previously the problem as the fields were created by hand (with the add_fields openupgradelib method)

See! That's why I shouldn't have deleted the `add_fields` 😜

pedrobaeza: hahaha

MiquelRForgeFlow:
Please @pedrobaeza update your review

partner_industry_secondary/migrations/11.0.1.0.0/post-migration.py
```diff
@@ -0,0 +1,67 @@
...
+
+def fill_res_partner_industry(cr):
+    cr.execute(
+        """
+        INSERT INTO res_partner_industry (name, active, create_uid,
+            create_date, write_uid, write_date, %s)
+        SELECT name, TRUE as active, create_uid, create_date, write_uid,
+            write_date, id
+        FROM res_partner_sector
+        """ % openupgrade.get_legacy_name('old_sector_id'),
```
pedrobaeza: You are injecting here SQL values. Proper syntax is:
```suggestion
        """, (AsIs(openupgrade.get_legacy_name('old_sector_id')), ),
```

partner_industry_secondary/migrations/11.0.1.0.0/post-migration.py
```diff
@@ -0,0 +1,67 @@
...
+    # corrected parent_id here:
+    cr.execute(
+        """
+        UPDATE res_partner_industry rpi
+        SET parent_id = rpi2.id
+        FROM res_partner_sector rps
+        LEFT JOIN res_partner_sector rps2 ON rps.parent_id = rps2.id
+        LEFT JOIN res_partner_industry rpi2 ON rps2.id = rpi2.%s
+        WHERE rpi.%s = rps.id
+        """ % (
```
pedrobaeza: The same about SQL injection

partner_industry_secondary/migrations/11.0.1.0.0/post-migration.py
```diff
@@ -0,0 +1,67 @@
...
+
+
+def update_res_partner_industry(cr):
+    cr.execute(
+        """
+        UPDATE res_partner rp
+        SET industry_id = rpi.id
+        FROM res_partner_industry rpi
+        WHERE rp.%s IS NOT NULL AND rp.%s = rpi.%s
+        """ % (
```
pedrobaeza: The same about SQL injection

partner_industry_secondary/migrations/11.0.1.0.0/post-migration.py
```diff
@@ -0,0 +1,67 @@
...
+def fill_res_partner_industry_secondary(cr):
+    cr.execute(
+        """
+        INSERT INTO res_partner_res_partner_industry_rel (res_partner_id,
+            res_partner_industry_id)
+        SELECT rel.res_partner_id, rpi.id
+        FROM res_partner_res_partner_sector_rel rel
+        LEFT JOIN res_partner_industry rpi
+            ON rel.res_partner_sector_id = rpi.%s
+        """ % openupgrade.get_legacy_name('old_sector_id')
```
pedrobaeza: The same about SQL injection

partner_industry_secondary/migrations/11.0.1.0.0/pre-migration.py
```diff
@@ -0,0 +1,26 @@
...
+def migrate(env, version):
+    cr = env.cr
+    if openupgrade.table_exists(cr, 'res_partner_sector'):
+        openupgrade.rename_columns(cr, _column_renames)
+        openupgrade.rename_xmlids(cr, _xmlid_renames)
+        cr.execute(
+            """
+            ALTER TABLE res_partner_industry
+            ADD COLUMN %s integer;
+            """ % openupgrade.get_legacy_name('old_sector_id')
```
pedrobaeza: The same about SQL injection

partner_industry_secondary/migrations/11.0.1.0.0/post-migration.py
pedrobaeza: And previously you have to import:
```python
from psycopg2.extras import AsIs
```

partner_industry_secondary/migrations/11.0.1.0.0/post-migration.py
```diff
@@ -0,0 +1,67 @@
...
+        FROM res_partner_sector
+        """ % openupgrade.get_legacy_name('old_sector_id'),
+    )
+    # corrected parent_id here:
+    cr.execute(
+        """
+        UPDATE res_partner_industry rpi
+        SET parent_id = rpi2.id
+        FROM res_partner_sector rps
+        LEFT JOIN res_partner_sector rps2 ON rps.parent_id = rps2.id
```
pedrobaeza: You don't need this second `res_partner_sector` table.

Just use this on the next `LEFT JOIN`: `ON rps.parent_id = rpi2.%s`

pedrobaeza (changes requested):

MiquelRForgeFlow:
Done.

partner_industry_secondary/migrations/11.0.1.0.0/post-migration.py
MiquelRForgeFlow: It's
```python
from psycopg2.extensions import AsIs
```

pedrobaeza: Agh, my memory has betrayed me!

partner_industry_secondary/migrations/11.0.1.0.0/post-migration.py
```diff
@@ -0,0 +1,67 @@
...
+        FROM res_partner_sector
+        """, (AsIs(openupgrade.get_legacy_name('old_sector_id')), ),
+    )
+    # corrected parent_id here:
+    cr.execute(
+        """
+        UPDATE res_partner_industry rpi
+        SET parent_id = rpi2.id
+        FROM res_partner_sector rps
+        LEFT JOIN res_partner_industry rpi2 ON rps.parent_id = rpi2.%s
```
pedrobaeza: Why `LEFT JOIN` here? I think this should be `INNER JOIN` for avoiding unneeded rewritings on empty values + be sure NULL values are not matched.

MiquelRForgeFlow: ok

MiquelRForgeFlow: done

pedrobaeza (approved):

PR #623, opened 2018-08-31
[7.0] [IMP] partner_relations_in_tab. Clean Code.

NL66278:
When defining a partner.relation.type where the left side should get its own tab, this does ot always work.

To find and rectify the problem, add debug logging and cleanup code.

Also move the code closer to 10.0, which also might need some cleanup.


NL66278:
Closed in favour of new approach here: https://github.com/OCA/partner-contact/pull/625

PR #625, opened 2018-09-06
[7.0] [IMP] partner_relations_in_tab. Clean Code.

NL66278:
Solves problem with tabs for left side relations.

Also:
- Add tests;
- No more need for _register_hook;
- Handles deletions and other changes in relation types directly on loading view.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,7 +23,6 @@ env:
 
   matrix:
   - LINT_CHECK="1"
-  - TRANSIFEX="1"
   - TESTS="1" ODOO_REPO="odoo/odoo"
   - TESTS="1" ODOO_REPO="OCA/OCB"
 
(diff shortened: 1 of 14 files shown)


rafaelbn (approved):

rafaelbn:
Hi @NL66278 , maybe @hbrunn codul review here ... this is v7 ! :atom:

NL66278:
@hbrunn Maybe have a look at this old one also. For a long time working now in production for Bisdom.

hbrunn:
/ocabot merge
i trust your approvals for a direct merge, but I'm curious how ocabot reacts to ancient code

NL66278:
I think we should give up on the ocabot merge for 7.0. It is failing now on modules that have nothing to do with this one, because of a version string format that got mandatory long after 7.0 was available, so nearly no module would be compliant.

rousseldenis:
> I think we should give up on the ocabot merge for 7.0. It is failing now on modules that have nothing to do with this one, because of a version string format that got mandatory long after 7.0 was available, so nearly no module would be compliant.

That's a possibility. Note that some processes are done by the bot (publishing to pypi, ...).

That depends if we want to invest time in such old branches (to fix such errors).

NL66278:
@rousseldenis I do not think people will invest time in this. So for PR's < 8.0 I will force merge if appropiate.

PR #626, opened 2018-09-17
[11.0][MIG] partner_ref_unique: Migrate to v11

cristinamartinrod:
This module allows add an unique constraint to partner ref field.

cc @tecnativa


partner_ref_unique/__manifest__.py
```diff
@@ -0,0 +1,23 @@
+# Copyright 2016 Antiun Ingenieria S.L. - Antonio Espinosa
+# Copyright 2017 Tecnativa - Vicent Cubells
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+{
+    "name": "Partner unique reference",
+    "summary": "Add an unique constraint to partner ref field",
+    "version": "11.0.1.0.0",
+    "category": "Customer Relationship Management",
+    "website": "http://www.tecnativa.com",
```
rafaelbn: Update website as commented in https://github.com/OCA/maintainer-tools/blob/master/template/module/__manifest__.py#L10

partner_ref_unique/README.rst
```diff
@@ -0,0 +1,59 @@
...
+
+Images
+------
+
+* Odoo Community Association: `Icon <https://github.com/OCA/maintainer-tools/blob/master/template/module/static/description/icon.svg>`_.
+
+Contributors
+------------
+
+* Antonio Espinosa <antonioea@antiun.com>
```
rafaelbn, line 43: Please change to antonio.espinosa@tecnativa.com which is the last correct email

partner_ref_unique/README.rst
```diff
@@ -0,0 +1,59 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+    :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+    :alt: License: AGPL-3
+
+========================
+Partner unique reference
+========================
+
+Add an unique constraint to partner ref field
```
rafaelbn, line 9: Please check it doesn't look it works

rafaelbn (changes requested):
I've tested in runbot @cristinamartinrod and it looks like constrain it's nor working. Please review! Thanks!

pedrobaeza:
Please squash together adjacent "OCA Transbot.." commits using https://github.com/OCA/maintainer-tools/wiki/Merge-commits-in-pull-requests

Check also Travis status

partner_ref_unique/models/__init__.py
```diff
@@ -0,0 +1,5 @@
+# Copyright 2016 Antiun Ingenieria S.L. - Antonio Espinosa
```
nikul-serpentcs: @cristinamartinrod Remove Copyright from all `__init__.py` file

partner_ref_unique/models/res_company.py
```diff
@@ -0,0 +1,15 @@
+# Copyright 2016 Antiun Ingenieria S.L. - Antonio Espinosa
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from openerp import models, fields
```
nikul-serpentcs: @cristinamartinrod IMO replace `openerp` to `odoo`

nikul-serpentcs (commented):
Small Change

cristinamartinrod:
All was fixed! Can you review @rafaelbn ?

christophe-hanon:
# Functional review – not ok - with 1 BUG
Thanks for your code, sorry to modestly disturb the process.

I have a few remarks :

1 Module Description and interface is not accurate “Add an unique constraint to partner ref field”
But for partner with the company flag it checks if the the customer flag is set (vs all partners).

2 Why handle differently individual customer and company customer ?

3 Looking at the way companies are checked is confusing because of asymmetric scenario :

3.1 Assume company1 is a customer  with reference ref1
Create a new customer2 with reference ref1 it will  fire a message – ok. 

3.2 Assume company1 not a customer  with reference ref1
Create a new customer2 with reference ref1 it will not fire a message – somehow strange. 


4 Modification of ‘is a customer flag’ is not checked – so if you create two non customer companies with the same ref then changing the ‘is customer’ flag will result into 2 companies customer with the same ref => BUG

Comes from @api.constrains('ref', 'is_company', 'company_id')

5 Similar to point 2. Check for ‘is customer’ makes sense but I think  ‘is supplier’ makes senses too. Let me know if there is a business reason for this. May be extra parameter is needed.

My 0.01 cent advise : either don’t check ‘is_customer’ or check both ‘is_customer’ and ‘is_supplier’

I hope this help, Ill be glad to review it again

partner_ref_unique/models/res_company.py
```diff
@@ -0,0 +1,15 @@
+# Copyright 2016 Antonio Espinosa
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import models, fields
```
hieulucky111: Please check import order: https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#imports

partner_ref_unique/models/res_partner.py
```diff
@@ -0,0 +1,33 @@
+# Copyright 2016 Antonio Espinosa
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import models, api, _
```
hieulucky111: Please check order import: https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#imports

hieulucky111 (changes requested):
Please check my review

christophe-hanon (changes requested):
Functional review – not ok - with 1 BUG
Thanks for your code, sorry to modestly disturb the process.

I have a few remarks :

1 Module Description and interface is not accurate “Add an unique constraint to partner ref field”
But for partner with the company flag it checks if the the customer flag is set (vs all partners).

2 Why handle differently individual customer and company customer ?

3 Looking at the way companies are checked is confusing because of asymmetric scenario :

3.1 Assume company1 is a customer with reference ref1
Create a new customer2 with reference ref1 it will fire a message – ok.

3.2 Assume company1 not a customer with reference ref1
Create a new customer2 with reference ref1 it will not fire a message – somehow strange.

4 Modification of ‘is a customer flag’ is not checked – so if you create two non customer companies with the same ref then changing the ‘is customer’ flag will result into 2 companies customer with the same ref => BUG

Comes from @api.constrains('ref', 'is_company', 'company_id')

5 Similar to point 2. Check for ‘is customer’ makes sense but I think ‘is supplier’ makes senses too. Let me know if there is a business reason for this. May be extra parameter is needed.

My 0.01 cent advise : either don’t check ‘is_customer’ or check both ‘is_customer’ and ‘is_supplier’

I hope this help, Ill be glad to review it again

pedrobaeza:
@cristinamartinrod can you review comments?

cristinamartinrod:
@christophe-hanon thanks a lot for your review. Looking back, I really don't know why "customer is true" condition was in v9, so I deleted it. The logical operation remains, as it is in the description of the readme. About your question 2 "Why handle differently individual customer and company customer?", now only difference between "all partners" and "only companies".

partner_ref_unique/models/res_partner.py
```diff
@@ -20,7 +19,6 @@ def _check_ref(self):
                 domain = [
                     ('id', '!=', partner.id),
```
pedrobaeza, line 20: When you are creating a record, this condition shouldn't be added. I think this can be tested if `partner.id` is False.

pedrobaeza (changes requested):

partner_ref_unique/models/res_partner.py
pedrobaeza: We have checked in an interactive session, and it's correct this way.

pedrobaeza (approved):
Extended use cases for tests

cristinamartinrod:
@rafaelbn, @christophe-hanon can you review, please? Thanks

christophe-hanon (approved):
Tested - the hidden check on the customer flag is gone. Works well and in sync with the description. Thanks !

rafaelbn (approved):
Tested in runbot :+1:

PR #628, opened 2018-09-20
[ADD] base_location_nuts_no_update

SimoRubi:
This modification allows to mark some NUTS regions as not updatable.

PR #631, opened 2018-09-27
[UPD] base_location_nuts: update based on flag

SimoRubi:
Forward porting https://github.com/OCA/partner-contact/pull/628


pedrobaeza:
Fast-tracking as it's the same as v10.

PR #632, opened 2018-09-27
[11.0][WIP] partner_group: New module

sergio-teruel:


partner_group/models/res_partner.py
```diff
@@ -0,0 +1,18 @@
+# Copyright 2018 Tecnativa - Sergio Teruel
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+from odoo import api, fields, models
+
+
+class ResPartner(models.Model):
+    _inherit = 'res.partner'
+
+    group_id = fields.Many2one(
```
yajo, line 9: Can't we use `parent_id` instead? :thinking:

sergio-teruel: No, jeje.. the records already has parent_id.
I want group companies in other parent_id.

yajo (approved):

partner_group/README.rst
```diff
@@ -0,0 +1,82 @@
...
+    :target: https://translation.odoo-community.org/projects/partner-contact-11-0/partner-contact-11-0-partner_group
+    :alt: Translate me on Weblate
+.. |badge5| image:: https://img.shields.io/badge/runbot-Try%20me-875A7B.png
+    :target: https://runbot.odoo-community.org/runbot/134/11.0
+    :alt: Try me on Runbot
+
+|badge1| |badge2| |badge3| |badge4| |badge5| 
+
+This module extends the functionality of partner to allow group partners into
+other partners, this modulo not use the parent company field.
```
pedrobaeza, line 29: s/this modulo not use the parent company field./. This module is an alternative to parent company field, as setting that field implies other business logic that wants to be avoided here as this is only for informative purposes.

partner_group/views/res_partner_views.xml
```diff
@@ -0,0 +1,30 @@
...
+
+    <record id="view_res_partner_filter" model="ir.ui.view">
+        <field name="model">res.partner</field>
+        <field name="inherit_id" ref="base.view_res_partner_filter"/>
+        <field name="arch" type="xml">
+            <filter name="group_country" position="after">
+                <filter name="group_group" string="Group" context="{'group_by': 'group_id'}"/>
+            </filter>
+            <field name="parent_id" position="after">
+                <field name="group_id" operator="child_of"/>
```
pedrobaeza: the operator `child_of` is not applicable if you don't have hierarchy on groups.

pedrobaeza (changes requested):

sergio-teruel:
Changes done

pedrobaeza (approved):

PR #638, opened 2018-10-08
[12.0 ][MIG] base_location: Migration to 12.0

aitorbouzas:
**Refactored**

This module has now been refactored to be more consistent with what base_address_city offers to the location management.
Now, every res.city record has a relation One2many to res.city.zip (old res.better.zip). This way, every zip has a realted city too.
Zips can be searched through city code, zip or city name (same as before).
Modified tests and deleted not needed tests.


pedrobaeza:
@aitorbouzas I'm tomorrow on holidays, but on Wednesday, we can meet for seeing the problem here + the refactoring that this module needs IMO since the introduction of `base_address_city`.

base_location/README.rst
```diff
@@ -0,0 +1,87 @@
...
+
+#. Access a partner record
+#. Fill the field *Location completion*
+#. Information about country, state, city and zip will be filled automatically
+
+
+
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/134/11.0
```
nikul-serpentcs: @aitorbouzas IMO Change target `12.0`

base_location/__manifest__.py
```diff
@@ -0,0 +1,31 @@
...
+    ],
+    'author': "Camptocamp,"
+              "ACYSOS S.L.,"
+              "Alejandro Santana,"
+              "Tecnativa,"
+              "AdaptiveCity,"
+              "Odoo Community Association (OCA)",
+    'license': "AGPL-3",
+    'summary': '''Enhanced zip/npa management system''',
+    'website': 'http://www.camptocamp.com',
```
nikul-serpentcs: Replace URL as per [this](https://github.com/OCA/maintainer-tools/blob/master/template/module/__manifest__.py#L10)

base_location/__manifest__.py
```diff
@@ -0,0 +1,31 @@
...
+    'data': ['views/better_zip_view.xml',
```
nikul-serpentcs: Follow this sequence https://github.com/OCA/maintainer-tools/blob/master/template/module/__manifest__.py#L27

nikul-serpentcs (commented):
Improve Code

aitorbouzas:
@pedrobaeza Ok Pedro, didn't bear in mind the new module when migrating this one. Just write to me through email and we'll fix an hour to meet.

aitorbouzas:
This module has now been refactored. Now it works more consistently with ```base_address_city``` (check commit info, PR's description or files changes for more info).

base_location/models/res_city.py
```diff
@@ -0,0 +1,16 @@
+# Copyright 2018 Aitor Bouzas <aitor.bouzas@adaptivecity.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class City(models.Model):
+    _inherit = 'res.city'
+
+    code = fields.Char(
```
pedrobaeza: Do you think this code brings too much value? Are there official codes for the cities?

base_location/models/res_city_zip.py
```diff
@@ -0,0 +1,47 @@
+# Copyright 2016 Nicolas Bessi, Camptocamp SA
+# Copyright 2018 Aitor Bouzas <aitor.bouzas@adaptivecity.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import api, fields, models
+
+
+class BetterZip(models.Model):
```
pedrobaeza: ResCityZip

base_location/models/res_city_zip.py
```diff
@@ -0,0 +1,47 @@
...
+    _description = __doc__
+    _order = "name asc"
+
+    name = fields.Char('ZIP')
+    city_id = fields.Many2one(
+        'res.city',
+        'City',
+        required=True,
+    )
+    latitude = fields.Float()
```
pedrobaeza: This would point only to a point, not defining an area. Do you think this adds enough value? I know they were previously on the module, but I doubt this has been useful any time.

base_location/models/res_city_zip.py
```diff
@@ -0,0 +1,47 @@
...
+    longitude = fields.Float()
+
+    @api.multi
+    @api.depends('name', 'city', 'state_id', 'country_id')
+    def name_get(self):
+        result = []
+        for rec in self:
+            name = []
+            if rec.name:
+                name.append('%(name)s' % {'name': rec.name})
```
pedrobaeza: Isn't easier to just do `name.append(rec.name)` (it's already a string)?

base_location/models/res_city_zip.py
```diff
@@ -0,0 +1,47 @@
...
+
+
+class BetterZip(models.Model):
+    """City/locations completion object"""
+
+    _name = "res.city.zip"
+    _description = __doc__
+    _order = "name asc"
+
+    name = fields.Char('ZIP')
```
pedrobaeza: I think it should be required

base_location/models/res_city_zip.py
```diff
@@ -0,0 +1,47 @@
...
+    latitude = fields.Float()
+    longitude = fields.Float()
+
+    @api.multi
+    @api.depends('name', 'city', 'state_id', 'country_id')
+    def name_get(self):
+        result = []
+        for rec in self:
+            name = []
+            if rec.name:
```
pedrobaeza: Remove this `if` being the name required

base_location/models/res_city_zip.py
```diff
@@ -0,0 +1,47 @@
...
+                name.append('%(name)s' % {'name': rec.name})
+            name.append('%(name)s' % {'name': rec.city_id.name})
```
pedrobaeza: `name.append(rec.city_id.name)`

base_location/models/res_city_zip.py
```diff
@@ -0,0 +1,47 @@
...
+            if rec.city_id.state_id:
+                name.append('%(name)s' % {'name': rec.city_id.state_id.name})
```
pedrobaeza: `name.append(rec.city_id.state_id.name)`

base_location/models/res_city_zip.py
```diff
@@ -0,0 +1,47 @@
...
+            if rec.city_id.country_id:
+                name.append('%(name)s' % {'name': rec.city_id.country_id.name})
```
pedrobaeza: `name.append(rec.city_id.country_id.name)`

base_location/models/res_city_zip.py
```diff
@@ -0,0 +1,47 @@
...
+            result.append((rec.id, ", ".join(name)))
```
pedrobaeza: I'm thinking that maybe you can add a computed stored field depending on these values for not overloading each `name_get` call, having to access 3 tables. This will rise the entries addition (for example, through `base_location_geonames_import`), but will speed up location search, which is the most habitual task. Being called `display_name`, it will be automatically done without overwriting this method.

base_location/models/res_city_zip.py
```diff
@@ -0,0 +1,47 @@
...
+        return result
+
+    @api.model
+    def name_search(self, name='', args=None, operator='ilike', limit=100):
```
pedrobaeza: I think changing `_rec_name` to `display_name` will perform the search over the new computed field, handling all these cases, and making it more performant.

base_location/models/res_company.py
```diff
@@ -57,12 +57,9 @@ def _onchange_zip_id(self):
         if self.zip_id:
             self.zip = self.zip_id.name
             self.city_id = self.zip_id.city_id
-            self.city = self.zip_id.city
-            self.country_id = self.zip_id.country_id
-            if self.country_id.enforce_cities:
-                self.state_id = self.city_id.state_id
-            else:
-                self.state_id = self.zip_id.state_id
+            self.city = self.zip_id.city_id.name
```
pedrobaeza: On v11, I remember this to be problematic, not filling correctly fields, but as it's only one time operation, I ignored the problem. Have you checked if this works?

base_location/models/res_partner.py
```diff
@@ -28,36 +28,35 @@ def _onchange_city_id(self):
...
             self.zip_id = False
         return res
 
     @api.onchange('zip_id')
     def _onchange_zip_id(self):
         if self.zip_id:
-            self.country_id = self.zip_id.country_id
-            if self.country_id.enforce_cities:
-                self.city_id = self.zip_id.city_id
+            self.country_id = self.zip_id.city_id.country_id
```
pedrobaeza: Only reset values of state and country if they are filled on the city.

base_location/models/res_partner.py
```diff
@@ -28,36 +28,35 @@ def _onchange_city_id(self):
 
     @api.onchange('state_id')
     def _onchange_state_id(self):
```
pedrobaeza: There are other onchange for this field on the bottom.

base_location/readme/CONFIGURE.rst
```diff
@@ -0,0 +1,13 @@
+#. Activate the developer mode in *Settings*.
+#. Go to *Settings / Technical / Locations Management / Locations*.
```
pedrobaeza: This menu no longer exists on v12, and I think that we should put Locations menu now on *Contacts > Configuration > Localization*, as we have that menu path for the rest of the localization elements (States, Countries...), and thus to add `contacts` as dependency, which is very little trade-off.

base_location/readme/CONTRIBUTORS.rst
```diff
@@ -0,0 +1,9 @@
+* Nicolas Bessi (Camptocamp)
+* Ignacio Ibeas (Acysos S.L.)
+* Pedro M. Baeza <pedro.baeza@serviciosbaeza.com>
```
pedrobaeza: Please remove my old email

base_location/security/ir.model.access.csv
```diff
@@ -1,3 +1,3 @@
 "id","name","model_id:id","group_id:id","perm_read","perm_write","perm_create","perm_unlink"
-"ir_model_access_betterzip0","res_better_zip group_user_all","model_res_better_zip",base.group_user,1,0,0,0
-"ir_model_access_betterzip1","res_better_zip group_user","model_res_better_zip","base.group_partner_manager",1,1,1,1
+"ir_model_access_cityzip0","res_city_zip group_user_all","model_res_city_zip",base.group_user,1,0,0,0
```
pedrobaeza: Not sure if we should grant read access to portal/website users as well.

base_location/views/res_city.xml
```diff
@@ -0,0 +1,60 @@
...
+<odoo>
+
+    <record id="view_city_tree_inherit" model="ir.ui.view">
+        <field name="model">res.city</field>
+        <field name="inherit_id" ref="base_address_city.view_city_tree"/>
+        <field name="arch" type="xml">
+            <tree position="attributes">
+                <attribute name="editable"/>
+            </tree>
+            <field name="zipcode" position="replace">
```
pedrobaeza: Please don't replace, but hide the element, or we expose to a possible error if other module takes this element as reference.

base_location/views/res_city_zip.xml
```diff
@@ -0,0 +1,64 @@
...
+                </group>
+            </form>
+        </field>
+    </record>
+
+    <record model="ir.ui.view" id="city_zip_tree">
+        <field name="name">res.city.zip.tree</field>
+        <field name="model">res.city.zip</field>
+        <field name="arch" type="xml">
+            <tree string="Zips">
```
pedrobaeza: Better to do this editable

pedrobaeza (changes requested):
Thank you very much for the changes! Excellent work

@rvalyi you'll be glad to know that with this, we normalize city DB structure, having only one record per city.

Several things to do for making it perfect:

-  ~100% coverage. It's a pity that some cases are not covered now while they were previously. 
- Views files are named as `model_views.xml`. This is for differentiating them from for example other files that adds data for that model when looking by file name.

base_location/models/res_city_zip.py
```diff
@@ -0,0 +1,47 @@
...
+
+class BetterZip(models.Model):
+    """City/locations completion object"""
+
+    _name = "res.city.zip"
+    _description = __doc__
+    _order = "name asc"
+
+    name = fields.Char('ZIP')
+    city_id = fields.Many2one(
```
pedrobaeza, line 17: Add an SQL constraint with `UNIQUE(city_id, name)`

base_location/models/res_city.py
```diff
@@ -0,0 +1,16 @@
+# Copyright 2018 Aitor Bouzas <aitor.bouzas@adaptivecity.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class City(models.Model):
+    _inherit = 'res.city'
```
pedrobaeza, line 8: We should add an SQL constraint with `UNIQUE(name, state_id, country_id)`

base_location/models/res_company.py
aitorbouzas: Yup, it works. Tried several times and it worked for me. Will try more times in case it fails.

aitorbouzas:
@pedrobaeza I think I made all the changes now. Added some tests for the constraints (will wait for codecov's report to add some more and reach 100%)

base_location/models/res_city.py
```diff
@@ -0,0 +1,19 @@
...
+
+    zip_ids = fields.One2many('res.city.zip', 'city_id',
+                              string="Zips in this city")
+
+    _sql_constraints = [
+        ('name_state_country_uniq',
+         'UNIQUE(name, state_id, country_id)',
+         'You already have a city with that name in the same state.'
```
simahawk, line 16: make the error translatable

base_location/models/res_city_zip.py
```diff
@@ -0,0 +1,40 @@
...
+        'res.city',
+        'City',
+        required=True,
+    )
+    display_name = fields.Char(compute='_compute_new_display_name',
+                               store=True, index=True)
+
+    _sql_constraints = [
+        ('name_city_uniq', 'UNIQUE(name, city_id)',
+         'You already have a zip with that code in the same city. '
```
simahawk, line 27: same

base_location/models/res_company.py
```diff
@@ -0,0 +1,67 @@
...
+        compute='_compute_address',
+        inverse='_inverse_zip_id',
+        oldname="better_zip_id",
+        help='Use the city name or the zip code to search the location',
+    )
+    # In order to keep the same logic used in odoo, fields must be computed
+    # and inversed, not related. This way, we can ensure that it works
+    # correctly on changes and inconsistencies cannot happen.
+    # When you make the fields related, the constrains added in res.partner
+    # will fail. because when you change the city_id in the company, you are
```
simahawk: `s/./,`

base_location/models/res_company.py
```diff
@@ -0,0 +1,67 @@
...
+    # effectively changing it in the partner. The constrains on the partner
+    # are evaluated before the inverse methods update the other fields (city,
+    # etc..). And we need constrains in the partner to ensure consistency.
```
simahawk: `s/). And [...]/) so we need then to ensure consistency.`

base_location/models/res_company.py
```diff
@@ -0,0 +1,67 @@
...
+    # So, as a conclusion, address fields are very related to each other.
```
simahawk: `# As a conclusion`

base_location/models/res_company.py
```diff
@@ -0,0 +1,67 @@
...
+    # Either you make them all related to the partner in company, or you
+    # don't for all of them. But mixing methods produces inconsistencies.
```
simahawk: `. Mixing both approaches...`

base_location/models/res_company.py
```diff
@@ -0,0 +1,67 @@
...
+
+    country_enforce_cities = fields.Boolean(
+        related='country_id.enforce_cities'
```
simahawk, line 39: :heart: this is readonly by default :smile:

base_location/models/res_company.py
```diff
@@ -0,0 +1,67 @@
...
+    )
+
+    def _get_company_address_fields(self, partner):
+        res = super(ResCompany, self)._get_company_address_fields(partner)
```
simahawk: `super()` and apply for all occurrences

base_location/models/res_partner.py
```diff
@@ -0,0 +1,71 @@
...
+    _inherit = 'res.partner'
+
+    zip_id = fields.Many2one('res.city.zip', 'ZIP Location')
+
+    @api.onchange('city_id')
+    def _onchange_city_id(self):
+        if not self.zip_id:
+            super(ResPartner, self)._onchange_city_id()
+        if self.zip_id and self.city_id != self.zip_id.city_id:
+            self.zip_id = False
```
simahawk:
```python
self.update({
    'zip_id': False,
    'zip': False,
    'city': False,
})
```

base_location/models/res_partner.py
```diff
@@ -0,0 +1,71 @@
...
+    def _onchange_country_id(self):
+        res = super(ResPartner, self)._onchange_country_id()
+        if self.zip_id and self.zip_id.city_id.country_id != self.country_id:
+            self.zip_id = False
+        return res
+
+    @api.onchange('zip_id')
+    def _onchange_zip_id(self):
+        if self.zip_id:
+            if self.zip_id.city_id.country_id:
```
simahawk, line 47: prepare all values and update them at once using `self.update`

base_location/models/res_partner.py
```diff
@@ -0,0 +1,71 @@
...
+                self.country_id = self.zip_id.city_id.country_id
+            if self.zip_id.city_id.state_id:
+                self.state_id = self.zip_id.city_id.state_id
+            self.city_id = self.zip_id.city_id
+            self.zip = self.zip_id.name
+            self.city = self.zip_id.city_id.name
+
+    @api.constrains('zip_id', 'country_id', 'city_id', 'state_id')
+    def _check_zip(self):
+        for rec in self.filtered('zip_id'):
```
simahawk: do not loop twice:
```python
for rec in self:
    if not rec.zip_id:
        continue
[...]
```

base_location/models/res_partner.py
```diff
@@ -0,0 +1,71 @@
...
+                    "The country of the partner %s differs from that in "
+                    "location %s") % (rec.name, rec.zip_id.name))
+            if rec.zip_id.city_id != rec.city_id:
+                raise ValidationError(_(
+                    "The city of partner %s differs from that in "
+                    "location %s") % (rec.name, rec.zip_id.name))
+
+    @api.onchange('state_id')
+    def _onchange_state_id(self):
+        if self.state_id.country_id:
```
simahawk, line 74: same, use `update(vals)`

base_location/tests/test_base_location.py
```diff
@@ -0,0 +1,315 @@
...
+
+from odoo.tests.common import TransactionCase
+from odoo.exceptions import ValidationError
+from odoo.tests import tagged
+from odoo.tools.misc import mute_logger
+import psycopg2
+
+
+@tagged('post_install', '-at_install')
+class TestBaseLocation(TransactionCase):
```
simahawk: 1. `SavepointCase` 
2. `setUp` -> `setUpClass`
3.  disable tracking everywhere
```python
# disable tracking test suite wise
cls.env = cls.env(context=dict(cls.env.context, tracking_disable=True))
```

base_location/tests/test_base_location.py
```diff
@@ -0,0 +1,315 @@
...
+
+    def setUp(self):
+        super(TestBaseLocation, self).setUp()
+        self.state_vd = self.env['res.country.state'].create({
+            'name': 'Vaud',
+            'code': 'VD',
+            'country_id': self.ref('base.ch'),
```
simahawk: `self.env.ref('base.ch').id`

base_location/tests/test_base_location.py
```diff
@@ -0,0 +1,315 @@
...
+        })
+        self.env['res.country'].browse(self.ref('base.es')).write({
```
simahawk: you already have the country recset: `self.env.ref('base.es').write(...)`

base_location/tests/test_base_location.py
```diff
@@ -0,0 +1,315 @@
...
+
+    def values_city_zip2(self):
+        return {
+            'name': 2000,
+            'city_id': self.city_bcn.id,
+        }
+
+    def test_onchange_partner_city_completion(self):
+        """Test whether all data in partner is filled correctly
+        after changing it's zip"""
```
simahawk: its

base_location/tests/test_base_location.py
```diff
@@ -0,0 +1,315 @@
...
+        }
+
+    def values_city_zip2(self):
+        return {
+            'name': 2000,
+            'city_id': self.city_bcn.id,
+        }
+
+    def test_onchange_partner_city_completion(self):
+        """Test whether all data in partner is filled correctly
```
simahawk: """Make sure partner data filled accordingly."""

base_location/tests/test_base_location.py
```diff
@@ -0,0 +1,315 @@
...
+        city_zip2.city_id.country_id.enforce_cities = True
+        partner1.zip_id = city_zip2
+        partner1._onchange_zip_id()
+        self.assertEqual(partner1.zip, city_zip2.name)
+        self.assertEqual(partner1.city, city_zip2.city_id.name)
+        self.assertEqual(partner1.state_id, city_zip2.city_id.state_id)
+        self.assertEqual(partner1.country_id, city_zip2.city_id.country_id)
+
+    def test_onchange_company_city_completion(self):
+        """Test whether all data in company is filled correctly
```
simahawk: same, you can shorten this docstrings and make them more PEP compliant (not mandatory as we don't enforce this yet in OCA modules)

base_location/tests/test_base_location.py
```diff
@@ -0,0 +1,315 @@
...
+            'state_id': city_zip2.city_id.state_id.id,
+            'city_id': city_zip2.city_id.id,
+        })
+
+        with self.assertRaises(ValidationError):
+            partner.city_id = self.city_lausanne
+
+    def test_partner_onchange_country(self):
+        """Test partner onchange country_id"""
+        country_es = self.browse_ref('base.es')
```
simahawk: `self.env.ref`
overall: I'd make all these `ref` usage consistent and just go for `self.env.ref` and then access `.id` when needed

base_location/tests/test_base_location.py
```diff
@@ -0,0 +1,315 @@
...
+        country_es.enforce_cities = True
+        city_zip1 = self.env['res.city.zip'].create(
```
simahawk: seems like you need these city_zip1/2 in every tests: create them in setUpClass.
Is useless and resource / time consuming when running tests ;)

aitorbouzas:
Fully green. All code covered by tests.

simahawk (changes requested):

base_location/models/res_city.py
pedrobaeza: I would say that it's translatable this way already... Odoo takes care of this. If not, you will have static translations at the server startup.

pedrobaeza:
@aitorbouzas great. I will review later and add migration scripts over your commits.

base_location/models/res_company.py
aitorbouzas: Sorry, I don't understand what you mean with this comment.

base_location/models/res_company.py
aitorbouzas: Can you elaborate on this?

base_location/models/res_partner.py
aitorbouzas: Does this have an impact in the performance?

base_location/models/res_partner.py
aitorbouzas: Well, this is usually running in a recordset with only one record... I understand your code has better performance but also it's less readable.

base_location/models/res_company.py
simahawk: rephrase like this: "...fields (city, etc..) so we need them to ensure consistency."
(you are already talking about the constrains)

base_location/models/res_partner.py
simahawk: yes, you write on it just once. Plus, `update` works on non existing (NewId) items too.

base_location/models/res_partner.py
simahawk: Actually, is more readable if you see that you are explicitly discarding records that do not have a zip. With "filtered" you are kind of hiding what your doing (especially to ppl not used to odoo's shortcuts BTW).
Also, using it in loops is a bad practice because in any case you loop twice and you (or whoever takes this code as an example) are going to reuse this approach all around by default "unconsciously".

base_location/models/res_company.py
simahawk: the dot is wrong, replace it w/ comma :)

aitorbouzas:
@simahawk Thanks for the great review. Changes has been done. :+1:

pedrobaeza:
@aitorbouzas when making changes on a reviewed PR, it's better to add an extra commit, but marked as a fixup, so that the review of these changes doesn't imply to review whole diff again. This is done with:
```
git commit --all --fixup HEAD
```
This way, and before merging, you can squash easily with
```
git rebase -i origin/12.0 --autosquash
```

aitorbouzas:
@pedrobaeza  Oops, i'm such a newbie yet. Thanks for the info.

base_location/__manifest__.py
```diff
@@ -0,0 +1,32 @@
...
+    'author': "Camptocamp,"
+              "ACYSOS S.L.,"
+              "Alejandro Santana,"
+              "Tecnativa,"
+              "AdaptiveCity,"
+              "Odoo Community Association (OCA)",
+    'license': "AGPL-3",
+    'summary': '''Enhanced zip/npa management system''',
+    'website': 'https://github.com/OCA/partner-contact',
+    'data': ['views/res_city_zip_view.xml',
```
nikul-serpentcs: @aitorbouzas change sequence as per [this]( https://github.com/OCA/maintainer-tools/blob/master/template/module/__manifest__.py#L29)

base_location/tests/test_base_location.py
```diff
@@ -0,0 +1,274 @@
...
+
+
+@tagged('post_install', '-at_install')
+class TestBaseLocation(common.SavepointCase):
+
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        cls.env = cls.env(context=dict(cls.env.context, tracking_disable=True))
+        cls.state_vd = cls.env['res.country.state'].create({
```
nikul-serpentcs: @aitorbouzas here `res.country.state` use multiple time, so you should you use global variable.

base_location/tests/test_base_location.py
```diff
@@ -0,0 +1,274 @@
...
+            'name': 'Madrid',
+            'code': '28',
+            'country_id': cls.env.ref('base.es').id,
+        })
+        cls.city_bcn = cls.env['res.city'].create({
+            'name': 'Barcelona',
+            'state_id': cls.state_bcn.id,
+            'country_id': cls.env.ref('base.es').id,
+        })
+        cls.city_madrid = cls.env['res.city'].create({
```
nikul-serpentcs: same here

base_location/tests/test_base_location.py
```diff
@@ -0,0 +1,274 @@
...
+            'name': 'Madrid',
+            'state_id': cls.state_madrid.id,
+            'country_id': cls.env.ref('base.es').id,
+        })
+        cls.city_lausanne = cls.env['res.city'].create({
+            'name': 'Lausanne',
+            'state_id': cls.state_vd.id,
+            'country_id': cls.env.ref('base.ch').id,
+        })
+        cls.lausanne = cls.env['res.city.zip'].create({
```
nikul-serpentcs: same here

base_location/tests/test_base_location.py
```diff
@@ -0,0 +1,274 @@
...
+        """Test partner 1 constraints"""
+        with self.assertRaises(ValidationError):
+            self.env['res.partner'].create({
+                'name': 'P1',
+                'zip_id': self.barcelona.id,
+            })
+
+    def test_constrains_partner_country(self):
+        """Test partner country constraints"""
+        partner = self.env['res.partner'].create({
```
nikul-serpentcs: same here

nikul-serpentcs (commented):
Improve Code

base_location/tests/test_base_location.py
aitorbouzas: I guess you mean to save the ref in a method-scope variable, right?

simahawk: @aitorbouzas just to clarify: `s/should/could` => is NOT mandatory.
It's actually a matter of taste + "ease creation" of other records w/out making the code redundant.
For instance: in my personal code I prefer that way as I'm lazy and I hate to copy/paste code :stuck_out_tongue:
but as you've seen I did not complain :wink: 
So all in all: nice to have but up to you.

aitorbouzas: @simahawk Oh well, I know that I must have my own criteria for this type of things. Thanks :)

nikul-serpentcs: > I guess you mean to save the ref in a method-scope variable, right?

Yes :+1:

pedrobaeza (approved):
Great work! Maybe you can add another demo data with "United States" or "Spain" that has states, so that the location is complete. Before merging, I want to make a bit of squashing for removing unneeded commits (like "Rename manifest file", etc), but I don't want to give you more work, so I'll do it when other reviewer approve this. We also need migration scripts, that I will add in that step.

simahawk (approved):
well done! thanks!

aitorbouzas:
I can squash them, but... Can I squash commits from different authors? I don't mind doing it. I can add some demo data too.

pedrobaeza:
I haved squashed commit history and add migration scripts. @aitorbouzas check if everything is correct for merging this.

aitorbouzas:
@pedrobaeza looks good :+1:

aitorbouzas:
With this scripts are ensured the relations between partners/companies and zips?
Edit: And what about if old database has some duplicated zip?

pedrobaeza:
You are right. That one is not preserved right now. I'm going to see if it's possible to be done easily, but it's also non critical, as data are dumped on onchange of the value.

pedrobaeza:
OK, check now that I have covered that part.

pedrobaeza:
The problem with these migration scripts is that right now, we are not testing them anywhere. Maybe they can be integrated on OpenUpgrade tests, but not right now. @StefanRijnhart what do you think about this?

aitorbouzas:
> OK, check now that I have covered that part.

:+1:

pedrobaeza:
Let's go ahead with this for now for being able to migrate other modules depending on this one.

emagdalenaC2i:
Should __openerp__.py be renamed as __manifest__.py?

pedrobaeza:
It's not mandatory for working, but desirable. I'm changing it now.

StefanRijnhart:
> The problem with these migration scripts is that right now, we are not testing them anywhere. Maybe they can be integrated on OpenUpgrade tests, but not right now. @StefanRijnhart what do you think about this?

I'd love to increase the scope of OpenUpgrade testing, but at the moment it's setup to work with a static database with just the core modules installed. I don't know how to set up scalable tests (given the Travis limitation of concurrent jobs) that load OCA modules on the fly and as a test, migrates them to later Odoo versions.

PR #639, opened 2018-10-09
[12.0][MIG] base_vat_sanitized

astirpe:
https://github.com/OCA/partner-contact/issues/630

### DIFF
--- a/base_vat_sanitized/models/res_partner.py
+++ b/base_vat_sanitized/models/res_partner.py
@@ -1,0 +1,25 @@
+# -*- coding: utf-8 -*-
+# © 2016 Akretion (http://www.akretion.com)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+# @author Alexis de Lattre <alexis.delattre@akretion.com>
+
+from openerp import models, fields, api
+import re
+
+
+class ResPartner(models.Model):
+    _inherit = 'res.partner'
+
+    sanitized_vat = fields.Char(
+        compute='compute_sanitized_vat', string='Sanitized TIN',
+        store=True, readonly=True,
+        help='TIN in uppercase without spaces nor special caracters.')
+
+    def _sanitize_vat(self, vat):
+        return vat and re.sub(r'\W+', '', vat).upper() or False
+
+    @api.multi
+    @api.depends('vat')
+    def compute_sanitized_vat(self):
+        for partner in self:
+            partner.sanitized_vat = self._sanitize_vat(partner.vat)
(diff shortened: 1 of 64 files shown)


base_vat_sanitized/models/res_partner.py
```diff
@@ -0,0 +1,25 @@
+# Copyright 2016 Akretion (http://www.akretion.com)
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+# @author Alexis de Lattre <alexis.delattre@akretion.com>
```
aitorbouzas, line 3: IMO (check [https://github.com/OCA/maintainer-tools/blob/master/template/module/models/model_name.py#L2](https://github.com/OCA/maintainer-tools/blob/master/template/module/models/model_name.py#L2)) you should add the author in the same line as the company.

base_vat_sanitized/tests/test_vat.py
```diff
@@ -0,0 +1,19 @@
+# Copyright 2016 Akretion (http://www.akretion.com)
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+# @author Alexis de Lattre <alexis.delattre@akretion.com>
```
aitorbouzas, line 3: Same note as above.

aitorbouzas (commented):
Runbot tested (adding sanitized_vat to the view) and code reviewed. LGTM but just made two annotations.

base_vat_sanitized/models/res_partner.py
astirpe: If I do that, the line will be more than 80 characters long, unless I remove some information.
I just wanted to avoid the removal of any information that was already present.

aitorbouzas: Oh, I see. Thanks for clarification.

alexis-via (approved):

base_vat_sanitized/models/res_partner.py
```diff
@@ -0,0 +1,25 @@
...
+
+import re
+
+from odoo import api, fields, models
+
+
+class ResPartner(models.Model):
+    _inherit = 'res.partner'
+
+    sanitized_vat = fields.Char(
```
rousseldenis, line 13: @alexis-via Even if we know the slowly computation of these kind of fields, shouldn't this be added to commercial_fields ?

https://github.com/odoo/odoo/blob/12.0/odoo/addons/base/models/res_partner.py#L435

rousseldenis: Nope, because vat is already a commercial field and should be ok on all same partners. I like answering on my own... :smile:

PR #642, opened 2018-10-12
[FIX] use provided name for firstname and lastname 

wpichler:
on copy if no firsrname and lastname is specified

### DIFF
--- a/partner_firstname/tests/test_copy.py
+++ b/partner_firstname/tests/test_copy.py
@@ -1,0 +1,54 @@
+# -*- coding: utf-8 -*-
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+"""Test copy function for models."""
+
+from odoo.tests.common import TransactionCase
+from .base import MailInstalled
+
+
+class UserCase(TransactionCase, MailInstalled):
+    """Test ``res.users``."""
+    model = "res.users"
+    context = {"default_login": "user@example.com"}
+
+    def setUp(self):
+        super(UserCase, self).setUp()
+        self.create_original()
+
+    def create_original(self):
+        self.original = self.env["res.users"].create({
+            "firstname": u"Firstname",
+            "lastname": u"Lastname",
+            "name": u"Firstname Lastname",
+            "login": u"firstname.lastname"
+        })
+
+    def compare(self, copy):
+        self.assertEqual(copy.lastname, u"Lastname2")
+        self.assertEqual(copy.firstname, u"Firstname2")
+        self.assertEqual(copy.name, u"Lastname2 Firstname2")
+
+    def test_copy_name(self):
+        """Copy original with default name set - firstname lastname not set."""
+        copy = self.original.copy(default={
+            "name": u"Lastname2 Firstname2"
+        })
+        self.compare(copy)
+
+    def test_copy_firstname_lastname(self):
+        """Copy original with default firstname and lastname set"""
+        copy = self.original.copy(default={
+            "firstname": u"Firstname2",
+            "lastname": u"Lastname2"
+        })
+        self.compare(copy)
+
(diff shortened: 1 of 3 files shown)


wpichler:
This PR does fix this problem:
With the partner_firstname module installed - when you do edit your template user - then you have to specify firstname and lastname to be able to save it. Then someone new does signup on your website - on signup the copy user function does get called. On copy name is set in defaults - but not firstname and lastname - so the copy function will use firstname and lastname from the template user - and will overwrite the provided name.

gdgellatly:
@wpichler Hi, can you write a test for this improvement please.  The travis failure seems unrelated.  I copied your code to #648 so will add the test there as well.

wpichler:
@gdgellatly I've now written a test - and found a bug with it ;-) But this is my first test case i every wrote for an odoo module (shame on me...) - so please take a look at, and check if it looks correct...

gdgellatly:
@wpichler what is the bug you found?  

I'm not the best on tests, but I've had to spend some time with the tests in this module.  The tests ultimately inherit from `tests/base.py` but I guess you worked that out.  There is lots of helpers in there, but this module has quite complicated tests compared to most.

I'm wondering if at least for some of this, setting the copy context is all that is required rather than manually setting the string, but I can see you've checked upstream so I won't second guess it.

wpichler:
@gdgellatly I did not imported _ ....
About the tests - I've implemented them - and they seem to work - and i found a bug with them - so they tests did their work quit good ;-)

gdgellatly:
@wpichler alright, fix these 2 lint errors and you'll be good for review

======== Testing test_flake8 ========
partner_firstname/models/res_users.py:43:80: E501 line too long (96 > 79 characters)
partner_firstname/models/res_users.py:44:80: E501 line too long (93 > 79 characters)

wpichler:
Fixed lint errors

partner_firstname/tests/test_copy.py
```diff
@@ -0,0 +1,57 @@
...
+
+    def create_original(self):
+        self.original = self.env["res.users"].create({
+            "firstname": u"Firstname",
+            "lastname": u"Lastname",
+            "name": u"Firstname Lastname",
+            "login": u"firstname.lastname"
+        })
+
+    def tearDown(self):
```
NL66278: No need for this method, as you do not add to the super method.

partner_firstname/tests/test_copy.py
```diff
@@ -0,0 +1,57 @@
...
+        super(UserCase, self).tearDown()
+
+    def compare(self, copy):
+        self.assertEqual(copy.lastname, u"Lastname2")
+        self.assertEqual(copy.firstname, u"Firstname2")
+        self.assertEqual(copy.name, u"Lastname2 Firstname2")
+
+    def test_copy_name(self):
+        """Copy original with default name set - firstname lastname not set."""
+        copy = self.original.copy({
```
NL66278: In your copy methods use the default= keyword. It works without, but it is clearer to call a keyword argument with a keyword.

NL66278 (changes requested):
See my other comments

partner_firstname/tests/test_copy.py
wpichler: done

partner_firstname/tests/test_copy.py
wpichler: done

rafaelbn:
Dear @wpichler , Thanks for this contribution. Could you please attend comments? Thanks! Travis is red :red_circle:

wpichler:
fixed the bug in the copy test case (typo) - and rebased it against origin/10.0

wpichler:
All checks are ok now - please merge

NL66278 (approved):
:+1: LGTM

PR #648, opened 2018-10-29
12.0 migrate partner first name

gdgellatly:
Changes:
- Changed default to first_last - plays nicer with existing data, especially demo data
- Changed visibility of recalculate names, as this button doesn't do what you think it does. If you click it after changing, it uses the old value as the new value is not saved yet.
- General cleanup
- Integration of PR bugfix on V11


partner_firstname/README.rst
```diff
@@ -0,0 +1,86 @@
...
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/134/11.0
+
+For further information, please visit:
+
+* https://www.odoo.com/forum/help-1
+
+
+Known issues / Roadmap
```
nikul-serpentcs, line 71: @gdgellatly IMO if not use then Remove it

partner_firstname/README.rst
```diff
@@ -0,0 +1,86 @@
...
+
+Bugs are tracked on `GitHub Issues
+<https://github.com/OCA/partner-contact/issues>`_. In case of trouble, please
+check there if your issue has already been reported. If you spotted it first,
+help us smashing it by providing a detailed and welcomed feedback.
+
+Credits
+=======
+
+Contributors
```
nikul-serpentcs: Adding Contributors member

partner_firstname/__init__.py
```diff
@@ -0,0 +1,5 @@
+# Copyright 2013 Nicolas Bessi (Camptocamp SA)
```
nikul-serpentcs: Remove Copyright from all `__init__` file

nikul-serpentcs (commented):
Improve Code

partner_firstname/README.rst
gdgellatly: @nikul-serpentcs This whole file gets auto replaced, no need to look at it

partner_firstname/tests/test_create.py
```diff
@@ -0,0 +1,80 @@
...
+    context = {"default_is_company": False}
+    model = "res.partner"
+
+    def setUp(self):
+        super(PersonCase, self).setUp()
+        self.good_values = {
+            "firstname": "Núñez",
+            "lastname": "Fernán",
+        }
+        self.good_values["name"] = "%s %s" % (self.good_values["lastname"],
```
njeudy: if you change default first / last name order you have to change test here :)

partner_firstname/tests/test_name.py
```diff
@@ -0,0 +1,88 @@
...
+
+    def test_update_firstname(self):
+        """Change firstname."""
+        self.expect(self.lastname, "newfïrstname")
+        self.original.name = self.name
+
+    def test_whitespace_cleanup(self):
+        """Check that whitespace in name gets cleared."""
+        self.expect("newlästname", "newfïrstname")
+        self.original.name = "  newlästname  newfïrstname  "
```
njeudy: And here (change order)
```suggestion
        self.original.name = "  newfïrstname newlästname  "
```

partner_firstname/tests/test_name.py
```diff
@@ -0,0 +1,88 @@
...
+    def test_company_inverse(self):
+        """Test the inverse method in a company record."""
+        name = "Thïs is a Companŷ"
+        self.expect(name, False, name)
+        self.original.name = name
+
+
+class UserCase(PartnerContactCase):
+    def create_original(self):
+        name = "%s %s" % (self.lastname, self.firstname)
```
njeudy: And here too :)

```suggestion
        name = "%s %s" % (self.firstname, self.lastname)
```

partner_firstname/tests/test_user_onchange.py
```diff
@@ -0,0 +1,52 @@
...
+            self.user.firstname = firstname
+            self.user._compute_name()
+
+            # Changes lastname, which triggers onchanges
+            self.user.lastname = lastname
+            self.user._compute_name()
+
+            self.assertEqual(self.user.lastname, lastname)
+            self.assertEqual(self.user.firstname, firstname)
+            self.assertEqual(self.user.name, " ".join((lastname, firstname)))
```
njeudy: And here

```suggestion
            self.assertEqual(self.user.name, " ".join((firstname, lastname)))
```

yajo (changes requested):
If changing the default value, add a migration script that creates the old default as the `ir.config_parameter`, if it didn't exist, so migrated databases do not see this altered.

gdgellatly:
@Yajo I don't really understand.  The default comes from ir.config_parameter if its already set so there should be no change in a migrated database?

But I don't really know how to do what you want anyway, can you point to an example?

yajo:
Yes, but what if it is not set? The previous user was relying on the previous default, which now changed, so the migrated database would have an unexpected behavior.

The solution is to add a migration script that creates the config parameter, only if missing. This way, the migrated DB will retain the old behavior, and that should avoid surprises.

See https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#migrations

gdgellatly:
@Yajo sorry that paragraph of text doesn't help me. Just point me to an example and I can do it.  I've never seen one outside of OpenUpgrade that I can recall, I didn't even know there was a built in migration function in Odoo.  I can't find a single one in Odoo source code.

yajo:
Yikes, sorry. [This is the most similar thing to docs we have](https://github.com/odoo/odoo/blob/fd27ef17653f225c955d487952f78568d65798ea/odoo/modules/migration.py#L54-L74), and [here you have an example](https://github.com/odoo/odoo/blob/12.0/addons/purchase/migrations/9.0.1.2/pre-create-properties.py).

gdgellatly:
@Yajo thanks, looks easy enough

gdgellatly:
@Yajo I have added the migration script.  I have asked on #642 for a test so I can get the coverage up there.  I notice the unicodedecode error isn't tested, but I am wondering if that is still relevant with Python 3.  If it is I'll write a test to cover it.

partner_firstname/migrations/12.0.1.0.0/pre-ir_config_param.py
```diff
@@ -0,0 +1,14 @@
+def store_ir_config_param(cr):
+    """Prior to version 12.0 the default order of partner
+    names was last_first.  In order to retain this behaviour we
+    store the config parameter if it is not present.
+    """
+    cr.execute("SELECT 1 FROM ir_config_parameter "
+               "WHERE name = 'partner_names_order'")
+    if not cr.fetchone():
+        cr.execute("INSERT INTO ir_config_parameter (key, value) VALUES "
+                   "('partner_names_order', 'last_first)")
```
yajo: :no_entry_sign: You're missing the closing `'`

yajo (changes requested):

partner_firstname/migrations/12.0.1.0.0/pre-ir_config_param.py
gdgellatly: Yh I caught that. I just did a big rebase and update. Lots of fixup commits in previous migrations, all between minutes and a few days apart.  All authors preserved.

partner_firstname/README.rst
eLBati: @gdgellatly if you want to generate it by yourself, you can use the `oca-gen-addon-readme` script from https://github.com/OCA/maintainer-tools#installation

partner_firstname/views/base_config_view.xml
```diff
@@ -23,7 +23,7 @@
                                     icon="fa-play"
                                     type="object"
                                     help="Recalculate names for all partners. This process could take so much time if there are more than 10,000 active partners"
-                                    attrs="{'invisible': [('partner_names_order_changed', '=', False)]}"/>
+                                    attrs="{'invisible': [('partner_names_order_changed', '=', True)]}"/>
```
eLBati: @gdgellatly why invisible if changed? Shouldn't this be as it was before?

eLBati:
For the rest, tested and LGTM

partner_firstname/views/base_config_view.xml
gdgellatly: @eLBati Because before it was broken and did the opposite of what was expected. It updated the partner names with the old value.  But you can't very well call set_values here as what if they changed something else, you can't write directly as that is counter to the rest of behaviour of settings options. so really the only way I could find to keep the function was to require an explicit write by the user and then an update.

eLBati: @gdgellatly could you document this with a comment?

gdgellatly: @eLBati in what way? it is in the commit message/PR already.

eLBati: @gdgellatly I meant a comment here in the XML file, to highlight the issue. Because I think it's not intuitive to allow to "Recalculate names for all partners" only when partners name order did not change

gdgellatly: @eLBati I ended up putting in an explicit write and reverting this view. 

This modules config.settings are not really well written, it really should have the config parameter as part of the module, with a noupdate, with a proper qualified name, and referenced by a config_parameter argument in the settings, Would get rid of a big percentage of the base settings code as well, but to keep backwards compatibility and same behaviour I just leave it as it is.  

I have no idea why it was decided to not set it in the first place and reimplement everything by hand. Maybe there is something non obvious I am missing.

eLBati: LGTM

eLBati (approved):

ddufresne (approved):
@gdgellatly Tested and works as expected. Thanks for the migration.

eLBati:
@Yajo what do you think?

mrcast (approved):
Tested and works as expected. Thanks

yajo (approved):

yajo:
Is any of this true?: https://github.com/OCA/partner-contact/pull/648#issuecomment-439791540

nikul-serpentcs (approved):
Code Review LGTM

eLBati:
> Is any of this true?: #648 (comment)

@Yajo they are old commits with old email addresses, see https://github.com/OCA/partner-contact/pull/648/commits/8b0bc3503f483656f7a9fce44126f6e0311bb5cd or https://github.com/OCA/partner-contact/pull/648/commits/efedffc8a0686ecdda9b8c8d0856b20f73e63d4b:
```
commit efedffc8a0686ecdda9b8c8d0856b20f73e63d4b
Author: Jairo Llopis <j.llopis@grupoesoc.es>
Date:   Fri May 15 13:59:59 2015 +0200
```

In order to fix them, @gdgellatly should rebase editing those commits using the up to date email addresses.

yajo:
OK, don't worry then. Thanks!

PR #652, opened 2018-11-13
12.0 mig partner_external_map

ernestotejeda:
Cc @Tecnativa


partner_external_map/__manifest__.py
```diff
@@ -0,0 +1,29 @@
...
+    'summary': 'Add Map and Map Routing buttons on partner form to '
+               'open GMaps, OSM, Bing and others',
+    'author': 'Akretion, '
+              'Tecnativa, '
+              'Odoo Community Association (OCA)',
+    'website': 'https://github.com/OCA/partner-contact',
+    'depends': [
+        'base',
+    ],
+    'data': [
```
nikul-serpentcs, line 20: @ernestotejeda Improve Data sequence should be [this](https://github.com/OCA/maintainer-tools/blob/master/template/module/__manifest__.py#L29)

nikul-serpentcs (commented):
Minor Change

partner_external_map/__manifest__.py
pedrobaeza: @nikul-serpentcs I wouldn't enforce that order as not always can be done that way due to XML-ID dependencies. There's no guideline about that either.

pedrobaeza (approved):

partner_external_map/readme/CONTRIBUTORS.rst
```diff
@@ -0,0 +1,6 @@
+* `Akretion <http://www.akretion.com>`_:
+  * Alexis de Lattre <alexis.delattre@akretion.com>
```
pedrobaeza, line 3: You should leave line breaks between this for proper rendering.

partner_external_map/__manifest__.py
nikul-serpentcs: Yes @pedrobaeza , But here `security` file is last, it's must be first

And we will try to make good sequence

pedrobaeza:
Can you add an address with all fields in the street for full module test coverage?

partner_external_map/__manifest__.py
pedrobaeza: Why? It doesn't matter. It's going to work exactly the same.

nikul-serpentcs: @pedrobaeza Because `security` files  it's must be first not last

pedrobaeza: No, where do you find that requirement?

nikul-serpentcs: @pedrobaeza  [here]( https://github.com/OCA/maintainer-tools/blob/master/template/module/__manifest__.py#L30)

but, Here not security XML file then It's Fine

pedrobaeza: No, that's a template where the order was not intended. Only an accident.

PR #654, opened 2018-11-14
12.0 mig partner multi relation

ddufresne:


pedrobaeza (approved):

partner_multi_relation/views/res_partner.xml
```diff
@@ -0,0 +1,57 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<odoo>
+
+        <record id="view_res_partner_filter" model="ir.ui.view">
+            <field name="name">partner_multi_relation.view_partner_filter</field>
+            <field name="inherit_id" ref="base.view_res_partner_filter" />
+            <field name="model">res.partner</field>
+            <field type="xml" name="arch">
+                <data>
```
nikul-serpentcs: @ddufresne Remove  `<data>`, not need

partner_multi_relation/tests/__init__.py
```diff
@@ -0,0 +1,6 @@
+# Copyright 2016 Therp BV
```
nikul-serpentcs: @ddufresne Remove Copyright from `init` file

nikul-serpentcs (commented):
Improve Code

NL66278 (approved):
:+1:  LGTM

partner_multi_relation/README.rst
```diff
@@ -1,83 +1,166 @@
...
-date.
+Partner Categories
+------------------
 
-More info
+You may use categories (tags) to further specify the type of partners.
+
+You could for example enforce the 'is member of' relation to accept only companies with the label 'Organization' on the right side.
+
+.. image:: readme/images/relation_type_form_category_filled.png
```
chienandalu: Shouldn't the images hang from `static`? otherwise the seem to be not reachable?

chienandalu (changes requested):
Just a little issue with the readme images. Otherwise ok :+1: Thanks!

partner_multi_relation/readme/USAGE.rst
```diff
@@ -0,0 +1,102 @@
+Relation Types
```
tarteo, line 1: https://github.com/OCA/maintainer-tools/blob/master/template/module/readme/USAGE.rst
"it MUST NOT contain reStructuredText sections only body text (paragraphs, lists, tables, etc)"

I'm not sure what the consequences are when you use sections in the usage, description etc. But is says here that you must not do it :S

chienandalu: Well, it depends on the paragraph level hierarchy. It would fail in CI if it wasn't correct

tarteo (approved):

partner_multi_relation/readme/USAGE.rst
pedrobaeza: The only level allowed here is `~~~~~~`

tarteo: Thanks!

tarteo: Oh so it must be changed?

pedrobaeza: I'm checking and it seems it's correctly rendered... @sbidoul `-----` titles are tolerated inside this section?

sbidoul: `~` is accepted yes, other may work or break anytime.

pedrobaeza: @ddufresne can you please switch to this title style for avoiding problems?

partner_multi_relation/README.rst
chienandalu: image

pedrobaeza: I'm able to see them:

seleccion_003

But the problem is that there's no good path for being displayed everywhere: apps store, README, Odoo viewer...

chienandalu: My screenshot was taken in the runbot instance app form wich seems to block any static resource linked outside the static folder.

pedrobaeza: It's not only that, but the path: although you put it inside static folder, you have to add the module name prefix, which screws then the visualization of the README on GitHub. I remember an issue about this, but I'm not able to find it. @elicoidal can you refresh my memory?

chienandalu (approved):

pedrobaeza:
@ddufresne please change title underline and for the images, there's a trick we can use: put the absolute URL that GitHub will give you when this got merged, and they will be shown properly on all the places.

You have some examples in mis-builder README: https://github.com/OCA/mis-builder/blob/10.0/mis_builder/README.rst

ddufresne:
@pedrobaeza I fixed the title underline and used your trick for images.

partner_multi_relation/tests/__init__.py
ddufresne: @nikul-serpentcs sorry, won't do that. I am not sure about OCA's policy about that and otherwise, I am not confortable removing someone's copyright from a python file.

Even though, I should not be the person commiting such change. Also, this should not block the PR.

partner_multi_relation/views/res_partner.xml
ddufresne: fixed

partner_multi_relation/tests/__init__.py
pedrobaeza: Well, it was a mistake to put copyright in a file that doesn't contain code, but it's not blocking either.

NL66278: We followed older guidelines to put copyright in ALL python files. I have no problem with removing it here and from any other ```__init__.py``` file.

pedrobaeza:
I'm going to squash some commits, amend copyright things after the grant of the author, and merge this.

pedrobaeza:
Thanks to all for reviewing and specially to @ddufresne for extending the README. Merging!

PR #655, opened 2018-11-16
[11.0][MIG] partner_employee_quantity: Migration to 11.0

pedrobaeza:
* Standard procedure
* Adapt dependencies
* Adapt menu
* Improve view layout
* README by fragments
* Rise to mature status with maintainer myself

@Tecnativa


partner_employee_quantity/models/res_partner.py
```diff
@@ -0,0 +1,24 @@
...
+# Copyright 2018 Tecnativa - Pedro M. Baeza
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import models, fields
+
+
+class ResPartner(models.Model):
+    _inherit = 'res.partner'
+
+    employee_quantity = fields.Integer()
```
nikul-serpentcs, line 12: @pedrobaeza IMO here need  `string="Employee quantity"`.

nikul-serpentcs (commented):
Minor Change

partner_employee_quantity/models/res_partner.py
pedrobaeza: That string is automatically got from the field name. In this case, it's "Employee Quantity", which is good enough.

nikul-serpentcs (approved):
Code Review LGTM

chienandalu (approved):
Tested on runbot. Looks good :+1:

PR #657, opened 2018-11-18
11.0 mig partner noncommercial

lfreeke:


partner_noncommercial/readme/ROADMAP.rst
```diff
@@ -0,0 +1,9 @@
+After installing this module a situation has been encountered where the new
```
NL66278: I think this whole section does not apply anymore, as we do no longer add a top menu item, but use the Contacts menu.

NL66278:
@lfreeke Now that we have the contacts main menu, it no longer makes sense to have a separate partner main menu entry. Please compare with https://github.com/NL66278/partner-contact/tree/10.0-partner_noncommercial-better-menu and check wether I miss something that you have. As I did not do a PR for my branch, keep al the needed changes in your branch and update the PR). One thing that needs still to be improved on my branch: configuration menu should go last.

lfreeke:
@NL66278 The separate partner menu is now gone and you can find the sub menu's in the contacts menu. The only difference between your  branch and this one is a couple of sequences and your configuration entry (I think that is needed for 10.0).

partner_noncommercial/views/menu.xml
```diff
@@ -0,0 +1,103 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+    <!--
```
NL66278: You can get rid of these comments as well, as we no longer use the sales_team stuff.

NL66278:
@lfreeke You also have an extra and unneeded level in the menu: Contacts==>Partners==>Partners (and the the menu options).

Merging my branch, that I have rebased on yours, will solve this. Also it renames the relation filter, which really is only for contacts that are partners, to contact. Thereby undoing a past mistake.

I can not make pull requests to your repository, so you need to change that or merge my branch yourself in yours.

See: https://github.com/NL66278/partner-contact/tree/11.0-mig-partner_noncommercial

lfreeke:
@NL66278 I merged your branch in. Is this correct now?

NL66278 (approved):
:+1: Looks good now. Only need some extra work for partner_multi_relation_tabs, but that can be done in a separate PR.

partner_noncommercial/views/menu.xml
```diff
@@ -0,0 +1,87 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+    <menuitem
+        id="menu_partner"
+        name="Partners"
+        parent="contacts.menu_contacts"
+        groups="base.group_user"
```
daramousk: Perhaps a `group_partner_manager` is needed here as in its parent?

daramousk (approved):
Just a comment

partner_noncommercial/views/menu.xml
NL66278: In this case might be better to not have a group at all, like on the other submenu's. Having base.group_user will not make a material difference in practice. But @lfreeke would you like to delete the groups= line?

gfcapalbo:
@lfreeke reviewed actions/menues and looks good. 
When you migrate something, add yourself to contributors manifest. and always write a line of description in MR.  :+1:

lfreeke:
@NL66278 I removed the group.

Issue #404, opened 2017-04-07
RFC - Keep partner data immutable for history purpose


bealdav:
As you Odoo res.partner concept don't keep history.

If you update a contact (address), older sales, picking and invoices address are also impacted by the update, even the oldiest.

I think we could find a way to keep history, at least for shipping address and invoice address.

### Approach 1
- when validating a Sale Order (same on picking and invoice) the linked address should be archived. This mean copying the existing address and creating an new one with active=False and not editable. This archived address will be used in the invoice, shipping...
- When validating an invoice or a shipping we do the same but if the address was already archived (from an sale order) we do not archived if again
- If you modify the delivery address after the sale order validation and before the shipping expedition a new archived address is created (We may use a wizard for this changed)
- when a Picking is validated we archived the address if not already archived
- when a Invoice is confirm we archived the address if not already archived
- we can have 3 modules, 1 for invoice, 1 for delivery and 1 for sale
- an archived address is always in readonly
- when creating an archived address we always look if an existing archived address already exist and always used an existing archived address instead of creating a new one

### Approach 2
- When editing an partner contact we always create a new contact with the new value and we archived the previous contact.
- When we edit a contact from a sale order, a picking or a invoice we should also replace many2one to use the new address created

legalsylvain:
Hi @bealdav. I'm not very in favor to create and archive new partners for that purpose. This is a reccurent problem that should be fixed with another approach, more lighter with a computed field, stored, that is invalidated when the order is confirmed.

### Approach 3

on purchase order, create a new field

```

delivery_adress_text = fields.Char(
    'delivery adress', compute=_compute_delivery_adress_text, store=True)

@api.depends('state')
def _compute_delivery_adress_text(self):
    for partner in self:
        if partner.state == 'confirmed':
            partner.delivery_adress_text = xxx
```
After, use this field instead of the partner adress.
That feat with the reality. When an order is confirmed, the delivery place is chosen and shouldn't change, even if the partner adress changes. (we can imagine a refresh button to update the field manually, calling the _compute_delivery_adress_text function)

It could be a simple community module to do.

what do you think ?

lasley:
This is one of my pain points with Odoo, and would be great to solve. I agree with @legalsylvain on not being in favor of creating new partners - it seems to me like we'd see a lot of edge cases.

[...]

PR #669, opened 2018-12-11
[10.0][ADD] module partner address version

bguillot:
Hi

This module sets versions on partners.
A versionned partner can't be modified.

It is a solution to the issue : https://github.com/OCA/partner-contact/issues/404

It is a base module, it should be used to apply the version when needed.
For the sale order the module is here : https://github.com/OCA/sale-workflow/pull/761


partner_address_version/__manifest__.py
```diff
@@ -0,0 +1,18 @@
...
+    'author': "Akretion, "
+              "Odoo Community Association (OCA)",
+    'website': 'http://www.akretion.com',
+    'category': 'CRM',
+    'license': 'AGPL-3',
+    'installable': True,
+    'depends': [
+        'base',
+    ],
+    'data': [
```
rousseldenis: Please remove blank entry

rousseldenis (changes requested):
@bguillot Could you maybe use https://github.com/OCA/maintainer-tools#readme-generator ? And complete manifest entry keys https://github.com/OCA/maintainer-tools/blob/master/template/module/__manifest__.py#L9

But I like the approach. Maybe you could add a little bit more explanation on how it works.

rvalyi:
I don't have an opinion on it yet, but here is some unofficial feedback from the rebellion https://twitter.com/cedrickrier/status/1072428174582136832

partner_address_version/models/res_partner.py
```diff
@@ -0,0 +1,60 @@
...
+                version[field] = self[field]
+        version_hash = hashlib.md5(str(version)).hexdigest()
+        return version_hash
+
+    @api.multi
+    def write(self, vals):
+        version_fields = self.get_version_fields()
+        written_versioned_fields = set(vals.keys()).intersection(
+            set(version_fields)
+        )
```
yvaucher, line 60: As you don't need the list of field put in written_versioned_fields. You could use a more efficient way by using `any`

```python
has_written_versioned_fields = any(
    (k in version_fields) for k in vals.keys()
)
```

Make sur that you use a generator (not using `[]` of comprehension list), it will be faster as it will exit the loop early.

partner_address_version/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1 @@
+This module allows to manage versions for partner addresses.
```
yvaucher: Can you write a sentence about the behavior when writting on a field that is versionned?

And maybe also why versionning is useful, for instance for e-commerce to track history.

partner_address_version/tests/test_address_version.py
```diff
@@ -0,0 +1,46 @@
...
+# Copyright 2018 Akretion - Benoît Guillot
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo.tests.common import TransactionCase
+from collections import OrderedDict
+import hashlib
+from odoo.exceptions import UserError
+
+
+class TestAddressVersion(TransactionCase):
```
yvaucher: You can use a SavepointCase here to create the partner only once.

yvaucher (approved):
LGTM, would help to document it a bit better, plus I added a performance suggestion on check of versionned fields.

partner_address_version/models/res_partner.py
```diff
@@ -0,0 +1,60 @@
...
+
+    @api.multi
+    def get_address_version(self):
+        """
+        Get a versioned partner corresponding to the partner fields.
+        If no versioned partner exists, create a new one.
+        """
+        self.ensure_one()
+        version_hash = self.get_version_hash()
+        versioned_partner = self.with_context(active_test=False).search(
```
yvaucher, line 24: What happens if there are multiple hits? Let's say 2 customers living together but for convenience make their delivery go to the work place of one. You would have the exact same hash but for 2 customers. You certainly want to have 1 shipping address even if the exact same on each customer, that way relation to parent_id is still correct.

yvaucher (changes requested):

bguillot:
Thank you @yvaucher !

I did the changes !

partner_address_version/__manifest__.py
```diff
@@ -0,0 +1,16 @@
+# -*- coding: utf-8 -*-
+# Copyright 2018 Akretion - Benoît Guillot
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+{
+    'name': 'Partner Address Version',
+    'version': '10.0.1.0.0',
+    'author': "Akretion, "
+              "Odoo Community Association (OCA)",
+    'website': 'http://www.akretion.com',
```
nikul-serpentcs: @bguillot Replace URL as per [this](https://github.com/OCA/maintainer-tools/blob/master/template/module/__manifest__.py#L11)

nikul-serpentcs (changes requested):
Minor Change

rousseldenis:
@bguillot Could you check travis ?

bguillot:
@rousseldenis I fixed the issue, travis is green now !

partner_address_version/models/res_partner.py
rousseldenis: @bguillot

bguillot: I added the parent_id in the version fields to handle this case.
 Yannick was with me at that time so we discussed it directly.

bguillot: I added the "parent_id" field in the version fields to handle this case.

rousseldenis (approved):
Code review

bguillot:
@yvaucher Is your request of changes satified ?

bguillot:
thanks @rousseldenis

nikul-serpentcs:
@bguillot Squash Commits

bguillot:
@nikul-serpentcs Squash done

nikul-serpentcs (approved):
only Code Review LGTM :+1:

PR #670, opened 2018-12-13
[12.0] mig partner ref unique

sbejaoui:


tbaden (approved):
code review: lgtm

tested in runbot

partner_ref_unique/models/res_partner.py
```diff
@@ -0,0 +1,32 @@
...
+                    (mode == 'companies' and partner.is_company))):
+                domain = [
+                    ('id', '!=', partner.id),
+                    ('ref', '=', partner.ref),
+                ]
+                if mode == 'companies':
+                    domain.append(('is_company', '=', True))
+                other = self.search(domain)
+
+                # active_test is False when called from
```
NL66278, line 27: Might there not be a lot of other situations where active_test is False? In all these cases the constraint will be disabled...

tbaden: as far as I can see the odoo core will only set active_test to false in base.partner.merge.automatic.wizard, like the comment stated. only 3rd party modules could bypass this.

NL66278 (approved):
:+1: Remark on ActiveTest not blocking, for the rest LGTM.

PR #672, opened 2018-12-14
[10.0] Add module nominatim_geolocalize

remi-filament:
This module is derived from base_geolocalize, but replacing Google API by Nominatim / OpenStreetMap one.


rafaelbn:
======================
Nominatim Geolocalize
======================

Get partner coordinate using Nominatim API

Alternative to Google API

pedrobaeza:
Module name should be `geolocalize_nominatim` for following guidelines.

nominatim_geolocalize/README.rst
```diff
@@ -0,0 +1,23 @@
...
+Get partner coordinate using Nominatim API
+
+Alternative to Google API
+
+
+
+Credits
+=======
+
+Contributors ------------
```
pedrobaeza: You can use readme by fragments for having this auto-populated. See template module: https://github.com/OCA/maintainer-tools/tree/master/template/module

remi-filament: Thanks @pedrobaeza, I renamed the module as per your suggestion, regarding the README, this is already done, only I kept README.rst for better readability of PR, but I understand this should be replaced by auto-populated one, no ?

pedrobaeza: Yes, sorry for not checking that part.

geolocalize_nominatim/__init__.py
```diff
@@ -0,0 +1,5 @@
+# -*- coding: utf-8 -*-
+# © 2018 Le Filament (<http://www.le-filament.com>)
```
NL66278: Please no copyright lines in init files.

geolocalize_nominatim/models/res_partner.py
```diff
@@ -0,0 +1,78 @@
...
+import urllib2
+
+from odoo import api, fields, models, tools, _
+from odoo.exceptions import UserError
+
+
+def geo_find(addr):
+    if not addr:
+        return None
+    url = 'https://nominatim.openstreetmap.org/search.php/'
```
NL66278: No need for three assignments. Do one, either using line continuation, or by putting the three partial strings between `(` and `)`.

geolocalize_nominatim/views/res_partner.xml
```diff
@@ -0,0 +1,24 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+        <record id="res_partner_nominatim_form" model="ir.ui.view">
```
NL66278: Decrease ident to 4.

NL66278 (changes requested):
:+1: Nice addition but needs some changes.

NL66278:
PR seems to be abandoned and is for no longer supported version.

remi-filament:
Thank you @NL66278 for reviewing this PR and making these changes
Sorry I was not able to be more reactive !
We are still using these functionalities in production for some customers that do not want to migrate to newer version, do you think it makes no sense to merge it in 10.0 branch ?
Best Regards

NL66278:
It makes sense. Just push the branch with the requested changes and I will re-open the PR and review again.

remi-filament:
It seems you have already pushed all the requested changes, or am I missing something ?

PR #673, opened 2018-12-14
12.0 mig partner company type mgs

GillesTephaneMeyomesse:
I think that the name of this module was not very representative of its functionality.
that's why I made some small changes to its migration


GillesTephaneMeyomesse:
@rousseldenis  @rmsolucoeseminformatica can you review please

rousseldenis:
@GillesTephaneMeyomesse the name was given to stay as generic as possible. But maybe it was too for you :smile:

sbidoul:
I personally like the new name, but that means a migration script. So better keep the other one for simplicity?

@GillesTephaneMeyomesse there is no need request review from someone who merely did a translation.

partner_legal_form/__manifest__.py
```diff
@@ -0,0 +1,23 @@
+# Copyright 2018 ACSONE SA/NV
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+{
+    'name': 'Partner Legal Form',
+    'summary': 'Adds a legal form to partner that are companies',
+    'version': '12.0.1.0.0',
+    'license': 'AGPL-3',
+    'author': 'ACSONE SA/NV,Odoo Community Association (OCA)',
+    'website': 'https://acsone.eu',
```
nikul-serpentcs: @GillesTephaneMeyomesse Replace URL as per [this](https://github.com/OCA/maintainer-tools/blob/master/template/module/__manifest__.py#L11)

nikul-serpentcs (commented):
Minor change

GillesTephaneMeyomesse:
Hello @rousseldenis ! As suggested @sbidoul  for simplicity I kept the name of the module ;)

partner_company_type/models/res_partner.py
```diff
@@ -1,5 +1,4 @@
-# -*- coding: utf-8 -*-
-# Copyright 2017 ACSONE SA/NV
+# Copyright 2018 ACSONE SA/NV
```
sbidoul: s/2018/2017-2018/

sbidoul (changes requested):
Can you check the runbot warnings?

GillesTephaneMeyomesse:
@sbidoul It is said "Two fields (partner_company_type_id, company_type) of res.partner() have the same label: Company Type." @rousseldenis that's why you used this migration script "'UPDATE res_partner SET partner_company_type_id = company_type_id'" ?

GillesTephaneMeyomesse:
@sbidoul @rousseldenis I think that renaming the label of the field "s/Company Type/Partner Company Type" it should solve the problem. Do you have a better idea ?

sbidoul:
Maybe the field should renamed to legal_form_id (with an oldname=...)? Or just change the label to Legal Form? Because, with the current situation a user doing an export will be confused with to field with a similar name (Company Type). @rousseldenis what do you think?

rousseldenis:
@sbidoul Ok, if the module is renamed as you wanted. Little OpenUpgrade script to change module name.

rousseldenis (approved):
Ok for me like that

partner_company_type/models/res_partner.py
```diff
@@ -0,0 +1,13 @@
...
+
+from odoo import fields, models
+
+
+class ResPartner(models.Model):
+    _inherit = "res.partner"
+
+    partner_company_type_id = fields.Many2one(
+        comodel_name='res.partner.company.type',
+        string='Partner Company Type',
```
sbidoul: @GillesTephaneMeyomesse As discussed, s/Partner Company Type/Legal Form/

partner_company_type/models/res_partner.py
nikul-serpentcs: @sbidoul IMO to change as `Copyright 2011-TODAY ACSONE SA/NV`

partner_company_type/models/res_partner.py
tbaden: @sbidoul seems like it's fixed. should be good to go after squash.

sbidoul (approved):

NL66278:
Please squash translation commits.

GillesTephaneMeyomesse:
yes, it is done

ThomasBinsfeld (approved):

rousseldenis:
@OCA/crm-sales-marketing-maintainers

PR #677, opened 2018-12-26
[12.0] partner_legal_name: Partner Legal Name module

DucTruongKomit:
_ Add a legal_name field
_Show legal_name into res_partner form and treeview


elicoidal (commented):
Thanks for the contribution!

I am not sure though of the functional interest of the module.
For me the legal name should always be in the field `name`.
Adding another field means that you need to potentially change a lot of places.

Is there some specific case where this module should apply (`name` could not contain the legal name for example)?

:-1:

DucTruongKomit:
@elicoidal  We created this field, because some company(for example in my country), they don't like to use the real name on the report because it's too long, so this field can solve that problem!

partner_legal_name/model/base/res_partner.py
```diff
@@ -0,0 +1,55 @@
...
+    @api.multi
+    def name_get(self):
+        res = []
+        for partner in self:
+            name_legal = partner._get_name()
+            res.append((partner.id, name_legal))
+        return res
+
+    def _get_name(self):
+        partner = self
```
NL66278, line 23: This is just creating confusion... No need for a different variable.

partner_legal_name/model/base/res_partner.py
```diff
@@ -0,0 +1,55 @@
...
+            name = name.replace('\n', '<br/>')
+        if self._context.get('show_vat') and partner.vat:
+            name = "%s - %s" % (name, partner.vat)
+        return name
+
+    @api.depends('is_company', 'legal_name', 'parent_id.name',
+                 'type', 'company_name')
+    def _compute_display_name(self):
+        for partner in self:
+            partner.display_name = partner.legal_name
```
NL66278, line 55: Why should the legal name be the display name?

NL66278 (changes requested):
This module seems to replace the actual name consistently with the legal name. Instead of having the legal name as an extra attribute for specific purposes. This IMHO should not be done like this.

NL66278:
No reaction to comments.

PR #678, opened 2018-12-27
[FIX] 11.0 partner_second_last_name required only if type = contact

j-zaballa:
Description of the issue/feature this PR addresses:

firstname, lastname and lastname2 fields should only be required if
contact `type` is `contact`. In partner_firstname module this is done
right with `('type', '=', 'contact')` leave in attrs required domain.
But in partner_second_last_name there is no such leave, so if the
contact is an address, the fields are mandatory too.

Current behavior:

Either firstname, lastname or lastname2 are mandatory when a contact is
not a company.

Desired behavior after PR is merged:

firstname, lastname or lastname2 are mandatory when the contact is of
type `contact`

### DIFF
--- a/partner_second_lastname/views/res_partner.xml
+++ b/partner_second_lastname/views/res_partner.xml
@@ -9,13 +9,13 @@
         <field name="arch" type="xml">
             <data>
                 <xpath expr="//field[@name='firstname']" position="attributes">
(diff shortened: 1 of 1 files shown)


pedrobaeza (approved):
Seems reasonable.

enriquemartin (approved):

PR #679, opened 2018-12-31
[12.0] [MIG] base_partner_merge

anandkansagra:
Issue [Reference](https://github.com/OCA/partner-contact/issues/630)


base_partner_merge/views/base_partner_merge.xml
```diff
@@ -0,0 +1,112 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<odoo>
+    <record model="ir.actions.act_window" id="action_partner_deduplicate">
```
tbaden: > When declaring a record in XML:
Place id attribute before model

id should be the first attribute in record

base_partner_merge/views/base_partner_merge.xml
```diff
@@ -0,0 +1,112 @@
...
+        <field name="name">Deduplicate Contacts</field>
+        <field name="res_model">base.partner.merge.automatic.wizard</field>
+        <field name="view_type">form</field>
+        <field name="view_mode">form</field>
+        <field name="target">new</field>
+        <field name="context">{'active_test': False}</field>
+    </record>
+
+    <record model="ir.ui.view" id="base_partner_merge_automatic_wizard_form">
```
tbaden: same as above:
id should be the first attribute in record

tbaden (changes requested):
thanks for the work. i did a code review, just some minor changes.

the record id's looked weird, but these are all id's from odoo core.

base_partner_merge/views/base_partner_merge.xml
anandkansagra: Done

tbaden (approved):
code review: lgtm

please squash your commits

LoisRForgeFlow (approved):
~Tested on runbot :+1: ~ EDIT: finally the module is not needed.

LoisRForgeFlow (changes requested):
Actually, this is now completly in the core base module: https://github.com/odoo/odoo/blob/12.0/odoo/addons/base/wizard/base_partner_merge_views.xml

So no need for this module anymore.

This can be closed.

LoisRForgeFlow:
Handled in OpenUpgrade here: https://github.com/OCA/OpenUpgrade/pull/1816

PR #680, opened 2019-01-03
[ADD]partner_ref_sequence

smangukiya:
@max3903 : Please review it. Thanks


partner_ref_sequence/__manifest__.py
```diff
@@ -0,0 +1,18 @@
...
+    "author": "Open Source Integrators, Odoo Community Association (OCA)",
+    "summary": "Generate Automatic customer reference number",
+    "category": "Customers",
+    "maintainer": "Open Source Integrators",
+    "website": "http://www.opensourceintegrators.com",
+    "depends": ["base"],
+    "data": [
+        "data/ir_sequence_data.xml",
+    ],
+    "installable": True,
```
max3903, line 17: @smangukiya You are missing the development_status and the maintainers keys. Website should be https://github.com/OCA/partner-contact

partner_ref_sequence/models/res_partner.py
```diff
@@ -0,0 +1,17 @@
+# Copyright (C) 2018 - TODAY, Open Source Integrators
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import api, models, fields
+
+
+class ResPartner(models.Model):
+    _inherit = 'res.partner'
+    _description = 'Add customer to customer number'
```
max3903: No need to reset the description

partner_ref_sequence/readme/CONFIGURE.rst
```diff
@@ -0,0 +1 @@
+* Install the module
```
max3903: To be removed.

partner_ref_sequence/readme/ROADMAP.rst
```diff
@@ -0,0 +1,2 @@
+The roadmap of the Partner contanct is documented on
```
max3903: `partner-contact`

partner_ref_sequence/readme/USAGE.rst
```diff
@@ -0,0 +1 @@
+* No specific steps for usage
```
max3903:
```
To use this module:

* Go to Contacts
* Create a partner
* The Customer Number field is automatically set with the next number in the sequence.

File: partner_ref_sequence/readme/DESCRIPTION.rst [hunk 6]
```diff
@@ -0,0 +1,2 @@
+This module will introduce automatic integer number generation for customer.
+And fill in ref field
```
max3903: `This module provides a sequence to generate an automatic number to set the customer number.`

max3903 (changes requested):

pedrobaeza:
Isn't this module the same as https://github.com/OCA/partner-contact/tree/11.0/base_partner_sequence?

smangukiya:
@max3903 : Source has been improved with all your comments. Thanks

smangukiya:
> Isn't this module the same as https://github.com/OCA/partner-contact/tree/11.0/base_partner_sequence?

@pedrobaeza : Yes, it seems with only difference that it will work for customer only. @max3903 : correct me if I am missing here. Thanks

pedrobaeza:
Then better to adapt that one and possibly expand it in a configurable way.

max3903:
Our customer is fine with the current feature set of base_partner_sequence.

PR #682, opened 2019-01-04
[IMP] partner_fax view

max3903:
Display the fax number after the mobile (instead of before the phone)
screenshot from 2019-01-04 14-08-54


bodedra (approved):
:+1:

osimallen (approved):
Approved

PR #684, opened 2019-01-07
[12.0][NEW] partner industry parent

sbejaoui:
I propose this module that add a parent relation to the partner industry


partner_industry_parent/__manifest__.py
```diff
@@ -0,0 +1,15 @@
...
+    'name': 'Partner Industry Parent',
+    'summary': """
+        This module add a parent relation to the partner industry""",
+    'version': '12.0.1.0.0',
+    'license': 'AGPL-3',
+    'author': 'ACSONE SA/NV,' 'Odoo Community Association (OCA)',
+    'website': 'https://github.com/OCA/partner-contact',
+    'depends': ["base"],
+    'data': ["views/res_partner_industry.xml"],
+    'demo': [],
```
nikul-serpentcs, line 14: @sbejaoui no need then remove it

partner_industry_parent/views/res_partner_industry.xml
```diff
@@ -0,0 +1,24 @@
...
+-->
+<odoo>
+    <!-- Inherit Form View to Modify it -->
+    <record id="res_partner_industry_view_form" model="ir.ui.view">
+        <field name="name">Industry (in partner_industry_parent)</field>
+        <field name="model">res.partner.industry</field>
+        <field name="inherit_id" ref="base.res_partner_industry_view_form"/>
+        <field name="arch" type="xml">
+
+            <xpath expr="//field[@name='name']" position="after">
```
nikul-serpentcs, line 14: @sbejaoui use a direct field, remove xpath.

nikul-serpentcs (changes requested):
Improve Code

partner_industry_parent/__manifest__.py
sbejaoui: @nikul-serpentcs , look [here](https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#modules).

>  The `website` key must be `https://github.com/OCA/<repo>`,
>     so as to provide the most relevant link to discover more information about the addon.
>     That link shows the repository README. Alternatively `https://github.com/OCA/<repo>/tree/<branch>/<addon>`
>     may be used, to provide a direct link to the addon README, which includes proper credits
>     (authors, contributors and theirs companies), and links to the relevant information on the OCA website.

partner_industry_parent/views/res_partner_industry.xml
sbejaoui: Thank you @nikul-serpentcs for reviewing this.

Why would you use direct field instead of xpath?

partner_industry_parent/__manifest__.py
nikul-serpentcs: @sbejaoui Yes, If not any file then remove it, no need.

It's not a big deal!! :smile:

ThomasBinsfeld (approved):

ThomasBinsfeld:
@nikul-serpentcs Since your review is about nitpicking's in code, could you approve the PR? It works well in production for months.

gva-acsone (approved):
Functional test +1

sbidoul (approved):

nikul-serpentcs (approved):
Only Code Review LGTM :+1:

partner_industry_parent/models/res_partner_industry.py
```diff
@@ -0,0 +1,35 @@
+# Copyright 2019 ACSONE SA/NV
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
```
NL66278: https:

partner_industry_parent/views/res_partner_industry.xml
NL66278: direct field is more easy to read, although in this case the difference is minimal

partner_industry_parent/models/res_partner_industry.py
```diff
@@ -0,0 +1,35 @@
...
+
+class ResPartnerIndustry(models.Model):
+
+    _inherit = 'res.partner.industry'
+    _parent_name = "parent_id"
+    _parent_store = True
+    _rec_name = 'complete_name'
+    _order = 'complete_name'
+
+    complete_name = fields.Char(
```
NL66278, line 15: You might consider to use the display_name instead of creating a separate field. If you rather use complete_name, should it not be visible in the form you create later? Because there might be multiple levels...

NL66278:
Code coverage decreases. Maybe add a test testing the computation of the complete name? Preferably with at least three levels...

NL66278:
@sbejaoui Could you take up some of the comments? Then I would be glad to merge this...

rafaelbn:
Hi @sbejaoui , please could you attend comments? https://github.com/OCA/partner-contact/pull/684#issuecomment-502881884

partner_industry_parent/models/res_partner_industry.py
sbejaoui: Complete_name is different than display_name. It contain the complete path for the current object to the first ancestor. Display name can contain the object name with its code for example.

Complete_name is used almost in every use case odoo standard set a parent store to true (hr.department, product.category..). So IMO, it's better to keep the same logic.

sbejaoui (commented):
> Code coverage decreases. Maybe add a test testing the computation of the complete name? Preferably with at least three levels...

Hi @NL66278 ,
Thank you for your review and sorry for the late response.

I added a unit test for the complete name compute method.

partner_industry_parent/models/res_partner_industry.py
NL66278: @sbejaoui OK, but that leaves my remark on showing somewhere the complete name, at least in the form.

sbejaoui: ok you are right, I missed that point. I pushed a new commit to add the complete name in the form view.
Thank you

NL66278 (approved):
:+1:  Thanks for taking up all comments!

LoisRForgeFlow:
@sbejaoui isn't this already covered here: https://github.com/OCA/partner-contact/blob/12.0/partner_industry_secondary/models/res_partner_industry.py#L18?

sbejaoui:
> @sbejaoui isn't this already covered here: https://github.com/OCA/partner-contact/blob/12.0/partner_industry_secondary/models/res_partner_industry.py#L18?

Now when I see it, yes it's the same. Thank you for your comment.
This kind of hidden features or non-exhaustive description leads to this duplication. The `partner_industry_secondary` module description doesn't mention it.

> This module adds secondary industries for partners.

sbidoul:
So what do we do here?

I propose to merge this one and for v13, remove the parent feature from partner_industry_secondary. Thoughts?

rafaelbn:
@sbidoul having a module in OCA that make something that already Odoo makes it's not good idea IMHO.

- Image of OCA: "They don't know that this is already in Odoo"
- People will use this module and mustn't as is already
- You have to take care in the migration
- Maybe someone make the migration of something is not needed

:thinking:

sbidoul:
> something that already Odoo makes

I'm confused, which part are you referring to?

NL66278:
@rafaelbn The parent feature is not in Odoo, but in another OCA module. I think having a secondary industry, and having an hierarchy in industries are separate concerns, that warrant separate modules. So would the solution proposed by @sbidoul then not be the best way to go?

sbejaoui:
> So what do we do here?
> 
> I propose to merge this one and for v13, remove the parent feature from partner_industry_secondary. Thoughts?

can we merge this PR and add a note in 13.0 migration issue to remove the parent feature from partner_industry_secondary

Issue #685, opened 2019-01-09
[12.0][base_partner_sequence] Same ref on write 


sbejaoui:
In the base_partner_sequence module, write on a recordset of partners with null ref give the same ref for all.
Run this test to reproduce:
 
```python
def test_unique_ref_on_write(self):
    vals = [
        {'name': "test1", 'email': "test@test.com"},
        {'name': "test2", 'email': "test@test.com"},
    ]
    partners = self.env['res.partner'].create(vals)
    self.assertFalse(partners[0].ref == partners[1].ref)
    partners.write({'ref': False})
    self.assertFalse(partners[0].ref)
    partners.write({})
    self.assertFalse(partners[0].ref == partners[1].ref)
```

NL66278:
Issue has been resolved.

PR #686, opened 2019-01-09
[12.0][base_partner_sequence] fix unique ref on write

sbejaoui:
fixes:
- [x] - #685

### DIFF
--- a/base_partner_sequence/tests/test_base_partner_sequence.py
+++ b/base_partner_sequence/tests/test_base_partner_sequence.py
@@ -33,3 +33,16 @@ class TestBasePartnerSequence(common.TransactionCase):
         self.assertEqual(
             self.partner.ref, contact.ref, "All it's ok as sequence doesn't "
                                            "increase.")
+
+    def test_unique_ref_on_write(self):
+        """Assert that create and write gives a different refs if we """
+        vals = [
+            {'name': "test1", 'email': "test@test.com"},
+            {'name': "test2", 'email': "test@test.com"},
+        ]
+        partners = self.env['res.partner'].create(vals)
+        self.assertFalse(partners[0].ref == partners[1].ref)
+        partners.write({'ref': False})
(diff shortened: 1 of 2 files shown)


pedrobaeza:
If you are notifying an issue at the same time you fix it with a PR, better to simply put the issue text in the PR.

base_partner_sequence/tests/test_base_partner_sequence.py
```diff
@@ -7,29 +7,49 @@
 
 
 class TestBasePartnerSequence(common.TransactionCase):
-
     def setUp(self):
         super(TestBasePartnerSequence, self).setUp()
 
         self.res_partner = self.env['res.partner']
-        self.partner = self.res_partner.create({
```
pedrobaeza: Don't do stylistic changes like this one. Always keep diff minimum if possible.

base_partner_sequence/tests/test_base_partner_sequence.py
```diff
@@ -7,29 +7,49 @@
...
+        self.partner = self.res_partner.create(
+            {'name': "test1", 'email': "test@test.com"}
+        )
 
     def test_ref_sequence_on_partner(self):
         # Test sequence on creating partner and copying it
         self.assertTrue(self.partner.ref, "A partner has always a ref.")
 
         copy = self.partner.copy()
-        self.assertTrue(copy.ref, "A partner with ref created by copy "
```
pedrobaeza: The same, and you change is not exactly Python correct

pedrobaeza (changes requested):

base_partner_sequence/tests/test_base_partner_sequence.py
sbejaoui: My bad, automatic style manager, fixed

pedrobaeza (approved):

astirpe (approved):

PR #688, opened 2019-01-15
[IMP] Added is_unique check

osi-scampbell:
Add is_unique field on a res.partner.relation.type form and added method to check if a relation being created or written is unique

### DIFF
--- a/partner_multi_relation/models/res_partner_relation.py
+++ b/partner_multi_relation/models/res_partner_relation.py
@@ -167,29 +167,21 @@ class ResPartnerRelation(models.Model):
 
     @api.multi
     def check_is_unique(self, vals):
-        type_id = self.env['res.partner.relation.type']
-        for line in self.env['res.partner.relation'].search([]):
-            if line.type_id.is_unique:
-                count = 0
-                if 'right_partner_id' in vals:
-                    if vals.get('right_partner_id') == \
-                       line.right_partner_id.id:
-                        count += 1
-                else:
-                    if self.right_partner_id == line.right_partner_id.id:
-                        count += 1
-                if 'type_id' in vals:
-                    if type_id.browse(vals.get('type_id')).name == \
-                       line.type_id.name:
-                        count += 1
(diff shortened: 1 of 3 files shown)


partner_multi_relation/models/res_partner_relation_type.py
```diff
@@ -73,6 +73,12 @@ class ResPartnerRelationType(models.Model):
              " that do not fit the new criteria.\n"
              "Specify how this situation should be handled.",
     )
+    is_unique = fields.Boolean(
+        string='Is Unique',
+        help="If a relation type is unique, then there cannot be 2"
+             "relations with the same parties and the same type.",
+        default=False
```
nikul-serpentcs: @osi-scampbell No need to `default` False

partner_multi_relation/models/res_partner_relation.py
```diff
@@ -154,3 +164,32 @@ def _check_relation_uniqueness(self):
...
+                    if (type_id.browse(vals.get('type_id')).is_unique
+                       or line.type_id.is_unique):
+                        count += 1
+                else:
+                    if self.type_id.name == line.type_id.name:
+                        count += 1
+                    if self.type_id.is_unique or line.type_id.is_unique:
+                        count += 1
+                if count == 3:
+                    return False
```
max3903: @osi-scampbell Why don't you just search if the same relation already exists?

Something like:
```
left = vals['left_partner_id'] or self.left_partner_id.id
type = vals['type_id] or self.type_id.id
right = vals['right_partner_id'] or self.right_partner_id.id
res = self.search([('left_partner_id', '=', left),
                    'type_id', '=', type),
                    'right_partner_id', '=', right)])
...
```

max3903 (approved):
LGTM. No tests.

osimallen (approved):
Approved!

wolfhall (approved):

pedrobaeza:
Why not doing this with an `@api.constrains`?

max3903:
@pedrobaeza Good point. #697

PR #700, opened 2019-01-31
[12.0] Migrate partner_disable_gravatar to v12

alexis-via:


partner_disable_gravatar/README.rst
```diff
@@ -0,0 +1,60 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
```
NL66278, line 2: Please change the url's on the license keys to https:

NL66278 (approved):
:+1:  One nag. But looks good to me. Could request the new readme structure, but I think that would be overkill for this module. Immediately going to activate it for all our customers...

partner_disable_gravatar/models/partner.py
```diff
@@ -0,0 +1,12 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import models
+
+
+class ResPartner(models.Model):
+    _inherit = 'res.partner'
+
+    def _get_gravatar_image(self, email):
+        return False
```
NL66278, line 12: You could put an ```# pragma: no cover``` at the end of this (trivial) line to prevent the checks from failing because of lowered coverage.

lmignon (approved):
Thank you @alexis-via

partner_disable_gravatar/README.rst
alexis-via: I DID convert the README.rst to the new format with the readme subdir, so it will be re-generated. I didn't delete the README.rst, but I can do it if it's what I'm supposed to do.

pedrobaeza: It's better to not remove it, as the diff raises. You can pre-generate ir however in the same commit for avoiding an extra one, but it's optional. It's recommended any way to mention that on the first comment for avoiding this kind of misunderstandings with the reviewers.

alexis-via: Thanks for this clarification @pedrobaeza

NL66278: @pedrobaeza Thanks also from my side. @alexis-via the remark on https instead of http still stands for the *.py files in the module.

elicoidal: additionally svg -> png

pedrobaeza: @elicoidal it applies the same: when README is put by fragments, no review should be done in it.

elicoidal: Sorry you are right! (automatic generation :shame:)

yajo (approved):
Please consider opening a PR to odoo master, where you add support for a `ir.config_parameter` that disables this upstream, and possibly consider adding that to the docs in the same PR.

Maintaining 1400+ LOC for this is absurd IMHO (not your fault!), unless strictly required.

Thanks!

alexis-via:
@Yajo Yes, I agree with that. But too many of my PR on odoo are ignored or rejected for bad reasons, so it kills my motivation to do so.

PR #704, opened 2019-02-13
[10.0] store field risk_exception

pedrocs-exo:
This way we can use the field in filters and quickly determine which partners are over their limits

### DIFF
--- a/partner_financial_risk/models/res_partner.py
+++ b/partner_financial_risk/models/res_partner.py
@@ -102,7 +102,8 @@ class ResPartner(models.Model):
     risk_exception = fields.Boolean(
         compute='_compute_risk_exception',
         string='Risk Exception',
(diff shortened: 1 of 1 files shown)


pedrobaeza:
@carlosdauden any reason for not being stored?

carlosdauden:
If store a compute hook is required when we install other related modules that extends this depends:
https://github.com/OCA/partner-contact/blob/4814218807e3ccc59378db7a46976f6362b711b2/partner_financial_risk/models/res_partner.py#L210

pedrobaeza (changes requested):
Thanks for the clarification, Carlos.

Pedro, you should add these hooks (in post-init-hook for example) for this to be valid. And then we will have a problem of performance while installing the modules...

NL66278:
PR seems to be abandoned by the developer.

PR #706, opened 2019-02-18
[10.0][ADD] partner_bank_sort_code: Add Sort Code on Banks

rousseldenis:


pedrobaeza:
What is the goal of this? Do you mean `short code` instead?

rousseldenis:
@pedrobaeza Nope. You didn't see this : https://github.com/OCA/partner-contact/pull/706/commits/ea6798937caf43b052c412c31d9981c69ddde1cb#diff-eec3b75ecd3331ec2ef5b65c5eb43154R30

pedrobaeza:
Nope, I didn't see that. It seems a bit localization specific... but I don't block anything. Too much opened fences, hehe.

rousseldenis:
That depends on countries, but as I know so far, UK, Ireland and Germany use these codes for interbanking transfers. That's the aim of this to modelize that feature.

partner_bank_sort_code/tests/test_sort_code.py
```diff
@@ -0,0 +1,41 @@
...
+        }
+        self.bank_2 = self.bank_obj.create(vals)
+
+    def test_sort_code_duplicate(self):
+        vals = {
+            'name': 'BANK 3',
+            'sort_code': '95-01-32',
+        }
+        with self.assertRaises(IntegrityError):
+            self.bank_obj.create(vals)
```
sbidoul, line 36: add a test with active=False that should not raise?

Nice trick with the unique constraint.

sbidoul (approved):
A small test improvement suggestion. Otherwise LGTM.

partner_bank_sort_code/tests/test_sort_code.py
rousseldenis: @sbidoul Done

partner_bank_sort_code/models/res_bank.py
```diff
@@ -0,0 +1,18 @@
...
+
+class ResBank(models.Model):
+
+    _inherit = 'res.bank'
+
+    sort_code = fields.Char()
+
+    _sql_constraints = [(
+        "sort_code_unique",
+        "EXCLUDE (sort_code WITH =) WHERE (active = True)",
```
lmignon, line 16: @rousseldenis  First time I read this syntax but it looks like you exclude the active record from the unique contrains. I've to check the postgresql documentation.

rousseldenis: Yes, the term is ambiguous. I've added tests and comments in order to be consistent

lmignon: @rousseldenis Great! Also tested on my side....

lmignon (approved):
LGTM (Code review)

Cedric-Pigeon (approved):
Code review :+1:

rousseldenis:
this is green :smile:  :green_apple:

PR #716, opened 2019-03-13
[12.0] FIX base_location blocking editing company address.

eLBati:
Steps:

 - Open company form
 - Set 'city completion' field

Get 'The state of the partner My Company differs from that in location X'

Disabling _check_zip while writing 'zip' fields from company, as incompatible with the sequence of write operations

### DIFF
--- a/base_location/__manifest__.py
+++ b/base_location/__manifest__.py
@@ -4,7 +4,7 @@
 
 {
     'name': 'Location management (aka Better ZIP)',
-    'version': '12.0.1.0.0',
+    'version': '12.0.1.0.1',
     'depends': [
         'base_address_city',
         'contacts',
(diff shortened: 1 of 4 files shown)


pedrobaeza:
Maybe this can be avoided having the same inverse method for both fields?

eLBati:
@pedrobaeza we should also have the same method for country and state, as they are checked in `_check_zip` too. But `_inverse_country` and `_inverse_state` are defined in `base` module.

Anyway if you have some alternative idea, I can test

pedrobaeza:
OK, it was just an idea. I haven't tried anything, so I suppose yours is a good solution. I would like any way to have a test for this use case. Is it possible?

eLBati:
@pedrobaeza test added:
line https://github.com/OCA/partner-contact/pull/716/files#diff-2bebd79e23619ec66684fc32c70ef871R138
would fail without this fix

pedrobaeza (approved):
Great, thanks for the test addition for avoiding regressions

mrcast (approved):
Functional test ok

PR #719, opened 2019-03-20
[11.0][mig] partner_academic_title

luismontalba:
Migration to 11.0.
This module was written to adds possibility to define some academic titles and
to add multiple titles on partners. there is a sequence on titles to display
them in an order on the partner.


tbaden (changes requested):
thanks for the work. please squash your commits.

code lgtm

luismontalba:
@tbaden commits squashed

tbaden (approved):

partner_academic_title/data/partner_academic_title_data.xml
```diff
@@ -0,0 +1,51 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Copyright 2015-2017 ACSONE SA/NV
+     License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl). -->
+<odoo>
+    <data noupdate="1">
```
NL66278: data element is no longer needed. noupdate attribute can be added to odoo element. Please remove data and decrease indents of records.

partner_academic_title/models/partner_academic_title.py
```diff
@@ -0,0 +1,16 @@
+# Copyright 2015-2017 ACSONE SA/NV (<http://acsone.eu>)
```
NL66278: Put https: on copyright and license lines.

partner_academic_title/models/res_partner.py
```diff
@@ -0,0 +1,38 @@
...
+
+
+class ResPartner(models.Model):
+    _inherit = 'res.partner'
+
+    def _get_separator(self):
+        return ', '
+
+    @api.depends('academic_title_ids', 'academic_title_ids.sequence')
+    @api.one
```
NL66278: @api.one is deprecated. Use self.ensure_one() as first line in method.

partner_academic_title/models/res_partner.py
```diff
@@ -0,0 +1,38 @@
...
+    def _compute_academic_title_display(self):
+        display_title = ""
+        separator = self._get_separator()
+        title_ids = self.academic_title_ids.sorted(lambda r: r.sequence)
+        for title in title_ids:
```
NL66278: More pythonic (and readable) to do something like separator.join([title.name for title in title_ids])

partner_academic_title/views/res_partner_view.xml
```diff
@@ -0,0 +1,23 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Copyright 2015-2017 ACSONE SA/NV
+     License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl). -->
+<odoo>
+    <data>
```
NL66278: Get rid of data tag here as well.

partner_academic_title/models/partner_academic_title.py
```diff
@@ -0,0 +1,16 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import models, fields
+
+
+class PartnerAcademicTitle(models.Model):
+    _name = 'partner.academic.title'
+
+    name = fields.Char(required=True,
```
NL66278: Please use the same identation style here as for the field definitions in res_partner.py

NL66278 (changes requested):
Nice additions, but code can use some improvements. See separate comments.

partner_academic_title/models/res_partner.py
luismontalba: Changes done, but ensure_one() causes an error when building the runbot instance because the method is applied on a set of records. Maybe iterating with something like "for rec in self" inside the function?

NL66278: Yes that would be the solution. I had not taken into account the possibility that this could be called for multiple records. ```for rec in self``` or ```for this in self``` as i prefer (but that is purely personal taste), is fine.

NL66278 (approved):

luismontalba:
@NL66278 changes done, can I squash my commits?

luismontalba:
OK, I didn't see the approval. Commits squashed

PR #721, opened 2019-03-22
[MIG] crm_deduplicate_acl > partner_deduplicate_acl: Migration to 12.0

victor-ditay:
This comes from `crm_deduplicate_acl` module in OCA/crm

cc @Tecnativa


pedrobaeza:
Please check error in menus:

```
ValueError: External ID not found in the system: crm.crm_menu_root
```

I think you can stripe out these menus. What I'm missing right now is the possibility of launching the wizard for finding duplicates, which is useful for several of our customers (and the base for other modules like old `crm_deduplicate_by_ref`), so it can be convenient to add that menu inside Contacts section.

victor-ditay:
>I think you can stripe out these menus. What I'm missing right now is the possibility of launching the wizard for finding duplicates, which is useful for several of our customers (and the base for other modules like old crm_deduplicate_by_ref), so it can be convenient to add that menu inside Contacts section.

@pedrobaeza I'll have a check

pedrobaeza (approved):

luismontalba (approved):
Tested on runbot

PR #723, opened 2019-03-25
[FIX] partner_phone_search: Return None when name is False/empty

igallart:
When we do not have a name, it returns None, and another module that read the result could fail.
For example, in the previous commit of this module: res + = super (ResPartner, self) .name_search ..., will fail.

### DIFF
--- a/partner_phone_search/models/res_partner.py
+++ b/partner_phone_search/models/res_partner.py
@@ -15,7 +15,7 @@ class ResPartner(models.Model):
             domain = ['|', '|', ('phone', operator, name),
                       ('mobile', operator, name), ('email', operator, name)
                       ]
-            partners = self.search(domain + args, limit=limit,)
+            partners = self.search(domain + args, limit=limit, )
             res = partners.name_get()
             if limit:
                 limit_rest = limit - len(partners)
@@ -23,6 +23,9 @@ class ResPartner(models.Model):
                 limit_rest = limit
             if limit_rest or not limit:
                 args += [('id', 'not in', partners.ids)]
-                res += super(ResPartner, self).name_search(
+                res += super().name_search(
                     name, args=args, operator=operator, limit=limit_rest)
             return res
+        return super().name_search(
+            name, args=args, operator=operator, limit=limit
+        )
--- a/partner_phone_search/README.rst
+++ b/partner_phone_search/README.rst
@@ -31,6 +31,7 @@ Contributors
 
 * Serpent Consulting Services Pvt. Ltd. <support@serpentcs.com>
 * Nikul Chaudhary <nikul.chaudhary.serpentcs@gmail.com>
+* Isaac Gallart <igallart@puntsistemes.es>
 
 Maintainer
 ----------
--- a/partner_phone_search/__manifest__.py
+++ b/partner_phone_search/__manifest__.py
@@ -4,7 +4,7 @@
 
 {
     'name': 'Search Partner Phone/Mobile/Email',
-    'version': '11.0.1.0.0',
+    'version': '11.0.1.0.1',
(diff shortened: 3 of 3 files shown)


tarteo (commented):
I think it's better to do the super before ```if name``` and than return the result at the end. And change the result of super in ```if name``` something like ```result = res + result```.

igallart:
> I think it's better to do the super before `if name` and than return the result at the end. And change the result of super in `if name` something like `result = res + result`.

If you does the super before name and name is not false, you has doing two super unnecessarily. And, the first super is not valid in 

> args += [('id', 'not in', partners.ids)]
> return res + super(ResPartner, self).name_search(
> name, args=args, operator=operator, limit=limit_rest)

because args are differents.

tarteo (approved):
@igallart Ah yes, makes sense. Thanks!

rafaelbn:
Hello @igallart , is it posiible to increase  codecov/patch — 66.66% of diff hit (target 97.69%) ? :smile:

igallart:
> Hello @igallart , is it posiible to increase codecov/patch — 66.66% of diff hit (target 97.69%) ? smile

@rafaelbn, I'm God! Mmmm, I don't know how is that, I only repair the error of module...

partner_phone_search/models/res_partner.py
```diff
@@ -15,14 +15,15 @@ def name_search(self, name, args=None, operator='ilike', limit=100):
...
             res = partners.name_get()
             if limit:
                 limit_rest = limit - len(partners)
             else:
                 limit_rest = limit
             if limit_rest or not limit:
                 args += [('id', 'not in', partners.ids)]
-                res += super(ResPartner, self).name_search(
-                    name, args=args, operator=operator, limit=limit_rest)
+                res += super().name_search(name, args, operator, limit_rest)
```
NL66278: You should call keyword arguments with keywords. So limit your change to removing the unneeded arguments from the super() call.

partner_phone_search/models/res_partner.py
```diff
@@ -15,14 +15,15 @@ def name_search(self, name, args=None, operator='ilike', limit=100):
...
             return res
+        else:
+            return super().name_search(name, args, operator, limit)
```
NL66278: Same here,

NL66278 (changes requested):
See comments: use keywords for arguments that are defined as keyword arguments.

NL66278 (changes requested):
The code itself looks mostly OK, but please add a test where you search on phone or mobile. Also test the situation that phone/mobile/email are not present.

To increase performance, you could check for digits in the name argument, before searching on phone/mobile. You could also check for '@' in the name argument before searching on email. Actually searching on phone/mobile and on email are mutually exclusive considering this.

In summary: if no digits or @ in name, just return super. If @ in name search on email. Else search on phone or mobile. Note that an email address can contain digits, but a phone number cannot contain a @.

igallart:
> The code itself looks mostly OK, but please add a test where you search on phone or mobile. Also test the situation that phone/mobile/email are not present.
> 
> To increase performance, you could check for digits in the name argument, before searching on phone/mobile. You could also check for '@' in the name argument before searching on email. Actually searching on phone/mobile and on email are mutually exclusive considering this.
> 
> In summary: if no digits or @ in name, just return super. If @ in name search on email. Else search on phone or mobile. Note that an email address can contain digits, but a phone number cannot contain a @.

@NL66278 , thanks by your comments, but I don't have time to implement all those improvements. I did the pull request to solve a fatal error under certain circumstances, therefore, if someone can make the proposed improvements it would be perfect, but I can't dedicate that time to a module that I don't think is essential either.

PR #724, opened 2019-03-29
[12.0][Add] partner_company_group

tschanzt:
Adds a company group field on the partner and a related field on lead, sale order and invoice to allow grouping and searching for the company group.

### DIFF
--- a/partner_company_group/__manifest__.py
+++ b/partner_company_group/__manifest__.py
@@ -1,0 +1,24 @@
+# Copyright 2019 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+{
+    'name': 'Partner Company Group',
+    'summary': 'Adds the possibility to add a company group to a company',
+    'version': '12.0.1.0.0',
+    'category': 'Sales',
+    'author': 'Camptocamp SA, Odoo Community Association (OCA)',
+    'license': 'AGPL-3',
+    'depends': [
+        'base',
+        'account',
+        'crm',
+        'sale',
+    ],
+    'website': 'https://github.com/OCA/partner-contact',
+    'data': [
+        'views/opportunity_view.xml',
+        'views/contact_view.xml',
+        'views/sale_order_view.xml',
+        'views/account_invoice_view.xml',
+    ],
+    'installable': True,
+}
--- a/partner_company_group/models/account_invoice.py
+++ b/partner_company_group/models/account_invoice.py
@@ -1,0 +1,13 @@
+# Copyright 2019 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import models, fields
+
+
+class AccountInvoice(models.Model):
+    _inherit = 'account.invoice'
+
+    company_group_id = fields.Many2one(
+        related='partner_id.company_group_id',
+        store=True
+    )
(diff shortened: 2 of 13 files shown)


partner_company_group/__manifest__.py
```diff
@@ -0,0 +1,24 @@
...
+    'category': 'Sales',
+    'author': 'Camptocamp',
+    'license': 'AGPL-3',
+    'depends': [
+        'base',
+        'account',
+        'crm',
+        'sale',
+    ],
+    'website': 'http://www.camptocamp.com',
```
nikul-serpentcs: `https://github.com/OCA/partner-contact`

nikul-serpentcs (changes requested):
netpicking

tschanzt:
@nikul-serpentcs updated.

partner_company_group/models/account_invoice.py
```diff
@@ -0,0 +1,13 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import models, fields
+
+
+class AccountInvoice(models.Model):
+    _inherit = 'account.invoice'
+
+    company_group_id = fields.Many2one(
+        related='partner_id.company_group_id',
```
mpanarin, line 11: readonly?

partner_company_group/models/crm_lead.py
```diff
@@ -0,0 +1,13 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import models, fields
+
+
+class CrmLead(models.Model):
+    _inherit = 'crm.lead'
+
+    company_group_id = fields.Many2one(
+        related='partner_id.company_group_id',
```
mpanarin, line 11: same

partner_company_group/models/sale_order.py
```diff
@@ -0,0 +1,13 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import models, fields
+
+
+class SaleOrder(models.Model):
+    _inherit = 'sale.order'
+
+    company_group_id = fields.Many2one(
+        related='partner_id.company_group_id',
```
mpanarin, line 11: same

mpanarin (changes requested):

partner_company_group/models/account_invoice.py
nikul-serpentcs: @mpanarin No need to `readonly = True` when a field is related.

Everywhere

mpanarin: You will be able to change the field and, therefore, changing the value on a different record completely. This is not very user-friendly and obscure, as well as should be avoided. 
So I still think they should be readonly

tschanzt: @mpanarin related fields are readonly=true as a default in v12. although i would have forgotten it.
https://github.com/odoo/odoo/blob/12.0/odoo/fields.py#L427

nikul-serpentcs: @mpanarin In Odoo Base version 12, related always `readonly = True` look like a `compute/ Functional` fields, so no need to added this type of attribute

simahawk (approved):

leemannd (approved):

leemannd (approved):

mpanarin (approved):

gurneyalex (approved):

PR #731, opened 2019-04-23
[FIX] make res.partner.relation.all#write work for multiple records

hbrunn:
accessing `self`'s fields in an `@api.model` function is dangerous, and we just need the model anyways. I changed this to always return a recordset, it will be falsy if empty


NL66278 (approved):
:+1:  LGTM

tbaden (approved):
lgtm

PR #732, opened 2019-05-02
[ADD] partner_priority module

patrickrwilson:
This module adds a priority field to contacts which can be configured within the configuration menu.


patrickrwilson:
Runbot states it has an error however I'm not sure why. I do see some warnings but I wouldn't think those would be the cause. If someone has some additional insight on it, that would be helpful. Thanks in advance.

partner_priority/data/partner_priority_data.xml
```diff
@@ -0,0 +1,27 @@
+<!-- Copyright 2019 Patrick Wilson <patrickraymondwilson@gmail.com>
+     License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl). -->
+
+<odoo>
```
nikul-serpentcs:
```suggestion
<odoo noupdate="1">
```

And remove `<data noupdate="1">`

nikul-serpentcs (changes requested):
Minor Change

partner_priority/data/partner_priority_data.xml
patrickrwilson: The reason I have the noupdate="1" is I think it's best practice to not re-import these when the module is updated, because if the company/user using this changes the values for their own use it shouldn't be changed back if its updated. I'm open to discussion though, I almost just excluded this data file altogether as it's not necessarily needed. Thoughts?

NL66278: @patrickrwilson I think you misread what @nikul-serpentcs is proposing. The attribute 'noupdate' can now be part of the odoo element. The data element is deprecated. So you can (and should) still use noupdate, but move it to the odoo element.

patrickrwilson: Yes, @NL66278 you are correct, I didn't realize you can put the 'noupdate' in the odoo element so I was misreading @nikul-serpentcs comment. Thanks. 

I've made this adjustment and squashed commits.

max3903 (approved):

partner_priority/README.rst
```diff
@@ -0,0 +1,25 @@
+.. image:: https://img.shields.io/badge/licence-LGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/lgpl-3.0-standalone.html
```
NL66278: Change license lines to https: throughout module.

NL66278:
With most OCA modules we have an AGPL license, instead of an LGPL. AGPL gives the user some more rights compared to LGPL. Is there a reason you use LGPL for this module?

EDIT: I see you have LGPL on the README.rst, but AGPL on other files. You have to be consequent. Would prefer AGPL everywhere.

partner_priority/__manifest__.py
```diff
@@ -0,0 +1,23 @@
...
+
+{
+    'name': 'Partner Priority',
+    'summary': "Adds priority to partners.",
+    'author': "Patrick Wilson, Odoo Community Association (OCA)",
+    'website': "https://github.com/OCA/partner-contact",
+    'category': 'Extra Tools',
+    'version': '12.0.1.0.0',
+    'license': 'AGPL-3',
+    'depends': ['base',
```
NL66278: Indent with four spaces, like on data keyword.

partner_priority/__manifest__.py
```diff
@@ -0,0 +1,23 @@
+# Copyright 2019 Patrick Wilson <patrickraymondwilson@gmail.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+{
+    'name': 'Partner Priority',
+    'summary': "Adds priority to partners.",
+    'author': "Patrick Wilson, Odoo Community Association (OCA)",
+    'website': "https://github.com/OCA/partner-contact",
+    'category': 'Extra Tools',
```
NL66278: The category Extra Tools does not really fit. Use "Customer Relationship Management" or "Partner Management".

partner_priority/models/partner_priority.py
```diff
@@ -0,0 +1,21 @@
...
+from odoo import models, fields, api
+
+
+class PartnerPriority(models.Model):
+    _name = 'partner.priority'
+    _order = 'sequence'
+    _description = 'Partner Priority'
+
+    name = fields.Char(string="Priority", required=True)
+    description = fields.Text(string="Description")
```
NL66278: You do not need the strings if the are the same as the field name (they will automatically start with a capital).

partner_priority/views/partner_priority.xml
```diff
@@ -0,0 +1,59 @@
+<!-- Copyright 2019 Patrick Wilson <patrickraymondwilson@gmail.com>
+     License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl). -->
+
+<odoo>
+    <!-- Sequences-->
```
NL66278: This should go in a separate file in the data directory.

NL66278 (changes requested):
Nice addition but needs some changes for conventions. See separate comments.

partner_priority/views/partner_priority.xml
```diff
@@ -0,0 +1,59 @@
...
+        <field name="arch" type="xml">
+            <tree string="Priority">
+                <field name="sequence" widget="handle"/>
+                <field name="name"/>
+                <field name="description"/>
+            </tree>
+        </field>
+    </record>
+
+    <!-- Form View -->
```
luisiniguezh: comments on this file seem kind of redundant

partner_priority/models/partner_priority.py
```diff
@@ -0,0 +1,21 @@
+# Copyright 2019 Patrick Wilson <patrickraymondwilson@gmail.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import models, fields, api
+
+
+class PartnerPriority(models.Model):
+    _name = 'partner.priority'
+    _order = 'sequence'
+    _description = 'Partner Priority'
```
luistorresm, line 10: Could you improve the description, please?

partner_priority/models/partner_priority.py
```diff
@@ -0,0 +1,21 @@
...
+    _name = 'partner.priority'
+    _order = 'sequence'
+    _description = 'Partner Priority'
+
+    name = fields.Char(string="Priority", required=True)
+    description = fields.Text(string="Description")
+    sequence = fields.Integer(string="Sequence")
+
+    @api.model
+    def create(self, vals):
```
luistorresm: Why not a default in the field, to avoid inheriting the create method

partner_priority/models/res_partner.py
```diff
@@ -0,0 +1,10 @@
+# Copyright 2019 Patrick Wilson <patrickraymondwilson@gmail.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import models, fields
+
+
+class ResPartner(models.Model):
+    _inherit = 'res.partner'
+
+    priority = fields.Many2one('partner.priority')
```
luistorresm:
```suggestion
    priority_id = fields.Many2one('partner.priority')
```

partner_priority/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1 @@
+This module adds priority levels on partners.
```
luistorresm: Could you improve the module description?

In wich case is necessary this module?

luistorresm (changes requested):

partner_priority/models/partner_priority.py
patrickrwilson: I'm not sure if I understand as the create is inherited for the sequence and i'm not setting any defaults. I'm not sure how to do that any other way actually.

partner_priority/models/partner_priority.py
patrickrwilson: I'm not sure what to change this to, do you have a suggestion you would like to see instead of what's there?

partner_priority/readme/DESCRIPTION.rst
patrickrwilson: I added some more to the description, my company will be using this to set a 'sensitivity' level for our customers. Originally I named this model as 'Partner Sensitivity' however priority used throughout Odoo so for consistency, I renamed it to 'Partner Priority' so others can use it for their own business needs.

patrickrwilson:
@NL66278 Thank you for your review, i've made all your suggested changes and marked your comments as resolved.

@luisiniguezh Thank you for your review, I've made some changes but see my comments on some of your other comments as i'm not quite sure how to implement your suggestions.

partner_priority/README.rst
```diff
@@ -0,0 +1,25 @@
+.. image:: https://img.shields.io/badge/licence-LGPL--3-blue.svg
```
NL66278: LGPL --> AGPL on lines 1 and 3 still todo.

partner_priority/README.rst
```diff
@@ -0,0 +1,25 @@
...
+   :target: https://www.gnu.org/licenses/agpl
+   :alt: License: LGPL-3
+
+===========================
```
NL66278: Missed this one on first review. The `===` lines above and below a title, should be exactly as long as the title.

partner_priority/security/ir.model.access.csv
```diff
@@ -0,0 +1,2 @@
+id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
+access_partner_priority_user,partner.priority.user,model_partner_priority,base.group_user,1,1,1,1,
```
NL66278: This gives any employee the possibility to edit the priorities, you might consider two separate entries, one for group base.group_partner_manager with all rights, and one for base.group_user with only read rights: 1,0.0.0.

partner_priority/data/partner_priority_data.xml
```diff
@@ -0,0 +1,25 @@
+<!-- Copyright 2019 Patrick Wilson <patrickraymondwilson@gmail.com>
+     License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl). -->
+
+<odoo noupdate="1">
```
NL66278, line 4: If you have priorities like 1 = Low,  3 = Normal, 5 = Elevated, 8 = Critical, you make it easier to add intermediate priorities, if somebody wants a more fine grained categorization.

partner_priority/models/partner_priority.py
NL66278: I have to agree with @luistorresm here. Using the sequence to get the next level of priority seems overly compicated. What if somebody wants to put in an intermediate priority? Would seem better to leave out the create method and the sequence. Just put default=1 (or 3), anyway the figure that stands for Normal as default.

partner_priority/models/partner_priority.py
```diff
@@ -0,0 +1,21 @@
...
+from odoo import models, fields, api
+
+
+class PartnerPriority(models.Model):
+    _name = 'partner.priority'
+    _order = 'sequence'
+    _description = 'Partner Priority'
+
+    name = fields.Char(string="Priority", required=True)
+    description = fields.Text()
```
NL66278: Make description and sequence required.

partner_priority/models/partner_priority.py
NL66278: It might also be a good idea to make the sequence unique. Otherwise you might have an ambiguous priority.

partner_priority/README.rst
patrickrwilson: Sorry I missed this, this has been fixed.

partner_priority/README.rst
patrickrwilson: Done

partner_priority/security/ir.model.access.csv
patrickrwilson: Good idea, i've added that manager permission.

partner_priority/models/partner_priority.py
patrickrwilson: Done.

partner_priority/data/partner_priority_data.xml
patrickrwilson: I changed the labels and sequence numbers on the imported data.

partner_priority/models/partner_priority.py
luistorresm: Maybe the model reason

partner_priority/models/partner_priority.py
patrickrwilson: I might be mis-understanding these comments, but the sequence lets you re-arrange the records into a proper order. When you create a record, it should automatically set the sequence to the next available sequence number, making it so the user doesn't have to figure out what sequence to put in (especially if it's a unique field). Adding a default value makes the whole purpose of the sequence a manual process for the user.

luistorresm: Or maybe like [here](https://github.com/odoo/odoo/blob/11.0/addons/stock/models/stock_quant.py#L336)

patrickrwilson: Excellent! I see what you mean now. I didn't know this was a possibility and I like it. I'll be building all my sequences like this going forward. Thank you. This has been changed.

partner_priority/models/partner_priority.py
patrickrwilson: I'm sorry, I just don't follow what you are looking for. I feel this is a fairly basic module and the description is stating exactly what it is, priority to the partners, but maybe i'm mis-understanding this too.

partner_priority/__manifest__.py
```diff
@@ -0,0 +1,25 @@
...
+{
+    'name': 'Partner Priority',
+    'summary': "Adds priority to partners.",
+    'author': "Patrick Wilson, Odoo Community Association (OCA)",
+    'website': "https://github.com/OCA/partner-contact",
+    'category': 'Customer Relationship Management',
+    'version': '12.0.1.0.0',
+    'license': 'AGPL-3',
+    'depends': [
+        'base',
```
luistorresm: You do not need this depends, base is in [contacts](https://github.com/odoo/odoo/blob/12.0/addons/contacts/__manifest__.py)

partner_priority/models/partner_priority.py
luistorresm: 👍

luistorresm (commented):
Sorry @patrickrwilson 

Only a small comment.

Regards

luistorresm (approved):
Technical review

partner_priority/__manifest__.py
patrickrwilson: Done, thanks

patrickrwilson:
Thanks everyone for your reviews. I think I hit all of them and squashed the commits. Hopefully it's ready for approval. Much appreciated.

NL66278:
Thanks for your contribution!

PR #736, opened 2019-05-07
[FIX] partner_email_unique: duplicating partner sets empty email

robinkeunen:
**before**: the check on unique email addresses makes it impossible to duplicate a partner (raises Validation Error).

**after**: the duplicated partner email address is set to False


pedrobaeza:
Please do it instead on the field:
```python
email = fields.Char(copy=False)
```

robinkeunen:
👍 done

pedrobaeza (approved):

NL66278 (approved):
:+1:

PR #739, opened 2019-05-14
[MIG] base_location_nuts: Migration to 12.0

Tardo:
Normal migration:
Now Odoo uses 'parent_path'.
https://www.odoo.com/documentation/12.0/reference/orm.html#odoo.models.Model.parent_path


base_location_nuts/models/res_partner_nuts.py
```diff
@@ -24,6 +24,7 @@ class ResPartnerNuts(models.Model):
     # Parent hierarchy
     parent_id = fields.Many2one(comodel_name='res.partner.nuts',
                                 ondelete='restrict')
+    parent_path = fields.Char(index=True)
```
pedrobaeza, line 27: If you switch this one to the new system, you have to:

* Remove `parent_left` field.
* Remove `parent_right` field.
* Add migration script for recomputing the field.

See in https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-12.0#tasks-to-do-in-the-migration full explanation, and ask me if you want about migration script.

base_location_nuts/wizard/nuts_import.py
```diff
@@ -166,7 +166,7 @@ def create_or_update_nuts(self, node):
         if not self._check_node(node):
             return False
 
-        nuts_model = self.env['res.partner.nuts']
+        nuts_model = self.env['res.partner.nuts'].sudo()
```
pedrobaeza: Why do you need `sudo`? The theory is that if you have access to the import menu, you should be able to import NUTS with your ACLs.

pedrobaeza (changes requested):
Not sure about tests failing in other module if it's caused by this one, but mostly.

Tardo:
Changes done, access rights changed.

base_location_nuts/security/ir.model.access.csv
```diff
@@ -1,2 +1,3 @@
 "id","name","model_id:id","group_id:id","perm_read","perm_write","perm_create","perm_unlink"
-"access_res_partner_nuts_user","res_partner_nuts group_user","model_res_partner_nuts","base.group_user",1,0,0,0
+"access_res_partner_nuts_user_all","res_partner_nuts group_user_all","model_res_partner_nuts","",1,0,0,0
+"access_res_partner_nuts_user","res_partner_nuts group_user","model_res_partner_nuts","base.group_user",1,1,1,1
```
pedrobaeza: This group is not correct. The one that has rights for creating states and countries is "Extra Rights / Contact Creation", so that one should be used here.

base_location_nuts/views/res_partner_nuts_view.xml
```diff
@@ -68,6 +69,7 @@
 
 <menuitem action="res_partner_nuts_action"
           id="res_partner_nuts_menu"
+          groups="base.group_user"
```
pedrobaeza: Use the same group as mentioned before.

pedrobaeza (changes requested):

base_location_nuts/security/ir.model.access.csv
pedrobaeza: `base.group_partner_manager`

Tardo: Done, sorry for the mistake... (i only saw permissions and not the group id)

base_location_nuts/views/res_partner_nuts_view.xml
Tardo: Done

pedrobaeza (approved):

chienandalu (approved):

carlosdauden (approved):

Issue #712, opened 2019-03-11
[BUG] partner_multi_relation: Migration from 8.0 to >= 10.0 


rousseldenis:
Before v10.0, Odoo stored default False boolean values as NULL.

Then, the convert_to_column function was introduced.
See https://github.com/odoo/odoo/commit/73bdc181ffb1c6c052b01450325de8a3f30885e2#diff-c1b8488f841717177876eb16877596afR1105

As this line https://github.com/OCA/partner-contact/blob/10.0/partner_multi_relation/models/res_partner_relation_type_selection.py#L115

returns False (NOT is_symmetric) if column value is NULL, the corresponding inverse is not well computed in view.

A migration query should be set as : 

` UPDATE res_partner_relation_type SET is_symmetric = False WHERE is_symmetric IS NULL;
`

@NL66278 @hbrunn 

Note: New False values are well stored in DB

PR #740, opened 2019-05-16
[10.0] [FIX] Issue #712 Need to set is_symmetric False where it is NULL.

NL66278:
Should solve issue: https://github.com/OCA/partner-contact/issues/712


max3903 (approved):

tbaden (approved):
LGTM

PR #742, opened 2019-05-23
12.0 mig partner_contact_job_position

ernestotejeda:
Normal migration

Cc @Tecnativa


partner_contact_job_position/models/res_partner.py
```diff
@@ -0,0 +1,33 @@
...
+
+from odoo import fields, models
+
+
+class ResPartner(models.Model):
+    _inherit = 'res.partner'
+
+    job_position_id = fields.Many2one(
+        "res.partner.job_position",
+        "Categorized job position",)
```
NL66278: If you put the closing ')' on the last line of the field definition (a style I definitely like, it goes well with the python philosophy on indenting), you are supposed not to have the trailing comma, and certainly not without following it by a space. So best to just leave the comma out.

partner_contact_job_position/security/ir.model.access.csv
```diff
@@ -0,0 +1,5 @@
+"id","name","model_id:id","group_id:id","perm_read","perm_write","perm_create","perm_unlink"
+"access_res_partner_job_position_public","res_partner_job_position group_public","model_res_partner_job_position","base.group_public",1,0,0,0
+"access_res_partner_job_position_portal","res_partner_job_position group_portal","model_res_partner_job_position","base.group_portal",1,0,0,0
+"access_res_partner_job_position_group_partner_manager","res_partner_job_position group_partner_manager","model_res_partner_job_position","base.group_partner_manager",1,1,1,1
```
NL66278: Leave out the 'group_' in the id and name columns in the last two lines, just like it was done on the first two lines, and as is according to conventions.

If you (optionally) want to follow conventions completely, change the name column contents to:
`access.<model name>.<group name>`, so elements separated with dots and group name without group.

partner_contact_job_position/views/res_partner_view.xml
```diff
@@ -0,0 +1,41 @@
...
+    <field name="model">res.partner</field>
+    <field name="inherit_id" ref="base.view_res_partner_filter"/>
+    <field name="arch" type="xml">
+        <field name="category_id" position="after">
+            <field name="job_position_id"/>
+        </field>
+        <filter name="salesperson" position="after">
+            <filter string="Job position"
+                    name="job_position_filter"
+                    domain="[('is_company','=', True)]"
```
NL66278: Question: why the domain on is_company? Maybe I am confused on the workings, but if we want to look for company contacts with a certain job position, will we not miss those contacts that are not companies? Also because above in the form you do not allow companies to have a job position.

partner_contact_job_position/README.rst
```diff
@@ -0,0 +1,98 @@
...
+.. |badge4| image:: https://img.shields.io/badge/weblate-Translate%20me-F47D42.png
+    :target: https://translation.odoo-community.org/projects/partner-contact-12-0/partner-contact-12-0-partner_contact_job_position
+    :alt: Translate me on Weblate
+.. |badge5| image:: https://img.shields.io/badge/runbot-Try%20me-875A7B.png
+    :target: https://runbot.odoo-community.org/runbot/134/12.0
+    :alt: Try me on Runbot
+
+|badge1| |badge2| |badge3| |badge4| |badge5| 
+
+This module extends the functionality of partners job position to support
```
NL66278, line 28: I do not quite get what the hierarchy for. Is it for instance to have a parent 'Technical people', with children 'programmer'  and 'database administrator'? An example in the Usage section would be helpfull I think.

NL66278 (dismissed):
Module looks very usefull. But I have some nags, and a question.  See other comments.

partner_contact_job_position/views/res_partner_view.xml
ernestotejeda: You are right, that is wrong, I' going to  fix it.

partner_contact_job_position/README.rst
pedrobaeza: The idea is to represent employee hierarchy, being parents as bosses, but it's hardly used, so maybe it can be removed.

partner_contact_job_position/views/res_partner_view.xml
```diff
@@ -31,7 +31,7 @@
         </field>
         <filter name="salesperson" position="after">
             <filter string="Job position"
-                    domain="[('is_company','=', True)]"
+                    name="job_position_filter" domain="[('is_company','!=', True)]"
```
chienandalu, line 34: Previous versions should be fixed as well then :sweat_smile:

chienandalu (changes requested):
Tested in runbot :+1: 

Anyway, I also think we can dismiss the hierarchy feature as it's not really used

pedrobaeza:
Requested changes?

chienandalu:
> Requested changes?

Yes, either to document it as @NL66278 asked for or simply get rid of it as you suggested.

ernestotejeda:
Changes done

chienandalu (approved):

pedrobaeza (approved):

PR #744, opened 2019-05-27
 [MIG] portal_partner_select_all: Migration to 12.0

chienandalu:
When a user wants invite a lot of custumers to the portal it can be a pain to
select them all one by one.

This module adds a helper in the wizard to toggle between all selected and only
the ones that already exist.

cc @Tecnativa


Tardo (approved):
LGTM, code review. Can rebase some commits...

chienandalu:
> Can rebase some commits...

There aren't successive commits of the same logic/author :thinking:

Tardo:
> There aren't successive commits of the same logic/author thinking

Is necessary know that oca-bot made these changes? I'm just asking, I'm still clarifying when it's useful to rebase.

pedrobaeza:
Although not mandatory, indeed the bot changes can be integrated in the previous migration commit, as they are "anonymous". We can't do that with other like nominal translations, that contains the author, or other translation stuff, as it has not been done by the migration author. But other stuff like updating the POT, the README or the icon, can be impersonated by the migration author for simplicity.

Any way, this is not blocking. I'm going to merge this for not slowing down this migration, and just take this into account for future PRs.

pedrobaeza (approved):

PR #752, opened 2019-06-28
[12.0][ADD] partner_secondary_address: new module

robinkeunen:
This module adds fields for a secondary address on partners.

### DIFF
--- a/partner_secondary_address/__init__.py
+++ b/partner_secondary_address/__init__.py
@@ -1,0 +1,1 @@
+from . import models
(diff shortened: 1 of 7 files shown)


pedrobaeza:
There's no README in the module + I'm not sure about the approach. You are denormalizing data structure, where each `res.partner` record represents an address, so you link to other partner if you need a secondary address instead of adding field.

For me, this is :-1:

robinkeunen:
I thought the readme would be generated by from the `readme` directory.

The use case: within a housing cooperative (similar to real estate), partners may have primary and secondary residencies.

pedrobaeza:
But that's reflected with a parent partner + a child contact with the second residency, indicating in the contact name exactly that. And you can have third residence, forth...

robinkeunen:
Ok, thank you for the suggestion, I'll have a look.

Issue #754, opened 2019-07-15
[ADD] partner_brand


max3903:
This module allows users to manage brands of the company.

# Usage
* Go to Contacts
* Search the current company
* Click on Edit and add an address
* Check the box "Brand" and enter the brand information: name, logo, address, website, phone

# Technical Specifications

- [x] Add an option "Brand" (brand) in the Type selection list
- [x] Add a filter "Brand" in Contacts
- [ ] Upon creating or saving a brand, check that the parent company and the company of the brand are the same as the company of the connected user.

sbidoul:
@max3903 @pedrobaeza I'm also interested in understanding the relationship of this module to `product_brand`. 

We have a use case for brands but not related to products: ie the same product can be sold under different brands. So the brand will be something to be selected on the SO/Invoice etc.

So I would say a base brand module, unrelated to product, makes sense. And product_brand can depend on it.

WDYT?

pedrobaeza:
OK, but for v12, the dependency should be `product_brand`, but for v13 we can split into `base_brand` (or `brand` only) + `product_brand`.

sbidoul:
cc @sbejaoui, @gva-acsone

sbidoul:
@pedrobaeza @max3903 ok for me with keeping product_brand at least in v12. It does not harm even if we don't set brand on any product.

FYI we are planning the following modules (current list is WIP):

- `brand_sale`: add brand field on SO + filter products proposed by brand
- `brand_account`: 

  - table with optional default receivable account per brand
  - brand field on invoice, proposes default receivable account

- `brand_contract`: brand field on contract, copied when generating invoices
- `brand_contract_sale`: glue module between `brand_sale` and `product_contract`

I'm wondering if we'd need a new repo for that.

We may use [report_subsitute](https://github.com/OCA/reporting-engine/pull/307) to help with generating report (although a different header/footer by brand may be enough).

max3903:
@sbidoul That is a good idea. FYI, you will find:

* `account_brand` in https://github.com/OCA/account-invoicing/issues/564
* `sale_brand` in https://github.com/OCA/sale-workflow/pull/906
* `sale_subscription_brand` in https://github.com/ursais/osi-addons/pull/158

@pedrobaeza `product_brand` was demonstrated to our customer and it does not answer their requirements. They manage different brands that sells the same products and services under the same company. They need the header and footer of the PDF reports to have the brand information: logo, name, address, email, website, phone of the brand, instead of the company one.

[...]

PR #755, opened 2019-07-20
[ADD] partner_brand (#754)

max3903:
#754


max3903:
Ping @RLeeOSI

partner_brand/views/report_template.xml
```diff
@@ -0,0 +1,80 @@
+<odoo>
+
+    <!-- Copyright (C) 2019 Open Source Integrators
+         License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl). -->
+
+    <template id="external_layout_standard_logo"
```
sbejaoui, line 6: I thought having a layout per brand would be easier to handle in use cases where the header and footer of the report would be completely different in look and content.

override [`web.external_layout`](https://github.com/odoo/odoo/blob/12.0/addons/web/views/report_templates.xml#L342]) to have something like this
```xml
<t t-if="'brand_id' in o.fields_get() and o.brand_id and o.brand_id.external_layout_id">
	<t t-call="{{o.brand_id.external_layout_id.key}}"/>
</t>
<t t-else="">
    <t 	t-if="company.external_report_layout_id" t-call="{{company.external_report_layout_id.key}}">
    	<t t-raw="0"/>		
    </t>
    <t t-else="else" t-call="web.external_layout_standard">
    	<t t-raw="0"/>
    </t>
</t>
```

max3903: @sbejaoui It would be nice to do that but Odoo has the header and footer and their elements in the same template. In your example, we would have to copy the whole template and adapt it in `o.brand_id.external_layout_id.key`. We chose a more surgical approach where we only inherits and modifies the elements we need. I guess it will make the maintenance and migration easier, we'll see.

sbejaoui: I think we can merge the two ideas:

- Override the `external_layout_standard` to handle brand details in the standard layout.
- Override the external_layout and call `o.brand_id.external_layout_id` if it is set in case we need to create a whole different design.

What do you think?

osi-scampbell (approved):

partner_brand/views/report_template.xml
max3903: #758

pedrobaeza:
I think this module should be called `partner_branding`

mgosai (approved):

pedrobaeza:
What about the renaming?

pedrobaeza:
And again other of my comments ignored... Why?

max3903:
@pedrobaeza 

1. `product_brand` is not named `product_branding`.
2. My customer don't have the time and budget for me to take care of all the OCA comments.
3. The issue #754 has been created 2 weeks ago. You didn't mention that renaming within those 2 weeks.

pedrobaeza:
The renaming doesn't take time. You are doing branding, not putting a brand. I didn't think in the confusion with `product_brand` module. People are going to think that this is the same, but for partners.

And at least you can answer before merging.

PR #757, opened 2019-07-29
[FIX] base_location_geonames_import: Handle duplicated city names

### LINKED ISSUE #749: Cities did not pull to correct states
Hi there,

I did my first import from geonames just yesterday and I'm noticing today that many of the cities that have the same name in multiple states did not get imported correctly.  For example, Worcester, the city, resides in MA, NY, PA, and VT.  But upon import, it's placed *ALL* cities named Worcester into VT.  (see screenshots for a few examples)...

cities_error_1
cities_error_2

pedrobaeza:
If there are several cities with the same name in different states, previous code doesn't
handle correctly this situation.

We amend this storing in the cached dictionary both city name and state.

Includes a test for checking this specific condition, got from real data in US.

Fixes #749

cc @Tecnativa


NL66278 (approved):
:+1: LGTM

alexey-pelykh (approved):
Code LGTM

PR #761, opened 2019-08-08
[IMP] partner_sale_risk: Sales orders in a state done for risk

rafamarpe:
We check the configuration to identify if you have to take into account the orders blocked in the financial risk

### DIFF
--- a/partner_sale_risk/models/res_partner.py
+++ b/partner_sale_risk/models/res_partner.py
@@ -24,10 +24,17 @@ class ResPartner(models.Model):
     def _compute_risk_sale_order(self):
         customers = self.filtered('customer')
         partners = customers | customers.mapped('child_ids')
-        orders_group = self.env['sale.order'].read_group(
-            [('state', '=', 'sale'), ('partner_id', 'in', partners.ids)],
-            ['partner_id', 'invoice_pending_amount'],
-            ['partner_id'])
+        if self.env['ir.values'].get_default('sale.config.settings', 'auto_done_setting'):
+            orders_group = self.env['sale.order'].read_group(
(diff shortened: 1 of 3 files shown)


partner_sale_risk/models/res_partner.py
```diff
@@ -24,10 +24,17 @@ class ResPartner(models.Model):
     def _compute_risk_sale_order(self):
         customers = self.filtered('customer')
         partners = customers | customers.mapped('child_ids')
-        orders_group = self.env['sale.order'].read_group(
-            [('state', '=', 'sale'), ('partner_id', 'in', partners.ids)],
-            ['partner_id', 'invoice_pending_amount'],
-            ['partner_id'])
+        if self.env['ir.values'].get_default('sale.config.settings', 'auto_done_setting'):
```
pedrobaeza, line 27: Please don't do spaghetti code and respect PEP8:
```python
states = ['sale']
# Isn't this the contrary?
if self.env['ir.values'].get_default(
        'sale.config.settings', 'auto_done_setting'):
    states.append('done')
orders_group = self.env['sale.order'].read_group([
    ('state', 'in', states),
    ...
```

partner_sale_risk/models/sale.py
```diff
@@ -17,7 +17,11 @@ class SaleOrder(models.Model):
     @api.depends('state', 'order_line.invoice_lines.invoice_id.amount_total')
     def _compute_invoice_amount(self):
         AccountInvoice = self.env['account.invoice']
-        for order in self.filtered(lambda x: x.state == 'sale'):
+        if self.env['ir.values'].get_default('sale.config.settings', 'auto_done_setting'):
```
pedrobaeza, line 20: The same comment as above

pedrobaeza (changes requested):

PR #763, opened 2019-08-14
[12.0] [ADD] module partner_postal_address

remi-filament:
Add a new type of address named "Postal Address", this way the main address on partner can remain administrative address


pedrobaeza:
Do you plan to use this new type in something later?

remi-filament:
Hi @pedrobaeza, we have had a few customers that needed to have clearly identified the postal address of their contacts (which can be different from administrative one when a company has multiple offices for instance).
Instead of having this added in specific module for each customer, I thought it would be better to have a generic module adding simply that.
Let me know in case you think this is not relevant...

This address will be used for sending mail through post office (out of Odoo) so I do not plan to make use of this field in another function, if that was your question.

NL66278:
@remi-filament @pedrobaeza I can confirm that this is a very common requirement.

pedrobaeza:
Well, what I mean is that this is disconnected from any other feature (like for example automatically put this address on the invoice report, or auto-fill sending address in invoices...). As I see it now, it's a new unconnected type that it can lead to side effects not being selectable in some m2o/m2m fields if they apply a domain over this type. When one of my customers tell me about this, I'm saying that they should put "Other address", and they can put in the "Contact Name" field an explanatory text like "Postal address".

remi-filament:
I understand what you are saying, although I have tried to search for places where the other type was used, and I could only find one hit on website_sale which seemed relevant (although I might have missed some).
Therefore from my point of view, it is easier for my customers to directly select a Postal address in the radio button rather that using Other and adding Postal in name (because it is not obvious, so many will never use it because they cannot find where to put it).

pedrobaeza:
OK, I'm not blocking, just giving my opinion :smiley:

bealdav (approved):

NL66278 (approved):
:+1: LGTM

Issue #765, opened 2019-08-16
[RFC] brand, the path forward


sbidoul:
Hi,

We currently have two independent approaches for brands in OCA:
- [partner_brand](https://github.com/OCA/partner-contact/tree/12.0/partner_brand), [account_brand](https://github.com/OCA/account-invoicing/tree/12.0/account_brand), [sale_brand](https://github.com/OCA/sale-workflow/tree/12.0/sale_brand) which basically mark some contacts as being brands and replace company information in report headers/footers with brand information is available
- [product_brand](https://github.com/OCA/product-attribute/tree/12.0/product_brand) which assign products to brand and adds that in some pivot tables

We have more requirements for brand (such as defining receivable/payable accounts by brand, or adding multi-company support on brands) for which neither approach is totally adequate. For that we need a separate `brand` object, since extending `res.partner` will be confusing and will not work in the long run.

So the first step we propose is as follow:
- update `partner_brand` to create a `brand` object with delegation inheritance on `res.partner`, with migration scripts
- update `sale_brand` and `account_brand`, with migration scripts
- add a dedicated menu for brands next to Companies in settings, with dedicated views

This is not perfect, but it's a first step to evolve what is merged today. Further down the road we can consider making `product_brand` depend on this new `brand` model, and possibly replace the delegation inheritance by something else.

@max3903 @pedrobaeza any opinion on this approach?

cc/ @sbejaoui

sbejaoui:
New model res.brand: #766

sbejaoui:
Update for sale_brand [\#920](https://github.com/OCA/sale-workflow/pull/920)

sbejaoui:
Update for account_brand [\#577](https://github.com/OCA/account-invoicing/pull/577)

pedrobaeza:
Should we create a new repo OCA/brand? And about the model name, why `res.brand`?

sbejaoui:
I thought that brands can be seen as system resources and in the same
registry like users, companies, partners, ..

Le ven. 16 août 2019 à 22:28, Pedro M. Baeza <notifications@github.com> a
écrit :

max3903:
Looks like we are doing the same as Odoo: We discuss a topic publicly but internally the decision seems to be already made and the job already done... :)

Anyway, I don't have any strong opinion about where and how the repo/module/object is named/developed. What matters is to maintain the same scope for our current customer. 

[...]

PR #766, opened 2019-08-16
[IMP] - Add new model res.brand

sbejaoui:
#765


rafaelbn:
@sbejaoui this is old I think, please take a look thanks

PR #767, opened 2019-08-17
[IMP] partner_multi_relation - assign relation to many partners.

fanzalika:
Add the functionality to assign many partners to a relation from the
contacts window list.


NL66278:
@fanzalika Could you describe better what the purpose would be? If you go to the relations list coming fro a partner using the Relations button, you can already quickly add contacts to a relation. What would this add?

NL66278:
@fanzalika Note also that you will need to do some cleanup of the code (no more data element needed in view, blank lines at end of code etc..). And solve travis error.

NL66278:
No reaction to question and comments.

PR #770, opened 2019-08-29
[10.0] [FIX] partner_email _check debug instead of error when logging

ntsirintanis:
Log a `debug` message instead of an `error` message when failing to import  `import validate_email`


hbrunn (approved):

partner_email_check/models/res_partner.py
```diff
@@ -11,7 +11,7 @@
 try:
     from validate_email import validate_email
 except ImportError:
-    _logger.error('Cannot import "validate_email".')
+    _logger.debug('Cannot import "validate_email".')
```
pedrobaeza, line 14: Please put info as other import errors.

hbrunn: I asked my colleague to follow the conventions: https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#importerror

pedrobaeza: OK, I miss the discussion that leads to this convention, as I would prefer info for seeing it always when starting my Odoo service. Any way, let's keep it as convention says.

hbrunn: thanks. I can't point to a discussion currently, but if I remember correctly the argument was that odoo's dependencies will disallow installing a module without the good libraries, and when debugging, you should turn on debug logging

pedrobaeza: Yeah, makes sense.

NL66278 (approved):
:+1: Approve although personally I also like info more... but let us follow conventions.

hbrunn:
then let's add a pr to the conventions and have a discussion there

PR #774, opened 2019-09-11
Add normalization and optional duplicate filter to partner_email_check

blyxxyz:
- `partner_email_check` now uses the `email-validator` package instead of `validate_email`
- Email addresses are normalized
- There is a setting to not allow multiple partners to use the same email address

### DIFF
--- a/partner_email_check/README.rst
+++ b/partner_email_check/README.rst
@@ -18,7 +18,7 @@ although validation and normalization are still supported in such cases.
 Configuration
 =============
 
-Install python package validate_email: ``sudo pip install email-validator``.
+Install python package email-validator: ``sudo pip install email-validator``.
 
 To not allow multiple partners to have the same email address, use the
 "Filter duplicate email addresses"/``partner_email_check_filter_duplicates``
--- a/partner_email_check/README.rst
+++ b/partner_email_check/README.rst
@@ -6,12 +6,23 @@
 Partner Email Check
 ===================
 
-This module validate the field ``email`` in the module ``res.partner``.
+This module validates and normalizes the field ``email`` in the module
+``res.partner``.
+
+As part of the normalization, email addresses are converted to lowercase.
+
+Optionally, multiple partners can not be allowed to have the same address.
+This will not work with multiple comma-separated email addresses in the field,
+although validation and normalization are still supported in such cases.
 
 Configuration
 =============
 
-Install python package validate_email: ``sudo pip install validate_email``.
+Install python package validate_email: ``sudo pip install email-validator``.
+
+To not allow multiple partners to have the same email address, use the
+"Filter duplicate email addresses"/``partner_email_check_filter_duplicates``
+setting.
 
 Usage
 =====
--- a/partner_email_check/models/__init__.py
+++ b/partner_email_check/models/__init__.py
@@ -2,4 +2,4 @@
 # Copyright 2017 Komit <http://komit-consulting.com>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
-from . import res_partner
+from . import base_config_settings, res_partner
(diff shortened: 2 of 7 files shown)


partner_email_check/README.rst
```diff
@@ -6,12 +6,23 @@
...
+
+Optionally, multiple partners can not be allowed to have the same address.
+This will not work with multiple comma-separated email addresses in the field,
+although validation and normalization are still supported in such cases.
 
 Configuration
 =============
 
-Install python package validate_email: ``sudo pip install validate_email``.
+Install python package validate_email: ``sudo pip install email-validator``.
```
hbrunn: rename the package too

partner_email_check/models/__init__.py
```diff
@@ -2,4 +2,4 @@
 # Copyright 2017 Komit <http://komit-consulting.com>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
-from . import res_partner
+from . import base_config_settings, res_partner
```
hbrunn: it's customary to have one import per line

partner_email_check/models/base_config_settings.py
```diff
@@ -0,0 +1,47 @@
...
+            'partner_email_check_filter_duplicates':
+                self.env['ir.config_parameter'].get_param(
+                    'partner_email_check_filter_duplicates', 'False'
+                ) == 'True'
+        }
+
+    partner_email_check_check_deliverability = fields.Boolean(
+        string="Check deliverability of email addresses",
+        help="Don't allow email addresses with providers that don't exist",
+    )
```
hbrunn: follow [OCA conventions](https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#models) about the order of members

partner_email_check/models/res_partner.py
```diff
@@ -9,30 +9,84 @@
...
+            return email
+
+        try:
+            result = validate_email(
+                email,
+                check_deliverability=self._should_check_deliverability(),
+            )
+        except EmailNotValidError as e:
+            raise ValidationError(
+                _("%s is an invalid email: %s") % (email.strip(), e.message)
```
hbrunn: how does `email_validator` handle translations? I doubt we can make this pick up Odoo's translation mechanism easily, so instead of having a mixed localized/non-localized string, I think we'll be better off not using its messages at all

partner_email_check/models/res_partner.py
```diff
@@ -9,30 +9,84 @@
...
+    def _should_check_deliverability(self):
+        conf = self.env['ir.config_parameter'].get_param(
+            'partner_email_check_check_deliverability', 'False'
+        )
+        return conf == 'True'
+
+    def _validate_no_duplicates(self, email):
+        is_existing_record = bool(self)
+        if is_existing_record:
+            self.ensure_one()
```
hbrunn: this will break when some code does a write on multiple records at once, something like `$model.browse([1,2,3]).write({'email': 'a@b.c'})`.

What's your reasoning not to implement the whole thing as an ORM constraint? Then you also don't need to care about corner cases, because you know this record exists in the current transaction

partner_email_check/models/res_partner.py
```diff
@@ -9,30 +9,84 @@
...
+
+        domain = [('email', '=', email)]
+        if is_existing_record:
+            domain.append(('id', '!=', self.id))
+        if self.search(domain, limit=1):
+            raise ValidationError(_("Email '%s' is already in use.") % email)
+
+    @api.model
+    def create(self, vals):
+        if vals.get('email', False):
```
hbrunn: no need to pass a default here, as otherwise you get `None` and that's falsy too, which is all you need

hbrunn (changes requested):
very nice! Some things about Odoo specifics, and given you made a company github handle, I suggest you close this PR and create a new one

partner_email_check/models/res_partner.py
blyxxyz: I removed the use of the exception message and added a separate case for non-deliverable email (which has a separate exception).

PR #776, opened 2019-09-12
Add normalization and optional uniqueness constraint to partner_email_check

janverb:
* `partner_email_check` now uses the `email-validator` package instead of `validate_email`
* Email addresses are normalized
* There is a setting to enforce uniqueness of partner email addresses
* There is a setting to check whether emails are deliverable (i.e. whether the domain resolves)

(originally #774)

### DIFF
--- a/partner_email_check/README.rst
+++ b/partner_email_check/README.rst
@@ -6,12 +6,27 @@
 Partner Email Check
 ===================
 
-This module validate the field ``email`` in the module ``res.partner``.
+This module validates and normalizes the field ``email`` in the model
+``res.partner``.
+
+As part of the normalization, email addresses are converted to lowercase.
+
+Optionally, multiple partners can not be allowed to have the same address.
+This will not work with multiple comma-separated email addresses in the field,
+although validation and normalization are still supported in such cases.
 
 Configuration
 =============
 
-Install python package validate_email: ``sudo pip install validate_email``.
+Install python package email-validator: ``sudo pip install email-validator``.
+
+To not allow multiple partners to have the same email address, use the
+"Filter duplicate email addresses"/``partner_email_check_filter_duplicates``
+setting.
+
+To validate that email addresses are deliverable (that the hostname exists),
+use the "Check deliverability of email addresses"/``partner_email_check_check_deliverability``
+setting.
 
 Usage
 =====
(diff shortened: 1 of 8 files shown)


partner_email_check/models/res_partner.py
```diff
@@ -9,30 +9,84 @@
...
+    def email_check(self, emails):
+        return ','.join(self._normalize_email(email.strip())
+                        for email in emails.split(','))
+
     @api.constrains('email')
-    def constrains_email(self):
-        for rec in self.filtered("email"):
-            self.email_check(rec.email)
+    def _check_email_unique(self):
+        if self._should_filter_duplicates():
```
hbrunn, line 33: no need to change this, tip for the future: return early in such cases where you only do something if a condition is met. This makes the code flow clearer and you save a level of indentation.

partner_email_check/tests/test_partner_email_check.py
```diff
@@ -32,3 +38,67 @@ def test_good_emails(self):
...
+    def test_multi_email_local_normalization(self):
+        """Test normalization of the local part of multiple addresses."""
+        self.test_partner.write({'email': 'You@mAiL.net,mE@mail.com'})
+        self.assertEqual(
+            self.test_partner.email,
+            'You'.lower() + '@mail.net,' + 'mE'.lower() + '@mail.com'
+        )
+
+    def disallow_duplicates(self):
+        self.env['ir.config_parameter'].set_param(
```
hbrunn: use the setup wizard to do this, then you also cover this code patch in the tests

hbrunn (changes requested):
please squash your commits and add a test to test for non-deliverability, then coverage should be fine too

hbrunn (approved):

partner_email_check/README.rst
```diff
@@ -6,12 +6,23 @@
 Partner Email Check
 ===================
 
-This module validate the field ``email`` in the module ``res.partner``.
+This module validates and normalizes the field ``email`` in the module
```
NL66278: in the model res.partner, not the module

NL66278 (approved):
:+1:  Very nice improvement

One comment, that was wrong already in the original code, can still be corrected, but that is not blocking for this PR.

pedrobaeza:
@NL66278 please use ocabot for merging PRs with possible manifest version change:

https://odoo-community.org/blog/the-oca-blog-1/post/the-oca-github-bot-can-now-merge-pull-requests-92

NL66278:
@pedrobaeza Thanks for the pointer. I was not aware I could use this process.

PR #777, opened 2019-09-13
[11.0][FIX] partner_vat_unique: Do not copy vat field

sergio-teruel:
cc @Tecnativa


pedrobaeza (approved):

NL66278 (approved):
:+1:  I think the vat field should not be copied anyway, but that is for Odoo SA to fix.

PR #784, opened 2019-09-30
[MIG] partner_firstname: Migration to 13.0

rven:


schout-it (approved):
lgtm

yucer:
Hello, 

You might some time trying to see why one of the unit tests for `test_onchange.py` was not running sucessfully.

I was using `python 3.7` locally to reproduce it, and I didn't get the error. 

After rebuilding we receive: 

`AssertionError: Outdated python version detected, Odoo requires Python >= 3.6 to run.`

I guess something in the config should change.

rven:
Hi @yucer locally the problem can be reproduced when first initiating a DB with options --init bus,base_setup,web_tour,web,base,base_address_city,mail,contacts

Then running the tests --test-enable --test-tags partner_firstname --init base_location,base_location_geonames_import,partner_fax,partner_firstname

A problem is raised on the message_follower_ids field, so I changed the test to use the view 'base.view_partner_simple_form' which does not include that field.

Now the test should be ok!

partner_firstname/tests/test_onchange.py
```diff
@@ -6,99 +6,95 @@
...
     def test_create_from_form(self):
-        """A user creates a company from the form."""
         name = "Sôme company"
-        with self.env.do_in_onchange():
-            # User presses ``new``
-            partner = self.new_partner()
-
-            # User sets a name, which triggers onchanges
-            partner.name = name
-            partner._onchange_name()
```
lmignon: @rven I don't see why you should use a Form to call the onchange. If you get an error when calling the onchange directly, it's maybe a bug into the Odoo or an other addon... We should at least understand the root cause of the issue into the original code. I already encountered some errors when calling an onchange method with computed fields when the compute method makes an SQL query with the `id` of the record as parameter without checking that the id is not a NewId....

lmignon (changes requested):

partner_firstname/tests/test_onchange.py
rven: Hi @lmignon,

I don't think you're reviewing my changes.
Your suggestion is also labeled **Outdated**
I don't have any calls to _onchange_name, that's why I changed it to a Form instance.

pedrobaeza:
Superseeded by #805

PR #790, opened 2019-10-07
[13.0] pre commit

sbidoul:


.flake8
```diff
@@ -0,0 +1,10 @@
+[flake8]
+max-line-length = 80
```
pedrobaeza, line 2: Are we going to keep this strict ?

.travis.yml
```diff
@@ -1,30 +1,43 @@
-sudo: false
-cache: pip
+language: python
+cache:
+  directories:
+  - $HOME/.cache/pip
```
pedrobaeza, line 4: Is this change needed for the new stuff?

base_location/__manifest__.py
```diff
@@ -3,32 +3,27 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
 {
-    'name': 'Location management (aka Better ZIP)',
-    'version': '13.0.1.0.0',
-    'depends': [
-        'base_address_city',
-        'contacts',
+    "name": "Location management (aka Better ZIP)",
```
pedrobaeza, line 6: Can we avoid this quote change?

.flake8
sbidoul: More info here: https://github.com/psf/black#line-length

.travis.yml
sbidoul: caching `$HOME/.cache/pre-commit` helps performance of the linter job.

base_location/__manifest__.py
sbidoul: Black has an option to not do string normalization yes. Personally I'm in favor of letting it normalize, since it's one less thing to think about.

.flake8
pedrobaeza: Yeah, but we talked about raising to 88 and maybe use the B check with tolerance, isn't it?

base_location/__manifest__.py
pedrobaeza: Well, I prefer not, as this change doesn't add anything valuable for homogenization (it looks pretty the same), and it can generate big diffs (as this case) and also the need to escape `"` characters inside the string.

sbidoul:
I just sent an e-mail to contributors ML with explanations. This PR is an illustration, as https://github.com/OCA/storage/pull/12 and https://github.com/OCA/queue/pull/175. 

I propose we keep the general discussion on the ML and detailed discussion on the config in https://github.com/OCA/maintainer-quality-tools/pull/613.

pedrobaeza:
OK

base_location/__manifest__.py
yvaucher: @pedrobaeza black will automatically use simple quotes when the string contains `"` when it can reduce the number of character used. With always priority to the use of "

`"\""` -> `'"'`
`'\''` -> `"'"`
`"\"'"` -> `"\"'"`
`'"\''` -> `"\"'"`

For mixed quotes it will choose the less use of `\`
```
'"""\''
"'''\""
```

pedrobaeza: Thanks for this info, Yannick

sbidoul:
Done at https://github.com/OCA/partner-contact/commit/a40176e0c925a0bd5481125795f886229797b29b

PR #795, opened 2019-10-25
[10.0] [FIX] Fix partner_email_check conflict with portal, bump version

janverb:
`portal`'s tests [set an invalid email address, `c@c`](https://github.com/OCA/OCB/blob/10.0/addons/portal/tests/test_portal.py#L41), that's rejected by the current version of `partner_email_check`. If `partner_email_check` is installed before `portal`'s tests run then they fail. To solve this, use [`base_manifest_extension`](https://github.com/OCA/server-tools/tree/10.0/base_manifest_extension) to force `portal`'s tests to run first.

The previous update to `partner_email_check` didn't bump the version number, so do that now too.

### DIFF
--- a/partner_email_check/__manifest__.py
+++ b/partner_email_check/__manifest__.py
@@ -4,7 +4,7 @@
 
 {
     'name': 'Email Format Checker',
-    'version': '10.0.1.0.2',
+    'version': '10.0.1.1.0',
     'summary': 'Validate email address field',
     'author': "Komit, Odoo Community Association (OCA)",
     'website': 'http://komit-consulting.com',
--- a/partner_email_check/models/res_partner.py
+++ b/partner_email_check/models/res_partner.py
@@ -5,6 +5,7 @@
 import logging
 from odoo import api, models, _
 from odoo.exceptions import ValidationError
+from odoo.tools import config
 
 _logger = logging.getLogger(__name__)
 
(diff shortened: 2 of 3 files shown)


pedrobaeza:
You can't introduce new dependencies in such old version. Try with post_install in tests.

hbrunn:
@pedrobaeza where does this come from?

pedrobaeza:
You will break all current installations. Reason should be very strong for forcing this.

janverb:
I'm not sure how `post_install` would help. The problem happens when the tests in `portal` run too late, after `partner_email_check` is installed. The current solution forces the tests in `portal` to run earlier. The tests in `partner_email_check` don't affect this.

But adding a new dependency does seem like a bad idea.

pedrobaeza:
What about disabling the behavior of the module when testing other modules. I do something like that in 

https://github.com/OCA/bank-payment/blob/96c502fbbd0798493c4969e148e5956e0af72036/account_payment_transfer_reconcile_batch/models/payment_order.py#L28-L32

NL66278 (approved):
:+1: LGTM

NL66278:
@pedrobaeza Could you have another look? As far as I can see your concerns have been dealt with.

pedrobaeza (approved):

pedrobaeza:
Great, thanks!

/ocabot merge patch

PR #798, opened 2019-10-30
[12.0] [ADD] multi-company management to partner_company_type

qgroulard:


sbejaoui (approved):

NL66278:
What would the usecase be for making company-type dependent on the company? And maybe you could put this information also in the USING.rst. Because if you define some legal forms possible for an organisation, it would apply to all partners I would think. Please explain.

qgroulard:
> What would the usecase be for making company-type dependent on the company? And maybe you could put this information also in the USING.rst. Because if you define some legal forms possible for an organisation, it would apply to all partners I would think. Please explain.

The use case would be an international company having different companies in Odoo for each of their subsidiaries. While working in a (subsidiary) company you don't want to be flooded with legal forms from other countries.

NL66278:
@qgroulard Would it then not be more logical to add a country_id field to the company type? Because in a lot of Odoo installations, partners from multiple countries are under one company (actually most Odoo installations are not multiple company). On the other hand you can have multiple companies all having customers in let us say Britain, France and Brazil. Then there are the many Odoo installations that do not attach partners to specific companies at all, even if the have multiple companies. In the end a company in Odoo is basically an entity having its own chart of accounts.

qgroulard:
> @qgroulard Would it then not be more logical to add a country_id field to the company type? Because in a lot of Odoo installations, partners from multiple countries are under one company (actually most Odoo installations are not multiple company). On the other hand you can have multiple companies all having customers in let us say Britain, France and Brazil. Then there are the many Odoo installations that do not attach partners to specific companies at all, even if the have multiple companies. In the end a company in Odoo is basically an entity having its own chart of accounts.

You are totally right, I open https://github.com/OCA/partner-contact/pull/836 and close this one.

PR #800, opened 2019-11-05
[13.0][MIG] base_location_nuts

astirpe:


.pylintrc-mandatory
```diff
@@ -39,7 +39,6 @@ enable=anomalous-backslash-in-string,
     method-required-super,
     method-search,
     missing-import-error,
-    missing-manifest-dependency,
```
pedrobaeza: Why removing this?

astirpe: The pylint was failing, even when adding the external-dependencies in manifest. I adopted the same solution as for `date_range`: https://github.com/OCA/server-ux/pull/99#issuecomment-543679437

astirpe: This workaround is not needed anymore since it's working fine after rebase.

base_location_nuts/readme/CONFIGURE.rst
```diff
@@ -0,0 +1,9 @@
+After installation, you must click at import wizard to populate NUTS items
+in Odoo database in:
+Sales > Configuration > Address Book > Import NUTS 2013
```
mamcode:
```suggestion
Contacts > Configuration > Localization > Import NUTS 2013
```

mamcode (changes requested):
Looks good, only a small change in readme/CONFIGURE.rst

base_location_nuts/readme/CONFIGURE.rst
astirpe: Done. Thanks!

rousseldenis (approved):
Code review

mamcode (approved):
LGTM. Thanks.

PR #804, opened 2019-11-06
[8.0][FIX] url download geonames

sergiocorato:


pedrobaeza:
The unavailability is temporal AFAIK. What you have in dump has other format.

pedrobaeza:
I confirm it: http://download.geonames.org/export/zip/ is now fully functional

sergiocorato:
> I confirm it: http://download.geonames.org/export/zip/ is now fully functional

ok, thanks!

PR #806, opened 2019-11-11
Fix partner contact name settings

Nooka10:
FIX bad inheritance of class `res.config.settings` which could break other settings wizards. It should not override the base class but create a new child class instead.

I also removed the unneeded UTF8 headers.

### DIFF
--- a/base_location_geonames_import/__init__.py
+++ b/base_location_geonames_import/__init__.py
@@ -1,4 +1,3 @@
-# -*- coding: utf-8 -*-
 
 from . import models
 from . import wizard
--- a/base_location_geonames_import/__manifest__.py
+++ b/base_location_geonames_import/__manifest__.py
@@ -1,4 +1,3 @@
-# -*- coding: utf-8 -*-
 # Copyright 2014-2016 Akretion (Alexis de Lattre
 #                     <alexis.delattre@akretion.com>)
 # Copyright 2014 Lorenzo Battistini <lorenzo.battistini@agilebg.com>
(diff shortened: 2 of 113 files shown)


astirpe (changes requested):
@Nooka10 with this PR, the settings for this module are not accessible anymore by end user.

I don't even understand what problem are you facing.

ecino:
@astirpe the inheritance of `res.config.settings` should be put into another class as stated here : https://github.com/odoo/odoo/blob/11.0/odoo/addons/base/res/res_config.py#L337

Otherwise other settings wizard that inherit the `res.config.settings` base class will have errors because they don't include the partner fields in their view.

I also don't understand why the enduser won't have access? It's a TransientModel so it doesn't have access rights.

astirpe:
@ecino try to execute the [configuration steps](https://github.com/OCA/partner-contact/blob/11.0/partner_firstname/README.rst) of this module. The settings for this module are now not accessible anymore. The functionality is broken.

Also if I look at the standard Odoo V11 code, I don't see any class with `_inherit = 'res.config.settings'` combined together with `_name = .... `. I guess the documentation you are mentioning is outdated.

astirpe:
> Nooka10 wants to merge 2 commits into OCA:11.0 from Nooka10:fix-partner-contact

Maybe this PR was supposed to be opened onto branch 10.0 instead of 11.0, is that the case?

pedrobaeza:
Documentation is correct, but it's document another thing: the way you can create the same kind of settings wizard, not to expand current one.

ecino:
Ok, it's different way of doing (include all settings at the same place vs having specific settings wizards).

PR #815, opened 2019-11-26
[13.0][MIG] partner_identification

Laurent-Corron:


mlaitinen:
@Laurent-Corron Please check the code coverage

ThomasBinsfeld (approved):

chrisdec (approved):
Functional tests

partner_identification/models/res_partner_id_number.py
```diff
@@ -22,8 +22,7 @@ def validate_id_number(self):
     name = fields.Char(
         string="ID Number",
         required=True,
-        help="The ID itself. For example, Driver License number of this "
-        "person",
+        help="The ID itself. For example, Driver License number of this " "person",
```
NL66278: You might merge the two strings (remove the double doublequotes between this and person)

NL66278 (approved):
:+1: One nag, but not blocking. Thanks for migrating this!

partner_identification/models/res_partner_id_number.py
Laurent-Corron: done! :)

ntsirintanis (approved):

NL66278:
This failes due to a test in the partner_firstname module. For some reason when checking the name 'Foo bis', bis is not the firstname, but also not seen as part of the lastname. Anybody any idea?

rousseldenis:
> This failes due to a test in the partner_firstname module. For some reason when checking the name 'Foo bis', bis is not the firstname, but also not seen as part of the lastname. Anybody any idea?

Maybe @lmignon as you worked on it.

@NL66278 You can just use bot merge command without major. If so, it will change version from 13.0.1.0.0 to 13.0.2.0.0

partner_identification/models/res_partner.py
```diff
@@ -0,0 +1,163 @@
...
+        Args:
+            field_name (str): Name of field to set.
+            category_code (str): Category code of the Identification type.
+        """
+        for record in self:
+            id_numbers = record.id_numbers.filtered(
+                lambda r: r.category_id.code == category_code
+            )
+            if not id_numbers:
+                continue
```
ntsirintanis, line 57: When we use this method as a compute method for a field, then things break if there's no `id_numbers`. We should do `record[field_name] = False` and then `continue`.

ntsirintanis (changes requested):

PR #821, opened 2019-12-05
13.0 mig base vat sanitized

StephanRozendaal:


lfreeke (approved):
Functional test :+1:

astirpe (approved):
Code review OK

CasVissers-360ERP (approved):
LGTM!

astirpe:
@pedrobaeza is it ok to merge this one?

alexis-via:
I think it was a mistake to port this module, because the feature is now native in v13. The native code that handles this feature is located in the base_vat module, in models/res_partner.py at the very end of the file: it inherits write and create and calls the compact method of the stdnum lib, cf https://github.com/odoo/odoo/blob/13.0/addons/base_vat/models/res_partner.py#L466

So, what shall we do ?
I think we should remove the module, but maybe some people have already installed it, so maybe we should just add a note in the README and keep it like that.
At least, we won't use it in OCA/edi. The only thing to handle is data migration for those who upgrade from v12... but I guess it is something to handle in openupgrade.

alexis-via:
I made a small PR to add a note in the README about it https://github.com/OCA/partner-contact/pull/862

PR #822, opened 2019-12-05
[13.0] [MIG] base country state translatable

poonlap:


StephanRozendaal (commented):
Thanks!
You can add yourself as a contributor, and maybe it's better to change the last commit to a 'migration' commit like so: [MIG] $MODULE: base_country_state_translatable to 13.0 ?

pedrobaeza:
There should be 2 commits in this branch migration: one for black, isort, and other for the rest of the stuff needed to be done for 13.0. See https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-13.0

poonlap:
@StephanRozendaal  @pedrobaeza  Thank you for the advices. 
I will close this pull request and retry again.

PR #828, opened 2019-12-12
[13.0][MIG] partner_bank_active

StephanRozendaal:
I've made a change to the way the module works because I noticed that in Odoo 13.0 activating / deactivating records is done differently in the Base Odoo code.


lfreeke (approved):
Functional test :+1:

ThomasBinsfeld (approved):

PR #830, opened 2019-12-19
[13.0] Fix partner_company_group filter priority

guewen:
Search on company group shouldn't overwrite the default invoice number search
Port of #773


pedrobaeza (approved):

gurneyalex (approved):

gurneyalex:
Apparently there's an issue with geonames.org : http://download.geonames.org/export/zip/ doesn't feature all the expected files. I suggest to wait for a couple of days.

pedrobaeza:
Yeah, not sure why Geonames is having such problems lately with the exported data files...

rafaelbn (approved):
:+1:

PR #833, opened 2019-12-21
[13.0][MIG] partner_contact_birthdate

newtratip:
Ready for review.


CarlosRoca13 (approved):
LGTM :+1:

newtratip:
@CarlosRoca13 Thank you for reviewed :)

newtratip:
@kittiu @ps-tubtim @Saran440 Can you review this ?

partner_contact_birthdate/__init__.py
```diff
@@ -0,0 +1,5 @@
+# Copyright 2014-2015  Grupo ESOC <www.grupoesoc.es>
```
NL66278: Remove copyright lines from init files.

partner_contact_birthdate/views/res_partner.xml
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+     Copyright 2015 Grupo ESOC Ingeniería de Servicios, S.L.U. -->
+
+<odoo>
+    <record id="view_personal_information_birthdate_date" model="ir.ui.view">
```
NL66278: The convention is to use the same id as in the inherited from module, but give a distinct name. Now the partner_personal_information_page did not follow this convention, so you have a choice of using personal_information here, or the original id. I would prefer the original id view_partner_form.

NL66278 (changes requested):
:+1: Looks good in general, but a few nags. See other comments.

partner_contact_birthdate/__init__.py
newtratip: @NL66278 
Please suggest me that why init files don't have copyright lines but other files such as res_partner.py have copyright lines.

Thank you.

partner_contact_birthdate/views/res_partner.xml
newtratip: Good idea.

newtratip: @NL66278 
Edited id to original id

Thank you for your review.

partner_contact_birthdate/__init__.py
NL66278: Actually the contributing guidelines are not speaking anymore about this. But the logic is that there is not really anything to claim copyright for...

newtratip: @NL66278 
I got it. This is new for me.

Thank you for reviewed :)

Saran440:
@newtratip please, rebase this pr to 2 commit -> isort and migrate.

NL66278 (approved):
:+1: Satisfied now. Just squash the last three commits into one.

newtratip:
@Saran440 and @NL66278 
I squash the last three commit into one already.

Thank you both for your suggest :)

PR #835, opened 2019-12-21
[13.0][MIG] partner_phone_extension

newtratip:
Ready for review.


partner_phone_extension/views/res_partner_view.xml
```diff
@@ -0,0 +1,46 @@
+<?xml version="1.0" encoding="utf-8"?>
```
NL66278, line 1: Not blocking, but xmlid's of views should be the same as the inherited views (they are where needed qualified by the module name anyway). The name can be used to indicate the purpose of the extention.

NL66278 (approved):
:+1: LGTM. One nag, see other comment.

partner_phone_extension/views/res_partner_view.xml
newtratip: I don't understand, Can you more explain this.

Sorry for delay replay.

rousseldenis:
@NL66278 We need to fix 13.0 branch as it is red. partner_firstname is failing

CarlosRoca13 (approved):

partner_phone_extension/views/res_partner_view.xml
NL66278: There is not much to explain. All the views should have the same xmlid in id= and in inherit_id=, except that inherit_id= will also have the name of the orginal module in it, and the xmlid in id= should be unqualified. The ir.ui.view name field can be used to describe what the extention or inheriting view's purpose is.

NL66278:
@CarlosRoca13 Could you review https://github.com/OCA/partner-contact/pull/854 ? It contains the needed fixes to get travis on on 13.0 PR's green.

newtratip:
Hello @NL66278 
I have change xmlid to same xmlid of inherit view, Could you check and merge again ?

PR #837, opened 2020-01-06
[12.0][FIX] partner_ref_unique: Trigger check on company update

grindtildeath:
Before this commit, the partner_ref_unique could be changed on the
company even if existing partners share the same internal ref.

Therefore, errors will appear afterwards as DB already contains
duplicates although the policy shouldn't allow it.

With this change, the error will appear as soon as the policy is
changed, so that the data must be fixed before activating the
constraint.

### DIFF
--- a/partner_ref_unique/models/res_partner.py
+++ b/partner_ref_unique/models/res_partner.py
@@ -11,7 +11,7 @@ class ResPartner(models.Model):
     # This related is needed in order to trigger the check when changing the
     # value on res.company
     partner_ref_unique = fields.Selection(
-        related='company_id.partner_ref_unique',
+        related='company_id.partner_ref_unique', store=True,
     )
 
     @api.multi
--- a/partner_ref_unique/models/res_partner.py
+++ b/partner_ref_unique/models/res_partner.py
@@ -1,18 +1,24 @@
 # Copyright 2016 Antonio Espinosa
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
-from odoo import _, api, models
+from odoo import _, api, fields, models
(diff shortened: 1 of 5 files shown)


Tonow-c2c (approved):

p-tombez (approved):

grindtildeath:
@antespi @pedrobaeza @rafaelbn @cubells  Can one of you guys please have a look ? :hugs:

partner_ref_unique/models/res_partner.py
```diff
@@ -1,18 +1,26 @@
...
 
 
 class ResPartner(models.Model):
     _inherit = "res.partner"
 
+    # This related is needed in order to trigger the check when changing the
+    # value on res.company
+    partner_ref_unique = fields.Selection(
+        related='company_id.partner_ref_unique',
+        store=True,
```
pedrobaeza: Don't store it for performance reason and readonly=True is the default

grindtildeath: @pedrobaeza I didn't have this commit (https://github.com/odoo/odoo/commit/62056e826a3b99186284cdfecc132cdfd7b1e8d7) in my code base so I post it here FTR.

pedrobaeza (approved):

pedrobaeza:
Please check Travis

grindtildeath:
@pedrobaeza Looks like the `store=True` is required or the `ValidationError` is not triggered :(

pedrobaeza:
Well, if store is needed, then please add a pre-init hook that creates the column and fill by SQL the data, or on a DB with a lot number of records the installation of the module will last a lot.

PR #840, opened 2020-01-12
[12.0][IMP] UX form view order first and last name

chrizzTs:
Change to order of the fields first and last name of partner view to improve UX.
Entering 1. "first name" and 2. "lastname" is the official order format in any western countries by improving this code, users do not mix up first and last name accidentally and avoid many data input errors.

### DIFF
--- a/partner_firstname/views/res_partner.xml
+++ b/partner_firstname/views/res_partner.xml
@@ -15,12 +15,12 @@
 
                 <xpath expr="//h1//field[@name='name']/.." position="before">
                     <group attrs="{'invisible': [('is_company', '=', True)]}">
-                        <field name="lastname" attrs=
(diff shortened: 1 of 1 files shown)


NL66278 (changes requested):
While it might be true that in daily use we call people by firstname + lastname, mostly in commercial or official communications we will use title + lastname. So I think this is very much a question of preference. In many cases the lastname will be known, but the fist not.

That makes that in my opinion it would be better to create a separate module, lets call it partner_firstname_first, that would change the order of the names, rather then force it on all present users of partner_firstname.

NL66278:
@chrizzTs What do you think of my proposal to make this into a separate module, so users will have a choice to change the order or not?

NL66278:
Questions are not taken up.

Issue #851, opened 2020-02-07
[13.0] partner_firstname: Tests errors


rousseldenis:
13.0 branch is red due to errors in partner_firstname.

I've checked and it could be a commit in core between 28/12/2019 and 23/01/2020.

@lmignon @NL66278

NL66278:
@rousseldenis I do not see much of commits on partner_firstname in this period. Maybe changes in some other piece of Odoo had a side-effect on partner_firstname. I will try to debug the problem.

rousseldenis:
> @rousseldenis I do not see much of commits on partner_firstname in this period. Maybe changes in some other piece of Odoo had a side-effect on partner_firstname. I will try to debug the problem.

Yes, that's what I said. I didn't explore but some commits on Odoo side have been done on that period.

rousseldenis:
@NL66278 I've found the commit that makes travis red:

https://github.com/odoo/odoo/commit/6c99fe3ffec3aeb8d210e3b943ccef81d62d44ce#diff-0a8f125a3f41f36c0507203a63cde9ad

rousseldenis:
See my comment : https://github.com/odoo/odoo/commit/6c99fe3ffec3aeb8d210e3b943ccef81d62d44ce#r37207800

NL66278:
See also this PR, needed to solve the `"Foó bis" != "Foó"` problem in the test of the partner form: https://github.com/odoo/odoo/pull/45355

NL66278:
Downstream PR is here: https://github.com/OCA/OCB/pull/909

PR #854, opened 2020-02-14
[13.0] [FIX] partner_firstname. Fix tests.

NL66278:
Actually there was nothing wrong with the tests by themselves, but rather with the
Odoo Form class in odoo/tests/common.py.

This commit disables one tests and works around the bugs in another.

See issue: https://github.com/OCA/partner-contact/issues/851


NL66278:
Note that without this (or an improved) fix no 13.0 PR can pass travis.

rousseldenis (approved):
Ok, let's go with this. When https://github.com/odoo/odoo/pull/45355 is merged we could revert it. As it's on test level I think there is minor risks.

CarlosRoca13 (approved):

PR #859, opened 2020-02-24
[10.0][ADD] partner_multi_name module

Capriatto:
This is our first contribution, all suggestions are welcome:

This module extends the "partner_second_lastname" functionality to have a middle name or other names in a separated way.

    Use Case:
    Partner Second Last Name: <Johannes Chrysostomus> <Theophilus> <Mozart>
    Partner Other Names: <Johannes> <Chrysostomus Wolfgangus> <Theophilus> <Mozart>


Capriatto:
ping: 
    @Tecnativa @Yajo @jgrandguillaume @rafaelbn

Capriatto:
ping: 
@Yajo  @jgrandguillaume @rafaelbn

pedrobaeza:
We are not using 10.0 anymore, so we are not interested in reviewing this. You should check any way contributing guidelines:

https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst

and other advises like:

* cleaned commit history
* only one module per PR
* if you want that someone reviews your PR, review you first others' PRs.

rousseldenis:
@Capriatto Can you take into account @pedrobaeza comments (split PR, ...) ? Thanks

Capriatto:
> @Capriatto Can you take into account @pedrobaeza comments (split PR, ...) ? Thanks

@rousseldenis Yes, as you can see we followed the recommendations. Plus, we added the 11.0 and 12.0 versions of the module at:

#864 
#865

Capriatto:
@rousseldenis we're done here. Does this PR needs any other fixing?

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+    "restructuredtext.confPath": ""
+}
```
janverb: Can you remove this file from git? I think `git rm --cached .vscode/settings.json` should work.

partner_other_names/models/base_config_settings.py
```diff
@@ -0,0 +1,43 @@
...
+            '&',
+            ('firstname', '!=', False),
+            ('lastname2', '!=', False),
+            '|',
+            '&',
+            ('othernames', '!=', False),
+            ('lastname', '!=', False),
+            '&',
+            ('othernames', '!=', False),
+            ('lastname2', '!=', False)])
```
janverb: This looks risky because you're not calling the version that's defined in `partner_second_lastname`. That one also looks for `'&', ('lastname', '!=', False), ('lastname2', '!=', False)`. Is that case left out on purpose, or by accident?

It might be better to take the results from the parent method and add onto that, like so:

```suggestion
        res = super(BaseConfigSettings, self)._partners_for_recalculating()
        res |= self.env['res.partner'].search([
            ('is_company', '=', False),
            '|',
            '&',
            ('othernames', '!=', False),
            ('lastname', '!=', False),
            '&',
            ('othernames', '!=', False),
            ('lastname2', '!=', False)])
        return res
```

partner_other_names/models/res_partner.py
```diff
@@ -0,0 +1,128 @@
...
+
+
+class ResPartner(models.Model):
+    """Adds other names."""
+    _inherit = "res.partner"
+
+    othernames = fields.Char("Other Names")
+
+    @api.model
+    def _get_computed_name(self, firstname, othernames, lastname, lastname2):
```
janverb: This method already exists in other modules, so it should probably stay compatible with those. That way if someone calls `._get_computed_name(last_name, first_name)` it'll still work if this module is installed.

It could be `def _get_computed_name(self, lastname, firstname, lastname2=None, othernames=None)`. That's similar to how `partner_second_lastname` does it: https://github.com/OCA/partner-contact/blob/3c6350d3/partner_second_lastname/models/res_partner.py#L18

partner_other_names/models/res_partner.py
```diff
@@ -0,0 +1,128 @@
...
+
+    @api.constrains("firstname", "othernames", "lastname", "lastname2")
+    def _check_name(self):
+        """Ensure at least one name is set."""
+        try:
+            super(ResPartner, self)._check_name()
+        except exceptions.EmptyNamesError:
+            for partner in self:
+                if not partner.othernames:
+                    raise
```
janverb: This can do the wrong thing if there are multiple partners.
If we run it on partner A and B, where A only has a `firstname` and B only has an `othernames`, then it raises because `A` doesn't have an `othernames` even though that's fine.
I think this would work better:
```suggestion
        for partner in self:
            try:
                super(ResPartner, partner)._check_name()
            except exceptions.EmptyNamesError:
                if not partner.othernames:
                    raise
```

janverb (commented):
This looks useful. I left a few comments.

Capriatto:
Hello Mr @janverb, thank you for your revision.  We will consider your suggestions very soon. For now, we rename the module because of the suggestion of Pedro Baeza in #865

.vscode/settings.json
Capriatto: TY for your revision: file removed as you suggested.

partner_other_names/models/base_config_settings.py
Capriatto: TY for you revision: Good catch! the line you mention was omitted by error, It was fixed too. We considered a big improvement your suggestion of " take the results from the parent method and add onto that". So, we did it.

partner_other_names/models/res_partner.py
Capriatto: TY for your revision: you're right, it was fixed.

partner_other_names/models/res_partner.py
Capriatto: TY for your revision: This method has been changed a bit in its logic to adapt to this module, for this reason other functions that do something similar cannot be used.

janverb:
It looks like f1a9c26993f23e15e34438971a09826651d314b8 accidentally undid most of the improvements, can you fix that?

Capriatto:
@rousseldenis hey!! how long, any other fix is required here? What coverage of code do we need to pass the codecov test? thanks.

rousseldenis:
> @rousseldenis hey!! how long, any other fix is required here? What coverage of code do we need to pass the codecov test? thanks.

For coverage, click on 'details' link (codecov/patch) here under and on the 'Diff' tab. The red lines are not tested. Maybe you can improve them in order to make it happy.

Capriatto:
> @Capriatto Can you take into account @pedrobaeza comments (split PR, ...) ? Thanks

@rousseldenis is still necessary the "needs fixing" tag?

partner_multi_name/models/base_config_settings.py
```diff
@@ -0,0 +1,43 @@
...
+        new_labels = {
+            'last_first': 'Lastname SecondLastname \
+                Firstname Othernames',
+            'last_first_comma': 'Lastname SecondLastname, \
+                Firstname Othernames',
+            'first_last': 'Firstname Othernames \
+                Lastname SecondLastname',
+        }
+        return [(k, new_labels[k]) if k in new_labels else
+                (k, v) for k, v in options]
```
janverb, line 29: You can also write it this way:
```suggestion
        return [(k, new_labels.get(k, v)) for k, v in options]
```

partner_multi_name/models/base_config_settings.py
```diff
@@ -0,0 +1,43 @@
...
+                        self)._partner_names_order_selection()
+
+        new_labels = {
+            'last_first': 'Lastname SecondLastname \
+                Firstname Othernames',
+            'last_first_comma': 'Lastname SecondLastname, \
+                Firstname Othernames',
+            'first_last': 'Firstname Othernames \
+                Lastname SecondLastname',
+        }
```
janverb, line 27: This puts a lot of spaces in the middle, though you can't see them on the web. You can write it like this instead:
```suggestion
        new_labels = {
            'last_first': 'Lastname SecondLastname '
                'Firstname Othernames',
            'last_first_comma': 'Lastname SecondLastname, '
                'Firstname Othernames',
            'first_last': 'Firstname Othernames '
                'Lastname SecondLastname',
        }
```

partner_multi_name/models/base_config_settings.py
```diff
@@ -0,0 +1,43 @@
...
+        res = super(BaseConfigSettings, self)._partners_for_recalculating()
+        res |= self.env['res.partner'].search([
+            ('is_company', '=', False),
+            '|',
+            '&',
+            ('othernames', '!=', False),
+            ('lastname', '!=', False),
+            '&',
+            ('othernames', '!=', False),
+            ('lastname2', '!=', False)])
```
janverb, line 42: I think this method has to find all partners with more than one name.

Then it should also look at combinations like `'&', ('othernames', '!=', False), ('firstname', '!=', False)`. But that's a lot of work. This is easier, and I think it's good enough:
```suggestion
        res |= self.env['res.partner'].search([
            ('is_company', '=', False),
            ('othernames', '!=', False)])
```

partner_multi_name/models/res_partner.py
```diff
@@ -0,0 +1,131 @@
...
+
+            if order == 'first_last':
+                if result['lastname2']:
+                    parts = result['lastname2'].split(" ", 1)
+                while len(parts) < 2:
+                    result['othernames'] = False
+                    return result
+                result['othernames'] = result['lastname']
+                result['lastname'] = parts[0]
+                result['lastname2'] = parts[1]
```
janverb, line 107: I think this is clearer if you write it like this:
```suggestion
                if len(parts) < 2:
                    result['othernames'] = False
                else:
                    result['othernames'] = result['lastname']
                    result['lastname'] = parts[0]
                    result['lastname2'] = parts[1]
```

partner_multi_name/views/res_partner_views.xml
```diff
@@ -0,0 +1,69 @@
...
+                <field name="othernames"
+                    attrs="{'required': ['|', '|', ('lastname', '!=', False), ('lastname2', '!=', False), ('lastname', '=', False), ('lastname2', '=', False), ('firstname', '=', False), ('is_company', '=', False)]}"/>
+            </xpath>
+
+            <xpath expr="//field[@name='lastname']" position="attributes">
+                <attribute name="attrs">{'required': ['|', '|', ('firstname', '!=', False), ('othernames', '!=', False), ('firstname', '=', False), ('othernames', '=', False), ('lastname2', '=', False), ('is_company', '=', False)]}</attribute>
+            </xpath>
+
+            <xpath expr="//field[@name='lastname2']" position="attributes">
+                <attribute name="attrs">{'required': ['|', '|', ('firstname', '!=', False), ('othernames', '!=', False), ('firstname', '=', False), ('othernames', '=', False), ('lastname', '=', False), ('is_company', '=', False)]}</attribute>
```
janverb, line 22: These look complicated, can you explain what they're supposed to do?

I think they don't work right, because `'|', '|', ('lastname', '!=', False), ('lastname2', '!=', False), ('lastname', '=', False)` is always true, because either `lastname == False` or `lastname != False`.

partner_multi_name/views/res_partner_views.xml
```diff
@@ -0,0 +1,69 @@
...
+                <field name="othernames"
+                    attrs="{'required': [('firstname', '=', False), ('is_company', '=', False)]}"/>
+            </xpath>
+
+            <xpath expr="//field[@name='child_ids']/form//field[@name='lastname']" position="attributes">
+                <attribute name="attrs">{'required': [('lastname2', '=', False), ('is_company', '=', False)]}</attribute>
+            </xpath>
+
+            <xpath expr="//field[@name='child_ids']/form//field[@name='lastname2']" position="attributes">
+                <attribute name="attrs">{'required': [('lastname', '=', False), ('is_company', '=', False)]}</attribute>
```
janverb, line 65: I don't think these are right. Compare to https://github.com/OCA/partner-contact/blob/10.0/partner_second_lastname/views/res_partner.xml#L31, then for the `firstname` field I think you'd want `{'required': [('othernames', '=', False), ('lastname', '=', False), ('lastname2', '=', False), ('is_company', '=', False)]}`. And so on for the others. So for each field you check the other three fields.

partner_multi_name/views/res_users_views.xml
```diff
@@ -0,0 +1,24 @@
...
+                <field name="firstname" attrs="{'required': [('othernames', '=', False)]}"/>
+                <field name="othernames" attrs="{'required': [('firstname', '=', False)]}"/>
+            </xpath>
+
+            <xpath expr="//field[@name='lastname']" position="attributes">
+                <attribute name="attrs">{'required': [('lastname2', '=', False)]}</attribute>
+            </xpath>
+
+            <xpath expr="//field[@name='lastname2']" position="attributes">
+                <attribute name="attrs">{'required': [('lastname', '=', False)]}</attribute>
```
janverb, line 20: The same here as my second comment on `res_partner_views.xml`

janverb (changes requested):
This is looking better! Sorry for taking a while. Just a few more issues, and some unimportant stuff about style.

NL66278:
Closing this as not maintained and 10.0 no longer supported.

PR #860, opened 2020-03-01
Fix migration when state_id is NULL

ypapouin:
Use "IS NOT DISTINCT FROM" instead of the equal operator since state_id could be NULL in some countries.

Note that I also applied it to country_id even if the previous query exclude null country_id, in order to be more consistent.

### DIFF
--- a/base_location/migrations/12.0.1.0.0/post-migration.py
+++ b/base_location/migrations/12.0.1.0.0/post-migration.py
@@ -36,7 +36,7 @@ def migrate(env, version):
         SET city_id = rc.id
         FROM res_city rc
         WHERE rc.name = rbz.city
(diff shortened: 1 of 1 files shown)


pedrobaeza:
I'm not sure about this, as data should be completed. Are you talking about data manually introduced?

ypapouin:
Data was imported from an older module [1], but in both modules, state_id is not required (contrary to country_id).

[1] https://bazaar.launchpad.net/~openerp-commiter/openobject-addons/trunk-extra-addons/files/head:/city

pedrobaeza:
OK, remove `country_id` change and it's good to go.

ypapouin:
Done

pedrobaeza (approved):

PR #864, opened 2020-03-05
[11.0][ADD] partner_multi_name module

Capriatto:
This is our first contribution, all suggestions are welcome:

This module extends the "partner_second_lastname" functionality to have a middle name or other names in a separated way.

    Use Case:
    Partner Second Last Name: <Johannes Chrysostomus>  <Theophilus> <Mozart>
    Partner Other Names: <Johannes><Chrysostomus Wolfgangus>< Theophilus><Mozart>


partner_other_names/models/res_partner.py
```diff
@@ -0,0 +1,128 @@
...
+
+
+class ResPartner(models.Model):
+    """Adds other names."""
+    _inherit = "res.partner"
+
+    othernames = fields.Char("Other Names")
+
+    @api.model
+    def _get_computed_name(self, firstname, othernames, lastname, lastname2):
```
joao-p-marques: You could add default argument values here to increase resilience.
```suggestion
    def _get_computed_name(self, firstname=None, othernames=None, lastname=None, lastname2=None):
```

NL66278 (changes requested):
Module name must be changed to not contain plural. See also corresponding review for 12.0 version.

partner_other_names/models/res_partner.py
joao-p-marques: You didn't change this...

joao-p-marques (approved):

partner_other_names/models/res_partner.py
Capriatto: You're right. We already did.

Capriatto:
> Module name must be changed to not contain plural. See also corresponding review for 12.0 version.

@NL66278 is still necessary the "needs fixing" tag?

partner_multi_name/.vscode/settings.json
```diff
@@ -0,0 +1,5 @@
+{
```
NL66278, line 1: This file should not be part of the PR

NL66278:
It would be good to add some unit tests...

partner_multi_name/.vscode/settings.json
grottas: @alejo-code @JoanMarin FYI

PR #866, opened 2020-03-07
[FIX] partner_multi_relation: Context Errors With Active ID

osi-scampbell:
When clicking through multiple buttons from different models the active_id will stay as the original id from the first button that was clicked. In the .xml we are overwriting the active_id which is good practice however we have had multiple reports of the 'search_default_this_partner_id' and 'default_this_partner_id' using the previous active_ids. 

Example: (Moduls installed: fieldservice, partner_multi_relation)
FSM Location -> Contacts
Contacts -> Relations
'search_default_this_partner_id' and 'default_this_partner_id' are using the id from the FSM Location.

The solution is to simply handmake the action in model itself. I believe this way we can easily inherit the action if we need it to be changed and we always can esure the proper default values are being set.

### DIFF
--- a/partner_multi_relation/models/res_partner.py
+++ b/partner_multi_relation/models/res_partner.py
@@ -176,9 +176,10 @@ class ResPartner(models.Model):
     def action_view_relations(self):
         for contact in self:
             relation_model = self.env['res.partner.relation.all']
-            relation_ids = relation_model.search(['|', 
-                ('this_partner_id', '=', contact.id),
-                ('other_partner_id', '=', contact.id)])
+            relation_ids = relation_model.\
+                search(['|',
+                       ('this_partner_id', '=', contact.id),
+                       ('other_partner_id', '=', contact.id)])
             action = self.env.ref(
                 'partner_multi_relation.action_res_partner_relation_all'
             ).read()[0]
--- a/partner_multi_relation/models/res_partner.py
+++ b/partner_multi_relation/models/res_partner.py
@@ -171,3 +171,23 @@ class ResPartner(models.Model):
         """
         self.ensure_one()
         return 'c' if self.is_company else 'p'
+
+    @api.multi
+    def action_view_relations(self):
+        for contact in self:
+            relation_ids = self.env['res.partner.relation.all'].\
+                search(['|', ('this_partner_id', '=', contact.id),
+                       ('other_partner_id', '=', contact.id)])
+            action = self.env.\
+                ref('partner_multi_relation.action_res_partner_relation_all').\
+                read()[0]
+            action['context'] = {}
+            action['domain'] = [('id', 'in', relation_ids.ids)]
+            action['context'].\
+                update({'search_default_this_partner_id': contact.id,
+                        'default_this_partner_id': contact.id,
+                        'active_model': 'res.partner',
+                        'active_id': contact.id,
+                        'active_ids': [contact.id],
+                        'active_test': False})
+            return action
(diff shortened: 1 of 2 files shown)


partner_multi_relation/views/res_partner.xml
```diff
@@ -21,25 +21,13 @@
...
-                            'default_this_partner_id': active_id,
-                            'active_model': 'res.partner',
-                            'active_id': id,
-                            'active_ids': [id],
-                            'active_test': False,
-                        }"
-                        name="%(action_res_partner_relation_all)d"
-                        icon="fa-users"
-                        >
-                        <field
```
NL66278, line 38: Leave the field element lines as is, there is nothing wrong with the existing formatting and indentation.

partner_multi_relation/models/res_partner.py
```diff
@@ -171,3 +171,23 @@ def get_partner_type(self):
...
+        for contact in self:
+            relation_ids = self.env['res.partner.relation.all'].\
+                search(['|', ('this_partner_id', '=', contact.id),
+                       ('other_partner_id', '=', contact.id)])
+            action = self.env.\
+                ref('partner_multi_relation.action_res_partner_relation_all').\
+                read()[0]
+            action['context'] = {}
+            action['domain'] = [('id', 'in', relation_ids.ids)]
+            action['context'].\
```
NL66278, line 187: Merge the existing context (self.env.context, with the updated values in one assignment to action['context'].

partner_multi_relation/models/res_partner.py
```diff
@@ -171,3 +171,23 @@ def get_partner_type(self):
         """
         self.ensure_one()
         return 'c' if self.is_company else 'p'
+
+    @api.multi
+    def action_view_relations(self):
+        for contact in self:
+            relation_ids = self.env['res.partner.relation.all'].\
```
NL66278: For style I would like to avoid the `'\'` line continuation, where not really needed.
If you assign `relation_model = self.env['res.partner.relation.all']` and put if for the loop, you can use
`relation_model.search(` and then can put the conditons on the next lines indented by 4 spaces.

partner_multi_relation/models/res_partner.py
```diff
@@ -171,3 +171,23 @@ def get_partner_type(self):
...
+                search(['|', ('this_partner_id', '=', contact.id),
+                       ('other_partner_id', '=', contact.id)])
+            action = self.env.\
```
NL66278: Also here you could do:
```
action = self.env.ref(
    'partner_multi_relation.action_res_partner_relation_all'
).read()[0]
```

NL66278 (changes requested):
Thanks for noting the possible problems and offering a fix. Just some styling requests. For the rest I would say it is good to go.

NL66278 (approved):
:+1: LGTM now. Could have prevented two more `\` line continuations, but not going to block over that. Thanks a lot for your contribution!

skeller1:
@osi-scampbell: Could you please provide a scenario on any OCA partner-contact runbot without using other addons. I'm understanding the code changes but can't reproduce the error locally or in runbot instance. Thank you

skeller1:
@osi-scampbell: There was a Odoo PR 1 month ago to fix a context problem.
This seems to be an issue related to your reported behaviour "surfing" through some views in Odoo

https://github.com/odoo/odoo/commit/328415ca88a1b9b53e80e0abb638832bc5ce1876

osi-scampbell:
@skeller1 I have not reproduced it on a partner-contact runbot I will attempt to do that. I believe "surfing" through views accurately describes the problem we encountered I am curious to see if that solved this problem (clicking through multiple buttons 3-4 times). Thank you very much for your feedback. I will investigate further.

max3903 (approved):

PR #867, opened 2020-03-11
[12.0][IMP] - filter city zip by country in partner form

sbejaoui:


rousseldenis (approved):
Code review.

base_location/models/res_partner.py
```diff
@@ -34,6 +34,17 @@ def _onchange_country_id(self):
...
             self.zip_id = False
+        if self.country_id:
+            city_zip_domain = {
+                "zip_id": [("city_id.country_id", "=", self.country_id.id)]
+            }
+            if isinstance(res, dict):
+                if "domain" in res and isinstance(res["domain"], dict):
+                    res["domain"].update(city_zip_domain)
+                else:
+                    res["domain"] = city_zip_domain
```
pedrobaeza: Is this part really needed?

base_location/models/res_partner.py
```diff
@@ -34,6 +34,17 @@ def _onchange_country_id(self):
         res = super()._onchange_country_id()
         if self.zip_id and self.zip_id.city_id.country_id != self.country_id:
             self.zip_id = False
+        if self.country_id:
+            city_zip_domain = {
+                "zip_id": [("city_id.country_id", "=", self.country_id.id)]
+            }
+            if isinstance(res, dict):
+                if "domain" in res and isinstance(res["domain"], dict):
```
pedrobaeza: Maybe better to use `zip_domain = res.setdefault('domain', {})`?

base_location/models/res_partner.py
sbejaoui: `_onchange_country_id` return None in `base` module but it may be inherited in any other module to return a dict that contain or not a domain key.

At this point I can't know in which case we are:
*  res == None
*  res == {'value': {}}
*  res == {'value': {}, 'domain': {}}

pedrobaeza: Yeah, but this case you are covering is:
```python
res == {'value': {}, 'domain': None}
```
which I think it's never going to happen.

sbejaoui: yes, but essentially  `res == {'value': {}}` (`else "domain" in res`).

Maybe I shouldn't add `isinstance(res["domain"], dict)` but i thought it's safer

pedrobaeza: That's why I insist in simplify code and avoid coverage drop for non existent cases. See my other suggestion.

base_location/models/res_partner.py
sbejaoui: ok Thnak you,

I applied your suggestion

gva-acsone (approved):
LGTM ! +1 Functional test

base_location/models/res_partner.py
```diff
@@ -27,13 +27,28 @@ def _onchange_city_id(self):
                     'zip_id': [('city_id', '=', self.city_id.id)]
                 },
             }
+        if self.country_id:
```
pedrobaeza, line 30:
```suggestion
        if self.country_id and self.contry_enforce_cities:
```

pedrobaeza (changes requested):

base_location/models/res_partner.py
sbejaoui: why you would like to add this condition?

pedrobaeza: This follows https://github.com/OCA/partner-contact/commit/64aaafa45c022116084675e6181d338a8c386f83 and it's for avoiding empty search, as `city_id` is not filled in that cases.

sbejaoui: We don't risk to get an empty search as we return zip based on the selected country. The data-model guarantee that a `zip.city` object have a related country.

IMO, it make sense that we have empty search if the user select a country that he didn't set zips for.

If for some use cases, we want to show all zips even the user select a country, I believe `enforce_cities` is not a good criteria. A new field `restrict_zip_to_country` make more sense.

pedrobaeza: OK, I didn't remember `city_id` is now required for zips in the new data model. OK then

pedrobaeza (approved):

pedrobaeza:
/ocabot merge patch

should this be forward-port it to v13?

PR #870, opened 2020-03-17
[13.0][MIG] partner_priority

newtratip:
Ready for review.


lfreeke (approved):
Functional test :+1:

patrickrwilson (approved):
LGTM

newtratip:
@Saran440 @ps-tubtim Could you review this ?

ps-tubtim (approved):

partner_priority/data/partner_priority_data.xml
```diff
@@ -0,0 +1,24 @@
...
+        <field name="name">Low</field>
+        <field name="description">Low Priority</field>
+        <field name="sequence">1</field>
+    </record>
+    <record id="res_partner_priority_normal" model="partner.priority">
+        <field name="name">Normal</field>
+        <field name="description">Normal Priority</field>
+        <field name="sequence">3</field>
+    </record>
+    <record id="res_partner_priority_elavated" model="partner.priority">
```
NL66278: There is a spelling error in the xml id: elavated instead of elevated.

partner_priority/models/partner_priority.py
```diff
@@ -0,0 +1,20 @@
...
+
+
+class PartnerPriority(models.Model):
+    _name = "partner.priority"
+    _order = "sequence"
+    _description = "Partner Priority"
+
+    name = fields.Char(string="Priority", required=True)
+    description = fields.Text(required=True)
+    sequence = fields.Integer(
```
NL66278, line 14: Having this field set from an ir.sequence instance looks a little bit to over-engineering to me.

NL66278 (approved):
Some remarks. But not blocking as this is a migration.

partner_priority/data/partner_priority_data.xml
newtratip: Edited.

rousseldenis:
@NL66278 The ocabot merge command has changed. The nobump option does not generate wheels for pypi. Even for migration PRs now, use major/minor/patch.

NL66278:
@rousseldenis In the past when I used major on a migration merge, I was told to leave out the major/minor/patch altogether. Recently the version argument became required, but the nobump was added, so I used that. So what am I supposed to use on a migration merge? major would seem to me the most logical...

rousseldenis:
Following the conversation in mailing list, the idea is to use patch. So, at the end, the module would have 13.0.1.0.1. But the wheels should be generated.

PR #872, opened 2020-03-25
new module crm_deduplicate_contact

jans23:
This module enables the dialogue to merge contacts which is hidden in current Odoo versions. The required functionality is part of Odoo already. You can find the dialogue under: Contacts -> Tools -> Deduplicate Contacts


pedrobaeza:
This is already done in https://github.com/OCA/partner-contact/tree/12.0/partner_deduplicate_acl

jans23:
I didn't noticed. Thank you for the information.

pedrobaeza:
It was previously hosted her and called `crm_deduplicate_acl` in version 11, because it required `crm` module. Now it's on `contacts` one, so that's why it's moved and renamed.

PR #873, opened 2020-03-30
[12.0][MIG] partner_label

hbrunn:


partner_label/tests/test_partner_label.py
```diff
@@ -0,0 +1,26 @@
+# Copyright 2017 Therp BV <http://therp.nl>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from openerp.tests.common import TransactionCase
```
NL66278: from odoo

partner_label/models/res_company.py
```diff
@@ -0,0 +1,41 @@
+# Copyright 2017-2019 Therp BV <http://therp.nl>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from openerp import fields, models
```
NL66278: from odoo

partner_label/models/res_config_settings.py
```diff
@@ -0,0 +1,66 @@
+# Copyright 2017-2019 Therp BV <http://therp.nl>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from odoo import api, fields, models
+
+
+class ResConfigSettings(models.TransientModel):
+    _inherit = 'res.config.settings'
+
+    partner_labels_width = fields.Float(
```
NL66278, line 9: Was there a reason not to put the settings directly in res.config.settings, but have them in company and then use related?

partner_label/__manifest__.py
```diff
@@ -0,0 +1,18 @@
+# Copyright 2019 Therp BV <http://therp.nl>
```
NL66278: Please change http: to https: in copyright and license lines troughout the module.

NL66278 (changes requested):
Small changes requested and one question.

partner_label/models/res_config_settings.py
hbrunn: `res.config.settings` is a transient model, we can't use this for permanent storage of our parameters. And making it related to the current company gives you storage very conveniently without having to write any {de,}serialization code.

Reading stuff saved on the current company is also very convenient, because you always can say `self.env.user.company_id`.

By now we've also a [convenient way](https://github.com/OCA/OCB/blob/12.0/odoo/addons/base/models/res_config.py#L371) to have a settings wizard write into config parameters, but those would be a little bit clumsy to use in https://github.com/OCA/partner-contact/pull/873/files#diff-adf307efb81bc5fecba0f8320c52202eR6

partner_label/models/res_company.py
NL66278: This one still needs to change.

partner_label/tests/test_partner_label.py
NL66278: And this one

hbrunn: thanks, for exactly this repo my checkers were botched and didn't actually check anything...

NL66278 (approved):
:+1: LGTM

ivantodorovich (approved):

ivantodorovich:
LGTM :+1: ! thanks!!

PR #876, opened 2020-04-02
[13.0] Add module partner_tz

grindtildeath:
This module removes timezone default value on res.partner and display the field
on form view.


jgrandguillaume (approved):
LGTM as far as I can tell

setup/partner_tz/odoo/addons/partner_tz
```diff
@@ -0,0 +1 @@
+../../../../partner_tz
```
Tonow-c2c, line 1: this work in every project scaffold ?

partner_tz/tools/tz_utils.py
```diff
@@ -0,0 +1,58 @@
...
+    :return: tz-naive datetime.datetime object
+    """
+    if isinstance(from_tz, str):
+        from_tz = pytz.timezone(from_tz)
+    if isinstance(to_tz, str):
+        to_tz = pytz.timezone(to_tz)
+    return from_tz.localize(date_time).astimezone(to_tz).replace(tzinfo=None)
+
+
+def tz_to_utc_naive_datetime(from_tz, date_time):
```
Tonow-c2c, line 26: it's odoo core method and you override them ?

partner_tz/__manifest__.py
```diff
@@ -0,0 +1,16 @@
+# Copyright 2020 Camptocamp SA
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
+{
+    "name": "Partner timezone",
+    "summary": "Remove partner timezone default value and display on form",
+    "version": "13.0.1.0.0",
+    "development_status": "Alpha",
+    "category": "Uncategorized",
```
Tonow-c2c, line 8: May be the category can be improve

setup/partner_tz/odoo/addons/partner_tz
grindtildeath: IIRC these setup files are regenerated after merges by the OCA git bot...

partner_tz/tools/tz_utils.py
grindtildeath: It is not, it's new helpers tools if one must do tz computations.

partner_tz/__manifest__.py
grindtildeath: Do you have a suggestion? It's only depending on base, and looking at other modules in this repo, I'm not sure there are much better fits :shrug:

LoisRForgeFlow (commented):
Looks good, could you check travis error?

LoisRForgeFlow:
@grindtildeath Hi, this module is a dependency for several modules you are proposing, if we get this merged it will ease testing and attract reviewers. Can you take a few minutes to fix travis and push this forward? :rocket:

partner_tz/tools/tz_utils.py
```diff
@@ -0,0 +1,58 @@
+# Copyright 2020 Camptocamp SA
```
NL66278, line 1: The functions presented here are very thin wrappers around the original functions from pytz. I think this is of questionable value. Better use the original pytz functions that are generally available and not depending on having this particular module installed.

NL66278 (changes requested):
Please limit the module to adding the timezone to the partner form.

But you might think about the suggested functionality of setting the default partner timezone based on the partners country...

HaraldPanten (approved):
Functional review

partner_tz/tools/tz_utils.py
grindtildeath: Thx for your review @NL66278 
The value of a function is quite subjective in any case, but to me they make tz conversions more explicit and avoid repeats. Moreover they are only tools that won't change anything when someone installs this module.
So I'm sorry but I don't see why having this file here, and using it in other modules having this module as a dependency would be a problem...

rousseldenis:
@grindtildeath I'd prefer explicit words in module name, partner_timezone is better

rousseldenis (approved):
Code review. Apart module name

rousseldenis:
And some tests for conversion functions should be great

jgrandguillaume:
Merging as Alpha version is set. Thank everyone for the review.

PR #879, opened 2020-04-03
[IMP] 13.0 Partner contact language

pegonzalezspesol:
On multicultural and multilanguage environments becomes more necessary to manage Contacts based on their language
This module extends standard Contact views with next features:
- Show the language in the narrowed contact form view. This, as example, makes possible to define several contacts with different language in the moment of the contact creation
- Include the language as filter in Contact search view
- Include the language as default group in Contact search view


pedrobaeza:
I think you have duplicated the features of https://github.com/OCA/partner-contact/tree/11.0/partner_contact_lang. Migrate/extend that one instead.

pegonzalezspesol:
> I think you have duplicated the features of https://github.com/OCA/partner-contact/tree/11.0/partner_contact_lang. Migrate/extend that one instead.

Thanks for your comment Pedro

Looking at v11 OCA module.. the behaviour is just to keep consistency in between the languages of parent contact and childs. This is the opposite to what my customer need.

The aim of my module, fitting my customer needs, is to have a main contact defined in a language, and to allow the childs to have a different one. As standard Odoo (For example, a main company contact with language EN, but the commercials with language DE or FR). Then I do visible the language in the view of the contact, to can easily select a different one.

By other hand, I do improve the search view, to find Customers by language and also to group them,  enabling the possibility to do whatever action I need, like massive mailing by language

What do you think?. Does in your mind worth to share this with the community?

pedrobaeza:
Yes, I see it interesting, but as part of the other module. They are complement and should be put all together for consistency. In previous versions, there were no need of putting the language in contact inner form as it was included. Now it seems it's needed.

pegonzalezspesol:
> Yes, I see it interesting, but as part of the other module. They are complement and should be put all together for consistency. In previous versions, there were no need of putting the language in contact inner form as it was included. Now it seems it's needed.

Hi @pedrobaeza 
Migration and functionality extended as you request in PR #882. I do close this PR.

PR #880, opened 2020-04-03
[13.0][IMP] base_location_geonames_import: Allows to import several countries altogether

manuelregidor:
Allows to import several countries from the same wizard instead of importing them one by one through different wizards.


pedrobaeza:
Please rebase over latest branch, as I pushed yesterday some fixes.

pedrobaeza:
Please don't perform merge commits in the PR that wastes the commit history. There's no need as well to bump manually module version, as now the mergebot performs it.

PR #881, opened 2020-04-03
Fix the naming of tab_left_id/tab_right_id so _update_right_vals kick…

Rad0van:
…s in on symmetric relations

Otherwise revrse partners were not shown in tabs.

### DIFF
--- a/partner_multi_relation_tabs/demo/res_partner_relation_type_demo.xml
+++ b/partner_multi_relation_tabs/demo/res_partner_relation_type_demo.xml
@@ -8,7 +8,7 @@
         <field name="name">has ceo</field>
         <field name="name_inverse">is ceo of</field>
         <field name="contact_type_left">c</field>
-        <field name="tab_left_id" ref="tab_board" />
+        <field name="tab_id_left" ref="tab_board" />
         <field name="contact_type_right">p</field>
         <field name="partner_category_right" ref="category_functionary"/>
     </record>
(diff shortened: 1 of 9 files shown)


NL66278:
@Rad0van Please describe a little bit more what goes wrong with the present code.

That said, I don't think this would be the good way to solve it, as it involves giving Many2one fields a non-standard name (not ending in _id). Also this will cause problems with existing installations as no migration is provided.

Rad0van:
> @Rad0van Please describe a little bit more what goes wrong with the present code.
> 
> That said, I don't think this would be the good way to solve it, as it involves giving Many2one fields a non-standard name (not ending in _id). Also this will cause problems with existing installations as no migration is provided.

This module is an extension of partner_multi_relation which allows to form relations between contacts. This bug involves symmetric kind of relations when used in tabs in this module (partner_multi_relation_tabs). If you configure a tab where symmetric kind of relation should provide items it works only from the left side of the relation. In main module the function _update_right_vals ensures that when you're editing symmetric relationship, all the properties named *_left are copied to their respective *_right counterparts. This is needed because the gui shows only left side. This works fine for partner_multi_relation but fails for partner_multi_relation_tabs, where tab id from left side is not copied to the right side. The reason being is those properties are named tab_left_id and tab_right_id - the _update_righ_vals does not catch up with these.

I am newbie with Python and Odoo so I was not aware of standard naming (nice) and existing installations. So alternative would be I guess to extend the _update_right_vals in partner_multi_relation_tabs to ensure proper copying of tab id from left to right.

NL66278:
I agree it will probably be better, and less invasive,  to adjust the _update_right_vals method.

Rad0van:
OK, closing this now. Will try different approach.

PR #883, opened 2020-04-05
[12.0] Fix tab functioning on symmetric relations

Rad0van:
This fixes setting up tab on symmetric relations because original method only handles properties having names ending with '_left' and we need to extend it to tab_left_id.

### DIFF
--- a/partner_multi_relation_tabs/models/res_partner_relation_type.py
+++ b/partner_multi_relation_tabs/models/res_partner_relation_type.py
@@ -57,3 +57,13 @@ class ResPartnerRelationType(models.Model):
                     tab_partner_category_id != rec.partner_category_right:
                 raise ValidationError(_(
                     "Partner category right not compatible with right tab"))
+
+    @api.multi
+    def _update_right_vals(self, vals):
+        """Make sure that on symmetric relations, right vals follow left vals.
+
+        @attention: original method only handles properties ending with _left
+                    and we need to update tab_right_id as well
+        """
+        vals['tab_right_id'] = vals.get('tab_left_id', self['tab_left_id'])
+        super(ResPartnerRelationType, self)._update_right_vals(vals)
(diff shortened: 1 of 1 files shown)


partner_multi_relation_tabs/models/res_partner_relation_type.py
```diff
@@ -57,3 +57,14 @@ def _check_tab_right(self):
                     tab_partner_category_id != rec.partner_category_right:
                 raise ValidationError(_(
                     "Partner category right not compatible with right tab"))
+
```
NL66278, line 60: Leave out one blank line (functions within a class should be separated with a single blank line), and travis test will be green.

NL66278:
Looks good mostly, but you need to solve the travis error - see other comment.

partner_multi_relation_tabs/models/res_partner_relation_type.py
Rad0van: Did that but Travis fails but it seems to me on something unrelated. I am quite newbie in hands-on experience with CI and stuff (git as well) so please bear with me. I also attempted to port this module as well at the additional one to work with Products - see here: https://github.com/OCA/product-attribute/pull/598 That for sure needs some work - I now found out what the tests directory is for so need to work on that there...

NL66278:
@Rad0van The new errors are I think unrelated to your changes. Code looks completely good now to me. Would you be able to merge the two commits into one?

Rad0van:
> @Rad0van The new errors are I think unrelated to your changes. Code looks completely good now to me. Would you be able to merge the two commits into one?

Would love to but not sure how to do that. Could you give me some quick pointer on how to do that? thanx.

Rad0van:
> @Rad0van The new errors are I think unrelated to your changes. Code looks completely good now to me. Would you be able to merge the two commits into one?

Not sure if it did what I wanted to... But I tried.

NL66278:
@Rad0van Well there are now four commits instead of one...

The "canonical" way to solve this would be to do an interactive rebase of your branch:
git -i rebase HEAD~4. Then mark the last three commits as a fixup of the first.

In case this is too complicated: save your change. Delete your branch. Create it again with the same name from the original (OCA 12.0). Restore your changes.

Whatever method you choose, you wil have to force push your branch to your repository again:
git push -f <remote name> <branch name>.

Tip: your branch is also called 12.0, like the original oca branch. Better next time to use a separate branch for each fix or feature you develop.

PR #884, opened 2020-04-06
[IMP] base_location_geonames_import: Import several countries through the same wizard.

manuelregidor:
Allow to import one or more countries information selecting them altogether in the same wizard.

### DIFF
--- a/base_location_geonames_import/tests/test_base_location_geonames_import.py
+++ b/base_location_geonames_import/tests/test_base_location_geonames_import.py
@@ -1,4 +1,5 @@
 # Copyright 2016-2019 Pedro M. Baeza <pedro.baeza@tecnativa.com>
+# Copyright 2020 Manuel Regidor <manuel.regidor@sygel.es>
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
 
 from odoo.exceptions import UserError
@@ -14,13 +15,22 @@ class TestBaseLocationGeonamesImport(common.SavepointCase):
             {"name": "Test city", "country_id": cls.country.id}
         )
         cls.wizard = cls.env["city.zip.geonames.import"].create(
-            {"country_id": cls.country.id}
+            {"country_ids": [(6, 0, [cls.country.id])]}
         )
         cls.wrong_country = cls.env["res.country"].create(
             {"name": "Wrong country", "code": "ZZYYXX"}
         )
         cls.wrong_wizard = cls.env["city.zip.geonames.import"].create(
-            {"country_id": cls.wrong_country.id}
+            {"country_ids": [(6, 0, [cls.wrong_country.id])]}
+        )
+        cls.country_2 = cls.env.ref("base.li")
+        cls.country_3 = cls.env.ref("base.sm")
+        cls.wizard_2 = cls.env["city.zip.geonames.import"].create(
+            {"country_ids": [(6, 0, [cls.country_2.id, cls.country_3.id])]}
+        )
+        cls.country_4 = cls.env.ref("base.ad")
+        cls.wizard_3 = cls.env["city.zip.geonames.import"].create(
+            {"country_id": cls.country_4.id}
         )
 
     def test_import_country(self):
@@ -146,7 +156,7 @@ class TestBaseLocationGeonamesImport(common.SavepointCase):
                 "4",
             ],
         ]
-        self.wizard._process_csv(parsed_csv)
+        self.wizard._process_csv(parsed_csv, country)
         cities = self.env["res.city"].search([("name", "=", "Auburn")])
         self.assertEqual(len(cities), 2)
         mapping = [
@@ -166,3 +176,17 @@ class TestBaseLocationGeonamesImport(common.SavepointCase):
                 state,
                 "Incorrect state for {} {}".format(state_name, zip_code),
             )
+
+    def test_import_countries(self):
+        max_import = 1
+        self.wizard_2.with_context(max_import=max_import).run_import()
+
+        # Look if there are imported zips
+        zip_country_2_count = self.env["res.city.zip"].search_count(
+            [("city_id.country_id", "=", self.country_2.id)]
+        )
+        zip_country_3_count = self.env["res.city.zip"].search_count(
+            [("city_id.country_id", "=", self.country_3.id)]
+        )
+        self.assertEqual(zip_country_2_count, max_import)
+        self.assertEqual(zip_country_3_count, max_import)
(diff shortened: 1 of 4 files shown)


base_location_geonames_import/__manifest__.py
```diff
@@ -17,6 +18,7 @@
         "Agile Business Group,"
         "Tecnativa,"
         "AdaptiveCity,"
+        "Sygel,"
```
pedrobaeza: This contribution doesn't qualify for co-authorship I'm afraid.

pedrobaeza:
Please avoid superfluous changes, like removing a related field and change variables across code. Keep the minimal changes for the new feature.

manuelregidor:
> Please avoid superfluous changes, like removing a related field and change variables across code. Keep the minimal changes for the new feature.

Thank you for your feedback. Please, could you indicate which changes you are refering to? Thank you.

base_location_geonames_import/wizard/geonames_import.py
```diff
@@ -25,13 +26,8 @@
 class CityZipGeonamesImport(models.TransientModel):
     _name = "city.zip.geonames.import"
     _description = "Import City Zips from Geonames"
-    _rec_name = "country_id"
 
-    country_id = fields.Many2one("res.country", "Country", required=True)
-    code_row_index = fields.Integer(
```
pedrobaeza, line 31: These changes for example

manuelregidor: Hi! Thank you for your answer. I removed the related fields code_row_index and name_row_index because country_id field has been converted into a Many2many field (it was a Many2one field before the changes I applied). That is also why _rec_name has been removed.

pedrobaeza: OK, understood. You're right.

manuelregidor:
Hi @pedrobaeza! I have applied some changes to the implementation and created a new test to check whether the single country import system still works under the new funcionality. Please, could you let me know if the changes performed solve the issues you pointed out? Thank you.

base_location_geonames_import/__manifest__.py
```diff
@@ -8,7 +8,7 @@
 
 {
     "name": "Base Location Geonames Import",
-    "version": "13.0.1.0.2",
+    "version": "13.0.1.0.1",
```
pedrobaeza: Don't change this version number

base_location_geonames_import/README.rst
```diff
@@ -23,7 +23,7 @@ Base Location Geonames Import
     :target: https://runbot.odoo-community.org/runbot/134/13.0
     :alt: Try me on Runbot
 
-|badge1| |badge2| |badge3| |badge4| |badge5| 
+|badge1| |badge2| |badge3| |badge4| |badge5|
```
pedrobaeza: Leave this file untouched or regenerate it with the tool.

base_location_geonames_import/tests/test_base_location_geonames_import.py
```diff
@@ -166,3 +176,93 @@ def test_import_duplicated_city_name(self):
...
+        max_import = 10
+        self.wizard_2.with_context(max_import=max_import).run_import()
+        # Look if there are imported states for the country
+        state_country_2_count = self.env["res.country.state"].search_count(
+            [("country_id", "=", self.country_2.id)]
+        )
+        state_country_3_count = self.env["res.country.state"].search_count(
+            [("country_id", "=", self.country_3.id)]
+        )
+        self.assertTrue(state_country_2_count)
```
pedrobaeza: Why these asserts? You have to test features of the module, not other things. Here you are asserting the previous state of the test DB, which is not desirable. And focus the test to make sure that cities in multiple imported countries are imported, not in other things.

base_location_geonames_import/tests/test_base_location_geonames_import.py
```diff
@@ -166,3 +176,93 @@ def test_import_duplicated_city_name(self):
...
+        self.assertEqual(city_country_2_count, city_2_count2)
+        self.assertEqual(city_country_3_count, city_3_count2)
+
+        zip_count = self.env["res.city.zip"].search_count(
+            [("city_id.country_id", "in", (self.country_2.id, self.country_3.id))]
+        )
+        self.assertEqual(zip_count, 20)
+
+    def test_import_country_single_value(self):
+        max_import = 5
```
pedrobaeza: No need to add this test, and more with the same invalid asserts than the previous. For checking this backward compatibility, just don't change in https://github.com/OCA/partner-contact/pull/884/files#diff-1718593fd8e26e4fec3112ad5a4657c4R120 the assigned value and everything should work the same.

base_location_geonames_import/wizard/geonames_import.py
```diff
@@ -79,8 +76,8 @@ def select_city(self, row, country, state_id):
             "WHERE name = %s AND country_id = %s AND state_id = %s LIMIT 1",
             (self.transform_city_name(row[2], country), country.id, state_id),
         )
-        row_city = self.env.cr.fetchone()
-        return (row_city[0], row_city[1]) if row_city else (False, False)
+        row = self.env.cr.fetchone()
```
pedrobaeza: Don't undo this change

pedrobaeza (changes requested):
Thanks for the efforts you are doing and the work for uncoupling the methods from the model. I request one last effort in the inline comments, and please also squash all your commits in one.

base_location_geonames_import/tests/test_base_location_geonames_import.py
manuelregidor: Sorry, I'm not sure if I've understood what you mean. country_2 and country_3 are only used at this test, which follows the steps followed by test_import_country. Should I do the test with more than two countries and focus just on the imported cities (and not the states and zips related to each country)? Thank you.

pedrobaeza: OK, I see, I got confused with so many counts. No need to test each object, as checking that there are zips, it's enough to test that the import has worked in this case. Also put `limit=1` for better performance.

HaraldPanten (approved):
Tested in local environment and it's working fine. Functional review 👍

ValentinVinagre (approved):

manuelregidor:
@pedrobaeza  is this PR ready to merge for you? Thank you.

base_location_geonames_import/wizard/geonames_import.py
```diff
@@ -25,13 +25,10 @@
 class CityZipGeonamesImport(models.TransientModel):
     _name = "city.zip.geonames.import"
     _description = "Import City Zips from Geonames"
-    _rec_name = "country_id"
```
pedrobaeza, line 28: If the model doesn't have `name` field, it requires a  `_rec_name` of an existing field. If you don't want `country_id`, then put `country_ids`

manuelregidor: @pedrobaeza Thank you for your feedback. The change has been applied.

pedrobaeza (approved):

PR #886, opened 2020-04-09
Fix symmetric tab - copy left tab id to right

Rad0van:
This fixes setting up tab on symmetric relations because original method only handles properties having names ending with '_left' and we need to extend it to tab_left_id.

Hope this is final attempt to do this right. Thanx @NL66278 for help.


NL66278 (approved):
:+1: Thanks for the fix! LGTM. Code review.

Rad0van:
>  Thanks for the fix! LGTM. Code review.

Thanx. Will study git(hub) more to do better work next time on bigger changes.

PR #891, opened 2020-04-26
[12.0][IMP] partner_multi_relation_tabs: improve res.partner browse performance

skeller1:
**Description of the issue/feature this PR addresses:**

Restrict partner relation tab loading to contact forms

**Current behavior before PR:**

- adding the partner_relation_tab xml code using the ORM `browse` method in `res.partner` model leads to performance problems on the whole system (browsing res.partner records is everywhere in Odoo)
- pg_stats_statement view after creating some sales orders (without using the contact form, there are a lot of tab related sql queries)
pg_stats_statement_tab

```sql
SELECT "res_partner_tab".id FROM "res_partner_tab" LEFT JOIN 
                (SELECT res_id, value FROM "ir_translation"
                 WHERE type=$1 AND name=$2 AND lang=$3 AND value!=$4)
             as "res_partner_tab__name" ON ("res_partner_tab"."id" = "res_partner_tab__name"."res_id") ORDER BY COALESCE("res_partner_tab__name"."value", "res_partner_tab"."name")
```

**Desired behavior after PR is merged:**

- use context `update_relation_tab: 1` to restrict loading of the partner relations only on related contact forms
- pg_stats_statement view after applied changes in code(no tab related sql queries)
pg_stats_statement_tab_after

### DIFF
--- a/partner_multi_relation_tabs/models/res_partner.py
+++ b/partner_multi_relation_tabs/models/res_partner.py
@@ -15,11 +15,12 @@ class ResPartner(models.Model):
 
     @api.multi
     def browse(self, arg=None, prefetch=None):
-        for tab in self._get_tabs():
-            fieldname = tab.get_fieldname()
-            if fieldname not in self._fields:
-                # Check this for performance reasons.
-                self.add_field(tab)
+        if "update_relation_tab" in self.env.context:
+            for tab in self._get_tabs():
+                fieldname = tab.get_fieldname()
+                if fieldname not in self._fields:
+                    # Check this for performance reasons.
+                    self.add_field(tab)
         return super(ResPartner, self).browse(arg=arg, prefetch=prefetch)
 
     @api.model
@@ -71,10 +72,11 @@ class ResPartner(models.Model):
     @api.depends('is_company', 'category_id')
     def _compute_tabs_visibility(self):
         """Compute for all tabs wether they should be visible."""
-        for tab in self._get_tabs():  # get all tabs
-            for this in self:
-                this[tab.get_visible_fieldname()] = \
-                    tab.compute_visibility(this)
+        if "update_relation_tab" in self.env.context:
+            for tab in self._get_tabs():  # get all tabs
+                for this in self:
+                    this[tab.get_visible_fieldname()] = \
+                        tab.compute_visibility(this)
 
     def _get_tabs(self):
         tab_model = self.env['res.partner.tab']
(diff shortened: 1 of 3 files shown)


partner_multi_relation_tabs/models/res_partner.py
```diff
@@ -71,10 +72,11 @@ def add_invisible_extra_field(view, extra_fields, fieldname):
     @api.depends('is_company', 'category_id')
     def _compute_tabs_visibility(self):
         """Compute for all tabs wether they should be visible."""
-        for tab in self._get_tabs():  # get all tabs
-            for this in self:
-                this[tab.get_visible_fieldname()] = \
-                    tab.compute_visibility(this)
+        if "show_relation_tab" in self.env.context:
```
NL66278: I think this change is not really needed (and therefore the change in test also not), because the change in the browse method already prevents this from being called unneeded.

partner_multi_relation_tabs/models/res_partner.py
```diff
@@ -15,11 +15,12 @@ class ResPartner(models.Model):
 
     @api.multi
     def browse(self, arg=None, prefetch=None):
-        for tab in self._get_tabs():
-            fieldname = tab.get_fieldname()
-            if fieldname not in self._fields:
-                # Check this for performance reasons.
-                self.add_field(tab)
+        if "show_relation_tab" in self.env.context:
```
NL66278: I think the context key should perhaps have a name like 'update_relation_tab'. The browse record could also be used programmatically to get at the partners that are related, in a way that places them on a tab. For instance suppose you have a "board tab" that lists chairperson, secretary and member relations for an organisation. You might use the partners on the board tab to send all of them an email.

NL66278 (changes requested):
Thank you for your contribution. As you have shown this can solve an important performance problem. Nevertheless I have a few small questions and remarks. See the previous comments.

partner_multi_relation_tabs/models/res_partner.py
skeller1: oh yes, removed it, thank you

partner_multi_relation_tabs/models/res_partner.py
skeller1: Changed context name it to `update_relation_tab`

skeller1:
Okay, if I remove the context check from the `_compute_tabs_visibility` method, the test fails because fields are not loaded before in browse. So I need to add the context to the test. Is this correct?

skeller1:
So the compute method is context dependent, this could be a perfect use case for Migration to Odoo 13 [@api.depends_context('update_relation_tab')](https://github.com/odoo/odoo/blob/13.0/odoo/api.py#L209)

NL66278 (approved):
:+1: LGTM

skeller1:
@NL66278: something to do: force push?

NL66278:
@skeller1 Try the force push. If it has no effect I will merge manually.

NL66278:
Going to merge manually, because of inexplicable failure in ocabot

PR #896, opened 2020-04-28
[12.0][IMP] partner_multi_relation: unlinking partner relations

skeller1:
**Description of the issue/feature this PR addresses:**

Allow unlinking of partner relation recordsets containing records with same res_id

**Current behavior before PR:**

1. Go to Contacts => Relations => Relations
2. Filter for Partner `Bart` (2 records)
3. Select both records in tree view (global checkbox)
4. Delete records
5. Error appears 

partner_relation_remove_relations

**Desired behavior after PR is merged:**

- unlink all checked partner relations in tree view without any error

### DIFF
--- a/partner_multi_relation/models/res_partner_relation_all.py
+++ b/partner_multi_relation/models/res_partner_relation_all.py
@@ -7,7 +7,7 @@ import logging
 from psycopg2.extensions import AsIs
 
 from odoo import _, api, fields, models
-from odoo.exceptions import ValidationError
+from odoo.exceptions import MissingError, ValidationError
 from odoo.tools import drop_view_if_exists
 
 
@@ -466,6 +466,9 @@ CREATE OR REPLACE VIEW %%(table)s AS
         """For model 'res.partner.relation' call unlink on underlying model.
         """
         for rec in self:
-            base_resource = rec.get_base_resource()
+            try:
+                base_resource = rec.get_base_resource()
+            except MissingError:
+                continue
             rec.unlink_resource(base_resource)
         return True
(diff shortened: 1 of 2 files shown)


NL66278 (changes requested):
Hi, while this change will work in this particular module, we also have custom extention modules that show for instance the relation between partner and contact as a relation between two partners. So in that case the underlying model is not res.partner.relation.

See here, if you are interested: https://github.com/NL66278/partner-contact/tree/10.0-partner_multi_relation_parent-turbo/partner_multi_relation_parent


I think the most safe and compatible way to prevent the error would be to keep a dictionary in the unlink method of base resources already seen/unlinked, and skip the unlink of the base-resource if it has already been handled.

skeller1:
@NL66278: Thank you for improving my understanding of this addon and the relation.

I started to implement it and I struggled with Cache Miss Error.
Finally I solved the problem using the rec.id field to extract the res_id.
The id of res.partner.relation.all records is build based on the res_id (see screenshot below, line 2 and 13 of the VIEW definition):
UNION_RELATION

At the end ugly code but the tests are green. :-D.

NL66278 (approved):
:+1: Nice solution!

skeller1:
@NL66278

Could this implementation be failure-prone if the `key_offset` (defined [here](https://github.com/OCA/partner-contact/blob/61ab3fc59a405b2ab90a200fb46465d3a0975be1/partner_multi_relation/models/res_partner_relation_all.py#L20)) has more than 1 digit. Is this possible through other extension addons ....
Still need to learn .....

Thank you

NL66278 (approved):
:+1: Probably the most foolproof method to solve the problem.

skeller1:
> +1 Probably the most foolproof method to solve the problem.

Readable and easy to understand. A long brain evolution by my side. Thank you for your support

rousseldenis (approved):
Nice!

PR #901, opened 2020-05-06
[IMP] partner_contact_birthday: Add computed age field

ajayjaviya:
Add field age to compute based on partner's birthdate
age

PR #902, opened 2020-05-06
[ADD] new module : partner_contact_age_range

ajayjaviya:
Add field age range compute automatically based on partner age and configured range

age_range

based on PR #901


partner_contact_age_range/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,7 @@
+This module allows to set up age ranges to categorize partners.
+
+The date range is computed based on the partner's age.
+
+The cron job will auto compute age range.
+
+.. image:: /partner_contact_age_range/static/description/age_range.png
```
ivantodorovich:
```suggestion
.. image:: ../static/description/age_range.png
```

partner_contact_age_range/readme/USAGE.rst
```diff
@@ -0,0 +1,7 @@
+Go to Contact / Configuration / Age Ranges
+
+Configure your Age Ranges
+
+.. image:: /partner_contact_age_range/static/description/configure_age_range.png
```
ivantodorovich:
```suggestion
.. image:: ../static/description/configure_age_range.png
```

partner_contact_age_range/models/res_partner_age_range.py
```diff
@@ -0,0 +1,37 @@
...
+                    % (rec.name, rec.age_from, rec.age_to)
+                )
+            range_id = rec.search(
+                [
+                    ("age_from", "<=", rec.age_to),
+                    ("age_to", ">", rec.age_from),
+                    ("id", "!=", rec.id),
+                ],
+                limit=1,
+            )
```
ivantodorovich, line 44: This constraint shouldn't allow this either:
image

A res.partner with age=45 would raise an error:

```
ValueError: Expected singleton: res.partner.age.range(4, 6, 7)
```

ivantodorovich (changes requested):
Thanks!

A few comments

partner_contact_age_range/__init__.py
```diff
@@ -0,0 +1,4 @@
+# Copyright (C) 2019-Today: Druidoo (<https://www.druidoo.io>)
```
NL66278: Please do not put copyright lines in __init__ files.

partner_contact_age_range/models/res_partner_age_range.py
```diff
@@ -0,0 +1,37 @@
...
+from odoo import _, api, fields, models
+from odoo.exceptions import ValidationError
+
+
+class ResPartnerDateRange(models.Model):
+    _name = "res.partner.age.range"
+    _description = "Partner Age Range"
+
+    name = fields.Char(string="Name", required=True)
+    age_from = fields.Integer(string="From", required=True)
```
NL66278: To make sure all age ranges are always covered, you could make age_from a computed field. So you could add Baby with age_to = 2, Toddler with age_to = 4, Child with age_to = 12... etc. Or any other division you want to make. age_from would be the age_to of the range with a lower age_to, or zero if there is no such record. This would also make the check for overlap unneeded, as there can not be any overlap if age_to must be unique. This would also give the user less work to maintain the ranges.

partner_contact_age_range/models/res_partner.py
```diff
@@ -0,0 +1,34 @@
...
+        "res.partner.age.range",
+        "Age Range",
+        compute="_compute_age_range_id",
+        store=True,
+    )
+
+    @api.depends("age")
+    def _compute_age_range_id(self):
+        age_ranges = self.env["res.partner.age.range"].search([])
+        for record in self:
```
NL66278, line 20: Put in a continue if there is no age (because there might not be a birthdate).

partner_contact_age_range/models/res_partner.py
```diff
@@ -0,0 +1,34 @@
...
+            age_range = age_ranges.filtered(
+                lambda age_range: age_range.age_from <= record.age <= age_range.age_to
+            )
+            if age_range:
+                record.age_range_id = age_range.id
```
NL66278: Only do the assignment if the range changes. Otherwise you will have lots of database updates where in fact nothing changes (except the write_date and wite_uid).

NL66278 (changes requested):
Nice functionality but a few changes would make it even better.

partner_contact_age_range/models/res_partner_age_range.py
ajayjaviya: @NL66278 : what will be the value for last range for eg. age for 2 Toddler , age 4 for child , 12 for young is the last range then for young what will be the value for to ?

partner_contact_age_range/models/res_partner.py
ajayjaviya: @NL66278 done

partner_contact_age_range/models/res_partner.py
ajayjaviya: @NL66278 done

partner_contact_age_range/__init__.py
ajayjaviya: @NL66278 : remove copyright lines from init files

partner_contact_age_range/models/res_partner_age_range.py
ajayjaviya: @ivantodorovich : I have fix constraint

partner_contact_age_range/readme/USAGE.rst
ajayjaviya: @ivantodorovich :done

partner_contact_age_range/readme/DESCRIPTION.rst
ajayjaviya: @ivantodorovich done

partner_contact_age_range/__manifest__.py
```diff
@@ -0,0 +1,20 @@
...
+    "name": "Contact's Age Range",
+    "version": "13.0.1.0.0",
+    "license": "AGPL-3",
+    "author": "Druidoo, Odoo Community Association (OCA)",
+    "website": "https://github.com/OCA/partner-contact",
+    "category": "Customer Relationship Management",
+    "summary": "Age field for Contact's",
+    "depends": ["partner_contact_birthdate"],
+    "data": [
+        "security/ir.model.access.csv",
```
rousseldenis, line 14: I'd rather do an xml to help inheritance.

partner_contact_age_range/models/res_partner_age_range.py
NL66278: So for Baby the computed age_from would be 0 (as there is no age range with a lesser age_to), for Toddler it would be 3 (the age_to from Baby + 1), for Child it would be 5 (age_to from Toddler +1) etc. For Adult, or for any other oldest age group you want to have, like Pensioner, you could just put a very large age_to, like 128 or if you like 969: https://en.wikipedia.org/wiki/Methuselah

partner_contact_age_range/models/res_partner.py
ivantodorovich: Actually, if there's no age we need to set the age_range_id to False, otherwise Odoo will not update the previous value

ie: let's say you remove the birthday of a partner, you expect the age_range to be removed as well

NL66278: @ajayjaviya @ivantodorovich is right. So in sum this method should be IMHO like this:
```
    @api.depends("age")
    def _compute_age_range_id(self):
        age_ranges = self.env["res.partner.age.range"].search([])
        for record in self:
            if not record.age:
                age_range = False
            else:
                age_range = age_ranges.filtered(
                    lambda age_range: age_range.age_from <= record.age <= age_range.age_to
                ) or False
            if record.age_range_id != age_range:
                record.age_range_id = age_range.id
```

ivantodorovich: btw @NL66278 , are you completely sure that we need to check if the value has changed before assignation?
I mean, isn't the ORM smart enough to avoid writing on the record if the field hasn't actually changed?

I'm asking because I honestly don't know :sweat_smile:

NL66278: @ivantodorovich You can check it in a shell session (I use bin/python_odoo from buildout):
```
>>> partner = partner_model.browse(8)
>>> partner.write_date
'2020-02-04 09:04:37'
>>> partner.customer
True
>>> partner.customer = True
>>> partner.write_date
'2020-05-07 11:58:27'
```

ajayjaviya: @NL66278  and @ivantodorovich  done

partner_contact_age_range/models/res_partner_age_range.py
ajayjaviya: @NL66278 : I have add default get to auto set age_from

partner_contact_age_range/__manifest__.py
ajayjaviya: @rousseldenis : we can also inherit this for eg.

access record in module:

`access_res_partner_age_range_user,access_res_partner_age_range_user,model_res_partner_age_range,base.group_user,1,0,0,0`

inherit it in other module:
`
partner_contact_age_range.access_res_partner_age_range_user,access_res_partner_age_range_user,model_res_partner_age_range,base.group_user,1,1,0,0`

ivantodorovich:
@ajayjaviya now that https://github.com/OCA/partner-contact/pull/901 is merged, can you rebase on 13.0?

partner_contact_age_range/__init__.py
```diff
@@ -0,0 +1,3 @@
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
```
NL66278: Make all license lines https:

partner_contact_age_range/models/res_partner.py
```diff
@@ -0,0 +1,42 @@
+# Copyright (C) 2019-Today: Druidoo (<https://www.druidoo.io>)
```
NL66278: get rid of the `(C)` here as well and replace Today:  with 2020

ajayjaviya:
> @ajayjaviya now that #901 is merged, can you rebase on 13.0?

@ivantodorovich done

partner_contact_age_range/models/res_partner.py
ajayjaviya: @NL66278 done

partner_contact_age_range/__init__.py
NL66278: This still needs to be taken up. Then we are good to go.

ivantodorovich (approved):

partner_contact_age_range/__init__.py
ajayjaviya: > This still needs to be taken up. Then we are good to go.

@NL66278 done

NL66278 (approved):
:+1: LGTM Thank you for your contribution.

ivantodorovich:
can we merge it?

ivantodorovich:
Apparently if failed with unrelated error:
image


would you please try again? :)

NL66278:
Merging this manually because of failing ocabot.

Issue #905, opened 2020-05-07
[10.0] base_location_geonames_import: Fix url


rousseldenis:
Tests are failing since all countries are included in one zip file.

https://github.com/OCA/partner-contact/blob/13.0/base_location_geonames_import/wizard/geonames_import.py#L116

For tests, IMHO we should not rely on external urls. Apart of that, I see they have WebServices, why not using them ?

@pedrobaeza 

EDIT: The base url is now https://download.geonames.org/export/dump/ I'll make a PR for that

rousseldenis:
It seems not to be sufficent as the format appears to have changed too...

pedrobaeza:
As explained in https://github.com/OCA/partner-contact/pull/906#issuecomment-625209511, this is a temporary glitch that Geonames suffers from time to time. The only solution is to wait for the fix, although I know it's an annoyance, it only happens once in 3 months or so, and it takes hours to be restored.

The format in dump doesn't serve, as it doesn't include enough information for all (if I don't remember bad, zipcodes are not expanded). Any way, it can suffer the same from temporary glitches.

Others have suggested to remove the online tests or mock them, but the module is just for importing information from Geonames, so it's totally suitable to test if that download and parsing can be done. If you mock or not tested at all, the important thing is that at the end, the module doesn't serve if you click on import and you get error, or Geonames decides to change the format.

About the webservice, last time I check was a paid service, but maybe that has changed and can be implemented that way. Can you please check?

pedrobaeza:
And Geonames works again perfectly after the less than 1-day glitch...

PR #906, opened 2020-05-07
[10.0][FIX] base_location_geonames_import: Fix url

rousseldenis:
Fix url and iso country code position

### DIFF
--- a/base_location_geonames_import/wizard/geonames_import.py
+++ b/base_location_geonames_import/wizard/geonames_import.py
@@ -124,7 +124,7 @@ class BetterZipGeonamesImport(models.TransientModel):
         country_code = self.country_id.code
         config_url = self.env['ir.config_parameter'].get_param(
             'geonames.url',
-            default='http://download.geonames.org/export/zip/%s.zip')
+            default='https://download.geonames.org/export/dump/%s.zip')
         url = config_url % country_code
         logger.info('Starting to download %s' % url)
         res_request = requests.get(url)
--- a/base_location_geonames_import/wizard/geonames_import.py
+++ b/base_location_geonames_import/wizard/geonames_import.py
@@ -66,7 +66,7 @@ class BetterZipGeonamesImport(models.TransientModel):
 
     @api.model
     def create_better_zip(self, row, country):
-        if row[0] != country.code:
+        if row[8] != country.code:
             raise UserError(
                 _("The country code inside the file (%s) doesn't "
                     "correspond to the selected country (%s).")


rousseldenis:
fixes #905

pedrobaeza:
URL is correct. The one you are mentioning has other different format. The problem is that from time to time, Geonames screws up its export scripts, and they are not available in some hours. It's an annoyance, but tests should be that way. I will expand more in the issue you have opened, but for now, let's close this.

rousseldenis:
> URL is correct. The one you are mentioning has other different format. The problem is that from time to time, Geonames screws up its export scripts, and they are not available in some hours. It's an annoyance, but tests should be that way. I will expand more in the issue you have opened, but for now, let's close this.

That's why we shouldn't rely on external url for tests. So, if you don't want this, please fix tests as main branch should go green

pedrobaeza:
Please read my full explanation, and no, tests can't be green as this is failing, although it relies in something external.

rousseldenis:
> Please read my full explanation, and no, tests can't be green as this is failing, although it relies in something external.

I did. Unit tests should be run in any cases as tests should be reproductible in any case. We should either mock requests library and provide a fake zip file...

See: https://en.wikipedia.org/wiki/Unit_testing

> To isolate issues that may arise, each test case should be tested independently. Substitutes such as method stubs, mock objects,[4] fakes, and test harnesses can be used to assist testing a module in isolation.

pedrobaeza:
I don't agree. Who says these are unit tests? These are tests for being sure the module works. Simply. And this is want I want to know: that today I'm not able to use this module.

rousseldenis:
> I don't agree. Who says these are unit tests? These are tests for being sure the module works. Simply. And this is want I want to know: that today I'm not able to use this module.

So, if geonames never fix it, we cannot run tests anymore on this repo... let me laugh

pedrobaeza:
No, we should fix the module for changing the new way the module should work. Imagine Geonames closes and this is no more available. What a mocked test serves? The module is useless, and then it should deactivated, or switch to other service, etc. The same applies if Geonames changes the format.

This temporary glitch is that, a temporary annoyance. If tomorrow this continues the same, then we can decide to do other action (like temporary excluding it in tests, or other thing), but the important thing is that test is serving its purpose: the feature of the module is not working (no matter what is the reason why it doesn't work).

This branch 10.0 is 3 years old and if this is the first time you notice this, this is not happening very frequent, isn't it?

rousseldenis:
> Who says these are unit tests?

https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#tests

NL66278:
I have seen this breaking before. I understand @pedrobaeza in the tests not making sense if no real connection is made. I can also understand @rousseldenis problem with having PR's randomly failing tests because some external website not being available. Would it be an option to move this module - at some stage - to its own repository?

rousseldenis:
> I have seen this breaking before. I understand @pedrobaeza in the tests not making sense if no real connection is made. I can also understand @rousseldenis problem with having PR's randomly failing tests because some external website not being available. Would it be an option to move this module - at some stage - to its own repository?

The general idea with tests is that they should run in isolated environment. You test your code, not the presence of the file... That's the responsability of the external provider(aka geonames) not yours. If you move the module to another repo, the problem should occur if there are pending PR's like today.

lmignon:
It's no the responsibility of OCA to check that the geoname server is available and working properly... Our responsibility is to check that our code works properly with the informations (the file) produced by the geoname service.  ~~IMO, the @rousseldenis proposal makes sens and should not be rejected.~~
IMO We should mock the call to provide the result expected by the call.

Mocking the call to an external resource help to improve the tests coverage since we can check how our code works if the external resources is unavailable or if the format received is not the one expected or ....

rousseldenis:
ocabot merge blocked due to geonames... What a pity

pedrobaeza:
https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#tests doesn't state anything about that needs to be what you call "unit tests". Any way, as said, you think about the mocking, and I don't want that as said. Just wait a bit and if tomorrow continues the same, then we take another decision. Can you wait 24 hours?

PR #912, opened 2020-05-11
[13.0] [MIG] partner_multi_relation

Rad0van:
This is my attempt to bring this to 13.0 However I still need some assistance because I wasn't able to solve these things:

- colors attribute in views/res_partner_relation_all.xml: ` <record id="tree_res_partner_relation_all" model="ir.ui.view">
            <field name="model">res.partner.relation.all</field>
                <tree
                    string="Partner Relations"
                    colors="gray:not active; blue:date_start &gt; current_date"
                    editable="top"
                    >`
- `with self.env.do_in_draft():` used in tests/test_partner_relation.py and tests/test_partner_relation_all.py

First one I guess only causes items not to get colored, the second one causes 2 tests to fail. With my limited Python / Odoo knowledge I couldn't find equivalents to these 2 so any help would be appreciated.


pedrobaeza:
Thanks for the contribution. Please preserve commit history following technical method explained in https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-13.0

NL66278:
What @pedrobaeza says. And then you should have one commit for black/isort to bring this up to date with present day conventions. Then add any needed changes specifically for 13.0 in the next commit.

Rad0van:
@pedrobaeza and @NL66278 wish I knew about the document before. What exactly should I do now? Cancel the PR and do that according to the document? If not, what are the steps?

pedrobaeza:
I suggest to start from scratch, but preserving current migration. Then, perform until last step of migrating the code to v13, and paste on the result your migrated module, committing only the difference.

PR #914, opened 2020-05-15
[13.0][ADD] partner_customer_supplier

Laurent-Corron:
A simple module that adds a customer and supplier field to res_partner


Laurent-Corron:
cc @ThomasBinsfeld

pedrobaeza:
Why not using customer_rank/supplier_rank?

ThomasBinsfeld:
> Why not using customer_rank/supplier_rank?

Boolean fields allow a strong definition of customer/supplier that `customer_rank` and `supplier_rank` don't. We may want having these fields to define restrictions on some `res.partner` fields or create record rules.

In the process of some migrations, we may want to keep these fields.

pedrobaeza:
As you prefer. I think it's swimming against current, but if you want to perpetuate that old classification...

ValentinVinagre:
I agree with @pedrobaeza . IMO this fields in < v.13, filtered the partners in different views and domain fields. If you add these fields again, they should affect customer_rank/supplier_rank fields to allow the classification used by Odoo in >= v.13

NL66278:
I do not quite see any need for this. It is already perfectly possible to search for customers or suppliers by checking for rank field > 0. Same tests can be done in record rules. What I am missing in partner form is easy way to set this field, but maybe I overlooked something.

PR #916, opened 2020-05-17
[13.0][MIG] partner_industry_secondary: Migration to 13.0

skeller1:
**Standard Migration from partner_industry_secondary to 13.0**

Changes/Improvement:
- [x] add uniq name constraint per parent (removed it from roadmap in docs)
- [x] allow copying of industries
- [x] add tests for copy and uniq name constraint per parent

IndustrySecondarySettings


pedrobaeza:
The problem with storing `complete_name` is that the translations are not effective.

skeller1:
> The problem with storing `complete_name` is that the translations are not effective.

Ok, understand this and changed back to the old implementation, thank you

HaraldPanten (approved):
Functional review 👍

skeller1:
add tests for copy and uniq name constraint per parent (open roadmap entry in docs)

tonygalmiche (approved):
+1 LGTM

mamcode (approved):
LGTM. @skeller1  I think you should be added to the list of contributors.

rousseldenis:
@NL66278 As I've already said, don't merge directly. Use ocabot. As you can see, module is not published on pypi : https://pypi.org/search/?q=odoo13-addon-partner-industry&o=

If you don't rely on it, maybe we could improve it or if you have questions about it, maybe I can answer. Thanks

PR #930, opened 2020-06-23
[11.0][IMP] partner_vat_unique: More rigorous check

pedrobaeza:
Two cases are not covered:

- If the partner with duplicated VAT is inactive
- If there are specific contact record rules that hides some of the results for that user.

We cover both with this modification.

@Tecnativa TT24455

### DIFF
--- a/partner_vat_unique/tests/test_vat_unique.py
+++ b/partner_vat_unique/tests/test_vat_unique.py
@@ -21,6 +21,14 @@ class TestVatUnique(SavepointCase):
                 'vat': 'ESA12345674'
             })
 
+    def test_duplicated_vat_creation_inactive(self):
+        self.partner.active = False
+        with self.assertRaises(ValidationError):
+            self.env['res.partner'].with_context(test_vat=True).create({
+                'name': 'Second partner',
+                'vat': 'ESA12345674'
+            })
+
     def test_duplicate_partner(self):
         partner_copied = self.partner.copy()
         self.assertFalse(partner_copied.vat)
--- a/partner_vat_unique/models/res_partner.py
+++ b/partner_vat_unique/models/res_partner.py
@@ -1,4 +1,5 @@
 # Copyright 2017 Grant Thornton Spain - Ismael Calvo <ismael.calvo@es.gt.com>
+# Copyright 2020 Tecnativa - Pedro M. Baeza
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
 from odoo import _, api, fields, models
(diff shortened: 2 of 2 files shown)


rafaelbn (approved):
Thanks :+1:

partner_vat_unique/tests/test_vat_unique.py
```diff
@@ -14,13 +14,21 @@ def setUpClass(cls):
             'vat': 'ESA12345674'
         })
 
-    def test_duplicated_vat_creation(self):
+    def _test_duplicated_vat_creation(self):
```
ernestotejeda: I can't see this method being used

partner_vat_unique/tests/test_vat_unique.py
```diff
@@ -14,13 +14,21 @@ def setUpClass(cls):
...
-    def test_duplicate_partner(self):
+    def test_duplicated_vat_creation_inactive(self):
+        self.partner.active = False
+        with self.assertRaises(ValidationError):
+            self.env['res.partner'].with_context(test_vat=True).create({
+                'name': 'Second partner',
+                'vat': 'ESA12345674'
+            })
+
+    def _test_duplicate_partner(self):
```
ernestotejeda: Neither this one

pedrobaeza:
Indeed I forgot to re-enable them. Fixed.

ernestotejeda (approved):

rousseldenis:
> Two cases are not covered:
> 
> * If the partner with duplicated VAT is inactive
> * If there are specific contact record rules that hides some of the results for that user.
> 
> We cover both with this modification.
> 
> @Tecnativa TT24455

In which case the inactive partner is bothering you ?

chienandalu (approved):

NL66278 (approved):
:+1: Code review. LGTM.

pedrobaeza:
> In which case the inactive partner is bothering you ?

@rousseldenis for example, for this use case:

- You archive a contact, thinking you are not going to work with that partner anymore.
- Sometime has past, and you work again with it.
- You try to create a new contact with that VAT.
- Without this PR, you will be able to create it, and have the partner history split in two.
- With the PR, you get the warning and then look for the other contact and enable it instead.

Imagine the same if you archive it by mistake.

I considered to change the error message to include the part (`maybe is it archived?`) or similar for giving a hint, but I didn't want to change the translations. Do you think it should be done?

rousseldenis:
> > In which case the inactive partner is bothering you ?
> 
> @rousseldenis for example, for this use case:
> 
> * You archive a contact, thinking you are not going to work with that partner anymore.
> * Sometime has past, and you work again with it.
> * You try to create a new contact with that VAT.
> * Without this PR, you will be able to create it, and have the partner history split in two.
> * With the PR, you get the warning and then look for the other contact and enable it instead.
> 
> Imagine the same if you archive it by mistake.
> 
> I considered to change the error message to include the part (`maybe is it archived?`) or similar for giving a hint, but I didn't want to change the translations. Do you think it should be done?

Ok, I see. Agree with that.

Indeed, IMHO messages should be as clear as possible for user. But understand your fear to change pot file. As you want

rousseldenis (approved):

HaraldPanten (approved):

pedrobaeza:
I have added as a separate translatable string the sentence about the possible archive of the partner, so we get both advantages. Thanks to all reviewers.

/ocabot merge patch

pedrobaeza:
Cherry-picked for v12 in f937e2c7e7712f69ced3210107b920bd822c3725

In v13 we are using the standard check, so we may need to do a PR to Odoo core.

pedrobaeza:
The Odoo PR for reflecting the same (and in v13 multi-company considerations is a new one) is here:

https://github.com/odoo/odoo/pull/53701

Please support me if you see it correct.

PR #934, opened 2020-06-25
[ADD] partner_contact_access_link: reach contact full form

yajo:
This module extends the functionality of the partners form
and to allow you to click on a button and open the full contact form
while watching a company form.

@Tecnativa TT24492 TT24498


chienandalu (approved):
Tested on runbot :+1:

ValentinVinagre (approved):
LGTM 👍  nice functionality

HaraldPanten (approved):
Functional review 👍

pedrobaeza (dismissed):
The chosen icon suggest that the form is going to be opened as popup, which is not true. I prefer the icon used in `web_tree_many2one_clickable`. I also have the doubt about what happen when you click on edit mode.

HaraldPanten:
> I also have the doubt about what happen when you click on edit mode.

Clicking edit mode is working fine.

rafaelbn (approved):
Functionally tested in runbot :+1:

yajo:
@pedrobaeza That icon could be intuitive in list views, but TBH I don't see it fits well in a form context like this. :thinking:  See how it looks:

Captura de pantalla de 2020-06-26 09-38-02

I think the previous one was better. I don't think most users relate that icon specifically to a floating view, but more with the fact that you can click on that icon and see the full details of related records. Would you reconsider this?

pedrobaeza:
Indeed it's not the best looking one, but I prefer that you find another one then, as the icon is clearly used in Odoo for opening in popup, so it's a mix of concepts. How does it look with 2/3 repetitions of the latter icon?

HaraldPanten:
> @pedrobaeza That icon could be intuitive in list views, but TBH I don't see it fits well in a form context like this. 🤔 See how it looks:
> 
> Captura de pantalla de 2020-06-26 09-38-02
> 
> I think the previous one was better. I don't think most users relate that icon specifically to a floating view, but more with the fact that you can click on that icon and see the full details of related records. Would you reconsider this?

IMHO I prefer the last one or a new one instead of this other.

yajo:
OK I found https://fontawesome.com/icons/external-link-square. I like it because it's similar to https://fontawesome.com/icons/external-link-alt but not exactly equal. That's our case: if you click there, you'll get a similar behavior but not exactly the same. It also looks good on the form:

Captura de pantalla de 2020-06-26 10-03-13

pedrobaeza:
Yeah, I also like that one!

HaraldPanten:
Looks good 👍

pedrobaeza (approved):

PR #948, opened 2020-08-16
[12.0][IMP] partner_contact_address_default: allow selecting all partner

skeller1:
By default only partner records with the same commercial_partner_id could be selected as default delivery or invoice address

With this PR this behaviour could be changed per configuration to allow selecting all partner contact records as delivery or invoice address. 

SettingsPartnerDefault
SelectAllPartner

### DIFF
--- a/partner_contact_address_default/models/res_config_settings.py
+++ b/partner_contact_address_default/models/res_config_settings.py
@@ -1,0 +1,12 @@
+from odoo import fields, models
+
+
+class ResConfigSettings(models.TransientModel):
+    _inherit = "res.config.settings"
+
+    group_address_default_allow_all_partner = fields.Boolean(
+        'Allow selecting from all partners as address default',
+        help='Allow selecting from all partners as address default',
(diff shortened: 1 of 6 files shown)


NL66278 (approved):
:+1: LGTM

partner_contact_address_default/models/res_config_settings.py
```diff
@@ -0,0 +1,12 @@
...
+
+
+class ResConfigSettings(models.TransientModel):
+    _inherit = "res.config.settings"
+
+    group_address_default_allow_all_partner = fields.Boolean(
+        'Allow selecting all partners as address default',
+        help='Allow selecting all partners as address default',
+        implied_group='partner_contact_address_default.group_allow_all_partner',
+        default=True,
```
pedrobaeza: This should be False by default for not changing current behavior.

pedrobaeza (changes requested):

pedrobaeza (approved):

PR #952, opened 2020-09-03

hugho-ad:
not firstname

name = 'Van-Eyck Jan'

this was converted to

firstname = False
lastname= 'Van-Eyck'
lastname2 = 'Jan'

and it should be like

firstname = 'Jan'
lastname= 'Van-Eyck'
lastname2 = False


hugho-ad:
replaced by 
https://github.com/OCA/partner-contact/pull/953

PR #953, opened 2020-09-03

hugho-ad:
not firstname

name = 'Van-Eyck Jan'

this was converted to

firstname = False
lastname= 'Van-Eyck'
lastname2 = 'Jan'

and it should be like

firstname = 'Jan'
lastname= 'Van-Eyck'
lastname2 = False

Before:
image

After
image

### DIFF
--- a/partner_second_lastname/models/res_partner.py
+++ b/partner_second_lastname/models/res_partner.py
@@ -69,32 +69,50 @@ class ResPartner(models.Model):
         - If the partner is a company, save it in the lastname.
         - Otherwise, make a guess.
         """
-        # Company name goes to the lastname
         result = {
             'firstname': False,
             'lastname': name or False,
             'lastname2': False,
         }
-        if not is_company and name:
-            order = self._get_names_order()
-            result = super(ResPartner, self)._get_inverse_name(
-                name, is_company)
-            parts = []
-            if order == 'last_first':
-                if result['firstname']:
-                    parts = result['firstname'].split(" ", 1)
-                while len(parts) < 2:
-                    parts.append(False)
-                result['lastname2'] = parts[0]
-                result['firstname'] = parts[1]
-            else:
-                if result['lastname']:
-                    parts = result['lastname'].split(" ", 1)
-                while len(parts) < 2:
-                    parts.append(False)
-                result['lastname'] = parts[0]
-                result['lastname2'] = parts[1]
+
+        # Company name goes to the lastname
+        if not name or is_company:
+            return result
+
+        order = self._get_names_order()
+        result.update(
+            super(ResPartner, self)._get_inverse_name(name, is_company))
+
+        if order in ('first_last', 'last_first_comma'):
+            parts = self._split_part('lastname', result)
+            if parts:
+                result.update({
+                    'lastname': parts[0],
+                    'lastname2': u" ".join(parts[1:]),
+                })
+        else:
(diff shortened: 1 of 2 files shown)


luiseevaquer:
LGTM :+1:

luiseevaquer:
@moylop260 

Could you review this PR, please?

moylop260:
I'm not able to check it

I prefer to request review of authors:
@Yajo @pedrobaeza 
Could you check it, please?

In fact,
@hugho-ad @lescobarvx 
The CI is red

hugho-ad:
@moylop260 
done, CI are green

moylop260:
@Yajo @pedrobaeza 
What do you think?

pedrobaeza (approved):
We are not using anymore this version nor this module, but seems OK

rousseldenis:
I would have added a test that highlight the problem.
@hugho-ad

partner_second_lastname/models/res_partner.py
```diff
@@ -85,8 +85,8 @@ def _get_inverse_name(self, name, is_company=False):
                     parts = result['firstname'].split(" ", 1)
                 while len(parts) < 2:
                     parts.append(False)
-                result['lastname2'] = parts[0]
-                result['firstname'] = parts[1]
+                result['firstname'] = parts[0]
```
yajo: I think this is wrong.

Check line 83: `if order == 'last_first':`. So, I understand lastname goes first with this setting.

marcelsavegnago (approved):

moylop260:
@hugho-ad @WRZ96 @lescobarvx 
Could you check the Jairo's comment, please?

partner_second_lastname/models/res_partner.py
hugho-ad: @Yajo 

could you review again please

WR-96:
@Yajo The method to split the last name have been refactored, could you review this PR again please

Cc: @marcelsavegnago @rousseldenis @pedrobaeza

WR-96:
@Yajo @marcelsavegnago @rousseldenis @pedrobaeza 
Is this PR still approved after the last changes? Could you review it please

moylop260 (approved):
:+1:

pedrobaeza:
@Yajo's concern is not satisfied AFAIK

pedrobaeza:
@Yajo do you see a blocking point for reverting this?

WR-96:
Hi @pedrobaeza, 

@Yajo's concern was about to the assign for the parts of the name according to the `partner_names_order` parameter, he said:

> I think this is wrong.
> Check line 83: if order == 'last_first':. So, I understand lastname goes first with this setting.

This is covered in the [line 93](https://github.com/OCA/partner-contact/pull/953/files#diff-e19e8b2b1ac28fd235743f835a140361R93) of `res_partner.py`, as the last name goes first we need to split the first name which contains the remaining part of the name

i.e. for a name written as 'last_name last_name2 first_name' like 'Wolowitz Joel Howard' when calling `_get_inverse_name` Odoo will return `{'lastname': 'Wolowitz', 'firstname': 'Joel Howard'}` by splitting the first name again we update the dict returned by `_get_inverse_name` which will now contain `{'lastname': 'Wolowitz', 'lastname2': 'Joel',  'firstname': 'Howard'}`

pedrobaeza:
OK, I let him to say if it's correct, as I don't know this.

PR #954, opened 2020-09-05
[FIX] partner_firstname: Avoid error with demo data.

luistorresm:
In the next case:

Module A dependencies:
  - partner_firstname
  - moduleB

If is installed partner_firstname and after moduleB loads a partner with
demo data, this returns the next error:

``"null value in column "name" violates not-null constraint"``

This, because the name is removed from vals, to get the correct name
with the compute method, but with demo data this not works, for this
reason, was improved to not remove the ``name`` in write method when is
the installed the module.

### DIFF
--- a/partner_firstname/models/res_partner.py
+++ b/partner_firstname/models/res_partner.py
@@ -46,7 +46,7 @@ class ResPartner(models.Model):
                     vals[key] = value
 
             # Remove the combined fields
-            if "name" in vals:
+            if "name" in vals and not self._context.get('install_mode'):
                 del vals["name"]
             if "default_name" in context:
                 del context["default_name"]


luiseevaquer:
LGTM :+1:

luiseevaquer (approved):

hugho-ad (approved):

hugho-ad:
LGTM 👍

luistorresm:
@moylop260

Could you help me to merge this PR if you agree, please?

Regards

moylop260:
Could you tag the authors please?

luistorresm:
Hello @alexis-via @pedrobaeza @nbessi 

What do you think about this improvement?

Regards

goliveirab:
LGTM +1

ovnicraft (approved):
LGTM

moylop260:
I didn't understand so well this change
But it has a lot of approvals

Please, if we are damaging something else, please ping us.

pedrobaeza:
Temporary glitch. Please wait until solved. I will scan it and tell you back.

moylop260:
The URL used is not working anymore:
 - http://download.geonames.org/export/zip/MC.zip
Screen Shot 2020-09-09 at 10 04 21

Should it use the following one?:
 - http://www.geonames.org/export/zip/MC.zip

pedrobaeza:
It's a problem that happens eventually with Geonames screwing up the automatic export. Data from `dump` folder doesn't serve, as information is not homogeneous. We only have to wait a bit. This indicates that the module `base_location_geonames_import` is not working right now (which is good to be warned), as someone has asked me through Twitter: https://twitter.com/karlggest/status/1303639589794578432?s=20

PR #955, opened 2020-09-07
[MIG][13.0] partner_ref_unique: Migration to v13.0 

joao-p-marques:
cc @pedrobaeza @Yajo 
TT25339


joao-p-marques:
changes made @pedrobaeza

partner_ref_unique/models/res_partner.py
```diff
@@ -1,38 +1,29 @@
...
-    @api.constrains("ref", "is_company", "company_id", "partner_ref_unique")
+    @api.constrains("ref", "is_company", "company_id")
     def _check_ref(self):
-        for partner in self:
-            mode = partner.partner_ref_unique
-            if partner.ref and (
-                mode == "all" or (mode == "companies" and partner.is_company)
-            ):
+        for partner in self.filtered("ref"):
+            # If the company is not defined, adopt default strategy for user company
```
pedrobaeza: This comment is outdated, you can update it to:
```suggestion
            # If the company is not defined in the partner, take current user company
```

partner_ref_unique/tests/test_res_partner_ref.py
```diff
@@ -9,49 +10,83 @@ class TestResPartnerRefUnique(common.SavepointCase):
...
-        cls.company = cls.env.ref("base.main_company")
         cls.partner_obj = cls.env["res.partner"]
-        cls.partner1 = cls.partner_obj.create({"name": "Partner1",})
-        cls.partner2 = cls.partner_obj.create({"name": "Partner2",})
+        cls.company_obj = cls.env["res.company"]
+        # look for possible already duplicated refs for being resilient
+        groups = cls.partner_obj.read_group([], ["ref"], ["ref"],)
+        for group in groups:
+            if group["ref_count"] > 1 and not group["ref"]:
+                cls.partner_obj.search(group["__domain"]).write({"ref": False})
```
pedrobaeza: We talked about resetting all refs:
```suggestion
        groups = cls.partner_obj.search([("ref", "!=", False)]).write({"ref": False})
```

partner_ref_unique/models/res_partner.py
```diff
@@ -1,38 +1,29 @@
...
+            if mode == "all" or (mode == "companies" and partner.is_company):
                 domain = [
                     ("id", "!=", partner.id),
                     ("ref", "=", partner.ref),
                 ]
                 if mode == "companies":
                     domain.append(("is_company", "=", True))
                 other = self.search(domain)
-
-                # active_test is False when called from
```
pedrobaeza: We haven't analyzed this. @Yajo do you remember why this dirty trick for merging. What is the problem merging?

partner_ref_unique/tests/test_res_partner_ref.py
```diff
@@ -9,49 +10,83 @@ class TestResPartnerRefUnique(common.SavepointCase):
...
+            self.company.partner_ref_unique = "companies"
+        # Let the situation without duplicate refs and apply global condition
+        partner3.unlink()
+        self.company.partner_ref_unique = "companies"
         with self.assertRaises(ValidationError):
             self.partner2.ref = "same_ref"
         with self.assertRaises(ValidationError):
             self.partner_obj.create(
-                {"is_company": True, "name": "other", "ref": "same_ref",}
+                {"name": "other", "ref": "same_ref", "is_company": True}
```
pedrobaeza: You can let this line without changes

partner_ref_unique/tests/test_res_partner_ref.py
```diff
@@ -9,49 +10,83 @@ class TestResPartnerRefUnique(common.SavepointCase):
...
             )
-        # Here there shouldn't be any problem
-        self.partner_obj.create(
```
pedrobaeza: This one is useful to be let as well.

pedrobaeza (changes requested):

pedrobaeza (approved):
Waiting the merge partner confirmation

partner_ref_unique/models/res_company.py
```diff
@@ -16,3 +17,17 @@ class ResCompany(models.Model):
...
+    def write(self, vals):
+        """Launch manually the constraint check in partners as current ORM
+        doesn't trigger the constraint on related fields.
+        """
+        res = super().write(vals)
+        if "partner_ref_unique" in vals:
+            partners = (
+                self.env["res.partner"]
+                .with_context(active_test=False)
+                .search([("company_id", "in", [False] + self.ids)])
```
yajo, line 30: I think this would be better.

```suggestion
                .search([("company_id", "in", self.ids)])
```

Otherwise different companies with different settings would collide one another.

partner_ref_unique/hooks.py
```diff
@@ -0,0 +1,33 @@
+# Copyright 2020 Camptocamp SA
```
pedrobaeza: Oh, we forget this! This can be fully removed and the hook removed in the manifest.

pedrobaeza (changes requested):

partner_ref_unique/models/res_partner.py
yajo: I don't remember, but it makes sense that one wants to merge archived partners too.

I think this code is just trying to avoid the validation if you're merging, maybe because during the merge there'll be a valid situation where 2 partners will share the same ref always? :thinking: 

This context doesn't seem to be the most reliable way of doing that, OTOH.

partner_ref_unique/README.rst
```diff
@@ -0,0 +1,102 @@
...
+.. |badge4| image:: https://img.shields.io/badge/weblate-Translate%20me-F47D42.png
+    :target: https://translation.odoo-community.org/projects/partner-contact-13-0/partner-contact-13-0-partner_ref_unique
+    :alt: Translate me on Weblate
+.. |badge5| image:: https://img.shields.io/badge/runbot-Try%20me-875A7B.png
+    :target: https://runbot.odoo-community.org/runbot/134/13.0
+    :alt: Try me on Runbot
+
+|badge1| |badge2| |badge3| |badge4| |badge5| 
+
+This module allows add an unique constraint to partner ref field.
```
HaraldPanten:
```suggestion
This module allows add a unique constraint to partner ref field.
```

partner_ref_unique/README.rst
```diff
@@ -0,0 +1,102 @@
...
+
+Usage
+=====
+
+#. Go to any partner in **Contacts > Contacts**
+#. Try to create two partners with the same ref.
+
+* If you selcted the option 'All partners' you can't create two partners with the same ref.
+* If you selcted the option 'Only companies' you can't create two companies with the same ref.
+* If you selcted the option 'None' you can create two partners with the same ref.
```
HaraldPanten:
```suggestion
* If you selected the option 'All partners' you can't create two partners with the same ref.
* If you selected the option 'Only companies' you can't create two companies with the same ref.
* If you selected the option 'None' you can create two partners with the same ref.
```

Typo

partner_ref_unique/hooks.py
joao-p-marques: Yes, thank you. Changes already made and pushed.

partner_ref_unique/README.rst
```diff
@@ -0,0 +1,102 @@
...
+
+.. contents::
+   :local:
+
+Configuration
+=============
+
+To configure this module, you need to:
+
+#. Go to **Settings > Companies**.
```
HaraldPanten:
```suggestion
#. Go to **Settings > Users & Companies > Companies**.
```

HaraldPanten (approved):
Functional review. Just some minor picky comments 😝👍

partner_ref_unique/models/res_company.py
pedrobaeza: No, it's correct as it is, as we want to specifically check both. Note that this is not executed with `sudo`, so the search is performed on the user environment (global contacts + current company contacts).

partner_ref_unique/models/res_partner.py
pedrobaeza: That context key was a trick for detecting that you are coming "from a merge operation", so indeed is not correct to check that (you can inject one context key if needed). @joao-p-marques you can check to put one partner with ref, and another without, activate the check, and try to merge one into another (try both ways) to see if there's such problem.

joao-p-marques:
Functional changes made as suggested by @HaraldPanten . Thank you!

partner_ref_unique/README.rst
pedrobaeza: It's not correct this way either:
```
* If you select the option 'None', you can...
```

pedrobaeza (approved):

partner_ref_unique/README.rst
pedrobaeza: It's correct as is, sorry.

partner_ref_unique/models/res_partner.py
pedrobaeza: Indeed there's a problem when merging that we have already handled. See 9a8c078

partner_ref_unique/models/res_company.py
yajo: Are you sure? AFAICS what could happen is that the user logs as company A, sets one policy, company-less contacts pass; then logs as company B, sets a different policy, and company-less contacts fail. :roll_eyes:

yajo (approved):

partner_ref_unique/models/res_company.py
pedrobaeza: I don't get you, but yes, I'm sure. If you don't, please expose detailed steps to reproduce a problem.

yajo: On my way, just hold on.

yajo: Yes, this creates an inconsistent UX:

1. Create Company A, with uniqueness check for all partners; create company B, with None: 
crear compañías
1. Go to contacts, switch to company B, create 2 contacts without company that have the same reference: 
sameref
   No problem because company B is allowed to do that.
1. Switch to company A, go to companies, edit company A, remove uniqueness constraint, add it again. Error: 
error

So, users are able to duplicate refs of partners that have no company if they are in the context of company B, but on under company A. Effectively, this means that the constraint is inconsistent, thus not very useful for companyless partners.

A solution would be to have an ICP for those partners that is consistently enforced across companies, or another solution would be that companyless partners inherit the most strict convention of all present companies.

pedrobaeza: We already thought about ICP, but it's more convoluted. The case you depict can happen the same with the ICP if the combination is not good, but it's not directly caused by including here the `False` part, and not including it will cause a bunch more of incorrect cases, so let's continue with this as is. Think that previously the duplications were worse, and it always force to have a company in the partners to work.

PR #959, opened 2020-09-11
[10.0] partner-contact - update res.country.state name on geonames import

fredzamoabg:
After geonames import was observed that field 'name' in res.country.state table is not changed based the downloaded file information. The module already found a record in DB based 'name', 'city' & 'country_id' from res.better.zip table and 
'country_id' & 'code' from es.country.state and is enough to return it without any check of other fields, particularly the 'name' one.

the modification will update the name field of res.country.state table based the information received from file imported if different from the one returned by DB.

### DIFF
--- a/base_location_geonames_import/__manifest__.py
+++ b/base_location_geonames_import/__manifest__.py
@@ -6,7 +6,7 @@
 
 {
     'name': 'Base Location Geonames Import',
-    'version': '10.0.1.1.1',
+    'version': '10.0.1.1.2',
     'category': 'Partner Management',
     'license': 'AGPL-3',
     'summary': 'Import better zip entries from Geonames',
--- a/base_location_geonames_import/wizard/geonames_import.py
+++ b/base_location_geonames_import/wizard/geonames_import.py
@@ -146,6 +146,7 @@ class BetterZipGeonamesImport(models.TransientModel):
         reader = unicodecsv.reader(data_file, encoding='utf-8', delimiter='	')
         for i, row in enumerate(reader):
             zip_code = self.create_better_zip(row, self.country_id)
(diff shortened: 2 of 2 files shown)


pedrobaeza:
For me, the name shouldn't be updated, as most of that names belongs to XML-ID with noupdate="0", and they will be overwritten next time you update Odoo base module.

fredzamoabg:
> For me, the name shouldn't be updated, as most of that names belongs to XML-ID with noupdate="0", and they will be overwritten next time you update Odoo base module.

Hi Pedro,
thanks for your reply. The modification is required in 10.0

fredzamoabg:
sorry closed by mistake

fredzamoabg:
> For me, the name shouldn't be updated, as most of that names belongs to XML-ID with noupdate="0", and they will be overwritten next time you update Odoo base module.

I will check updating the module, thanks

fredzamoabg:
> > For me, the name shouldn't be updated, as most of that names belongs to XML-ID with noupdate="0", and they will be overwritten next time you update Odoo base module.
> 
> I will check updating the module, thanks

@pedrobaeza I checked res/res.country.state.csv file in base module but I can not find any UK entries on it.

pedrobaeza:
But this module is not only for UK, but for a lot of countries. And I don't get that if you don't have data in Odoo base, now you have them incorrect. If a bad manipulation by your users are performed, this module shouldn't modify that.

fredzamoabg:
> But this module is not only for UK, but for a lot of countries. And I don't get that if you don't have data in Odoo base, now you have them incorrect. If a bad manipulation by your users are performed, this module shouldn't modify that.

ok agree. I will close the PR.

PR #960, opened 2020-09-11
[FIX] geoname import: wizard: fix download url as it changed.

remyz17:
note: this fix should be reported to v10 v11 and v12


pedrobaeza:
No, it hasn't changed. It just that sometimes gets broken (there's already three in a week, which is unusually frequent), but dump doesn't serve as it has another format and not homogeneous on each country. Just wait a bit until fixed.

PR #961, opened 2020-09-21
[12.0][FIX] Update partner tabs for all partner actions.

NL66278:
Existing code only adds tab to partner form, when this is reached through the main Contacts action. But in practice there are many more views that lead to the partner form.


NL66278:
Closed in favor of this one: https://github.com/OCA/partner-contact/pull/979

PR #966, opened 2020-09-28
[13.0][FIX] partner_multi_relation:  Fixing problems with migration of computed fields with search f…

BT-rmartin:
…unction and no computed function defined

Since v13 all computed functions have to assign a value in all cases to the received self records. 
Before, when defined a computed field just in order to have the search function, it was enough with defining the computed function as compute=lambda self: None, but this is not longer allowed since v13. 
This PR is to fix that error in several fields of the partner_multi_relation module in v13

@NL66278 @gurneyalex @pedrobaeza @dufresnedavid


NL66278 (approved):
:+1: Thank you for your contribution. Looks good to me!

BT-cserra (approved):

PR #967, opened 2020-09-29
[12.0] partner noncommercial streamline

NL66278:
- Remove window action / menu to show all contacts, as it duplicates the main Contacts action;
- Make sure general contact configuration menu is shown after added menu's;
- Move some selections from context to domain, to ensure view always reflects its purpose;
- Add some filters / groupings nearly always needed in non commercial settings (but usefull in  general).

### DIFF
--- a/partner_noncommercial/readme/USAGE.rst
+++ b/partner_noncommercial/readme/USAGE.rst
@@ -1,3 +1,5 @@
-This module creates a main menu All partners. There it gives special
-options to work with persons or organisationss.
+This module changes some strings for partners to make them more appropiate
+for non-commercial context, especially renaming companies to organisations. It also
+adds some filters that are often used in non-commercial settings, although they can
+be usefull within companies too.
 
--- a/partner_noncommercial/views/res_partner.xml
+++ b/partner_noncommercial/views/res_partner.xml
@@ -4,14 +4,9 @@
     <record id="view_res_partner_filter" model="ir.ui.view">
         <field name="inherit_id" ref="base.view_res_partner_filter" />
         <field name="model">res.partner</field>
+        <field name="mode">extension</field>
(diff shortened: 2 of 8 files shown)


lfreeke (approved):
LGTM Functional review

NL66278:
Updated this after comments from customer, clearly showing there is no need from a separate menu, apart from the main contacts menu-option. The extra filters and non-commercial terminology are satisfactory.

partner_noncommercial/readme/USAGE.rst
```diff
@@ -1,3 +1,3 @@
-This module creates a main menu All partners. There it gives special
+This module creates a main menu Partners. There it gives special
```
thomaspaulb: This is technically not true any more, we should maybe change this to:

"This module adds specific filters, groupings and modifies some naming to make the existing main menu Contacts easier to use for non-commercial entities".

partner_noncommercial/views/res_partner.xml
```diff
@@ -4,24 +4,52 @@
...
+                    string="City"
+                    context="{'group_by': 'city'}"
+                    />
+                <filter
+                    name="group_organisation"
+                    string="Organisation"
+                    context="{'group_by': 'commercial_partner_id'}"
+                    />
+            </filter>
+
```
thomaspaulb, line 52: Can you do an interactive rebase to get the whole bunch of commits down to two: adding of the filters/grouping, and removal of the menus?

thomaspaulb (changes requested):

partner_noncommercial/readme/USAGE.rst
NL66278: I adapted the usage.

thomaspaulb (approved):

PR #969, opened 2020-10-02

luistorresm:
not firstname

name = 'Van-Eyck Jan'

this was converted to

firstname = False
lastname= 'Van-Eyck'
lastname2 = 'Jan'

and it should be like

firstname = 'Jan'
lastname= 'Van-Eyck'
lastname2 = False

### DIFF
--- a/partner_second_lastname/models/res_partner.py
+++ b/partner_second_lastname/models/res_partner.py
@@ -69,32 +69,50 @@ class ResPartner(models.Model):
         - If the partner is a company, save it in the lastname.
         - Otherwise, make a guess.
         """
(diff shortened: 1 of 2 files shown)


luistorresm:
Hello @moylop260 @pedrobaeza 

This is a fowardport from v10, is necessary to get the approbations?

Regards

moylop260 (approved):

moylop260:
It is a pre-approved change in other branch

So, maybe we can fast-track it

But let's wait for review

NL66278 (approved):
:+1: LGTM

PR #976, opened 2020-10-08
[10.0][IMP] partner_firstname: Improve performances

rousseldenis:
When batch modifications are done, the recomputation can be very
long.

* One query to ir.config.parameter when computing names, useful
when multi recordset.
* On inverse, updating both fields at once and only when needed.

### DIFF
--- a/partner_firstname/models/res_partner.py
+++ b/partner_firstname/models/res_partner.py
@@ -94,11 +94,12 @@ class ResPartner(models.Model):
             'partner_names_order', self._names_order_default())
 
     @api.model
-    def _get_computed_name(self, lastname, firstname):
+    def _get_computed_name(self, lastname, firstname, order=None):
         """Compute the 'name' field according to splitted data.
         You can override this method to change the order of lastname and
         firstname the computed name"""
-        order = self._get_names_order()
+        if order is None:
+            order = self._get_names_order()
         if order == 'last_first_comma':
             return u", ".join((p for p in (lastname, firstname) if p))
         elif order == 'first_last':
@@ -110,9 +111,10 @@ class ResPartner(models.Model):
     @api.depends("firstname", "lastname")
     def _compute_name(self):
         """Write the 'name' field according to splitted data."""
+        order = self._get_names_order()
         for record in self:
             record.name = record._get_computed_name(
-                record.lastname, record.firstname,
+                record.lastname, record.firstname, order=order
             )
 
     @api.multi
@@ -157,7 +159,7 @@ class ResPartner(models.Model):
         return name
 
     @api.model
-    def _get_inverse_name(self, name, is_company=False):
+    def _get_inverse_name(self, name, is_company=False, order=None):
         """Compute the inverted name.
 
         - If the partner is a company, save it in the lastname.
@@ -170,12 +172,15 @@ class ResPartner(models.Model):
         When this method is called, :attr:`~.name` already has unified and
         trimmed whitespace.
         """
+        # After this, order can be void too, but this to avoid forcing parameter AND
+        # allowing it when looping for instance (one call to ir.config.parameter)
+        if order is None:
+            order = self._get_names_order()
         # Company name goes to the lastname
         if is_company or not name:
             parts = [name or False, False]
         # Guess name splitting
         else:
-            order = self._get_names_order()
             # Remove redundant spaces
             name = self._get_whitespace_cleaned_name(
                 name, comma=(order == 'last_first_comma'))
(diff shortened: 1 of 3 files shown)


partner_firstname/models/res_partner.py
```diff
@@ -94,11 +94,10 @@ def _get_names_order(self):
             'partner_names_order', self._names_order_default())
 
     @api.model
-    def _get_computed_name(self, lastname, firstname):
+    def _get_computed_name(self, lastname, firstname, order):
```
NL66278: To ensure compatibility with other modules calling _get_computed_name (like we have in the Dutch localization, but I assume there will be others), make order a keyword argument: order=None.

Then do not remove the assignment to order but make it: order = order if order else self._get_names_order() in case order was not passed into the method.

partner_firstname/models/res_partner.py
```diff
@@ -192,11 +192,23 @@ def _get_inverse_name(self, name, is_company=False):
...
-        """Try to revert the effect of :meth:`._compute_name`."""
+        """
+            Try to revert the effect of :meth:`._compute_name`.
+            Generate one write if values are really updated
+        """
         for record in self:
             parts = record._get_inverse_name(record.name, record.is_company)
-            record.lastname = parts['lastname']
-            record.firstname = parts['firstname']
+            to_update = {}
```
NL66278: use values instead of to_update, as this is also the actual argument name in to_update.

NL66278 (changes requested):
:+1: Good to improve performance but some changes needed. Also look at calls in unit tests (see travis errors).

partner_firstname/models/res_partner.py
rousseldenis: Didn't catch what you mean, but done

partner_firstname/models/res_partner.py
```diff
@@ -110,9 +111,10 @@ def _get_computed_name(self, lastname, firstname):
     @api.depends("firstname", "lastname")
     def _compute_name(self):
         """Write the 'name' field according to splitted data."""
+        order = self._get_names_order()
         for record in self:
             record.name = record._get_computed_name(
-                record.lastname, record.firstname,
+                record.lastname, record.firstname, order
```
NL66278: Use order=order here as well.

partner_firstname/models/res_partner.py
```diff
@@ -192,11 +197,24 @@ def _get_inverse_name(self, name, is_company=False):
...
+        """
+        order = self._get_names_order()
         for record in self:
-            parts = record._get_inverse_name(record.name, record.is_company)
-            record.lastname = parts['lastname']
-            record.firstname = parts['firstname']
+            parts = record._get_inverse_name(record.name, record.is_company, order)
+            values = {}
+            if parts['lastname'] != record.lastname:
+                values.update({
```
NL66278, line 210: Why not just: `values["lastname"] = parts["lastname"]`?

And the same below for firstname...

partner_firstname/models/res_users.py
```diff
@@ -28,9 +28,10 @@ def default_get(self, fields_list):
     @api.onchange("firstname", "lastname")
     def _compute_name(self):
         """Write the 'name' field according to splitted data."""
+        order = self.env["res.partner"]._get_names_order()
         for rec in self:
             rec.name = rec.partner_id._get_computed_name(
-                rec.lastname, rec.firstname)
+                rec.lastname, rec.firstname, order)
```
NL66278: order=order: keyword argument

rousseldenis:
@NL66278 Done.

IMHO, the action button to recompute all partners is too much heavy (85000 partners recompute takes 15 minutes - if you want to maintain a low cpu time limit...). Queue job is a good solution for that. Either we add a dependency here, either we externalize that process in another module. what do you think ?

NL66278:
@rousseldenis Maybe finish this PR first and then have a new one adding the job / queue?

In general I agree that a front-end button that will need to work on thousands or even tens of thousands of records, while the user is waiting, and waiting some more.... is a very bad idea. So the button should submit a job to a queue. We will need some safeguard against double submissions, or recomputations that will not change anything.

rousseldenis:
> @rousseldenis Maybe finish this PR first and then have a new one adding the job / queue?
> 
> In general I agree that a front-end button that will need to work on thousands or even tens of thousands of records, while the user is waiting, and waiting some more.... is a very bad idea. So the button should submit a job to a queue. We will need some safeguard against double submissions, or recomputations that will not change anything.

Indeed. This one is ready

partner_firstname/models/res_partner.py
NL66278: Please reply to this remark...

rousseldenis:
@NL66278

NL66278 (approved):
:+1: LGTM

rousseldenis:
@gva-acsone

PR #977, opened 2020-10-08
Change connection to ec.europa.eu from http to https

hugosantosred:
ec.europa.eu site has enabled the https protocol by default.

This commit is for update base_location_nuts to use this URL with https.


rousseldenis (approved):

pedrobaeza (approved):

pedrobaeza:
Please forward-port it to v12

PR #978, opened 2020-10-09
[12.0] Change connection to ec.europa.eu from http to https

hugosantosred:
12.0 version of PR #977


pedrobaeza:
Fast-tracking this:

/ocabot merge patch

PR #979, opened 2020-10-09
[12.0] partner multi relation improvements

NL66278:
- Black code, to enable backport/forward port fixes to/from 13.0 and beyond;
- Correct handling of inactive;
- Correct color coding of view entries;
- View fixes;
- Ensure tabs shown on all partner forms;
- Better integration into standard odoo Contacts app.


partner_multi_relation/models/res_partner.py
```diff
@@ -170,28 +167,26 @@ def get_partner_type(self):
...
-                        'default_this_partner_id': {0},
-                        'active_model': 'res.partner',
-                        'active_id': {0},
-                        'active_ids': [{0}],
-                        'active_test': False""".format(contact.id)]
-            context = '{' + ', '.join(elements + to_add) + '}'
-            action['context'] = context
-            return action
+        """Handle button to show relations for partner."""
+        self.ensure_one()
```
patrickrwilson, line 176: I could be wrong but isn't ensure_one deprecated?

partner_multi_relation/views/res_partner.xml
```diff
@@ -1,43 +1,53 @@
 <?xml version="1.0" encoding="UTF-8"?>
```
patrickrwilson, line 1: These are no longer needed since v11 I believe but don't impact functionality leaving them there.

patrickrwilson (approved):
Works great, especially the archive change. I did add only a few comments but no stoppers at all. Good work. LGTM.

partner_multi_relation/views/res_partner.xml
rousseldenis: > These are no longer needed since v11 I believe but don't impact functionality leaving them there.

Don't misunderstand python files encoding and xml document encoding which is not equal!

partner_multi_relation/models/res_partner.py
rousseldenis: Nope, ensure_one() is not removed as it can be useful when you want to ... ensure one record :-)

partner_multi_relation_tabs/models/ir_actions_act_window.py
```diff
@@ -0,0 +1,28 @@
+# Copyright 2020 Therp BV <https://therp.nl>.
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+from odoo import api, models
+
+
+class IrActionsActWindow(models.Model):
+    """Make sure tabs visibility updated when loading partner forms."""
```
skeller1: Brilliant, perfect solution using ir stuff for this.

Maybe you have an idea to fix the problem that the relation tab is not added to res.partner form after:

**1. Click on external link in many2one widget**
RelationTabExternalLinkListViewRelation


**2. Open contact/partner using channel follower list links (chatter)**
RelationTabOpenPartnerthorughChannelFollowerList


The redirect in follower.js is realized using this.do_action({})  https://github.com/odoo/odoo/blob/8325a0ec3f4ab1e7c92599789bfee5ca9bc3da6d/addons/mail/static/src/js/followers.js#L392

Maybe this is a further improvement.

Great work LGTM

NL66278:
To solve the problem of the tabs not adjusting in some other forms, I added improvements for both usability and code quality that I extensively tested in 10.0, and that is already used in production for two years. Unfortunately the web_domain_field module, used for this was not ported yet to OCA/web 12.0. The needed PR is here: https://github.com/OCA/web/pull/1738

scigghia (approved):

rvalyi:
@NL66278 is it OK for merging for you?

rvalyi:
@NL66278 is it OK for merging for you?

NL66278:
@rvalyi Certainly, please go ahead!

PR #982, opened 2020-10-13
[MIG] partner_second_lastname: Migration to 14.0

luistorresm:


luistorresm:
Hello @moylop260

Could you merge this PR, please?

Regards

partner_second_lastname/views/+
```diff
@@ -0,0 +1,86 @@
+<?xml version="1.0" encoding="utf-8" ?>
```
emagdalenaC2i: Delete this file

emagdalenaC2i (changes requested):

partner_second_lastname/views/+
luistorresm: Why?

luistorresm: Done @emagdalenaC2i 

Thanks

emagdalenaC2i: The file named "partner_second_lastname/views/+" because it is not used and I think is the same as "partner_second_lastname/views/res-partner.xml"

PR #989, opened 2020-10-15
[13.0][MIG] partner_email_check: Migration to 13.0

qgroulard:


kos94ok-3D:
Not allowed duplicate partner with filled email so maybe need set `copy=False` or if `_should_filter_duplicates` remove `email` from `copy_data`

partner_email_check/models/res_config_settings.py
```diff
@@ -0,0 +1,43 @@
+from odoo import api, fields, models
+
+
+class ResConfigSettings(models.TransientModel):
+    _inherit = "res.config.settings"
+
+    partner_email_check_filter_duplicates = fields.Boolean(
+        string="Filter duplicate partner email addresses",
+        help="Don't allow multiple partners to have the same email address.",
```
simahawk:
```suggestion
        help="Don't allow multiple partners to have the same email address.",
        config_parameter="partner_email_check_filter_duplicates"
```

partner_email_check/models/res_config_settings.py
```diff
@@ -0,0 +1,43 @@
...
+    )
+
+    partner_email_check_check_deliverability = fields.Boolean(
+        string="Check deliverability of email addresses",
+        help="Don't allow email addresses with providers that don't exist",
```
simahawk:
```suggestion
        help="Don't allow email addresses with providers that don't exist",
        config_parameter="partner_email_check_check_deliverability"
```
and trash get/set_values overrides ;)

BTW IMO this should be configured on the company, then you can get rid of the params.

partner_email_check/models/res_partner.py
```diff
@@ -0,0 +1,99 @@
...
+    _inherit = "res.partner"
+
+    @api.model
+    def email_check(self, emails):
+        if config["test_enable"] and not self.env.context.get(
+            "test_partner_email_check"
+        ):
+            return emails
+        return ",".join(
+            self._normalize_email(email.strip()) for email in emails.split(",")
```
simahawk:
```suggestion
            self._normalize_email(email.strip()) for email in emails.split(",") if email.strip()
```

partner_email_check/models/res_partner.py
```diff
@@ -0,0 +1,99 @@
...
+            )
+        return result["local"].lower() + "@" + result["domain_i18n"]
+
+    def _should_filter_duplicates(self):
+        conf = (
+            self.env["ir.config_parameter"]
+            .sudo()
+            .get_param("partner_email_check_filter_duplicates", "False")
+        )
+        return conf == "True"
```
simahawk: use `tools.str2bool` :wink:

partner_email_check/tests/test_partner_email_check.py
```diff
@@ -0,0 +1,147 @@
...
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from unittest.mock import patch
+
+from odoo.exceptions import UserError, ValidationError
+from odoo.tests.common import TransactionCase
+from odoo.tools.misc import mute_logger
+
+
+class TestPartnerEmailCheck(TransactionCase):
```
simahawk: pls use `SavepointCase` and use `cls.env = cls.env(context=dict(cls.env.context, tracking_disable=True))` to disable tracking globally.

simahawk (changes requested):

rousseldenis:
@qgroulard Could you attend comments?

qgroulard:
> @qgroulard Could you attend comments?

Yes I am trying to squeeze this in today's agenda but I'm afraid it will be for another day.
I'll work on this asap.

partner_email_check/models/res_partner.py
```diff
@@ -24,14 +24,24 @@
 class ResPartner(models.Model):
     _inherit = "res.partner"
 
+    def copy_data(self, default=None):
+        res = super(ResPartner, self).copy_data(default=default)
+        if self._should_filter_duplicates():
+            for copy_vals in res:
+                if "email" in copy_vals:
+                    copy_vals["email"] = ""
```
NL66278: Why not delete the value from the dictionary, or set it to False, so email will have a NULL value in the database.

NL66278 (approved):
:+1: LGTM

rousseldenis:
@simahawk Could you update your review?

partner_email_check/models/res_partner.py
```diff
@@ -0,0 +1,99 @@
...
+
+class ResPartner(models.Model):
+    _inherit = "res.partner"
+
+    def copy_data(self, default=None):
+        res = super(ResPartner, self).copy_data(default=default)
+        if self._should_filter_duplicates():
+            for copy_vals in res:
+                if "email" in copy_vals:
+                    del copy_vals["email"]
```
simahawk: not a blocker but you could do:
```suggestion
            for copy_vals in res:
                copy_vals.pop("email", None)
```

partner_email_check/models/res_partner.py
```diff
@@ -0,0 +1,99 @@
...
+        return res
+
+    @api.model
+    def email_check(self, emails):
+        if config["test_enable"] and not self.env.context.get(
```
simahawk: do we really need this? What's the main reason?

partner_email_check/tests/test_partner_email_check.py
```diff
@@ -0,0 +1,141 @@
...
+from odoo.tests.common import SavepointCase
+from odoo.tools.misc import mute_logger
+
+
+class TestPartnerEmailCheck(SavepointCase):
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        cls.env = cls.env(context=dict(cls.env.context, tracking_disable=True))
+        # Checks are disabled during tests unless this key is set
```
simahawk: I don't get why we need this ctx key :thinking: 
Checks should be disabled by the flags set on the company.
If you want to keep them ON by default you can set them OFF in tests for the rest of the world by adding demo data :wink:

qgroulard: This is actually the same comment as https://github.com/OCA/partner-contact/pull/989#discussion_r665111764 since this key is only used there.

I think I get it now...

There are actually 3 checks performed:
-  Check email unique: only if set on company;
-  Check deliverability: only if set on company, performed with https://pypi.org/project/email-validator/ thus can't run without checking emails syntax as well;
- Check syntax: currently always performed unless in test mode without this key (test_partner_email_check), performed with https://pypi.org/project/email-validator/

So I guess the idea is not to break other modules' tests since not having this context key set (and being in test mode) is the only way not to check emails syntax.

Maybe the solution would be to add a third flag for the syntax check.
It should be off by default otherwise other modules' test may fail. Thus users must be aware that behavior changes with v13 since beforehand this check used to always be performed.

What do you think ?

partner_email_check/migrations/13.0.1.0.0/post-migration.py
```diff
@@ -0,0 +1,24 @@
+# Copyright 2021 ACSONE SA/NV
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
```
NL66278: Make it https:

partner_email_check/tests/test_partner_email_check.py
simahawk: nope, the solution is simpler and I reported it above :wink: 

Just do this:

1. add demo data file
2. in that file call write function on all companies
3. set the flags to false
4. in your test class enable the flags

This way for the rest of the test suites (referred as "rest of the world" above) won't be impacted.

simahawk: of course, the alternative is that you make the flag false by default and you state in the README that it should be enabled

simahawk (approved):
thanks for your changes!

PR #1002, opened 2020-10-23
[ADD] partner_route: New module

chienandalu:
This module serves as a baseline to define the partner routes that the members
of an organization need to do to acomplish a task over customers regularly

For example:

    - Commercial routes.
    - Delivery routes.
    - Salesmen routes.
    - Repair routes.

The module comes without any of this specific capacities, that should be further
developed.

image

cc @Tecnativa TT25235


chienandalu:
@carlosdauden @sergio-teruel can you review?

carlosdauden (approved):
I miss some filter from route to see in which route a partner is or from partners to determine which partners are in a route, but nothing that is not solved with an advanced search.

partner_route/models/partner_route.py
```diff
@@ -0,0 +1,100 @@
...
+    user_id = fields.Many2one(
+        comodel_name="res.users",
+        string="Responsible",
+        help="Which user is in charge of the route",
+    )
+    route_type = fields.Selection(
+        selection=[
+            ("crm", "CRM"),
+            ("sale", "Sales"),
+            ("delivery", "Delivery"),
```
sergio-teruel, line 23: Maybe CRM and sale is the same...

partner_route/models/partner_route.py
```diff
@@ -0,0 +1,100 @@
...
+        related="partner_id.street",
+    )
+    city = fields.Char(
+        related="partner_id.city",
+    )
+    zip = fields.Char(
+        related="partner_id.zip",
+    )
+    state_id = fields.Many2one(
+        related="partner_id.state_id",
```
sergio-teruel, line 96: I am not sure... but comodel_name is not needed??

sergio-teruel: And the field strings??

sergio-teruel (changes requested):

partner_route/models/partner_route.py
pedrobaeza: On related fields, if not specified, they are taken from the source, so in summary, no need.

sergio-teruel (approved):

marcelsavegnago (approved):

PR #1003, opened 2020-10-24
[14.0][MIG] base_location: Migration to 14.0

pedrobaeza:
- Standard procedure
- `invisible` > `attrs` in view
- Changed method

@Tecnativa TT26286


pedrobaeza:
@newtratip @NL66278 maybe you are interested in reviewing this.

rousseldenis (approved):

lmignon (approved):

HaraldPanten (approved):

pedrobaeza:
I'm trying here to switch to computed writable, so please don't merge yet.

base_location/models/__init__.py
```diff
@@ -0,0 +1,6 @@
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
```
NL66278, line 1: Would be best to have all license lines https://

NL66278 (approved):
:+1: One nag, but LGTM

desdelinux (approved):
Functional test is ok

fmdl (approved):

newtratip (approved):

OsoTranquilo:
Correct functionality test

alexis-via:
Sorry if my question is stupid... I haven't followed the development of base_location since v12. Why do we need the res.city.zip object ? I really don't understand the role of this object: res.city is enough, isn't it ?

alexis-via:
Another question: res.city object has a zipcode field (cf https://github.com/odoo/odoo/blob/14.0/addons/base_address_city/models/res_city.py#L13), and that field existed since the beginning of base_address_city, but its value is not set by base_location_geonames_import, cf https://github.com/OCA/partner-contact/blob/13.0/base_location_geonames_import/wizard/geonames_import.py#L98
So it would mean that we don't want to use zipcode on res.city, but we use res.city.zip to store the zipcode ? Maybe it's the answer to my previous question...

alexis-via:
FYI, as geonames.org is down (https://download.geonames.org/export/zip/ is empty), I can't test easily the base_location + base_location_geonames_import modules... which makes things more difficult.

pedrobaeza:
@alexis-via Geonames is up again. About the data model, it was the same in v12. The idea of the refactoring of this module was to have a normalized data model, where there's an specific model that represents a city, not a city + zip. Your colleague @rvalyi claimed for it sometime ago in v7/v8. Then, we have a cardinality of 1..N from city to zips linked to that city. There's an ongoing conversation in #1020 for relaxing current constraint for having a hybrid data model.

I'm finishing the conversion to computed writable for having it as example for things like https://github.com/OCA/odoo-community.org/pull/50. I will write back when finished.

alexis-via:
I'll study it more in depth and talk with @rvalyi about it. Now that geonames is back up and running, I will be easier for me to test. I'll see if I arrive to the conclusion that res.city.zip is needed or if it should be dropped.

alexis-via:
Just to facilitate the debate:
- in v10, we had a single object res.better.zip with zip + city name + state_id + country_id, cf https://github.com/OCA/partner-contact/blob/10.0/base_location/models/better_zip.py#L17
- starting from v12, there is a new official addon base_address_city with an object res.city with the same information cf https://github.com/odoo/odoo/blob/14.0/addons/base_address_city/models/res_city.py#L12

pedrobaeza:
I have adapted the code to new computed writable fields for depicting the possibilities.

Please take a look to this new "style", and if you give your blessings, I finally merge this.

base_location/README.rst
```diff
@@ -0,0 +1,115 @@
...
+    :alt: Translate me on Weblate
+.. |badge5| image:: https://img.shields.io/badge/runbot-Try%20me-875A7B.png
+    :target: https://runbot.odoo-community.org/runbot/134/14.0
+    :alt: Try me on Runbot
+
+|badge1| |badge2| |badge3| |badge4| |badge5| 
+
+This module introduces a zip model that allows you to manage locations in a better way.
+
+The zips will allow the users to complete automatically all address-related fields by just filling the zip.
```
NL66278, line 30: Actually not all address related fields. You will not be able to fill in house_number and door_number from the zip. In most countries also not the street. Maybe replace 'all' with 'many'?

pedrobaeza: AFAIK, that fields are added through extension modules, so it's OK to say that with current dependency set.

NL66278: It is only a minor point anyway. Very glad with the module as is. But street is also not filled automatically, and this is definitely also address related.

pedrobaeza: Yeah, I get your point about the street, although it's impossible to fill a not known data derived from the ZIP. If I read the text, I understand that the "address-related" text refers exactly to this: the address fields that are linked with the ZIP, not the non related ones, and street (and the others) is not related, as a ZIP code contains a lot of streets.

OsoTranquilo (approved):
It works well

pedrobaeza:
Can I merge?

HaraldPanten:
> Can I merge?

👍

alexis-via:
I talked with @rvalyi about res.city.zip object, and he told me that we doesn't use base_location in Bazil, so he's not concerned about the issue.
As I still think that base_location is now useless, I developed this afternoon a new module base_address_city_geonames_import, that does NOT depend on base_location. KISS concept. I wonder if I should propose this module in OCA/partner-contact or keep it in some Akretion Github project.

pedrobaeza:
OK, I still see valuable this structure as base instead of the base_address_city one. You can propose if you want the other module to OCA for having both options. 

Merging this one:

/ocabot merge nobump

alexis-via:
Unless if somebody explicitly says in this thread that he's interested in removing the res.city.zip object and in the associated simplification of the code, I think I'll publish my module base_address_city_geonames_import for v14 under akretion/odoo-usability (I'll do that probably tomorrow). The extra work needed for publishing in OCA is not worth it if I'm the only person interested.

alexis-via:
I just published my new module base_address_city_geonames_import in a dedicated Github project:
https://github.com/akretion/geonames-import-simple
I'm very happy with the result:
base_location + base_location_geonames_import = 208 + 212 lines of python (tests excluded) = 420 lines total
base_address_city_geonames_import = 186 lines of python (tests excluded)
55% less lines of python code. KISS.

As a consequence, I'll stop maintaining l10n_fr_base_location_geonames_import in OCA/l10n-france (I'll move the equivalent code elsewhere).
If one day the strategy changes and the object res.city.zip is dropped, I may propose my module base_address_city_geonames_import to OCA or go back to using base_location + base_location_geonames_import.

fmdl:
@alexis-via I interessed to have this module in OCA.

PR #1008, opened 2020-11-04
[FIX] partner_contact_address_default: can't adapt type 'res.partner'

ioans73:
Update with the same type of value as the original function (ID instead of BrowseRecord). As it is now, in runbot it gives an error, and in some of our clients it also fails

**Runbot error: create invoice from a sale order**
OCA_partner_contact_address_default

### DIFF
--- a/partner_contact_address_default/models/res_partner.py
+++ b/partner_contact_address_default/models/res_partner.py
@@ -26,5 +26,5 @@ class ResPartner(models.Model):
             for addr_type in default_address_type_list:
                 default_address_id = partner['partner_{}_id'.format(addr_type)]
                 if default_address_id:
-                    res[addr_type] = default_address_id
+                    res[addr_type] = default_address_id.id
         return res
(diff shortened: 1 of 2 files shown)


ivantodorovich (approved):
LGTM :+1: 
The overloaded method returns ids instead of recordsets, so this method should do the same.

however tests are failing, they need fixing, as they weren't coded to check on ids:

```
----------
FAIL
FAIL: test_contact_address_default (odoo.addons.partner_contact_address_default.tests.test_partner_contact_address_default.TestPartnerContactAddressDefault)
`   File "/home/travis/build/OCA/partner-contact/partner_contact_address_default/tests/test_partner_contact_address_default.py", line 35, in test_contact_address_default
`     self.assertEqual(res['delivery'], self.partner)
` AssertionError: 50 != res.partner(50,)
FAILED
```

ioans73:
@ivantodorovich tests fixed

blopa98-S73 (approved):

ypapouin (approved):

pedrobaeza:
Is this needed in v13 as well?

pedrobaeza:
Cherry-picked for v13 in 0e5f6c064d352fc6e4e9831277f69d5b90fe1824

pedrobaeza:
And also for v14 in 135f393f9c88b51bff6e671bb6a4f9ab5b203785

PR #1017, opened 2020-11-16
[13.0][FIX] partner_firstname: byte encoded string from ldap

astirpe:
This PR fixes the following error:

```
  File "/path_to/partner_firstname/models/res_partner.py", line 38, in create
    self._get_whitespace_cleaned_name(name),
  File "/path_to/partner_firstname/models/res_partner.py", line 131, in _get_whitespace_cleaned_name
    name = " ".join(name.split()) if name else name
TypeError: sequence item 0: expected str instance, bytes found - - -
```

The error occurs when the user name coming from LDAP (eg.: FreeIPA) is a byte encoded string.

### DIFF
--- a/partner_firstname/models/res_partner.py
+++ b/partner_firstname/models/res_partner.py
@@ -127,6 +127,11 @@ class ResPartner(models.Model):
 
         Removes leading, trailing and duplicated whitespace.
         """
+        if isinstance(name, bytes):
+            # With users coming from LDAP, name can be a byte encoded string.
+            # This happens with FreeIPA for instance.
+            name = name.decode("utf-8")
+
         try:
             name = " ".join(name.split()) if name else name
         except UnicodeDecodeError:


partner_firstname/models/res_partner.py
```diff
@@ -127,6 +127,11 @@ def _get_whitespace_cleaned_name(self, name, comma=False):
 
         Removes leading, trailing and duplicated whitespace.
         """
+        if isinstance(name, bytes):
```
NL66278, line 130: Is this module the right place to solve this problem? It should not be possible to set a name on a partner to bytes in the first place I think...

astirpe: Standard Odoo automatically converts the vals from bytes to strings while creating/writing Char fields. This is done [here](https://github.com/odoo/odoo/blob/14.0/odoo/tools/pycompat.py#L35-L36) when invoking `odoo.tools.pycompat.to_text` to converts bytes to string.

In this particular case, method `_get_whitespace_cleaned_name()` is executed on the vals before the bytes are converted to strings, which is done later when invoking the `super()`. So I think this is the right place to solve this problem, unless someone else wants to refactor the module itself.

NL66278 (approved):
:+1: LGTM. Accepting this as a stopgap solution. The canonical way would be to do the conversion of bytes to string at the point where the data enters the system, not where the data is used.

HaraldPanten (approved):

PR #1027, opened 2020-12-01
[12.0][FIX] base_location: Update name when changing city's state or country

ypapouin:
Update name when changing city's state or country otherwise computed named is not correct
image


pedrobaeza (approved):

marcelsavegnago (approved):

HaraldPanten (approved):

pedrobaeza:
Can you please check in v13 and v14?

PR #1030, opened 2020-12-02
[14.0][FIX] base_location: Update name when changing city's state or country

ypapouin:
Forward port from #1027


HaraldPanten (approved):
LGTM as well

marcelsavegnago (approved):

PR #1034, opened 2020-12-14
[12.0] [MIG] partner street number

NL66278:
The old partner_street_number was mostly, but not completely replaced by base_address_extended in Odoo standard.

However they forgot to adapt the contact addresses in the main partner form. This module remains to fill that gap.


partner_street_number/models/res_partner.py
```diff
@@ -1,98 +0,0 @@
...
-        lead to silent data corruption should the synchronisation API
-        ever change.
-        """
-        res = super(ResPartner, self)._address_fields()
-        if 'street' in res and not (
-                self._context.get('display_address')):
-            res.remove('street')
-        return res + ['street_name', 'street_number']
-
-    street_name = fields.Char('Street name')
```
rousseldenis: @NL66278 If you let the view, you need to keep those fields

PR #1036, opened 2020-12-14
[14.0][MIG] partner_employee_quantity: Migration to 14.0

newtratip:
Standard Migration, Ready for review.


rousseldenis (approved):

newtratip:
@ecosoft-odoo Could you review this ?

rousseldenis (approved):
Code review

partner_employee_quantity/views/res_partner_view.xml
```diff
@@ -0,0 +1,37 @@
...
+    <record id="view_partner_form" model="ir.ui.view">
+        <field name="name">Partner form with employees</field>
+        <field name="model">res.partner</field>
+        <field name="inherit_id" ref="base.view_partner_form" />
+        <field name="arch" type="xml">
+            <page name="internal_notes" position="inside">
+                <group>
+                    <group
+                        name="employee_quantity"
+                        string="Employee quantity"
```
NL66278, line 12: Why specify string here? Would be the same as on the field?

NL66278:
It should not be to much work to add the minimum and maximum quantity to the range, and make the range on partner a computed field based on the number of employees. Given that the module is designated as "Mature", this would be kind of expected?

HaraldPanten (approved):

PR #733, opened 2019-05-02
[12.0] base_location: allow to correctly save company form

eLBati:
Reverting 70fc4395740e9df4c8056adc877ab784c1c6c21c and removing _check_zip as preventing to correctly save company data.

FIX the following cases:

 - open company form
 - set zip_id
 - save
 - user is blocked

and

 - open company form
 - set state_id
 - save
 - state_id is not saved

PR #1041, opened 2020-12-15
[12.0] base_location: allow to correctly save company form

eLBati:
From https://github.com/OCA/partner-contact/pull/733

Reverting 70fc4395740e9df4c8056adc877ab784c1c6c21c and removing _check_zip as preventing to correctly save company data.

FIX the following case

 - open company form
 - set state_id
 - save
 - state_id is not saved

base_location

### DIFF
--- a/base_location/models/res_company.py
+++ b/base_location/models/res_company.py
@@ -47,21 +47,11 @@ class ResCompany(models.Model):
 
     def _inverse_city_id(self):
         for company in self:
-            company.with_context(
-                skip_check_zip=True).partner_id.city_id = company.city_id
+            company.partner_id.city_id = company.city_id
 
     def _inverse_zip_id(self):
         for company in self:
-            company.with_context(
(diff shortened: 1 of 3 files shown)


pedrobaeza:
Please split in separate commits the revert + the new fix

eLBati:
@pedrobaeza done

pedrobaeza:
@JordiBForgeFlow I remember you put these constraints. What was the reason?

@eLBati is that steps reproducible as is in runbot?

eLBati:
@pedrobaeza I made the gif in runbot

pedrobaeza:
Not sure the solution should be to remove the constraint, but make it to work.

eLBati:
@pedrobaeza that's right, I tried but didn't succeed

pedrobaeza:
Let's see if @JordiBForgeFlow can tell us the reasoning for the constraints and the problem.

PR #1044, opened 2020-12-27
[13.0][FIX] partner_firstname: allow saving company name

### LINKED ISSUE #964: [13.0] partner_firstname: Wrong company name
### Description of the issue
Contact has wrong name after entering both firstname/lastname for individual and company name.

### Version
Odoo - 13.0 from the odoo repository
partner_firstname - 13.0.1.0.1

### How to reproduce

1. Open contact creating form
2. Switch to individual contact type
3. Enter firstname and lastname e.g. Adam Smith
4. Switch to company contact type
5. Clear name field and enter new name e.g. TestCompany
6. Click on 'Save' button

**Expected result:**
Contact with name 'TestCompany' was created.

**Actual result:**
Contact with name 'Adam Smith' was created.

pankk:
Fixes #964
Without this the name is always deleted and only the "Individual" type naming works.

### DIFF
--- a/partner_firstname/tests/test_create.py
+++ b/partner_firstname/tests/test_create.py
@@ -30,7 +30,6 @@ class PersonCase(TransactionCase):
         )
         for key, value in self.good_values.items():
             self.assertEqual(self.record[key], value, "Checking key %s" % key)
-
         super(PersonCase, self).tearDown()
 
     def test_no_name(self):
@@ -38,18 +37,18 @@ class PersonCase(TransactionCase):
         del self.values["name"]
 
     def test_wrong_name_value(self):
-        """Wrong name value is ignored, name is calculated."""
+        """Wrong name value is ignored, name is calculated for persons."""
         self.values["name"] = "BÄD"
 
     def test_wrong_name_context(self):
-        """Wrong name context is ignored, name is calculated."""
+        """Wrong name context is ignored, name is calculated for persons."""
         del self.values["name"]
         self.context["default_name"] = "BÄD"
 
     def test_wrong_name_value_and_context(self):
-        """Wrong name value and context is ignored, name is calculated."""
-        self.values["name"] = "BÄD1"
-        self.context["default_name"] = "BÄD2"
+        """Wrong name value and context is ignored, name is calculated for persons."""
+        self.values["name"] = "BÄD"
+        self.context["default_name"] = "BÄD"
 
 
 class CompanyCase(PersonCase):
@@ -58,8 +57,11 @@ class CompanyCase(PersonCase):
     context = {"default_is_company": True}
 
     def setUp(self):
+        """For companies only the name is relevant. Lastname is added by
+        _get_inverse_name.
+        """
         super(CompanyCase, self).setUp()
-        self.good_values.update(lastname=self.values["name"], firstname=False)
+        self.good_values.update(lastname="BÄD", firstname=False, name="BÄD")
         self.values = self.good_values.copy()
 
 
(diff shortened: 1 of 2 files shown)


pankk:
I'll check and update the failing tests later ~~today~~ :sweat_smile: .

pankk:
The changes to the test are because the firstname/lastname should not at all be used when the created partner is a company. At least, I deduce this from the fact that in the UI `firstname` and `lastname` are hidden when the `res_partner.company_type` is `company` instead of `person`. `lastname` is still a duplicate of `name`.
Tested the original bug report as well as the reverse case (saving the individual instead of the company).

skeller1 (approved):
:+1: LGTM functional and code review (tested locally, runbot is not available)
Thank you for supporting OCA

HaraldPanten (approved):

rousseldenis:
@lmignon

partner_firstname/tests/test_create.py
```diff
@@ -30,26 +30,25 @@ def tearDown(self):
...
-        """Wrong name context is ignored, name is calculated."""
+        """Wrong name context is ignored, name is calculated for persons."""
         del self.values["name"]
         self.context["default_name"] = "BÄD"
 
     def test_wrong_name_value_and_context(self):
-        """Wrong name value and context is ignored, name is calculated."""
-        self.values["name"] = "BÄD1"
-        self.context["default_name"] = "BÄD2"
+        """Wrong name value and context is ignored, name is calculated for persons."""
```
lmignon, line 49: @pankk What' wrong with this these tests? Why do you modify the `setUp` method into the `CompanyCase` class? I don't see any specific case for the changes you introduce (at least it's not transparent into the code). The changes into `res_partner` make sense but not the changes into the test class. For example, two different values must be assigned to the name and the default_name into the context in `test_wrong_name_value_and_context`. This allow to detect what part of the code is broken if the the test fails.

lmignon (changes requested):

partner_firstname/tests/test_create.py
pankk: I tried to explain it in my earlier message. From memory: the unmodified tests will fail with the changes to the code, as the tests assume there will always be a firstname and a lastname component. However, when the partner type is a company only the name field is used (and lastname is computed using it, but firstname stays empty). It comes down to that for a company the name really does not need to be calculated, as it's already the only thing filled in, and firstname and lastname are not relevant in this case.

lmignon: @pankk `test_wrong_name_value_and_context`should not be modified. Only the definition of `good_values` into `CompanyCase` should be modified. Moreover, the expected `good_values` must be based on `self.values["name"]`. `self.good_values.update(lastname=self.values["name"], firstname=False, name=self.values["name"])`

pankk: I didn't see an easy way to do it, hence my initial comment. My approach was to have something that worked without refactoring a large part of the module. Maybe it makes more sense to just replace the entire "company" case from the tests. The only thing in that case that should be tested is if the compute method does not touch the name.
I have no urgency to get this merged, yet it would be great if someone else could pick this up.

lmignon: I prefer separate test cases for company... IMO it's the only way to properly test cases related to company.

pankk: Ellipsis not required. We're in agreement.

dreispt:
@lmignon Is this blocked for merging?

PR #1048, opened 2021-01-15
[12.0][ADD] partner_pricelist_search: New module

carlosdauden:
This module adds search by pricelist field option in partners.
Also add a new Smart Customers button in price list form view to show partner
with that pricelist.

@Tecnativa TT27659


partner_pricelist_search/README.rst
```diff
@@ -0,0 +1,80 @@
...
+    :target: https://translation.odoo-community.org/projects/partner-contact-12-0/partner-contact-12-0-partner_pricelist_search
+    :alt: Translate me on Weblate
+.. |badge5| image:: https://img.shields.io/badge/runbot-Try%20me-875A7B.png
+    :target: https://runbot.odoo-community.org/runbot/134/12.0
+    :alt: Try me on Runbot
+
+|badge1| |badge2| |badge3| |badge4| |badge5| 
+
+This module adds search by pricelist field option in partners.
+Also add a new Smart Customers button in price list form view to show partner
```
pedrobaeza:
```suggestion
Also add a new smart button called "Customers" in pricelist form view to show partners
```

partner_pricelist_search/README.rst
```diff
@@ -0,0 +1,80 @@
...
+with that pricelist.
+
+With many partner records the search can be slow because all partners are
```
pedrobaeza: To ROADMAP.rst

partner_pricelist_search/__manifest__.py
```diff
@@ -0,0 +1,19 @@
+# Copyright 2021 Tecnativa - Carlos Dauden
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl)
+
+{
+    "name": "Partner pricelist search",
+    "version": "12.0.1.0.1",
```
pedrobaeza:
```suggestion
    "version": "12.0.1.0.0",
```

partner_pricelist_search/models/res_partner.py
```diff
@@ -0,0 +1,33 @@
...
+    _inherit = "res.partner"
+
+    property_product_pricelist = fields.Many2one(
+        search="_search_property_product_pricelist")
+
+    @api.model
+    def _search_property_product_pricelist(self, operator, value):
+        if operator == "=":
+            def filter_func(partner):
+                return partner.property_product_pricelist.id == value
```
pedrobaeza, line 18: For clearer code, you can do here lambda function:
```suggestion
            filter_func = lambda x: x.property_product_pricelist.id == value
```

partner_pricelist_search/tests/test_partner_pricelist_search.py
```diff
@@ -0,0 +1,72 @@
...
+            "property_product_pricelist": cls.pricelist_1,
+        })
+        cls.customer_2 = cls.env["res.partner"].create({
+            "name": "Test customer 2",
+            "property_product_pricelist": cls.pricelist_2,
+        })
+
+    def test_partner_pricelist_search(self):
+        ResPartner = self.env["res.partner"]
+
```
pedrobaeza: Don't put empty lines inside a method

partner_pricelist_search/tests/test_partner_pricelist_search.py
```diff
@@ -0,0 +1,72 @@
...
+        # Test search '='
+        partners = ResPartner.search([
```
pedrobaeza: You can put each of these checks in a different test method for not having interrupted the execution in case one of them failed, and at the end, there's no initialization needed that forces to share the same method (except the model handler that you can store in a variable in the setup).

partner_pricelist_search/tests/test_partner_pricelist_search.py
```diff
@@ -0,0 +1,72 @@
...
+        # Test search not implemented
+        with self.assertRaises(UserError):
+            ResPartner.search([
+                ("property_product_pricelist", "ilike",
+                 "pricelist xx")
+            ])
+
+    def test_show_pricelist_partners(self):
+        res = self.pricelist_1.show_pricelist_partners()
+        self.assertTrue(res["domain"])
```
pedrobaeza: Test the real result
```suggestion
        self.assertEqual(ResPartner.search(res["domain"]), self.customer_1)
```
and the same for the other one

partner_pricelist_search/README.rst
pedrobaeza: And add as well that no grouping is allowed.

partner_pricelist_search/models/res_partner.py
carlosdauden: I had used lambda:
https://github.com/OCA/partner-contact/compare/79ea65cd67e7f3282e6c44ac95c7beb07db168cd..cf47c5e3cdca8e5b6563a927f8a4670607b3c075
But travis turned red for PEP8 E731:
https://travis-ci.com/github/OCA/partner-contact/jobs/472143017#L322-L326

partner_pricelist_search/tests/test_partner_pricelist_search.py
carlosdauden: I don't want add complexity in the test to avoid future problems if a new módule adds conditions in this domain.
I am satisfied that the method returns an action with domain.

carlosdauden:
Changes done

HaraldPanten (approved):
Good job, this one is really useful 👍 

Just a question:

When you're working with several currencies, pricelists are adding the (Currency code) next to the pricelist name.

It makes searching by pricelist (in contacts) "harder", as search can't be done by "ilike" and users would normally search by name, not by name + (currency code).

Do you think that it could be improved somehow to get a better UX?

partner_pricelist_search/models/res_partner.py
pedrobaeza: OK, I see https://www.flake8rules.com/rules/E731.html as reasoning for tracebacks, but don't really agree with the resulting code for such simple thing.

partner_pricelist_search/tests/test_partner_pricelist_search.py
pedrobaeza: Well, this test is executed right after the installation of this module, so it's literally impossible that this fails although you do any extension. The only way this can fail is if you define the tests with the tag post-install.

pedrobaeza:
@HaraldPanten I don't get your comment.

partner_pricelist_search/tests/test_partner_pricelist_search.py
carlosdauden: Ok. Changes done.

carlosdauden:
@HaraldPanten The ability to search by ilike name (with the search method) is not implemented. I started to implement it but had to take into account uppercase, lowercase, accents, etc. and I discarded it.

Although I think you mean something else that is outside the scope of this module.
Take a look at:
https://github.com/odoo/odoo/blob/6b3c81bb5eebaf9be82c9a838cdafd9d3da44787/addons/product/models/product_pricelist.py#L45-L82

pedrobaeza (approved):

HaraldPanten:
Maybe I didn't explain that use case well enough.

I meant that:

1- When I was searching by Pricelist = Public Pricelist the search had no results.

2- I had to search by Pricelist = Public Pricelist (USD) to find some results.

3- User must search by Pricelist = Pricelist name + (currency) to find the partners linked to a pricelist.

But the name of the pricelist is only Public Pricelist, not Public Pricelist (USD).

IMO it's not the best UX. As "ilike" is not possible (I understand, no problem. I didn't suggest that), it's a little bit "tedious" for "average users" to search by "Public Pricelist (USD)" because (USD) is not a part of the Pricelist name. Do you know what I mean?

carlosdauden:
@HaraldPanten can you take a look at the fourth point of this?:
https://github.com/OCA/web/blob/12.0/web_advanced_search/README.rst#id4

:wink:

HaraldPanten:
> @HaraldPanten can you take a look at the fourth point of this?:
> https://github.com/OCA/web/blob/12.0/web_advanced_search/README.rst#id4
> 
> 😉

Yep, I know that one. That could be an option, sure, but I had to review your new ADD. Even though, I approved the PR. 👍

PR #1058, opened 2021-02-12
[MIG] partner_identification: Migration to 14.0

phuctranfxvn:
Standard as-is migration


partner_identification/views/res_partner_view.xml
```diff
@@ -0,0 +1,23 @@
...
+        <field name="priority" eval="10" />
+        <field name="arch" type="xml">
+            <page name='sales_purchases' position="after">
+                <page string="ID Numbers" name="id_numbers">
+                    <field
+                        name="id_numbers"
+                        colspan="4"
+                        nolabel="1"
+                        widget="one2many_list"
+                        context="{'default_partner_id': active_id}"
```
kittiu, line 17: This indicate to have default partner when we add a line, but it doesn't show up when I test it.

kittiu: image

kittiu:
This module suppose to allow use to use python validation_code to validate. But couldn't find the way.

image

Note:

* I am surprised. I found that the validation_code field is not shown since version 7, but should we open this field?

kittiu:
Since, it never has, it would be nice, if you could add demo sample of validation code for couple of case, i.e., Number only, or number of digits etc.

partner_identification/views/res_partner_view.xml
phuctranfxvn: @kittiu Thanks for your feedbacks, fixed in https://github.com/OCA/partner-contact/pull/1058/commits/2d8aef4bf5cd1c766b1b7c16a845b93e9dd83976

phuctranfxvn:
> This module suppose to allow use to use python validation_code to validate. But couldn't find the way.
> 
> image
> 
> Note:
> 
> * I am surprised. I found that the validation_code field is not shown since version 7, but should we open this field?

> Since, it never has, it would be nice, if you could add demo sample of validation code for couple of case, i.e., Number only, or number of digits etc.

@kittiu I agree with you that we should add it back in this commit https://github.com/OCA/partner-contact/pull/1058/commits/2d8aef4bf5cd1c766b1b7c16a845b93e9dd83976. Sample for python code validation has been added also.

bosd (approved):
Functionel test, LGTM

StefanRijnhart (approved):
Thanks!

StefanRijnhart:
@kittiu is this one alright for you now?

HaraldPanten (approved):

PR #1066, opened 2021-02-18
[14.0][MIG] partner_contact_address_default: Migration to 14.0 

Zar21:
Standard migration Odoo 14.0


HaraldPanten (approved):
Functional review

vanderperre (approved):

vanderperre (changes requested):
It will be wonderful if the domain proposed is not only limited by myself (address) or by my childs but also of all contacts of the delivery/invoice type.  A search read would be able to place/sort my favourites contacts (me and my childs) and after the other ...  Also i have see a bug when you create/update a partner: you can select myself as delivery or invoice default but if you change the company type of the partner, this selection become empty ...

PR #1068, opened 2021-02-21
[12.0] partner_multi_relation: Check reverse overlapping symmetric relations

JonathanStein:
If Bart is sibling to Lisa, you can make an overlapping sibling relation from Lisa to Bart. This commit adds a reverse left/right check if the relation type is symmetric.


NL66278 (approved):
:+1: LGTM

PR #1069, opened 2021-02-24
[13.0][MIG] partner_pricelist_search: Migration to v13.0 

carlosdauden:
@Tecnativa TT28442


pedrobaeza (approved):

HaraldPanten:
Am I missing something?

Grabación de pantalla 2021-03-04 a las 9 42 09

carlosdauden:
@HaraldPanten if this is the same question this is the same answer:
https://github.com/OCA/partner-contact/pull/1048 :wink: 

Feel free to add a PR with that improvement supporting accents and not penalizing performance.
It works for us by simply installing this:
https://github.com/OCA/web/blob/13.0/web_advanced_search/README.rst#advanced-search

HaraldPanten:
Hi @carlosdauden maybe I didn't explain me enough.

I meant that I couldn't make a functional review. Is the module working properly for you? I just tried to search by "test" pricelist but I didn't find any result.

I tried to search it by test (USD) and I couldn't find any result, as well.

partner_pricelist_search/models/product_pricelist.py
```diff
@@ -0,0 +1,21 @@
...
+            domain = [("property_product_pricelist", "=", self.id)]
+        else:
+            domain = [("property_product_pricelist", "in", self.ids)]
+        partners = self.env["res.partner"].search(domain)
+        action = self.env.ref("base.action_partner_form")
+        res = action.read()[0]
+        res["domain"] = [
+            ("id", "in", partners.ids),
+        ]
+        return res
```
kos94ok-3D, line 21: I think it worth
```suggestion
    def show_pricelist_partners(self):
        action = self.env.ref("base.action_partner_form").read()[0]
        if len(self) == 1:
            res["domain"] =  [("property_product_pricelist", "=", self.id)]
        else:
            res["domain"] = [("property_product_pricelist", "in", self.ids)]
        return action
```

victoralmau (approved):
Code review OK

partner_pricelist_search/models/product_pricelist.py
carlosdauden: Sorry, I didn't see the suggestion.
Right now I don't remember why I did like this.
I will try that in the next PR or if you prefer you can do your PR and I approve it

PR #1070, opened 2021-02-26
[UPD]partner_fax:set field position after mobile

diggy128:
Having fax as the first field of communication data is confusing and
not as it was back when it was in core Odoo.
This changes the fax position to be right after mobile


dreispt (approved):

bodedra (commented):
LG :+1:

bodedra (approved):

HaraldPanten (approved):

PR #1076, opened 2021-03-10
14.0 mig partner multi relation

mostafaala2:


fernandahf:
@mostafaala2 

Is there a problem with this PR? or What is the reason it is closed?

Regards.

mostafaala2:
@fernandahf it does not passed the test cases

fernandahf:
@mostafaala2 

It's good know it, thanks.

PR #1077, opened 2021-03-11
[13.0][OU-MIG] partner_company_group:  Rename field in partner_group module merge into this one

sergio-teruel:
cc @Tecnativa
Script for rename group_id field from partner_group module to merge with partner_company_group

Merge module in "apriori"
https://github.com/OCA/OpenUpgrade/pull/2608


pedrobaeza (approved):

MiquelRForgeFlow:
I see a problem here. If you have in v12 installed both `partner_company_group` and `partner_group`, then you would have a conflict in this rename as the `company_group_id` column already exists.

MiquelRForgeFlow:
Done in https://github.com/OCA/partner-contact/pull/1100.

PR #1079, opened 2021-03-15
[FIX] partner_industry_secondary: remove a migration script

kos94ok-3D:


pedrobaeza:
Why are you removing this? It's needed.

kos94ok-3D:
Core does not do it?

pedrobaeza:
Not for this

kos94ok-3D:
Maybe it was fixed in core.
I tried installing the module on clean DB - all OK
I tried to migrate from 12 to 13 - all OK

pedrobaeza:
`parent_path` is only computed on save, but it doesn't work like a computed field. Maybe you have retriggered the computation through any action (or a record creation), but it' not done by regular ORM. Anyway, this is not very heavy, so what's the problem?

PR #1085, opened 2021-03-16
[CHG] partner_address_version: revamp functionality

kevinkhao:
This changes the functionality of https://github.com/OCA/partner-contact/commit/3ee4689fe00be63dc8a8d53c1a7600d33533a31c

Previously, we would specify table names and excluded column names (keys): it seems very easy to have side-effects from it, because we anytime we add a new res.partner M2O, then it will be automatically impacted by this module. It seems less error-prone instead of specifying excluded columns, to specify instead which columns to use.

For instance we will have problems if we also use this: 
https://github.com/OCA/sale-workflow/tree/14.0/sale_commercial_partner
The above module adds a new field on sale orders "commercial_partner_id" that checks for the parent company of the partner_id. This allows us to do charts and reports (e.g Mr. X and Ms. Y work at Agrolait, I don't care about Mr. X and Mr. Y, on my report I want to quickly see which sales were made to Agrolait and its children contact)

This PR proposes that we explicitly state which models, on which fields, we want to version.

This would make it much harder to get side effects.

@Cedric-Pigeon

### DIFF
--- a/partner_address_version/models/res_partner.py
+++ b/partner_address_version/models/res_partner.py
@@ -44,6 +44,23 @@ class ResPartner(models.Model):
         version_hash = hashlib.md5(str(version).encode("utf-8")).hexdigest()
         return version_hash
 
+    def _version_impacted_tables(self):
+        """
+        :return:
+            - list of tables to update in case of address versioning
+        """
+        return []
+
+    def _version_exclude_keys(self):
+        """
+        :return:
+            - dict:
+                key = table name
+                value = list of columns to ignore in case of address
+                        versioning
+        """
+        return {}
+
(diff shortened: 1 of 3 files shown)


kevinkhao:
I made a new commit to add compatibility for previous functionality. Now we should specify EITHER:
* impacted tables and/or excluded keys (columns), OR
* impacted columns (in another format)


If you implement both at the same time, you will get an error because it doesn't make sense.

kevinkhao:
This PR has grown beyond its original scope; there are some helper functions and some more tests.

kevinkhao:
@sebastienbeau

kevinkhao:
@sebastienbeau

PR #1092, opened 2021-03-19
[13.0][IMP] partner_identification duplicate

TDu:
At the moment, it is possible to create duplicate of number ID
for the same category.
Which not something that make sense.

So this change adds a constraint on the combo (category, id_number).

### DIFF
--- a/partner_identification/models/res_partner_id_number.py
+++ b/partner_identification/models/res_partner_id_number.py
@@ -68,3 +68,11 @@ class ResPartnerIdNumber(models.Model):
         ]
     )
     active = fields.Boolean(string="Active", default=True)
+
+    _sql_constraints = [
+        (
+            "unique_number_and_category",
+            "UNIQUE(category_id, name)",
+            "The combo category and ID number must be unique",
+        )
+    ]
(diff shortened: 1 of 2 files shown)


NL66278:
@TDu Please check travis errors. There are black and isort errors. You can prevent this by installing a pre-commit hook as explained here: https://odoo-community.org/blog/the-oca-blog-1/post/black-isort-pre-commit-97

TDu:
Pre-commit fixed, but now problem with the test, I will look at it

NL66278:
@TDu You have two problems:
- the module partner_identification_gln has its own check for duplicates. This one optionally allows duplicates if those are inactive; As your constraint throws a postgres error instead of a ValidationError, the test fails.
- the module partner_coc tests wether duplicates are allowed. I do not know why by the way.

So it will be difficult to get the SQL constraint through testing. If you still want the general constraint, I think it would be best to refactor the constraint if partner_identification_gln into a general constraint, only valid for the identification categories (will need an extra boolean field), that enable it.

TDu:
@NL66278 Thanks for your reviews.
Finally, maybe not the best approach to had this constraint, like this.
Will keep this open until I find a better solution, for what I need.

TDu:
This change was not a good idea see above comments and Travis results.
So I created this instead #1093

PR #1099, opened 2021-04-20
[10.0][BACKPORT] partner_company_group

LoisRForgeFlow:
Standard backport from v13.

@ForgeFlow


partner_company_group/models/account_invoice.py
```diff
@@ -0,0 +1,12 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import fields, models
+
+
+class AccountInvoice(models.Model):
+    _inherit = "account.invoice"
+
+    company_group_id = fields.Many2one(
+        related="partner_id.company_group_id", store=True
```
MiquelRForgeFlow, line 11:
```suggestion
        related="partner_id.company_group_id", store=True, readonly=True,
```
apply the same in other files

by default in v10, related and compute were readonly=False.

PR #1100, opened 2021-04-20
[13.0][FIX] partner_company_group: assure no error in migration

MiquelRForgeFlow:
Just in case you already have partner_company_group in previous versions.

### DIFF
--- a/partner_company_group/migrations/13.0.1.0.0/pre-migration.py
+++ b/partner_company_group/migrations/13.0.1.0.0/pre-migration.py
@@ -9,4 +9,14 @@ field_renames = [
 
 @openupgrade.migrate()
 def migrate(env, version):
-    openupgrade.rename_fields(env, field_renames)
+    if openupgrade.column_exists(env.cr, "res_partner", "group_id"):
+        if not openupgrade.column_exists(env.cr, "res_partner", "company_group_id"):
+            openupgrade.rename_fields(env, field_renames)
+        else:
+            openupgrade.logged_query(
+                env.cr,
+                """
+                UPDATE res_partner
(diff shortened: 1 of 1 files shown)


partner_company_group/migrations/13.0.1.0.0/pre-migration.py
```diff
@@ -9,4 +9,12 @@
...
 def migrate(env, version):
-    openupgrade.rename_fields(env, field_renames)
+    if openupgrade.column_exists(env.cr, "res_partner", "group_id"):
+        if not openupgrade.column_exists(env.cr, "res_partner", "company_group_id"):
+            openupgrade.rename_fields(env, field_renames)
+        else:
+            openupgrade.logged_query(
+                env.cr, """
+                UPDATE res_partner
+                SET company_group_id = group_id
```
pedrobaeza, line 20: The target model is the same?

MiquelRForgeFlow: yes, both are res.partner

MiquelRForgeFlow: I could add a query to check both field target same model, but that would make dirty the code

pedrobaeza: OK, just want to make sure of that.

pedrobaeza:
Please check pre-commit

MiquelRForgeFlow:
Updated :)

pedrobaeza (approved):

NL66278 (approved):
:+1: LGTM

PR #1106, opened 2021-05-18
[13.0][FIX] partner_manual_rank: call compute when rank is incremented

MiquelRForgeFlow:
The _increase_rank updates the rank by sql instead of using orm, so api.depends doesn't work.

I have tested the fix.

### DIFF
--- a/partner_manual_rank/models/res_partner.py
+++ b/partner_manual_rank/models/res_partner.py
@@ -47,3 +47,8 @@ class ResPartner(models.Model):
                 partners._increase_rank("supplier_rank")
             else:
                 partners.supplier_rank = 0
+
+    def _increase_rank(self, field):
+        super()._increase_rank(field)
+        if self.ids and field in ["customer_rank", "supplier_rank"]:
+            self.modified([field])


HviorForgeFlow (approved):

AaronHForgeFlow (approved):
LGTM

partner_manual_rank/models/res_partner.py
```diff
@@ -47,3 +47,12 @@ def _inverse_is_supplier(self):
...
                 partners.supplier_rank = 0
+
+    def _increase_rank(self, field):
+        super()._increase_rank(field)
+        map_compute_ranks = {
+            "customer_rank": "_compute_is_customer",
+            "supplier_rank": "_compute_is_supplier",
+        }
+        if self.ids and field in map_compute_ranks.keys():
+            getattr(self, map_compute_ranks[field])
```
LoisRForgeFlow: maybe use `modified`?

MiquelRForgeFlow: done

LoisRForgeFlow (approved):

JordiBForgeFlow (approved):

HaraldPanten (approved):

PR #1107, opened 2021-05-19
[IMP] partner_tier_validation (Auto Active and New def for "is_company")

patrickrwilson:
- Added new optional tier definition for is_company = true.
- Added auto active if no validation needed during create.
- Added reset validation during write for specific fields.

### DIFF
--- a/partner_tier_validation/data/tier_definition.xml
+++ b/partner_tier_validation/data/tier_definition.xml
@@ -1,12 +1,12 @@
 <odoo noupdate="1">
-    <record id="partner_tier_definition" model="tier.definition">
-        <field name="name">Partner Validation</field>
+    <record id="partner_tier_definition_company_only" model="tier.definition">
+        <field name="name">Partner Validation (Company)</field>
         <field name="model_id" ref="base.model_res_partner" />
         <field name="review_type">group</field>
         <field name="reviewer_group_id" ref="base.group_user" />
         <field name="definition_type">domain</field>
         <field
             name="definition_domain"
-        >["&amp;",["parent_id","=",False],"|",["active","=",True],["active","=",False]]</field>
+        >["&amp;",["is_company","=",True],"|",["active","=",True],["active","=",False]]</field>
     </record>
 </odoo>
--- a/partner_tier_validation/readme/DESCRIPTION.rst
+++ b/partner_tier_validation/readme/DESCRIPTION.rst
@@ -1,5 +1,12 @@
 Adds an approval workflow to Partners.
-The default rule requires new parent Contacts to be approved
+The default rule requires new company Contacts to be approved
 before they can be used.
+
+The rule can be extended to new non-company contact,
+but beware that may cause issues with automatically created new contacts,
+such as the ones generated when processing incoming emails.
+
+If the 'Is Company' or 'Parent' field changes then the contact is Request
+for approval.
 
 For this, the new Contact record is kept as "Archived" until it is approved.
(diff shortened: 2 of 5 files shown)


patrickrwilson:
Ping @dreispt

rousseldenis:
Maybe some tests should avoid regression ?

partner_tier_validation/data/tier_definition.xml
```diff
@@ -9,4 +9,15 @@
             name="definition_domain"
         >["&amp;",["parent_id","=",False],"|",["active","=",True],["active","=",False]]</field>
     </record>
+    <record id="partner_tier_definition_company_only" model="tier.definition">
+        <field name="name">Partner Validation (Company)</field>
+        <field name="model_id" ref="base.model_res_partner" />
+        <field name="review_type">group</field>
+        <field name="reviewer_group_id" ref="base.group_user" />
+        <field name="active">False</field>
```
dreispt: I feel it is OK to leave this, and have users explicitly enable it after installation.

partner_tier_validation/data/tier_definition.xml
```diff
@@ -9,4 +9,15 @@
             name="definition_domain"
         >["&amp;",["parent_id","=",False],"|",["active","=",True],["active","=",False]]</field>
     </record>
```
dreispt: @patrickrwilson This rule is known to cause issues, whenever Odoo automatically creates a new contact out of an email address. I still think is is not wise to include it.

partner_tier_validation/readme/DESCRIPTION.rst
```diff
@@ -2,4 +2,10 @@ Adds an approval workflow to Partners.
 The default rule requires new parent Contacts to be approved
 before they can be used.
 
+An optional rule requires new company Contacts to be approved
+before they can be used.
```
dreispt:
```suggestion
The default rule requires new company Contacts to be approved
before they can be used.

The rule can be extended to new non-company contact,
but beware that may cause issues with automatically created new contacts,
such as the ones generated when processing incoming emails.
```

dreispt (commented):
See comments.
Tests would be nice, maybe this can help: https://github.com/OCA/sale-workflow/blob/14.0/sale_tier_validation/tests/test_tier_validation.py

partner_tier_validation/data/tier_definition.xml
patrickrwilson: I will remove this rule then.

patrickrwilson: I'm fine with making this the disabled rule and the new one the default rule with your suggested changes to the description,

partner_tier_validation/data/tier_definition.xml
patrickrwilson: Since the other rule will be disabled, I will make this one the active one.

partner_tier_validation/data/tier_definition.xml
patrickrwilson: eh, I just decided to remove it like you originally suggested.

patrickrwilson:
> Maybe some tests should avoid regression ?

I will be working on tests, stay tuned.

patrickrwilson:
@dreispt @rousseldenis This is ready for review. For some reason codecov is complaining due to coverage on a different module so i'm not sure why it's showing red on this PR, below is a screenshot of my local instance coverage test.
image

dreispt (commented):
Thank you for adding the tests and for squashing commits.
I testes it and it works great!

dreispt (approved):

PR #1108, opened 2021-05-20
[14.0] [ADD] Added dependencies repo in oca-dependencies file.

Chandresh-OSI:
- `partner_tier_validation` module is depend on `base_tier_validation`, that is inside `server-ux` repo.

### DIFF
--- a/oca_dependencies.txt
+++ b/oca_dependencies.txt
@@ -1,1 +1,2 @@
 # See https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#oca_dependencies-txt
+server-ux


rousseldenis:
> * `partner_tier_validation` module is depend on `base_tier_validation`, that is inside `server-ux` repo.

Hi @Chandresh-OSI, the test flow >= v14 has changed as we use pip now to install modules (see https://github.com/OCA/partner-contact/blob/14.0/.travis.yml#L30)

If the module is already merged(and is on pypi), nothing more is required.

dreispt:
@rousseldenis
I had the idea that the oca_dependencies.txt should still be maintained, because there are deployment pipelines out there relying on it.
If it isn't, why is it there at all?

rousseldenis:
> @rousseldenis
> I had the idea that the oca_dependencies.txt should still be maintained, because there are deployment pipelines out there relying on it.
> If it isn't, why is it there at all?

I think (after discussion with Stephane) we are waiting to see if people are complaining about its 'voidness' before making it disappear completely

NL66278:
@Chandresh-OSI Do you still need this merged, or can this be closed?

Chandresh-OSI:
No need for merge. Closing PR. Thanks

PR #1119, opened 2021-07-08
[14.0][IMP+FIX] base_location

ivantodorovich:
A few usability improvements on `base_location`

Also fixes a bug introduced here: https://github.com/OCA/partner-contact/commit/5fd93fb8d138cb9a650410d49745b0777be43be0, more details on the commit messages

### DIFF
--- a/base_location/models/res_company.py
+++ b/base_location/models/res_company.py
@@ -33,7 +33,6 @@ class ResCompany(models.Model):
         inverse="_inverse_zip_id",
         help="Use the city name or the zip code to search the location",
     )
-
     country_enforce_cities = fields.Boolean(
         related="partner_id.country_id.enforce_cities"
     )
@@ -48,24 +47,20 @@ class ResCompany(models.Model):
         return res
 
     def _inverse_city_id(self):
-        for company in self:
-            company.with_context(
-                skip_check_zip=True
-            ).partner_id.city_id = company.city_id
+        for company in self.with_context(skip_check_zip=True):
+            company.partner_id.city_id = company.city_id
 
     def _inverse_zip_id(self):
-        for company in self:
-            company.with_context(skip_check_zip=True).partner_id.zip_id = company.zip_id
+        for company in self.with_context(skip_check_zip=True):
+            company.partner_id.zip_id = company.zip_id
 
     def _inverse_state(self):
-        return super(
-            ResCompany, self.with_context(skip_check_zip=True)
-        )._inverse_state()
+        self = self.with_context(skip_check_zip=True)
+        return super(ResCompany, self)._inverse_state()
 
     def _inverse_country(self):
-        return super(
-            ResCompany, self.with_context(skip_check_zip=True)
-        )._inverse_country()
+        self = self.with_context(skip_check_zip=True)
+        return super(ResCompany, self)._inverse_country()
 
     @api.onchange("zip_id")
     def _onchange_zip_id(self):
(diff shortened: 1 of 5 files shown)


pedrobaeza:
Please indicate the steps to reproduce the problem before the PR.

ivantodorovich:
> Please indicate the steps to reproduce the problem before the PR.

Ohh yes, sorry :sweat_smile: 

1. Configure a country with `enforce_cities`
2. Create a new partner, and choose the country you've configured

Expected: City should be a one2many (city_id) and not a char field (city).
Result: Not that


image

image



As explained on the commit message, this is due to the fact that we're adding city_id onto the view xml, and `_fields_view_get_address` in `base_address_city` won't do its magic if that's the case:

https://github.com/odoo/odoo/blob/e06aaecf9e8a590604818b09d23fe1de410e6762/addons/base_address_city/models/res_partner.py#L27-L33

base_location/models/res_company.py
```diff
@@ -48,24 +47,22 @@ def _get_company_address_field_names(self):
...
+        self = self.with_context(skip_check_zip=True)
         for company in self:
-            company.with_context(
-                skip_check_zip=True
-            ).partner_id.city_id = company.city_id
+            company.partner_id.city_id = company.city_id
 
     def _inverse_zip_id(self):
+        self = self.with_context(skip_check_zip=True)
         for company in self:
```
pedrobaeza:
```suggestion
        for company in self.with_context(skip_check_zip=True):
```

base_location/models/res_company.py
```diff
@@ -48,24 +47,22 @@ def _get_company_address_field_names(self):
         return res
 
     def _inverse_city_id(self):
+        self = self.with_context(skip_check_zip=True)
         for company in self:
```
pedrobaeza: Save one line and avoid the overwriting of the variable:
```suggestion
        for company in self.with_context(skip_check_zip=True):
```

base_location/models/res_company.py
```diff
@@ -48,24 +47,22 @@ def _get_company_address_field_names(self):
...
-            company.with_context(skip_check_zip=True).partner_id.zip_id = company.zip_id
+            company.partner_id.zip_id = company.zip_id
 
     def _inverse_state(self):
-        return super(
-            ResCompany, self.with_context(skip_check_zip=True)
-        )._inverse_state()
+        self = self.with_context(skip_check_zip=True)
```
pedrobaeza, line 58: This is not correct, as it's a local variable and it's not pass upstream. Please revert.

base_location/models/res_company.py
```diff
@@ -48,24 +47,22 @@ def _get_company_address_field_names(self):
...
+        return super()._inverse_state()
 
     def _inverse_country(self):
-        return super(
-            ResCompany, self.with_context(skip_check_zip=True)
-        )._inverse_country()
+        self = self.with_context(skip_check_zip=True)
```
pedrobaeza, line 62: The same.

base_location/models/res_company.py
ivantodorovich: It is passed :)

pedrobaeza: Well, I prefer not to do such things that depends on the Python implementation. Python zen: explicit is better than implicit.

ivantodorovich: IMO it's explicit but ok, it's changed

pedrobaeza (approved):

simahawk (approved):
LG

PR #1122, opened 2021-07-22
[14.0][ADD] partner_identification_stock.

Chanakya-OSI:


Chanakya-OSI:
Moved the field 'color' to the module 'partner_identification'.

patrickrwilson:
Depends on #1127

patrickrwilson:
@Chanakya-OSI Just an adjustment to the tests and it should be good after that.

patrickrwilson:
@Chanakya-OSI can you fix travis and make this green? maybe this just needs a rebase?

partner_identification_stock/models/stock_picking.py
```diff
@@ -0,0 +1,43 @@
...
+                    and i.valid_until
+                )
+                if any(
+                    line.valid_until >= fields.Date.today()
+                    and line.status in ["open", "pending"]
+                    for line in ffl_lines
+                ):
+                    rec.valid_id = True
+
+    require_id = fields.Boolean(string="Require ID?", compute="_compute_id_requirement")
```
rousseldenis: Fields names should be more specific in order to avoid collision.

require_partner_identification and valid_partner_identification should be more convenient.

partner_identification_stock/models/stock_picking.py
```diff
@@ -0,0 +1,43 @@
...
+    def action_confirm(self):
+        for rec in self:
+            if (
+                rec.require_id
+                and rec.state not in ["draft", "close"]
+                and not rec.valid_id
+            ):
+                raise UserError(
+                    _(
+                        "Cannot confirm a Sales Order without an active license."
```
rousseldenis: IMHO, you mixed terms in same flow ('Valid ID' in fields and 'license' here).

partner_identification_stock/models/stock_picking.py
```diff
@@ -0,0 +1,43 @@
...
+                    line.valid_until >= fields.Date.today()
+                    and line.status in ["open", "pending"]
+                    for line in ffl_lines
+                ):
+                    rec.valid_id = True
+
+    require_id = fields.Boolean(string="Require ID?", compute="_compute_id_requirement")
+    valid_id = fields.Boolean(string="Valid ID?", compute="_compute_id_requirement")
+
+    def action_confirm(self):
```
rousseldenis, line 35: As you require valid partner identification for 'confirmation', this should appears in module naming.

So partner_identification_stock_picking_confirmation is longer but better.

rousseldenis (changes requested):
Some things have to be clarified here.

Moreover, as this change stock flow, shouldn't it be in stock-logistics-workflow? But no blocking on this.

partner_identification_stock/models/stock_picking.py
Chanakya-OSI: We should not change the module name, as it might affect other modules (if any module depends on this one).

rousseldenis: > We should not change the module name, as it might affect other modules (if any module depends on this one).

This module is not yet merged. So, this can still be done

PR #1123, opened 2021-07-22
[14.0][ADD] partner_identification_sale.

Chanakya-OSI:


partner_identification_sale/README.rst
```diff
@@ -0,0 +1,74 @@
...
+.. |badge4| image:: https://img.shields.io/badge/weblate-Translate%20me-F47D42.png
+    :target: https://translation.odoo-community.org/projects/partner-contact-14-0/partner-contact-14-0-partner_identification_sale
+    :alt: Translate me on Weblate
+.. |badge5| image:: https://img.shields.io/badge/runbot-Try%20me-875A7B.png
+    :target: https://runbot.odoo-community.org/runbot/134/14.0
+    :alt: Try me on Runbot
+
+|badge1| |badge2| |badge3| |badge4| |badge5| 
+
+Adds Partner ID requirements to products with the option to prevent confirming sales orders if there is no valid ID.
```
rousseldenis, line 28: @Chanakya-OSI From your description, I think pulling the stock module is a little bit too much no  ?

patrickrwilson: @rousseldenis the stock module adds the page/sections to the product so this module would add to that which is why it's in the depends. Future modules for manufacturing and such would do the same.

rousseldenis: I understand, what I wanted to says, is this module that is named ..._sale should depend on stock ?

As Odoo did a sale_stock glue module, shouldn't we do the same and have a partner_identification_sale_stock?

We can imagine someone that has only sale installed and want to restrict sale confirmation with Ids (and does not want to install stock as it requires new fees for instance).

patrickrwilson: @rousseldenis I understand what you are saying now and agree this is a better approach. @Chanakya-OSI lets move forward with this suggestion. I think it would just be a matter of removing the stock dependency and changing the product views to be similar to the stock module where if you install this alone then the options would be on the products. Then create a new module called product.template.attribute.value changing the view so they coincide together. Please reach out to me if you have questions.

patrickrwilson:
Depends on #1127

patrickrwilson:
@Chanakya-OSI works better with last commit. Just need to fix the test as it looks like it's still referring to a stock field and we should be good to go.

patrickrwilson (approved):
This works well, thank you. LGTM once #1127 merges.

patrickrwilson:
@rousseldenis do you mind reviewing?

rousseldenis:
Same remarks as for https://github.com/OCA/partner-contact/pull/1122

PR #1127, opened 2021-08-06
[14.0][IMP] Added the field 'color' to partner_id_category.

Chanakya-OSI:


patrickrwilson:
@Chanakya-OSI Please fix pre-commit then this should be ready for review. Thanks for adding this.

patrickrwilson (approved):
LGTM

patrickrwilson:
@rousseldenis do you mind reviewing this?

ajaniszewska-dev (approved):

NL66278:
@Chanakya-OSI How and where is this color used? I see no way to set or to display it...

partner_identification/models/res_partner_id_category.py
```diff
@@ -18,6 +20,10 @@ class ResPartnerIdCategory(models.Model):
     _description = "Partner ID Category"
     _order = "name"
 
+    def _get_default_color(self):
+        return randint(1, 11)
+
+    color = fields.Integer(string="Color Index", default=_get_default_color)
```
rousseldenis, line 26: Better to use a lambda for default in order to allow override.

dreispt: IMO Better use a method to allow extending.

dreispt:
@Chanakya-OSI This PR is blocked because it is expecting ans answer from you. Thank you.

Chanakya-OSI:
> @Chanakya-OSI How and where is this color used? I see no way to set or to display it...



> @Chanakya-OSI How and where is this color used? I see no way to set or to display it...

This color field is used in the modules 'partner_identification_sale' as well as 'partner_identification_stock'.

partner_identification/models/res_partner_id_category.py
Chanakya-OSI: Yes keeping method for allowing extending.

PR #1130, opened 2021-08-31
[14.0][ADD] DEA and Medical License.

Murtaza-OSI:


andhit-r:
@Murtaza-OSI IMHO you should create a separate module to add different kind of identification. Take a look of this OCA module https://github.com/OCA/partner-contact/tree/13.0/partner_identification_gln

NL66278:
@Murtaza-OSI I agree with @andhit-r here. There are lots of possible license and identification type. One common module to manage them, with extra modules to add new types, will be the way to go.

Murtaza-OSI:
Created Extend Module

PR #1136, opened 2021-09-15
[13.0][partner_affiliate][imp] modify partner views to allow you to edit the address and vat of an affiliate.

JordiBForgeFlow:
Currently the address and VAT fields are read-only by the time you choose a parent. That makes sense when you are linking a contact to a parent company, but not when you are linking a company to a parent company.

@ForgeFlow


NL66278 (approved):
:+1: LGTM

MiquelRForgeFlow:
Forward ported to https://github.com/OCA/partner-contact/pull/1165.

PR #1138, opened 2021-09-23
[14.0][MIG] partner_pricelist_search: Migration to 14.0

victoralmau:
Migration to 14.0.

Please @carlosdauden and @joao-p-marques can you review it?

@Tecnativa TT31807


partner_pricelist_search/models/res_partner.py
```diff
@@ -0,0 +1,43 @@
...
+class Pricelist(models.Model):
+    _inherit = "res.partner"
+
+    property_product_pricelist = fields.Many2one(
+        search="_search_property_product_pricelist"
+    )
+
+    @api.model
+    def _search_property_product_pricelist(self, operator, value):
+        if operator == "=":
```
pedrobaeza, line 17: I think you can use the same technique as used in other module. @CarlosRoca13 can you point to the module with the operator import?

partner_pricelist_search/readme/CONTRIBUTORS.rst
```diff
@@ -0,0 +1,3 @@
+* `Tecnativa <https://www.tecnativa.com>`_:
+
+  * Carlos Dauden
```
pedrobaeza, line 3: Put yourself here

partner_pricelist_search/models/res_partner.py
CarlosRoca13: Yes, sure, here you have the PR were was added that logic https://github.com/OCA/stock-logistics-warehouse/pull/1246

pedrobaeza: After seeing it, it's not as easy and this code is enough.

pedrobaeza (approved):

joao-p-marques (approved):

carlosdauden (approved):

PR #1103, opened 2021-05-03
[13.0][MIG] partner_create_by_vat

NL66278:
Migration done by @feketemihai. Just squashed the translation commits.

PR #1140, opened 2021-09-26
[FIX BUILD!] base_location_geonames_import: apparently Geonames zip files changed download location

dreispt:
Resolve build error https://app.travis-ci.com/github/OCA/partner-contact/jobs/539640838#L1574,
found on https://github.com/OCA/partner-contact/pull/1103.

### DIFF
--- a/base_location_geonames_import/readme/CONFIGURE.rst
+++ b/base_location_geonames_import/readme/CONFIGURE.rst
@@ -2,5 +2,5 @@ To access the menu to import city zip entries from Geonames
 you must be part of the group *Administration / Settings*.
 
 If you want/need to modify the default Geonames URL
-(http://download.geonames.org/export/zip/), you can set the *geonames.url*
+(http://download.geonames.org/export/dump/), you can set the *geonames.url*
 system parameter.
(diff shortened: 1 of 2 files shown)


pedrobaeza:
This is a recurrent thing, that Geonames screws up the automatic file generation, and there is some time where it's not available, but dump files has nothing to do with the needed files for location, as they contain different information. You just need to wait until the files are up again. You can fill an issue to Geonames for not removing files until they are properly generated, for not having the directory empty meanwhile.

dreispt:
Having tests fail intermittently is not a good thing.
IMO tests should then be fixed to use a mockup of the service, or similar.

pedrobaeza:
This debate has taken place a lot of times. Mocking this will mean that you are tricking for having green CI, but the reality is that if you use the module for downloading names, it will fail. I know it's an annoyance when it's not good due to external reasons (that's why I think we should push Geonames for not having this recurrent problem), but mocking in test for me is just a shortcut to having a non fully tested system. If a module serves to download from Geonames the locations, and this is not working, why tests should go green?

pedrobaeza:
I have written in the Geonames mailing list to see if another method can be put for not emptying the directory. Anyways, I don't want to impose my criteria, so if most of you wants to mock it, go ahead.

PR #1142, opened 2021-09-29
[14.0][FIX] partner_tier_validation: Draft partners not inactive

RLeeOSI:


RLeeOSI:
@dreispt  Changing partners to inactive is breaking core Odoo functionality. If a partner has a user linked, setting the partner to inactive raises an error, making it functionally impossible to change any revalidation fields on a partner that is also a user. Itstead, I would suggest that this module should be extended for each application, adding a domain filter on fields with state = "confirmed" that will prevent unvalidated partners from being selected on those fields.

partner_tier_validation/__manifest__.py
```diff
@@ -3,7 +3,7 @@
 {
     "name": "Partner Tier Validation",
     "summary": "Support a tier validation process for Contacts",
-    "version": "14.0.2.0.0",
+    "version": "14.0.2.1.0",
```
rousseldenis: You don't need to do that but you have migration scripts for instance.

It will be done through ocabot

RLeeOSI: Fixed, thank you

dreispt:
@RLeeOSI I'm not sure I'm comfortable with creating an extension for each application.
What if the inactivation only happens for Partner record meeting the "Is Customer" or "Is Supplier" criteria (this is doable).
Other partner records, such as users and contacts wouldn't be affected.

RLeeOSI:
@dreispt  That might work, I can test that. What do you think about the possibility of making this an optional configuration? We could add a bool in ir.config.parameters that decides whether unvalidated partners should be made inactive

dreispt:
@RLeeOSI @rousseldenis FYI setting the `active`flag was not a good idea. So #1216 does a refactor removing that (supersedes this PR)

RLeeOSI:
@dreispt  I will close this PR then and switch customer code where appropriate. I should have a chance to get to this Monday 2/7

rafaelbn:
@dreispt @RLeeOSI @Freni-OSI hello! Are you going to go ahead with this PR o can we close it? Thank you!

RLeeOSI:
@rafaelbn  Thanks for the followup. PR is closed for #1216

PR #1143, opened 2021-09-30
[FIX] partner_multi_relation: Avoid access error on Relations button

tobiaszehntner:
In Odoo v14, ir.actions.act_window have only access rights for admin while a user had read access in previous versions. Odoo provides now a function (`_for_xml_id`) that returns the action read with sudo.

To reproduce: As standard user that has a relation, click on the "Relations" magic button on the contact form. Without this fix it throws an access error:
```
You are not allowed to access 'Action Window' (ir.actions.act_window) records.
This operation is allowed for the following groups:
	- Administration/Settings
```

### DIFF
--- a/partner_multi_relation/models/res_partner.py
+++ b/partner_multi_relation/models/res_partner.py
@@ -187,9 +187,9 @@ class ResPartner(models.Model):
                     ("other_partner_id", "=", contact.id),
                 ]
             )
-            action = self.env.ref(
+            action = self.env["ir.actions.act_window"]._for_xml_id(
                 "partner_multi_relation.action_res_partner_relation_all"
-            ).read()[0]
+            )
             action["domain"] = [("id", "in", relation_ids.ids)]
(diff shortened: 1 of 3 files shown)


theangryangel:
Thanks @tobiaszehntner, I was literally going to make the same PR this afternoon <3

Tested and LGTM.

NL66278 (approved):
:+1: LGTM

NL66278:
@tobiaszehntner If you can add a simple test, calling action_view_relations() on the model, we can get rid of the coverage message. If you do this with a user that is not admin, we will have a valid test for this functionality...

tobiaszehntner:
@theangryangel @NL66278 thank you for the speedy response, that was smooth! 👌

NL66278:
Thanks for your contribution. On the last run I noticed that buildout had failed for no reason and with no logging - something that has surprised me before. As the fix was quite clear, I decided to merge right away.

PR #1144, opened 2021-09-30
Disable sms on Fax-fields

bertstomphorst:
Fax-numbers are not used for SMS, so showing the sms-button ahead of it is a bit confusing for end-users.


HaraldPanten:
Hi @bertstomphorst you should check pre-commit. 👍

bertstomphorst:
Hi @HaraldPanten thanx, I just did, and committed a pretty-print change. Not really sure if this is what's expected?

bertstomphorst:
I give up, for me it's totally unclear why pre-commit fails on pretty-print-task.

PR #1148, opened 2021-10-08
[14.0][MIG] partner_deduplicate_filter: Migration to 14.0

victoralmau:
Migration to 14.0

Please @pedrobaeza and @joao-p-marques, can you review it?

@Tecnativa TT30292


rousseldenis:
@victoralmau Could you check tests?

partner_tier_validation/hooks.py
```diff
@@ -0,0 +1,15 @@
+# Copyright 2021 Tecnativa - Víctor Martínez
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+from odoo import SUPERUSER_ID
+from odoo.api import Environment
+from odoo.tools import config
+
+
+def post_init_hook(cr, pool):
+    if not config["test_enable"]:
```
pedrobaeza, line 13: Explain why are we doing this.

pedrobaeza (commented):
First line commit message should be very short, and then explain in detail on the rest of the lines.

https://www.odoo.com/documentation/15.0/developer/misc/other/guidelines.html#commit-message-structure

Example for here:

```
[FIX] partner_tier_validation: Defer rule activation for compatibility

Having this rule activated by default in tests environments makes other tests to fail due to the additional constraint not expected by them, so what we are doing is:

- Having the rule deactivated by default.
- On a post-init hook, we activate it only if we are not in a test environment.
- On own modules tests, we activate it as well

TT...
```

pedrobaeza:
I see that you don't put the TT as well on the migration commit.

victoralmau:
Changes done.

pedrobaeza (approved):

joao-p-marques (approved):

PR #1150, opened 2021-10-12
[14.0][FIX] base_location: address field sync zip_id

ivantodorovich:
`_address_fields` are used by `update_address` [^1], which is in turn called by `_fields_sync` and `_children_sync`.

This mechanism is in charge of synchronizing parent/children address fields for contacts of type == "contact", as they don't really have an address of their own. They simply copy the address of their parent.

Without this fix, not only the zip_id is not sync'ed, but also in some cases an exception may be raised by `_check_zip_id`, due to only some fields being sync'ed but not all, creating inconsistencies between zip_id and other address fields.

[^1]: https://github.com/odoo/odoo/blob/20648ef21/odoo/addons/base/models/res_partner.py#L429

### DIFF
--- a/base_location/tests/test_base_location.py
+++ b/base_location/tests/test_base_location.py
@@ -204,6 +204,29 @@ class TestBaseLocation(common.SavepointCase):
         self.company._onchange_state_id()
         self.assertEqual(self.company.country_id, self.company.state_id.country_id)
 
+    def test_partner_address_field_sync(self):
+        """Test that zip_id is correctly synced with parent of contact addresses"""
+        parent = self.env["res.partner"].create(
+            {
+                "name": "ACME Inc.",
+                "is_company": True,
+                "street": "123 Fake St.",
+                "city": "Springfield",
+                "state_id": self.barcelona.state_id.id,
+                "country_id": self.barcelona.country_id.id,
+                "zip_id": self.barcelona.id,
+            }
+        )
+        contact = self.env["res.partner"].create(
+            {
+                "name": "John Doe",
+                "type": "contact",
+                "parent_id": parent.id,
+            }
+        )
+        parent.zip_id = self.lausanne
+        self.assertEqual(contact.zip_id, self.lausanne, "Contact should be synced")
+
     def test_display_name(self):
         """Test if the display_name is stored and computed properly"""
         self.assertEqual(
--- a/base_location/models/res_partner.py
+++ b/base_location/models/res_partner.py
@@ -146,3 +146,7 @@ class ResPartner(models.Model):
         for node in doc.xpath("//field[@name='zip_id']"):
             node.attrib["domain"] = self._zip_id_domain()
         return etree.tostring(doc, encoding="unicode")
+
+    @api.model
+    def _address_fields(self):
+        return super()._address_fields() + ["zip_id"]


simahawk (approved):

base_location/tests/test_base_location.py
```diff
@@ -204,6 +204,14 @@ def test_company_onchange_state(self):
         self.company._onchange_state_id()
         self.assertEqual(self.company.country_id, self.company.state_id.country_id)
 
+    def test_partner_address_field_sync(self):
+        """Test that zip_id is correctly synced with parent of contact addresses"""
+        partner = self.env.ref("base.res_partner_address_1")
```
pedrobaeza: Don't use demo data in tests.

base_location/tests/test_base_location.py
```diff
@@ -204,6 +204,14 @@ def test_company_onchange_state(self):
...
+        parent = partner.parent_id
+        self.assertEqual(partner.type, "contact")
```
pedrobaeza: You shouldn't test features outside the module, and this is also needed because of using demo data.

pedrobaeza (changes requested):

base_location/tests/test_base_location.py
simahawk: There's nothing wrong w/ using demo data in tests unless you expect this data to be "fragile" (eg: can be trashed).
PRO: you don't create any other record (better for perf)
CON: test might be broken (but in this case is unlikely)

base_location/tests/test_base_location.py
simahawk: I don't think he's testing anything here. Such statements can be used simply to ensure that the state of an obj is what you expect and let the test fail if the original record changed. See above for the pro and con.

base_location/tests/test_base_location.py
ivantodorovich: ☝️ we gotta save the penguins

pedrobaeza: Yes, they are constantly broken through versions, and make tests fail without being a real problem on the module when you migrate such module. If Odoo has finally removed this burden, please don't introduce it again here, at least on the modules I maintain.

base_location/tests/test_base_location.py
pedrobaeza: Any assert must be for testing module features. Requiring this for ensuring previous data state is a consequence of using demo data or not preparing correctly your test data. Please go for the other approach.

pedrobaeza: Look a direct problem I have found for asserting things that are not of the module itself:

https://github.com/OCA/l10n-spain/commit/a205e3f80148d871c9544b1106553d5e304c9d20#diff-3b8f173a5205124604ea5cef331b3e906af2e0a743921c59dd5d78a63ed3ca5cL95

This was re-testing things of the main module (because it's the expected thing), and when this main module changes due to OCA/bank-payment#860, then it breaks without being a problem of this module.

ivantodorovich:
@pedrobaeza tests updated

base_location/tests/test_base_location.py
```diff
@@ -204,6 +204,29 @@ def test_company_onchange_state(self):
...
+        """Test that zip_id is correctly synced with parent of contact addresses"""
+        parent = self.env["res.partner"].create(
+            {
+                "name": "ACME Inc.",
+                "is_company": True,
+                "street": "123 Fake St.",
+                "city": "Springfield",
+                "state_id": self.barcelona.state_id.id,
+                "country_id": self.barcelona.country_id.id,
+                "zip_id": self.barcelona.id,
```
pedrobaeza, line 217: I think all of these fields are not needed to be filled for the purpose of the test, but I don't want to annoy you more.

pedrobaeza (approved):

base_location/tests/test_base_location.py
simahawk: Don't get me wrong: I do think tests' data should be independent. In this simple cases IMO you can ignore this.
Anyway, as you are the maintainer you have the right to ask/advise for such changes.
Thanks for pointing this out and providing references :wink:

pedrobaeza: Yeah, but what you think it can be simple, it can go complicated depending on what Odoo decides to do. It's not the first time Odoo changes contact demo data, and as the names are simply numbered references, on next version `base.res_partner_address_1` is a contact with no parent and the test blows away. The cost of creating the 2 records in this case is very low IMO. I would doubt for another more complex data, but in the end, my experience having to repair tests (even in the middle of the version) due to the demo data says that it always compensates.

PR #1151, opened 2021-10-14
[14.0][FIX] base_location: security - all users should be able to read

ivantodorovich:
For some reason, this access was removed during the migration to 12.0 (67baf9196).
But this security is perfectly valid and aligned to Odoo core's security for `res.city`[^1].

Without this, you might get bitten while trying to implement website features that
require reading or setting partner's zip_id, for example.

[^1]: https://github.com/odoo/odoo/blob/f5a5d9616/addons/base_address_city/security/ir.model.access.csv


pedrobaeza:
Although core is that way, don't you think is better to limit at least to internal users?

ivantodorovich:
> Although core is that way, don't you think is better to limit at least to internal users?

Well.. it's not sensible data, and core is like this for `res.country.group`, `res.country`, `res.country.state`, `res.city`.. I think it's good to be consistent here, as that's what you'd expect ~as a developer~ IMO.

Also in our use case we really need public access, not only internal users, as you would for website features

pedrobaeza (approved):
OK

simahawk (approved):

PR #1157, opened 2021-10-28
[12.0][FIX] partner recursion check

sergiocorato:
Current behavior: it is possible to select the partner itself as its company group
Desired behavior: recursion is not allowed

### DIFF
--- a/partner_company_group/models/res_partner.py
+++ b/partner_company_group/models/res_partner.py
@@ -1,7 +1,8 @@
 # Copyright 2019 Camptocamp SA
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
-from odoo import models, fields
+from odoo import api, models, fields, _
+from odoo.exceptions import ValidationError
 
 
 class Contact(models.Model):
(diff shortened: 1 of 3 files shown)


partner_company_group/models/res_partner.py
```diff
@@ -13,5 +14,11 @@ class Contact(models.Model):
...
     )
 
+    @api.constrains('company_group_id')
+    def _check_company_group_id(self):
+        if not self._check_recursion('company_group_id'):
+            raise ValidationError(
+                _('Error! You cannot create recursive partner company group'))
+
     def _commercial_fields(self):
         return super(Contact, self)._commercial_fields() + ['company_group_id']
```
NL66278: Just use super(), this is python 3.x, repeating the class name with self is no longer needed.

NL66278 (approved):
:+1: LGTM. You can either take up my other comment, or ignore it, as it is outside of the scope of what you want to fix.

partner_company_group/models/res_partner.py
sergiocorato: Ook, I was not changing that part, but I will do, thanks

sergiocorato:
> +1 LGTM. You can either take up my other comment, or ignore it, as it is outside of the scope of what you want to fix.

+1

PR #1159, opened 2021-10-29
[11.0][FIX] partner_tag_smart_assignation: invalid constrains.

NL66278:
@constrains does not support dotted field names.


NL66278:
If no body objects I will merge the first week of november, because the patch is quite straightforward, and this is causing PR's for 11.0 to get a red cross....

ecino (approved):

PR #1160, opened 2021-10-30
[14.0] Add module partner_email_duplicate_warn

alexis-via:
I love these warning banners: they are a great tool to fight duplicates while avoiding to completely block the creation of a new partner in case another partner has the same email.
partner_warn_banner


partner_email_duplicate_warn/models/res_partner.py
```diff
@@ -0,0 +1,48 @@
...
+                partner_id = partner._origin.id
+                if partner_id:
+                    domain += [
+                        ("id", "!=", partner_id),
+                        "!",
+                        ("id", "child_of", partner_id),
+                        "!",
+                        ("id", "parent_of", partner_id),
+                    ]
+                search_partners = self.with_context(active_test=False).search(domain)
```
sebastienbeau, line 40: Maybe I am wrong but in the search_partners result all email will match and will have a value ?
if yes why not just using 

```
self.with_context(active_test=False).search(domain, limit=1)
```

mourad-ehm: Thus 
`partner.same_email_partner_id = self.with_context(active_test=False).search(domain, limit=1)`  ?

acalmett (approved):
Used in production, no code review

partner_email_duplicate_warn/models/res_partner.py
rousseldenis: Same remark as in #1161

alexis-via: When you think about it, you have to take into account the fact that you can have a space before and/or after the email in the 'email' field in the DB, because Odoo doesn't auto-strip char fields upon write(). That's why your proposal with a simple self.with_context(active_test=False).search(domain, limit=1) doesn't work. The '%' char in the domain matches any character and I'm not aware of any special char in the domain that would match only spaces (I would be very useful BTW).
Imagine you have 2 partners with emails :
1) '<space>alexis@via.fr'
2) 'alexis@via.fr.com'
With the code you propose, Odoo would show the duplicate warning. With my code, it won't show the duplicate warning.

alexis-via:
@sebastienbeau @mourad-ehm @rousseldenis Are you ok with my explainations? If yes, could you approve the PR?

sebastienbeau (approved):
Lgtm code review

alexis-via:
Anybody around with the super-powers ?

PR #1161, opened 2021-10-30
[14.0] Add module partner_mobile_duplicate_warn

alexis-via:
Did I already say that I love these warning banners ?
partner_warn_banner


alexis-via:
I developed an equivalent module for email duplicates in https://github.com/OCA/partner-contact/pull/1160 The code is not exactly the same, because each field has its specifics (for example: I consider that several employees of the same company can share the same email, but they can't share the same mobile). Also, in the email field, the content is now always stripped ; for the Mobile field, with phone_validation, the content is always stripped.

mourad-ehm (approved):
Code review

acalmett (approved):
Used in production, no code review.

partner_mobile_duplicate_warn/models/res_partner.py
```diff
@@ -0,0 +1,40 @@
...
+                if partner.company_id:
+                    domain += [
+                        "|",
+                        ("company_id", "=", False),
+                        ("company_id", "=", partner.company_id.id),
+                    ]
+                partner_id = partner._origin.id
+                if partner_id:
+                    domain.append(("id", "!=", partner_id))
+                same_mobile_partner = self.with_context(active_test=False).search(
```
rousseldenis, line 36: I'm wondering if a read_group before the loop would be faster ?

rousseldenis: @alexis-via

alexis-via: I wouldn't have partner._origin.id via a read_group, would I ?
I don't see how a read_group() would be faster here because "self" already contains the recordsets.

NL66278, line 36: Why the `active_test=False`? Maybe unlikely that a phone number might be reused by really another person / company in the same odoo database, but phonenumber can be reused. What is even more likely is that somebody changes his identity (was a single person, now has a company, or the other way around, but keeps the phone). Then not really blocking as only a warning is given.

alexis-via: My idea was to be able to identify a partner that was archived, but now needs to be unarchived instead of creating a new partner. But your arguments can also be accepted, so if other people also think we should remove with_context(active_test=False), I'm ok to remove it.

NL66278: @alexis-via Now I understand the use case I am quite satisfied with leaving it as is.

NL66278 (approved):
:+1: LGTM

alexis-via:
@rousseldenis Are you ok on this PR now? Can you approve/merge it ?

rousseldenis (approved):

PR #1165, opened 2021-11-02
[14.0][IMP] partner_affiliate: forward port improvement

MiquelRForgeFlow:
Forward port of https://github.com/OCA/partner-contact/pull/1136.

### DIFF
--- a/partner_affiliate/models/res_partner.py
+++ b/partner_affiliate/models/res_partner.py
@@ -2,7 +2,7 @@
 # Copyright 2018 brain-tec AG - Raul Martin
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
-from odoo import fields, models
+from odoo import api, fields, models
 
 
 class ResPartner(models.Model):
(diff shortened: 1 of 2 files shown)


fredzamoabg (approved):

partner_affiliate/models/res_partner.py
```diff
@@ -22,3 +22,10 @@ class ResPartner(models.Model):
         string="Affiliates",
         domain=[("active", "=", True), ("is_company", "=", True)],
     )
+
+    @api.onchange("parent_id")
```
fredzamoabg, line 26: explicit `api.onchange `here is not needed afaik.

partner_affiliate/models/res_partner.py
```diff
@@ -22,3 +22,10 @@ class ResPartner(models.Model):
...
+    def onchange_parent_id(self):
+        # If it is a company and it has a parent, do not change the address.
+        if self.parent_id and self.is_company:
+            return {}
+        return super(ResPartner, self).onchange_parent_id()
```
fredzamoabg, line 31: you can directly call `super()` without params I mean

PR #1166, opened 2021-11-04
[14.0][ADD] phone_validation_configure

remi-filament:
This module extends the functionality of phone_validation default Odoo module.
phone_validation is configured by default to use international phone numbers and to not raise exceptions in case phone number is incorrect.

This modules allows you to configure both phone format and whether or not to raise exceptions.


leemannd (approved):

jordancoest (approved):

phone_validation_configure/__manifest__.py
```diff
@@ -0,0 +1,18 @@
+{
+    "name": "Phone Validation Configuration",
+    "summary": "Configure format and exceptions for phone numbers",
+    "version": "14.0.1.0.0",
+    "development_status": "Production/Stable",
+    "website": "https://github.com/OCA/partner-contact",
+    "author": "Le Filament, Odoo S.A., Odoo Community Association (OCA)",
+    "maintainers": ["remi-filament"],
+    "license": "GPL-3",
```
rousseldenis, line 9: Either AGPL or LGPL

rousseldenis (commented):
Minor comment @remi-filament

phone_validation_configure/__manifest__.py
remi-filament: Hi @rousseldenis, thank you for your comments, I put GPL-3 here on purpose, since I am reusing LGPL code from Odoo (this function I am overriding : https://github.com/odoo/odoo/blob/2632e9e553cc50e4273a0150bef67818b83ba8de/addons/phone_validation/models/phone_validation_mixin.py#L17-L27)

According to https://www.gnu.org/licenses/gpl-faq.en.html#AllCompatibility I can redistribute LGPLv3 code as GPLv3 but I am not sure whether I can directly use AGPL-3 - although I see the final sentence before the compatibility matrix stating : 

> Each place that the matrix states GPLv3, the same statement about compatibility is true for AGPLv3 as well.

I prefer to use either GPL or AGPL as less permissive than LGPL, but I am sure this has already been done in OCA, so any guidance here would be appreciated !
In OCA documentation I can only find sentence about backporting [here](https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#id60) but not about rewriting...

Best Regards,

JulianaPoudou (approved):

phone_validation_configure/__manifest__.py
rousseldenis: I see this: https://odoo-community.org/page/faq

The question come as it's quite unusual to have GPL in OCA modules. AGPL or LGPL is more common.

PR #1167, opened 2021-11-05
partner_identification_gln: migrate to v14

simahawk:


partner_identification_gln/tests/test_gln.py
```diff
@@ -6,19 +6,20 @@
...
 class TestGLN(SavepointCase):
-    def setUp(self):
-        super(TestGLN, self).setUp()
-        self.partner = self.env["res.partner"].create({"name": "TestGLN"})
-        self.partner2 = self.env["res.partner"].create({"name": "TestGLN2"})
-        pc = self.env.ref(
-            "partner_identification_gln." "partner_identification_gln_number_category"
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
```
TDu, line 11: `cls.env = cls.env(context=dict(cls.env.context, tracking_disable=True))`
Maybe can be added as well ?

TDu (approved):

partner_identification_gln/tests/test_gln.py
simahawk: good tip :wink:

mmequignon (approved):

partner_identification_gln/data/partner_identification_gln.xml
```diff
@@ -0,0 +1,25 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<!-- Copyright 2016-2017 ACSONE SA/NV (<http://acsone.eu>)
+     License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html). -->
+<odoo>
+    <record
+        id="partner_identification_gln_number_category"
+        model="res.partner.id_category"
+    >
+        <field name="name">GLN Identification Number</field>
+        <field name="code">gln_id_number</field>
```
simahawk, line 10: @rousseldenis @sbejaoui question: why this code is not `GLN`? Seems weird and likely it won't automatically support common scheme IDs like `schemeID="GLN"` (eg: ubl) which might be used by `partner_identification_import`.

If you are supposed to add your own records and these here only serve as an example, I'd move them to demo.

simahawk: Extract from PEPPOL docs

image

sebalix (approved):

rousseldenis (approved):

partner_identification_gln/data/partner_identification_gln.xml
simahawk: @rousseldenis what about this comment? :)

PR #1176, opened 2021-11-10
Test pre-commit fix from https://github.com/OCA/oca-addons-repo-template/pull/89

joao-p-marques:


joao-p-marques:
Seems to work! Closing then...

ping @pedrobaeza @sbidoul

PR #1177, opened 2021-11-12
[15.0][MIG] migration partner_firstname

bjornbillen:


bjornbillen:
Can someone tell me why the builds are still failing?

pedrobaeza:
Don't worry about runbot failing. It's not your fault and it's not taken into account either.

partner_firstname/hooks.py
```diff
@@ -0,0 +1,9 @@
+# Copyright 2017 LasLabs Inc.
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import SUPERUSER_ID, api
+
+
+def post_init_hook(cr, registr):
```
BT-rmartin: Why adding registr as second parameter and not using anymore with api.Environment.manage()?

partner_firstname/models/res_partner.py
```diff
@@ -0,0 +1,222 @@
...
+                if not vals.get(key) or context.get("copy"):
+                    vals[key] = value
+
+            # Remove the combined fields
+            if "name" in vals:
+                del vals["name"]
+            if "default_name" in context:
+                del context["default_name"]
+        # pylint: disable=W8121
+        return super(ResPartner, self.with_context(context)).create(vals)
```
BT-rmartin, line 51: self.with_context(**context)) might work instead of disabling the pylint check?

partner_firstname/tests/test_create.py
```diff
@@ -0,0 +1,80 @@
...
+            self.good_values["firstname"], self.good_values["lastname"]
+        )
+        if "default_is_company" in self.context:
+            self.good_values["is_company"] = self.context["default_is_company"]
+        self.values = self.good_values.copy()
+
+    def tearDown(self):
+        # pylint: disable=W8121
+        self.record = (
+            self.env[self.model].with_context(self.context).create(self.values)
```
BT-rmartin: Does it make any sense here replacing the whole context by itself?
I would remove with_context(self.context) more than disabling the pylint check

partner_firstname/tests/test_delete.py
```diff
@@ -0,0 +1,39 @@
...
+        """Test what happens if recomputed after unlinking.
+
+        This test might seem useless, but really this happens when module
+        ``partner_relations`` is installed.
+
+        See https://github.com/OCA/partner-contact/issues/154.
+        """
+        data = {"name": "Söme name"}
+        # pylint: disable=W8121
+        record = self.env[self.model].with_context(self.context).create(data)
```
BT-rmartin: with_context(**self.context) to avoid the pylint check issue more than disabling the check

partner_firstname/tests/test_empty.py
```diff
@@ -0,0 +1,74 @@
...
+    """Test ``res.partner`` when it is a company."""
+
+    model = "res.partner"
+    context = {"default_is_company": True}
+
+    def tearDown(self):
+        try:
+            data = {"name": self.name}
+            # pylint: disable=W8121
+            model = self.env[self.model].with_context(self.context)
```
BT-rmartin: with_context(**self.context) to avoid the pylint check issue more than disabling the check, or even directly removing the with_context. Why is it needed here?

partner_firstname/tests/test_name.py
```diff
@@ -0,0 +1,90 @@
...
+
+        # Need this to refresh the ``name`` field
+        self.original.invalidate_cache(["name"])
+
+
+class PartnerCompanyCase(BaseCase):
+    def create_original(self):
+        res = super(PartnerCompanyCase, self).create_original()
+        self.original.is_company = True
+        return res
```
BT-rmartin, line 58: You are adding this return for avoiding the pylint check missing-return, but the parent create_original has no return. 
In this case and the others inside this class I would prefer to disable the pylint check

partner_firstname/views/res_partner.xml
```diff
@@ -0,0 +1,79 @@
+<odoo>
+    <record id="view_partner_form" model="ir.ui.view">
```
BT-rmartin, line 34: Why did you delete the old extension of view_partner_simple_form? I think it's still needed as well as this other extension

partner_firstname/views/res_partner.xml
```diff
@@ -0,0 +1,79 @@
...
+        <field name="model">res.partner</field>
+        <field name="inherit_id" ref="base.view_partner_form" />
+        <field name="priority" eval="99" />
```
BT-rmartin: Why do you need a priority?

partner_firstname/views/res_partner.xml
```diff
@@ -0,0 +1,79 @@
...
+        <field name="arch" type="xml">
+                <xpath
+                expr="//div[hasclass('oe_title')]//field[@name='name'][1]"
```
BT-rmartin: Replacing and using [1] are both bad practices in general. 
Here the problem is the core view added now 2 times the name, with id individual and company. 
The point is there updating both, not removing one. 
This is needed to set the field as readonly, so that when name is given a value through the computed field as soon as you are changing the firstname or lastname, it doesn't arrive to the create, that would call to the reverse function ending up in a wrong assignment of firstname and lastname. 
`<xpath expr="//field[@id='individual']" position="attributes">
                    <attribute name="attrs">
                        {
                        'readonly': [('is_company', '=', False)],
                        'required': [('type', '=', 'contact'), ('is_company', '=', True)]
                        }
                    </attribute>
                </xpath>
                <xpath expr="//field[@id='company']" position="attributes">
                    <attribute name="attrs">
                        {
                        'readonly': [('is_company', '=', False)],
                        'required': [('type', '=', 'contact'), ('is_company', '=', True)]
                        }
                    </attribute>
                </xpath>`

BT-rmartin (changes requested):
I was about to contribute when I realized you did already the PR with the migration. 
I see some space for improvements, especially in the res_partner.xml.

bjornbillen:
@BT-rmartin : thanks for your review. I will try to respond to the corresponding suggestions a.s.a.p.

partner_firstname/models/res_partner.py
bjornbillen: Unfortunately, I tried that. Some test were failing after parsing the context as kwargs. Will investigate further and try to make it work.

partner_firstname/tests/test_create.py
bjornbillen: I will try to remove the with_context and run the tests again. This came with the original source code by the way, although that doesn't mean we don't have to look into it. My bad.

partner_firstname/tests/test_delete.py
bjornbillen: Will try to make it work. Again, this came with the original source code.

partner_firstname/tests/test_empty.py
bjornbillen: Will try to make it work with kwargs. Again, this came with the original source code.

partner_firstname/tests/test_name.py
bjornbillen: I agree, will add the pylint disable

partner_firstname/views/res_partner.xml
bjornbillen: You are right. I forgot to add this back again to the PR. Same problem here though, as with the view_partner_form. Read more below and please suggest a solution for this.

partner_firstname/views/res_partner.xml
bjornbillen: Pre-commit was throwing a warning on this : "Inheritance without a priority is dangerous."

partner_firstname/views/res_partner.xml
bjornbillen: The original form in Odoo 15 consists of 2 name fields, one for companies and one for individuals, with the only difference : the placeholder. As we don't need the individual name to be displayed anymore, I replaced it. The firstname and lastname will appear for individuals, the name field will appear for companies. In any case, the name field will still be available on the form (just invisible), I don't see why the inverse function would go wrong? The tests are running fine.

Can you please give me directions for a better solution according to you? I could definitely target both fields with the ID, that's for sure. With this code, I even might have targeted the wrong name field (with the wrong placeholder). Only I don't see why I can't replace one of the two, because the computation will still happen on the other name field.

`<field id="company" name="name" default_focus="1" placeholder="e.g. Lumber Inc" attrs="{'required' : [('type', '=', 'contact')], 'invisible': [('is_company','=', False)]}"/>`

` <field id="individual" name="name" default_focus="1" placeholder="e.g. Brandom Freeman" attrs="{'required' : [('type', '=', 'contact')], 'invisible': [('is_company','=', True)]}"/>`

partner_firstname/models/res_partner.py
BT-rmartin: I did try and I have all test passing

partner_firstname/tests/test_create.py
BT-rmartin: I did remove it and all tests continue passing

partner_firstname/views/res_partner.xml
BT-rmartin: No, replace without priority is dangerous. And that's aligned with what I said that using replace is a bad practice. Please, remove the priority

partner_firstname/views/res_partner.xml
BT-rmartin: Sorry I forgot to put it as code. I had already given you the solution in my previous input.  
That's needed in both views. And tests will run
For individuals you also need the name, it is computed from the firstname and lastname, that's how this module was working in previous versions. 
If you have some name field as not readonly, as soon as you changed firstname/lastname, the name is computed and gets a  value. When you save, it comes as a value, and then in the create it tries to split it with the same function as the inverse to get the firstname and lastname, assigning them in the wrong order if you for instance just gave firstname and not lastname.

bjornbillen: Thanks for the explanation

partner_firstname/models/res_partner.py
bjornbillen: Update : after adding the **context, these tests are failing :
- test_wrong_name_context
- test_wrong_name_value
- test_wrong_name_value_and_context

partner_firstname/hooks.py
bjornbillen: Update : runboat build now fails after removing the registr parameter.

2021-11-17 11:22:45,387 177 CRITICAL bf88fa468-066b-497a-95ed-00e4444d3040 odoo.service.server: Failed to initialize database `bf88fa468-066b-497a-95ed-00e4444d3040`. 
  File "/opt/odoo/odoo/service/server.py", line 1246, in preload_registries
    registry = Registry.new(dbname, update_module=update_module)
  File "/opt/odoo/odoo/modules/registry.py", line 87, in new
    odoo.modules.load_modules(registry, force_demo, status, update_module)
  File "/opt/odoo/odoo/modules/loading.py", line 474, in load_modules
    processed_modules += load_marked_modules(cr, graph,
  File "/opt/odoo/odoo/modules/loading.py", line 363, in load_marked_modules
    loaded, processed = load_module_graph(
  File "/opt/odoo/odoo/modules/loading.py", line 240, in load_module_graph
    getattr(py_module, post_init)(cr, registry)
TypeError: post_init_hook() takes 1 positional argument but 2 were given

partner_firstname/tests/test_name.py
BT-rmartin: After doing some checks, maybe your approach was better. Even when the parent function doesn't return anything, it's implicitly returning a None. So please, come back with your previous approach and sorry for disturbing you

partner_firstname/hooks.py
BT-rmartin: I'm not telling to remove it, but in the past it was _. What I miss is more the with api.Environment.manage() as first line of the method

partner_firstname/tests/test_delete.py
BT-rmartin: That's not a reason to keep it. On every migration we should improve things if possible

partner_firstname/tests/test_empty.py
BT-rmartin: That's not a reason to keep it. On every migration we should improve things if possible

partner_firstname/hooks.py
bjornbillen: According to the new guidelines for migration to Odoo 15.0 :

"If you write a hook or any code that requires to create a new environment, there's no need of using the with Environment.manage(): context statement anymore."

partner_firstname/views/res_partner.xml
BT-rmartin: I realized I had a mistake in my proposal. This would be the right one, as otherwise the invisible part from the core was broken
`<xpath expr="//field[@id='individual']" position="attributes">
                    <attribute name="attrs">
                        {
                        'readonly': [('is_company', '=', False)],
                        'required': [('type', '=', 'contact'), ('is_company', '=', False)],
                        'invisible': [('is_company','=', True)]
                        }
                    </attribute>
                </xpath>
                <xpath expr="//field[@id='company']" position="attributes">
                    <attribute name="attrs">
                        {
                        'readonly': [('is_company', '=', False)],
                        'required': [('type', '=', 'contact'), ('is_company', '=', True)],
                        'invisible': [('is_company','=', False)]
                        }
                    </attribute>
                </xpath>`

partner_firstname/models/res_partner.py
BT-rmartin: Not in my local, and here the github actions are not initiated for any reason

partner_firstname/views/res_partner.xml
bjornbillen: Already did that. I noticed the invisible attrs was missing in your proposal.

partner_firstname/models/res_partner.py
bjornbillen: To be investigated further. The testcases involve a local context param on top in the class definition, which is updated in every single method. Probably been written in the past to avoid having to parse a new context every time and decrease the amount of lines of code. Not wrong to do that in my opinion. But for some reason, parsing the context with kwargs in this particular method, messes up a couple of test cases mentioned above.

bjornbillen: @BT-rmartin : any feedback from your side? I can't make it work with parsing the context as kwargs.

bjornbillen:
Can we please have a second review on this PR? @pedrobaeza @BT-rmartin

partner_firstname/views/res_user.xml
```diff
@@ -0,0 +1,27 @@
+<odoo>
+    <record id="view_users_form" model="ir.ui.view">
+        <field name="name">Add firstname and surnames</field>
+        <field name="model">res.users</field>
+        <field name="inherit_id" ref="base.view_users_form" />
+        <field name="arch" type="xml">
+            <data>
```
ajaniszewska-dev:
```suggestion
```

partner_firstname/views/res_user.xml
```diff
@@ -0,0 +1,27 @@
...
+                            name="lastname"
+                            attrs="{'required': [('firstname', '=', False)]}"
+                        />
+                        <field
+                            name="firstname"
+                            attrs="{'required': [('lastname', '=', False)]}"
+                        />
+                    </group>
+                </xpath>
+            </data>
```
ajaniszewska-dev:
```suggestion
```

ajaniszewska-dev (approved):
minor comments

partner_firstname/models/res_partner.py
gurneyalex: @BT-rmartin @bjornbillen for the record: calling with_context with keyword arguments has a different effect than calling with_context with a dictionary: the first call will extend the current context while the second overwrites the current context with the dictionary. Since the code here is building a new context dictionary with values *removed* from self.env.context, you need to use the dictionary arg rather than the keyword argument form, otherwise the removed values will still be present. 

my 0.02€ ;)

gurneyalex (approved):
LGTM. Can you please squash your two last commits?

ph-obs:
Hey guys. Sorry to disturb. Any idea on how this will be ready? Thanks so much.

bjornbillen:
> Hey guys. Sorry to disturb. Any idea on how this will be ready? Thanks so much.

Hey there. Waiting for final approval. I guess, in the mean time we'll have to be patient

houssine78 (approved):

houssine78:
hi @gurneyalex seems that runbot his failling but I can't see why. Nothing related to the roll out on runboat ?

houssine78:
Hi @pedrobaeza this PR seems to be stuck in ci/runbot and pending in runboat. Do you have an idea why ? Or maybe I'm missing something?

pedrobaeza:
runbot should be ignored, as it's no longer working, and no idea about runboat.

houssine78:
thanks @pedrobaeza do you know what is missing to have this PR merged ? It seems to have the two reviews needed.

pedrobaeza:
CIs are red, so you have to care first about that. And then, a PSC must authorize the merge.

bjornbillen:
> CIs are red, so you have to care first about that. And then, a PSC must authorize the merge.

Test of base_location_geonames_import is failing?

2022-02-04 15:48:43,579 349 ERROR odoo odoo.modules.loading: Module **base_location_geonames_import**: 0 failures, 1 errors of 7 tests 

2022-02-04 15:48:38,072 349 INFO odoo odoo.modules.loading: Module partner_firstname loaded in 4.17s (incl. 3.59s test), 662 queries (+4330 test)

pedrobaeza:
Can you please rebase then to see the CI after fresh update and new build?

bjornbillen:
> Can you please rebase then to see the CI after fresh update and new build?

Done!

houssine78:
@pedrobaeza I think it all green now :)

PR #1179, opened 2021-11-22
[12.0][FIX] base_location: address field sync zip_id

victoralmau:
FWP from 14.0: https://github.com/OCA/partner-contact/pull/1150

`_address_fields` are used by `update_address` [1], which is in turn
called by `_fields_sync` and `_children_sync`.

This mechanism is in charge of synchronizing parent/children address fields
for contacts of type == "contact", as they don't really have an address of
their own. They simply copy the address of their parent.

Without this fix, not only the zip_id is not sync'ed, but also in some cases
an exception may be raised by `_check_zip_id`, due to only some fields being
sync'ed but not all, creating inconsistencies between zip_id and other address
fields.

Please @pedrobaeza @joao-p-marques can you review it?

@Tecnativa TT33047

[1]: https://github.com/odoo/odoo/blob/20648ef21/odoo/addons/base/models/res_partner.py#L429

### DIFF
--- a/base_location/models/res_partner.py
+++ b/base_location/models/res_partner.py
@@ -100,3 +100,7 @@ class ResPartner(models.Model):
                 'city': False,
             })
         self.update(vals)
+
+    @api.model
+    def _address_fields(self):
+        return super()._address_fields() + ["zip_id", "city_id"]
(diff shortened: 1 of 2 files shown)


HaraldPanten (approved):

base_location/models/res_partner.py
```diff
@@ -100,3 +100,7 @@ def _onchange_state_id(self):
                 'city': False,
             })
         self.update(vals)
+
+    @api.model
+    def _address_fields(self):
+        return super()._address_fields() + ["zip_id", "city_id"]
```
pedrobaeza, line 106: Curious that `base_address_city` doesn't add already `city_id`.

pedrobaeza (approved):

PR #1180, opened 2021-11-22
[13.0][FIX] base_location: address field sync zip_id

victoralmau:
FWP from 13.0: https://github.com/OCA/partner-contact/pull/1150

`_address_fields` are used by `update_address` [1], which is in turn
called by `_fields_sync` and `_children_sync`.

This mechanism is in charge of synchronizing parent/children address fields
for contacts of type == "contact", as they don't really have an address of
their own. They simply copy the address of their parent.

Without this fix, not only the zip_id is not sync'ed, but also in some cases
an exception may be raised by `_check_zip_id`, due to only some fields being
sync'ed but not all, creating inconsistencies between zip_id and other address
fields.

Please @pedrobaeza @joao-p-marques can you review it?

@Tecnativa TT33047

[1]: https://github.com/odoo/odoo/blob/20648ef21/odoo/addons/base/models/res_partner.py#L429

### DIFF
--- a/base_location/models/res_partner.py
+++ b/base_location/models/res_partner.py
@@ -79,3 +79,7 @@ class ResPartner(models.Model):
         if self.zip_id and self.state_id != self.zip_id.city_id.state_id:
             vals.update({"zip_id": False, "zip": False, "city": False})
         self.update(vals)
+
+    @api.model
+    def _address_fields(self):
+        return super()._address_fields() + ["zip_id", "city_id"]
--- a/base_location/tests/test_base_location.py
+++ b/base_location/tests/test_base_location.py
@@ -4,7 +4,7 @@
 import psycopg2
 
 from odoo.exceptions import ValidationError
-from odoo.tests import common, tagged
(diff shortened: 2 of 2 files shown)


HaraldPanten (approved):

pedrobaeza (approved):

pedrobaeza:
I suppose this is needed in v14 and v15

victoralmau:
> I suppose this is needed in v14 and v15
I think it is not totally necessary because in v14 and v15 `city_id` is redefined as a compute.

pedrobaeza:
But for `zip_id` ? And how that compute affects to this module?

victoralmau:
This module re-defines the `city_id` field as a compute since v14 (https://github.com/OCA/partner-contact/pull/1003/commits/4a71faea5c76c2f062934cde9365a4f3b19e22b6)

pedrobaeza:
Not really, they are computed writable fields, so they act as onchanges, so no propagation is done that way AFAIK.

PR #1187, opened 2021-11-29
[14.0][MIG]purchase_supplier_rank: Migration to 14.0

GuillemCForgeFlow:
Standard migration
@ForgeFlow


purchase_supplier_rank/__init__.py
```diff
@@ -0,0 +1,11 @@
...
+
+from . import models
+from odoo import api, SUPERUSER_ID
+
+
+def post_init_hook(cr, registry):
+    env = api.Environment(cr, SUPERUSER_ID, {})
+    partners = env["purchase.order"].search([]).mapped("partner_id")
+    partners |= partners.mapped("commercial_partner_id")
+    partners._increase_rank("supplier_rank")
```
ChrisOForgeFlow: You could update the ranking in a more specific way, increasing the number of purchases that the supplier has, the function accepts a parameter n to be able to specify it

https://github.com/odoo/odoo/blob/d80ec72696010d668047c8fb1d9665f6aa1b2ee6/addons/account/models/partner.py#L495

ChrisOForgeFlow (changes requested):

purchase_supplier_rank/__init__.py
GuillemCForgeFlow: Thanks for the reply! Hope that works! @ChrisOForgeFlow

ChrisOForgeFlow (approved):
+1

SerM11:
We want to test this pr but runbot/runboat is not working.

GuillemCForgeFlow:
> We want to test this pr but runbot/runboat is not working.

Should be working now! @SerM11

PR #1198, opened 2021-12-20
[15.0][mig] partner_company_group

damdam-s:


dhx:
Hi @damdam-s while this PR seems to be merged, unfortunately the partner_company_group module seems to not have made it into 15.0.

PR #1199, opened 2021-12-21
[IMP][14.0] Added a cron job functionlityl

AmmarOfficewalaSerpentCS:
**There should be a cron job that runs daily that checks the expiration date on res.partner.id_number and if the Valid Until date is earlier than today, and if the license is in the "open" or "pending" states, it should move to the "close" state.**


ajaniszewska-dev:
pls correct typo in commit message: functionlityl -> functionality

PR #844, opened 2020-01-24
[10.0][ADD] partner_contact_type_end_user

rousseldenis:
image

PR #1202, opened 2021-12-23
[10.0][ADD] partner_contact_type_end_user

rousseldenis:


rousseldenis:
Reopen this https://github.com/OCA/partner-contact/pull/844

@NL66278 Comments about https solved.

No need to apply black as there is no configuration in 10.0. And for migration, following the migration guide will do that automatically.

PR #1209, opened 2022-01-07
[FIX] write in one time

jco-subteno-it:
Limit bdd call


partner_firstname/models/res_partner.py
```diff
@@ -187,8 +187,10 @@ def _inverse_name(self):
         """Try to revert the effect of :meth:`._compute_name`."""
         for record in self:
             parts = record._get_inverse_name(record.name, record.is_company)
-            record.lastname = parts["lastname"]
-            record.firstname = parts["firstname"]
+            record.write({
+                'lastname': parts["lastname"],
+                'firstname': parts["firstname"]
+            })
```
dreispt, line 193: DB writes are now delayed to end of method, so this is not relevant anymore.
The original code is more compact, so I'm not for this change.

PR #1215, opened 2022-01-31
[14.0][FIX] Fix access right issues when using actions

sebalix:


NL66278 (approved):
:+1: LGTM

i-vyshnevska (approved):

animal_owner/models/res_partner.py
```diff
@@ -17,7 +17,8 @@ def _compute_animal_count(self):
     )
 
     def action_view_animals(self):
-        action = self.env.ref("animal.action_animal").read()[0]
+        xmlid = "animal.action_animal"
+        action = self.env["ir.actions.act_window"]._for_xml_id(xmlid)
```
dreispt, line 21: I'm not comfortable with the usage of the private method `_for_xml_id`.
Why is there an access right issue reading the action?
If it can't be solved properly, would using sudo() be a better solution?

i-vyshnevska: sudo changes user and tracking of initial user is lost

pedrobaeza: > sudo changes user and tracking of initial user is lost

That's not really true since v13 (it was in <=v12).

Anyways, Odoo uses this formula as the default one. See for example:

https://github.com/odoo/odoo/blob/a7f7233e0eae8ee101d745a9813cba930fd03dcb/addons/sale_crm/models/crm_lead.py#L37

Here is the explanation of why Odoo has changed the way to control this:

https://github.com/odoo/odoo/commit/6156f98288b16f6c84d36bacf16d5a66a661abb3

simahawk (approved):

PR #1224, opened 2022-02-15
[13.0][IMP] base_location_geonames_import: Try/except is added to remove cities and postal codes

victoralmau:
Try/except is added to remove cities and postal codes.

Please @pedrobaeza and @chienandalu can you review it?

@Tecnativa TT33464

### DIFF
--- a/base_location_geonames_import/wizard/geonames_import.py
+++ b/base_location_geonames_import/wizard/geonames_import.py
@@ -191,6 +191,22 @@ class CityZipGeonamesImport(models.TransientModel):
             self._process_csv(parsed_csv, country)
         return True
 
+    def _action_remove_old_records(self, model_name, old_records, country):
+        model = self.env[model_name]
+        items = model.browse(list(old_records))
+        try:
+            logger.info("removing %s entries" % model._name)
+            items.unlink()
+            logger.info(
+                "%d entries deleted for country %s" % (len(old_records), country.name)
+            )
+        except Exception:
+            for item in items:
+                try:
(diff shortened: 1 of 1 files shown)


chienandalu (approved):
Not sure on why runboat is failiing

pedrobaeza:
`External ID not found in the system: contacts.res_partner_menu_config` in `partner_contact_age_range`. No `contacts` dependency put, but the main problem is that there's no Travis here, and I don't know why.

pedrobaeza:
I don't think this is the solution to the main problem, which is possible cities not being able to be removed. Maybe you can include a flag for not removing them, or try to remove them in global with a try, and if not possible, remove one by one if possible.

victoralmau:
> I don't think this is the solution to the main problem, which is possible cities not being able to be removed. Maybe you can include a flag for not removing them, or try to remove them in global with a try, and if not possible, remove one by one if possible.

One possibility would be to add an `is_custom` field (boolean) both to the cities and zip codes and to the domain, add
`('is_custom', '=', False)` to not remove those. What do you think about it?

pedrobaeza:
I prefer the other options.

victoralmau:
Changes done, it's ready to review again.

pedrobaeza (approved):

pedrobaeza:
Cherry-picked for v14 in 1709a66c9e7badf965728ef98ca4c2b0af2a753d and in v15 in 759f40a13bbdf04e9f4109771367ac3798ef5352

PR #1227, opened 2022-02-16
[FIX]partner_vat_unique: Make vat constraint case insensitive.

jesusVMayor:


pedrobaeza:
This affects a lot the performance, so it's something to avoid. On v13+, the VAT is sanitized by default.

jesusVMayor:
And making the constraint against the field sanitized_vat of base_vat_sanitized?

pedrobaeza:
That can be another option, but for that, you need a glue module.

Issue #1225, opened 2022-02-15
[13.0] base_location : unable to add a contact to a partner with zip_id


sanrav:
## Module
base_location

## Describe the bug
When I add a contact to a partner, on which the field zip_id is filled in, ODOO displays an error message when I save the partner form.

## To Reproduce
Version : V13

1. Create a new partner (a company) and select an address in the field zip_id (configuration in the menu Contacts/Configuration/Zips) and save

![Step 1 - create partner with zip_id](https://user-images.githubusercontent.com/39826575/154085084-a2aac643-913a-4135-8052-c543d414d115.JPG)

2. Edit the partner and add a contact (field type = contact), fill in the lastname and firstname and "Save and close"

![Step 2 - add contact](https://user-images.githubusercontent.com/39826575/154085088-25be2165-1925-4388-ba46-b62405c53e89.JPG)

3. Save the partner => ODOO shows an error "The city of partner Thomas Dupont differs from that in location 1000"

![Step 3 - error message](https://user-images.githubusercontent.com/39826575/154085095-93c69ce9-23e4-4eb3-80c9-59e2e6398ad1.JPG)

**Expected behavior**
We should be able to add a contact to a partner when the field zip_id is filled in on the main partner.

pedrobaeza:
I'm not able to replicate the problem in runbot, as you can see here:

http://oca-partner-contact-13-0-523c7073a284.runboat.odoo-community.org/web#id=47&action=140&model=res.partner&view_type=form&cids=&menu_id=102

Maybe the problem is just in your DB, and it's due to old data (I have found this problem in one of my DBs migrated to v13). A little tip for fixing the situation: run this SQL:

```sql
UPDATE res_partner rp SET city_id = rcz.city_id FROM res_city_zip rcz WHERE rp.city_id IS NULL AND rp.zip_id = rcz.id;
```

I will add it to the migration scripts for others doing the same migration.

sanrav:
Pedro,

I did a test with the link you mentioned and I have the same issue :

![image](https://user-images.githubusercontent.com/39826575/154682272-6697a8f9-7900-46da-b4ec-78fc8cfb3fc6.png)

I don't know you created your contact without having the issue...

pedrobaeza:
Can you please put a video with all the steps?

sanrav:
Hi Pedro,
Here is the video :

[...]

PR #1232, opened 2022-02-24
[13.0][FIX] base_location - fix save of company with zip with a child

baimont:
see https://github.com/OCA/partner-contact/issues/1225

### DIFF
--- a/base_location/models/res_partner.py
+++ b/base_location/models/res_partner.py
@@ -82,4 +82,4 @@ class ResPartner(models.Model):
 
     @api.model
     def _address_fields(self):
-        return super()._address_fields() + ["zip_id"]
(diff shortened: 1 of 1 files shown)


pedrobaeza:
But this shouldn't be needed:

https://github.com/odoo/odoo/blob/13.0/addons/base_address_city/models/res_partner.py#L30

and this module depends on base_address_city.

NL66278:
I have to agree with @pedrobaeza here. The city_id field must, following the chain of dependencies, already be in the address fields.

NL66278 (changes requested):
:-1: This PR is invalid. Because base_location depends on base_address_city, city_id will already be in the address fields, unless you somehow succeeded in having base_location installed without base_address_city....

PR #1234, opened 2022-02-25
[FIX] [14.0] Removed external dependencies already defined in Odoo base

Shide:
Removed external dependencies already defined in Odoo base

https://github.com/odoo/odoo/blob/ae6f414ba41eee4e14235a1e82fd5ebd237dd97f/requirements.txt#L48

MT-282 @moduon

### DIFF
--- a/partner_identification_gln/__manifest__.py
+++ b/partner_identification_gln/__manifest__.py
@@ -10,7 +10,6 @@
     "license": "AGPL-3",
     "author": "Acsone S.A.,Odoo Community Association (OCA)",
     "website": "https://github.com/OCA/partner-contact",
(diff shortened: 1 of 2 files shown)


NL66278 (approved):
:+1: LGTM

yajo (approved):

pedrobaeza (changes requested):
This should be removed instead, as it's part of standard Odoo requirements:

https://github.com/odoo/odoo/blob/ae6f414ba41eee4e14235a1e82fd5ebd237dd97f/requirements.txt#L48

pedrobaeza (approved):

PR #1236, opened 2022-02-27
[15.0] Migrating partner contact tradename module

patovala:


partner_contact_tradename/models/res_partner.py
```diff
@@ -0,0 +1,37 @@
...
+
+    tradename = fields.Char(
+        size=300,
+    )
+
+    @api.model
+    def name_search(self, name="", args=None, operator="ilike", limit=100):
+        args = args or []
+        try:
+            recs = self.search(
```
cormaza, line 19: Should be good super call for other implementations of this feature or another modules

cormaza (changes requested):
Please check super call, because this class is used very much with implementations on name_search

partner_contact_tradename/models/res_partner.py
patovala: Yes, will do the fix asap, thanks

cormaza:
Can be good idea check this implementation:

https://github.com/OCA/l10n-spain/blob/96ce681c8ff0d0c5e53aea2e6734acb8f87dd73f/l10n_es_partner/models/res_partner.py#L39-L57

PR #1250, opened 2022-04-01
[FIX] base_location_geonames_import: some bottlenecks

yajo:
- Instead of doing one `.search()` per state, do one per batch.
- Instead of searching in lists with `in` operator, use a set.

Thousands of search queries and useless list loop lookups saved. Importing Spain states went down from 4:00 minutes to 2:30.

Still having a big bottleneck when creating the states, but [that is explained in odoo][1] and outside of this fix scope.

Speedscopes for the bottlenecks:

1. [_create_states.speedscope.json](https://github.com/OCA/partner-contact/files/8396690/_create_states.speedscope.json)
2. [_create_cities.speedscope.json](https://github.com/OCA/partner-contact/files/8396694/_create_cities.speedscope.json)

[1]: https://github.com/odoo/odoo/blob/c80d7a99236dfb6b99cca724d76632a82f6b605e/odoo/models.py#L4141-L4161

@moduon MT-426


rafaelbn (approved):
Great 👍

Shide (approved):

PR #1251, opened 2022-04-01
[10.0][FIX] Merging contact failed on installation with abstract model

flotho:
When an abstract modl is defined like in https://github.com/OCA/connector-prestashop/blob/10.0/connector_prestashop/models/binding/common.py#L101 then the merge wizard failing because it's trying to search on non existing fileds there : https://github.com/OCA/OCB/blob/10.0/addons/crm/wizard/base_partner_merge.py#L249
My first idea was first to propose an OCB patch for https://github.com/OCA/OCB/blob/10.0/addons/crm/wizard/base_partner_merge.py#L240 with something like : 
```
Model = self.env[record.
field = Model._fields[record.name]
if Model._abstract:
   continue
```

Yet I think that OCB is not the proper way to deal with something that mostly depends from other OCA modules.

Thanks for your review @pedrobaeza @hbrunn  @yvaucher


flotho:
Hi  @pedrobaeza @hbrunn @yvaucher
any chance to have your feedback for the solution on this.
additionnaly, can't see why travis is failing. 
Thanks in advanced

pedrobaeza:
Sorry, too old version.

flotho:
hum... I'll test in most recent version

PR #1252, opened 2022-04-04
[13.0][IMP] partner_company_group: Add advises when company_group pricelist is different than member pricelists

CarlosRoca13:
cc @Tecnativa TT35300

please @carlosdauden @sergio-teruel review this

### DIFF
--- a/partner_company_group/models/res_partner.py
+++ b/partner_company_group/models/res_partner.py
@@ -1,7 +1,7 @@
 # Copyright 2019 Camptocamp SA
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
-from odoo import fields, models
+from odoo import _, api, fields, models
 
 
 class Contact(models.Model):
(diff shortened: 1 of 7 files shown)


partner_company_group/models/res_partner.py
```diff
@@ -18,3 +18,64 @@ class Contact(models.Model):
...
+                "message": _(
+                    "The company group %s has the pricelist %s, that is different than"
+                    " the pricelist set on this contact"
+                    % (
+                        self.company_group_id.display_name,
+                        self.company_group_id.property_product_pricelist.display_name,
+                    )
+                ),
+            }
+        elif (
```
carlosdauden: You could extract the common code to a new method to not repeat the same code ([DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself)).
Or in this case call direcly to _onchange_company_group_id
```suggestion
        res = self._onchange_company_group_id()
        if not res:
```

carlosdauden (approved):

CarlosRoca13:
Changed!

sergio-teruel (changes requested):
@CarlosRoca13 please add some test to coverage this behavior

sergio-teruel:
ping @CarlosRoca13

CarlosRoca13:
Sorry for the delay, added

sergio-teruel (approved):

rousseldenis (commented):
LGTM

rousseldenis (approved):

PR #1255, opened 2022-04-08
[13] FIX - partner_external_map  fix view to better look Update res_partner_view.xml

sergioeix:
this way smart buttons looks better on partner form


sergioeix:
@alexis-via @pedrobaeza @ernestotejeda are you agree with this small change to merge?

pedrobaeza (approved):

ernestotejeda (approved):

alexis-via:
According to the logs, the pb was on VIES. Maybe we can try again later and it will become green.

alexis-via:
OK, it seems we have to rely on "the real mans" for this merge.

PR #1259, opened 2022-04-15
14.0 mig partner_coc

bosd:
standard migration


bosd:
This module was originally developed to include the dutch Chamber of Commerce Number.
A.K.A KvK nummer.

From V15, odoo has added a filed for this in the dutch localization.
So if no other counrties have a need for this module. This could be the last port.
What do you think? @astirpe @NL66278 

https://github.com/odoo/odoo/blob/69dc1f561f4d1740d3d558c68b9006dfba3ab5c3/addons/l10n_nl/models/res_partner.py#L8

astirpe:
In my opinion the Chamber of Commerce Number could be considered a generic field, not just limited to the dutch localization.

bosd:
Ok, so well continue with this one.

Im also thinking of creating some modules to comply to dutch law.
All company communication should include kvk/coc number.
So it should be added to the invoices, website and optionally email messages.
Obviously that should go in the l10n-netherlands repo.

Not sure if this should be based on this module for v14 or a backport of core.

NL66278:
@astirpe I agree the coc number could be used in a more generic way, but in practice if in Odoo 15.0 the kvk number is present in standard Odoo, I think this is what we should use. If other localizations also have something as a chamber of commerce with a unique number for each company, they can create their own modules based on partner_identification. @bosd We will still need a small stub module in 15.0 to support migration of the data to the standard Odoo field.

bosd:
:pray: while at it, would you like to leave a review for this one?

astirpe (approved):

NL66278 (approved):
:+1: Sraight migration and LGTM

PR #1264, opened 2022-04-26
[13.0][FIX] partner_company_group: Terms not translated

CarlosRoca13:
cc @Tecnativa TT35300

please @pedrobaeza @victoralmau review it

### DIFF
--- a/partner_company_group/models/res_partner.py
+++ b/partner_company_group/models/res_partner.py
@@ -32,10 +32,10 @@ class Contact(models.Model):
                 "message": _(
                     "The company group %s has the pricelist %s, that is different than"
                     " the pricelist set on this contact"
-                    % (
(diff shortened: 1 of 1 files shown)


partner_company_group/models/res_partner.py
```diff
@@ -55,12 +55,13 @@ def _onchange_property_product_pricelist(self):
             )
             members_str = ""
             for member in company_members:
-                members_str += _("\t- %s\n" % member.display_name)
+                members_str += _("\t- %s\n") % member.display_name
```
pedrobaeza: I don't think this should be translatable, isn't it?

CarlosRoca13:
Removed

pedrobaeza (approved):

carlosdauden (approved):

PR #1269, opened 2022-05-09
[14.0][IMP] partner_contact_address_default ux

CarlosRoca13:
- Filter delivery and invoice addreses properly
- Show it only in contact addresses
- Default to the address the addresses set in the commercial partner if
any.
- If a contact is archived, unlink it from the default addresses

TT34631

forward port of 

- [x] https://github.com/OCA/partner-contact/pull/1228

ping @chienandalu @carlosdauden


chienandalu (approved):
Thanks :)

pedrobaeza (approved):

PR #1276, opened 2022-05-24
[REF] partner_firstname,partner_second_lastname: Avoid _inverse_name when recalculating names

payen000:
Currently, pressing the Recalculate Names button in the 'Partner Names Order' setting
triggers the _inverse_name method which can modify the firstname, lastname and lastname2
of a partner depending on the 'partner_names_order' setting and the number of names
in each field.

Example video: https://www.youtube.com/watch?v=vZ5FeMoTqRU

The modification to action_recalculate_partners_name prevents the _inverse_name method from 
triggering when pressing Recalculate Names, leaving the firstname, lastname and lastname2 
fields intact.

Example after applying changes: https://youtu.be/8WRSU2G86O0

Additionally, added related unittest to ensure names are computed correctly without
modifying firstname, lastname or lastname2.

### DIFF
--- a/partner_firstname/models/base_config_settings.py
+++ b/partner_firstname/models/base_config_settings.py
@@ -62,7 +62,10 @@ class ResConfigSettings(models.TransientModel):
         )
         partners = self._partners_for_recalculating()
         _logger.info("Recalculating names for %d partners.", len(partners))
-        partners._compute_name()
+        # Use add_to_compute instead of _compute_name to avoid triggering
+        # _inverse_name_after_cleaning_whitespace, which can
+        # modify a partner's firstname, lastname and lastname2
+        self.env.add_to_compute(self.env["res.partner"]._fields["name"], partners)
         self.partner_names_order_changed = False
         self.execute()
         _logger.info("%d partners updated.", len(partners))
--- a/partner_second_lastname/tests/test_name.py
+++ b/partner_second_lastname/tests/test_name.py
@@ -197,3 +197,33 @@ class UserCase(PersonCase, MailInstalled):
         # Skip if ``mail`` is installed
         if not self.mail_installed():
             super(UserCase, self).tearDown()
+
+
+class TestRecalculateNames(TransactionCase):
+    def setUp(self):
+        super().setUp()
+        self.config_settings = self.env["res.config.settings"].create({})
+
+    def test_recalculate_names(self):
+        firstname = "Xavier De Jesús"
+        lastname = "Payen"
+        lastname2 = "Sandoval"
+        correct_names = {
+            "first_last": f"{firstname} {lastname} {lastname2}",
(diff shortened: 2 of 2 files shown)


partner_second_lastname/models/res_config_settings.py
```diff
@@ -33,3 +36,18 @@ def _partners_for_recalculating(self):
                 ("lastname2", "!=", False),
             ]
         )
+
+    def action_recalculate_partners_name(self):
+        # Override of partner_firstname's action_recalculate_partners_name
```
hugho-ad: move this to a docstring

```python
def action_recalculate_partners_name(self):
    """DOC"""

File: partner_second_lastname/models/res_config_settings.py [hunk 2]
```diff
@@ -33,3 +36,18 @@ def _partners_for_recalculating(self):
...
+        # to avoid triggering _inverse_name when pressing Recalculate Names button
+        self.env["ir.config_parameter"].sudo().set_param(
+            "partner_names_order", self.partner_names_order
+        )
+        partners = self._partners_for_recalculating()
+        _logger.info("Recalculating names for %d partners.", len(partners))
+        self.env.add_to_compute(self.env["res.partner"]._fields["name"], partners)
```
hugho-ad: this is the main change right? 

why not doing it in original method definition?

payen000: I was trying to keep this change and the unittest in the same module, but since partner_second_lastname depends on partner_firstname, there will be no dependency problems, so there's no reason to override after all, thank you!

payen000:
Hello again @hugho-ad, I've applied your suggestions, would you review again, please? :)

hugho-ad (approved):

partner_second_lastname/tests/__init__.py
```diff
@@ -2,4 +2,5 @@
 
 from . import test_name
 from . import test_config
+from . import test_recalculate_names
```
luisg123v: This could be included into `test_name` IMHO

partner_firstname/models/base_config_settings.py
```diff
@@ -62,7 +62,7 @@ def action_recalculate_partners_name(self):
         )
         partners = self._partners_for_recalculating()
         _logger.info("Recalculating names for %d partners.", len(partners))
-        partners._compute_name()
+        self.env.add_to_compute(self.env["res.partner"]._fields["name"], partners)
```
luisg123v, line 68: A comment explaining why computation is delayed instead of manually calling compute method would be great.

luisg123v (changes requested):

luisg123v:
This should be into separated commits I think. When migrated, `partner_firstname` is usually migrated before `partner_second_lastname`. So when they will be migrated, this commit will be duplicated.

PR #1279, opened 2022-05-26
[ADD] partner_customer_supplier: Easily set up a partner as a customer or supplier

frahikLV:
Reviewers:

@luisg123v


partner_customer_supplier/__manifest__.py
```diff
@@ -0,0 +1,25 @@
...
+    "license": "LGPL-3",
+    "category": "Hidden/Dependency",
+    "website": "https://github.com/OCA/partner-contact",
+    "version": "15.0.1.0.0",
+    "maintainers": [
+        "luisg123v",
+        "frahikLV",
+    ],
+    "depends": [
+        "contacts",
```
luisg123v: Not actually required

partner_customer_supplier/__manifest__.py
```diff
@@ -0,0 +1,25 @@
...
+        "account",
+    ],
+    "data": [
+        "views/res_partner.xml",
```
luisg123v: `_views`

partner_customer_supplier/__manifest__.py
```diff
@@ -0,0 +1,25 @@
+{
+    "name": "Partner Customer/Supplier",
+    "summary": "Adds the fields 'customer' and 'supplier' to the res.partner model.",
+    "author": "Vauxoo, Odoo Community Association (OCA)",
+    "license": "LGPL-3",
+    "category": "Hidden/Dependency",
```
luisg123v, line 6: It's hidden but also marked as an application

partner_customer_supplier/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,2 @@
+This module restores the functionality of <= 12.0 to set contacts as customer or
+supplier, and allows users to create custom filters more user-friendly.
```
luisg123v:
```suggestion
vendor, and allows users to create custom filters on a more user-friendly manner.
```

partner_customer_supplier/models/res_partner.py
```diff
@@ -0,0 +1,60 @@
...
+    _inherit = "res.partner"
+
+    customer = fields.Boolean(
+        string="Is Customer",
+        compute="_compute_partner_is_customer",
+        inverse="_inverse_partner_is_customer",
+        search="_search_partner_is_customer",
+    )
+    supplier = fields.Boolean(
+        string="Is supplier",
```
luisg123v: vendor

luisg123v (changes requested):
- Missing translation file. You could add only .pot file so this module becomes translatable in Weblate
- Commit description is empty. You could copy the same text as in README's description

frahikLV:
@luisg123v could you review again?

luisg123v (approved):
LGTM :+1: 

@moylop260 what do you think?

CC @desdelinux @hugho-ad

desdelinux:
LGTM 

Functional review is ok

desdelinux (approved):
LGTM

Functional review is ok

partner_customer_supplier/models/res_partner.py
```diff
@@ -0,0 +1,60 @@
...
+        string="Is Vendor",
+        compute="_compute_partner_is_supplier",
+        inverse="_inverse_partner_is_supplier",
+        search="_search_partner_is_supplier",
+    )
+
+    @api.depends("customer_rank")
+    def _compute_partner_is_customer(self):
+        for partner in self:
+            partner.customer = partner.customer_rank > 0
```
hugho-ad, line 31: but this will be inconsistent, since the _increase_rank is done by sql, so that do not triggers the compute leading on an inconsistency between the value on DB and the value on your new field right?

hugho-ad:
Also I'm thinking that we should adapt to the new Odoo's way to identify the partners

partner_customer_supplier/models/res_partner.py
luisg123v: No, because this field is not stored, so its computed each time is consulted

frahikLV:
@hugho-ad 
Odoo Users usually aren't technically persons, `customer_rank > 0` it's less friendly that `customer = True`, it's confusing when they try to create filters, also, some times, the users need revert when they don't want sale anymore (set `customer_rank = 0`), Odoo don't have any method to decrease the `customer_rank` or set it to 0, so this module gives the user more control to theirs partners.

This module it's not for coding with it, it's for simplify the user process and give better experience

luisg123v:
X2 to @frahikLV's comment.

partner_customer_supplier/tests/test_res_partner.py
```diff
@@ -0,0 +1,81 @@
+from odoo.tests import common, tagged
+
+
+@tagged("res_partner")
+class TestResPartner(common.TransactionCase):
+    def setUp(self):
+        super().setUp()
+        self.partner = self.env["res.partner"].create({"name": "Microsoft Corporation"})
+        self.partner_2 = self.env["res.partner"].create({"name": "Apple Inc."})
```
moylop260, line 9: you want to see the world burn! 🔥 hahaha

```suggestion
        self.partner = self.env["res.partner"].create({"name": "Microsoft Corporation"})
        self.partner_2 = self.env["res.partner"].create({"name": "Apple Inc."})
```

partner_customer_supplier/models/res_partner.py
```diff
@@ -0,0 +1,60 @@
...
+
+    def _inverse_partner_is_customer(self):
+        self.filtered(lambda p: not p.customer).write({"customer_rank": 0})
+        self.filtered(lambda p: p.customer and not p.customer_rank).write(
+            {"customer_rank": 1}
```
moylop260, line 36: Question:

Shouldn't be counting the current customer account moves posted?

I mean, the main focus to use a rank is to have the best customer at the top of the views (e.g. m2o)

But if the user set to False then to True it will be reset to 1 so it will be at the bottom of the list

Something like `move.read_group([domain_is_sale + posted + partner], fields=id:count, group_by=partner_id)`

WDYT?

moylop260 (approved):
The module looks very smooth (no new stored fields) and smart (searching with domain instead of ids and inverse)

👍 

Only I had a question about new *_rank value

partner_customer_supplier/models/res_partner.py
luisg123v: This is expected to me. If I manually unset a customer, it probably means I don't trust them anymore. If I reset it to True, trust needs to be re-earned :)

frahikLV: @moylop260 

As I see it, if Odoo user remove a partner from the customers (set `customer_rank = 0`), maybe he has many overdue debts and the responsible do not want to continue selling to him, so if at some point he settles the debt and later wants to buy again, he will have to earn back the "trust" (starting with `customer_rank = 1`).

cc: @luisg123v @tomeyro

moylop260: Ok, good point

I even disagree but you have a point

So, resolved

We can re-discuss it after

moblesmarti:
partner_customer_supplier
Hello, has there been any problem with this module that in the v.15 directory no longer appears?
Thanks

luisg123v:
Hi @moblesmarti,

The module `partner_customer_supplier` has been replaced by `partner_manual_rank`, as the latter is older (we wasn't aware)

PR #1281, opened 2022-05-27
[14.0][MIG] partner_socialmedia: migrate partner_social module

mauriciomatos96:
Change the version on the manifest.py and used pre-commit


rousseldenis:
@mauriciomatos96 Thanks for this.

Could you follow migration guide ? Especially the full process part ? Thanks

https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-14.0

Two other migrations exists for v13 and v15. Better to merge v13 first

PR #1282, opened 2022-05-27
[MIG] partner_manual_rank: Migration to V15

frahikLV:
Migrate `partner_manual_rank` module and Merge with `partner_customer_supplier`

Reviewers: 
@MiquelRForgeFlow 
@luisg123v
@moylop260


moylop260:
@MiquelRForgeFlow @JordiBForgeFlow

Could you review it, please?

sebalix (approved):

partner_manual_rank/models/res_partner.py
```diff
@@ -0,0 +1,64 @@
...
+        inverse="_inverse_is_customer",
+        search="_search_is_customer",
+        string="Is a Customer",
+    )
+    is_supplier = fields.Boolean(
+        compute="_compute_is_supplier",
+        inverse="_inverse_is_supplier",
+        search="_search_is_supplier",
+        string="Is a Supplier",
+    )
```
celm1990, line 30: I prefered fields `store=True` as previous version https://github.com/OCA/partner-contact/blob/14.0/partner_manual_rank/models/res_partner.py#L13

celm1990 (changes requested):

partner_manual_rank/models/res_partner.py
frahikLV: This field does not need to be stored, since the information it contains is identical to the information in the *`_rank` field, so it is redundant information. As it has the search method, it behaves identically to a stored field.

What did you think @luisg123v

luisg123v: I think those fields should not be stored for two reasons:

- Since ranks are updated [via SQL](https://github.com/odoo/odoo/blob/120895356a5517ae47146a41539842f0d466256f/addons/account/models/partner.py#L560), such changes won't be identified by dependencies, so computed fields won't be updated.
- Even if they were updated, that would increase chances of concurrency issues. The reason Odoo moved such increasing to SQL is to avoid concurrency issues. But if the "customer" and "supplier" fields are updated as standard fields (by the ORM), the avoided concurrency will come back with those fields

Besides, as @frahikLV mentioned, almost all features are preserved as if the field were stored, as it's searchable.

Regards,

celm1990: thanks for your comments

partner_manual_rank/readme/CREDITS.rst
```diff
@@ -0,0 +1,2 @@
+ForgeFlow, S.L.
+Vauxoo
```
luisg123v: To the RST syntax, a new line is equivalent to a whitespace. If you want to add several items, you should use a list.

You may check how the resulting README was generated

luisg123v (changes requested):
I see a few issues regarding commits wording:
- On the removal commit, missing an explanation. Removing a module is not a trivial nor self-explanatory change, so you should mention why, i.e. deprecated infavor of `partner_manual_rank`
- Instead of "module merging", you should say "remove module"
- Since the migration includes some refactor, you should mention on the migration commit what changes are performed, i.e. fields are not stored anymore to avoid computed fields getting outdated and concurrency issues

moylop260:
Hi @MiquelRForgeFlow @JordiBForgeFlow

Friendly reminder to review it, please

frahikLV:
@luisg123v Your suggestions were applied, could you look again?

luisg123v (approved):

celm1990 (approved):

moylop260 (approved):
👍

moylop260:
@JordiBForgeFlow

If you have a concern about this PR let me know, please

For now, we will merge it since that it was approved 2 days ago

PR #1283, opened 2022-06-02
[REF] partner_firstname,partner_second_lastname: Avoid _inverse_name when recalculating names

payen000:
Currently, pressing the Recalculate Names button in the 'Partner Names Order' setting
triggers the _inverse_name method which can modify the firstname, lastname and lastname2
of a partner depending on the 'partner_names_order' setting and the number of names
in each field.

The modification to action_recalculate_partners_name prevents the _inverse_name method from 
triggering when pressing Recalculate Names, leaving the firstname, lastname and lastname2 fields intact.

Demo video (left window is before fix; right window is after applying fix): https://www.youtube.com/watch?v=5yAVLBajTbo

Additionally, added related unittest to ensure names are computed correctly without
modifying firstname, lastname or lastname2.

### DIFF
--- a/partner_second_lastname/tests/test_name.py
+++ b/partner_second_lastname/tests/test_name.py
@@ -197,3 +197,33 @@ class UserCase(PersonCase, MailInstalled):
         # Skip if ``mail`` is installed
         if not self.mail_installed():
             super(UserCase, self).tearDown()
+
+
+class TestRecalculateNames(TransactionCase):
+    def setUp(self):
+        super().setUp()
+        self.config_settings = self.env["res.config.settings"].create({})
+
+    def test_recalculate_names(self):
+        firstname = "Xavier De Jesús"
+        lastname = "Payen"
+        lastname2 = "Sandoval"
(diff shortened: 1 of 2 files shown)


payen000:
Hello @hugho-ad, would you review my code again, please? This is the same fix as in [this PR for v14](https://github.com/OCA/partner-contact/pull/1276) but for v15.

hugho-ad (approved):

partner_second_lastname/tests/test_name.py
```diff
@@ -197,3 +197,33 @@ def tearDown(self):
         # Skip if ``mail`` is installed
         if not self.mail_installed():
             super(UserCase, self).tearDown()
+
+
+class TestRecalculateNames(TransactionCase):
```
luisg123v, line 202: Why a separate class but same file?

If it's just a forward-port, then leave it as it is, but take it into account for future changes

luisg123v (changes requested):

PR #1284, opened 2022-06-02
 [13.0][MIG] partner_socialmedia: migrate partner_social module

mauriciomatos96:
Do the pre-commits and change the manifest.py version


rousseldenis:
@mauriciomatos96 Thanks for this.

Could you follow migration guide ? Especially the full process part ? Thanks

https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-13.0

Two other migrations exists for v14 and v15. Better to merge this first

rousseldenis:
Module exists in v12.

PR #1287, opened 2022-06-07
[ADD] search method for partner age

victor-champonnois:
This PR adds a `_search_age` method for partner's `age` computed field. Tests are also added.

### DIFF
--- a/partner_contact_birthdate/models/res_partner.py
+++ b/partner_contact_birthdate/models/res_partner.py
@@ -27,7 +27,7 @@ class ResPartner(models.Model):
             record.age = age
 
     def _search_age(self, operator, value):
-        if operator not in ("=", "!=", "<", "<=", ">", ">=", "in", "not in"):
+        if operator not in ("=", "!=", "<", "<=", ">", ">="):
             return []
         # pylint: disable=sql-injection
         # the value of operator is checked, no risk of injection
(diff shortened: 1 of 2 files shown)


robinkeunen (approved):

partner_contact_birthdate/models/res_partner.py
```diff
@@ -23,3 +25,20 @@ def _compute_age(self):
...
+        # the value of operator is checked, no risk of injection
+        query = """
+            SELECT id
+            FROM res_partner
+            WHERE extract(year from age(CURRENT_DATE, birthdate_date))
+              {operator} %s
+            """.format(
+            operator=operator
+        )
+        self.env.cr.execute(query, (value,))
```
carmenbianca: I don't think `in` and `not in` are valid SQL. They should be removed from the list of operators.

carmenbianca (changes requested):

partner_contact_birthdate/models/res_partner.py
victor-champonnois: It exists in [PostgreSQL's doc ](https://www.postgresql.org/docs/current/functions-subquery.html#FUNCTIONS-SUBQUERY-IN), why do you think it's not valid?

carmenbianca: Huh, you may be right. Can you maybe add a test for them?

carmenbianca: On second inspection: I think they don't work. The operators require a subquery.

victor-champonnois: OK let's just remove them.

carmenbianca (approved):
LGTM! :tada:

PR #1290, opened 2022-06-10
[15.0][MIG] partner_phone_extension

dhx:
Standard migration


lfreeke (approved):
:+1:  Functional test

leemannd (approved):

partner_phone_extension/views/res_partner_view.xml
```diff
@@ -0,0 +1,45 @@
...
+                expr="//div[hasclass('oe_kanban_details')]/div[@t-if='record.phone.raw_value']"
+                position="after"
+            >
+                <div t-if="record.phone_extension.raw_value">
+                    Ext:
+                    <field name="phone_extension" />
+                </div>
+            </xpath>
+            <xpath
+                expr="/form/sheet/notebook/page/field/form/sheet/group/group/field[@name='phone']"
```
NL66278, line 38: If you make your xpath a little bit more generic, it will be less vulnerable of code reorganization, or other modules that might insert groups or divs etc.
Maybe use: `expr="//page[@name='contact_addresses']/field[@name='child_ids']//field[@name='phone']"`

Same goes for some of the other xpaths that sum up the exact elements encountered when navigating to the point in the xml you want to modify.

NL66278, line 38: If you make your xpath a little bit more generic, it will be less vulnerable of code reorganization, or other modules that might insert groups or divs etc.
Maybe use: `expr="//page[@name='contact_addresses']/field[@name='child_ids']//field[@name='phone']"`

Same goes for some of the other xpaths that sum up the exact elements encountered when navigating to the point in the xml you want to modify.

dhx: Thanks @NL66278 for your comment, this is just a standard migration without touching the xpath expression here though (which remains unchanged going back to v9, see also: 05882718fc78546d1c4b57b76b7ba157a4e9b8b0). Maybe you want to create a separate PR to contribute your improvement?

NL66278: I think migrations are also an opportunity for improvement, but I know not everybody agrees. I will merge your PR as is for now. Thanks for your effort!

NL66278 (approved):
:+1: Standard migration

PR #1294, opened 2022-06-13
[ADD] New module for USPS api integration for address validation

ckolobow:


pedrobaeza:
OCA modules mustn't have branding, and this being something very related to USA, should be in l10n-usa IMO.

PR #1304, opened 2022-07-15
[15.0][IMP]partner_deduplicate_acl

dsolanki-initos:
@jans23 FYI

### DIFF
--- a/partner_deduplicate_acl/views/base_partner_merge_view.xml
+++ b/partner_deduplicate_acl/views/base_partner_merge_view.xml
@@ -14,7 +14,7 @@
     <menuitem
         id='partner_merge_automatic_menu'
         action='base.action_partner_deduplicate'
-        groups='group_automatically'
+        groups='group_manually'
         parent='base_tools'
     />
 </odoo>


dsolanki-initos:
@CarlosRoca13 Can you please review this PR?

ikapasi-initos (approved):
Code review and functional test.
LGTM

pedrobaeza (changes requested):
This is not `[ADD]`, but `[FIX]` or `[IMP]`, and the commit message must include a short text about the change and the reason to change it.

dsolanki-initos:
@sergio-teruel could you please review this PR?

jans23:
@pedrobaeza We would really like to see this fix re-opened and merged eventually. All feedback had been incorporated. Could you re-open it, please?

pedrobaeza:
And if you want this, why didn't you review it? Anyway, there's no explanation on why these changes. Steps to reproduce the problem or similar. I'm reopening, but it misses this.

dsolanki-initos:
Hi @pedrobaeza, In the partner_deduplicate_acl module, has a menu item that is accessible for the users having the group "Automatically".

So group "Manual" user was not able to access the menu item, that's why we replaced the group "Automatically"

pedrobaeza:
And why you don't put this from the beginning? It should be also on the commit message, and the pre-commit reference doesn't make sense.

Anyway, seeing the explanation, this has nothing to do with adding the group `base.group_system`.

dsolanki-initos:
> And why you don't put this from the beginning? It should be also on the commit message, and the pre-commit reference doesn't make sense.
> 
> Anyway, seeing the explanation, this has nothing to do with adding the group `base.group_system`.

base.group_system  is because the admin user can access the menu without any groups.

pedrobaeza:
No, it shouldn't. Please undo that change, change the commit message and just leave the manual one.

dsolanki-initos:
Okay @pedrobaeza

dsolanki-initos:
@pedrobaeza, I have done the changes as per your suggestion, Could you please review this PR?

partner_deduplicate_acl/views/base_partner_merge_view.xml
```diff
@@ -14,7 +14,7 @@
     <menuitem
         id='partner_merge_automatic_menu'
         action='base.action_partner_deduplicate'
-        groups='group_automatically'
+        groups='partner_deduplicate_acl.group_manually'
```
pedrobaeza: Put relative references:

```suggestion
        groups='group_manually'
```

dsolanki-initos: I've done the changes. Please check it.

pedrobaeza:
The commit message is incorrect:

- It shouldn't contain the version. You put the version on the PR, not on the commit message.
- It doesn't contain the module name.
- It doesn't explain the problem.
- The first line is very long.

Please put this:

```
[FIX] partner_deduplicate_acl: Change menu group

The menu for deduplicating was only accessible for the users having the
group "Automatically", so the group "Manual" users weren't able to access
it.

This commits changes to this second group for both being able to access.
```

dsolanki-initos:
> partner_deduplicate_acl: Change menu group
> 
> The menu for deduplicating was only accessible for the users having the
> group "Automatically", so the group "Manual" users weren't able to access
> it.
> 
> This commits changes to this second group for both being able to access

Thank you so much @pedrobaeza

pedrobaeza (approved):
/ocabot merge minor

PR #1310, opened 2022-07-27
[15.0][UPD] partner_deduplicate_acl. Set bigger value of a sequence menu

cesar-tecnativa:
In order to prevent that the wizard "Deduplicate Contacts" is automatically launched by just clicking on "Contacts", the sequence of this menu is increased.

@Tecnativa TT37296


cesar-tecnativa:
@pedrobaeza could you review please?

pedrobaeza:
But is this placing the menu between "Contacts" and "Configuration"?

`[UPD]` label doesn't exist. It should be either `[FIX]` or `[IMP]`. In this case, it seems more a fix.

pedrobaeza:
Closing in favor of #1335

PR #1315, opened 2022-08-08
[14.0][IMP] partner_tz: dev status Alpha -> Beta

simahawk:


pedrobaeza (approved):
Please remember to follow commit message guidelines (with `[IMP]`):

https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#71commit-message

and on the PR title, to add the version at the beginning

simahawk:
All good but switching dev status is not an improvement by itself from POV :wink: 
Also, you didn't leave me the time to change the title :stuck_out_tongue_winking_eye:

pedrobaeza:
I consider as such, as you are improving the status of the module, and as such, you have made any improvement on the module itself to get to that status. I know that in this case, it's only to raise the manifest, but it's still an improvement.

pedrobaeza:
What is clear is that it must have a label, whatever you take, and the `[IMP]` one is the more suitable.

simahawk:
I disagree as long as the commit message IS crystal clear (like in this case).
Most of the times is redundant and just wastes chars. 
Matter of taste, I guess.
And again, this is not an improvement, sorry. TBH is probably a patch because it should have been updated since a while and it was not (reason for which I used the patch command).

Enough for this thread: we can talk about this at the OCA days :wink:

pedrobaeza:
OK, but the guideline is the guideline: you MUST put the tag. Don't try to save 5 chars in this case.

PR #1319, opened 2022-08-16
[13.0][ADD] partner_country_lang: New addon.

victoralmau:
New addon.

Please @pedrobaeza and @chienandalu can you review it?

@Tecnativa TT38559


partner_country_lang/models/res_partner.py
```diff
@@ -0,0 +1,30 @@
...
+
+
+class ResPartner(models.Model):
+    _inherit = "res.partner"
+
+    @api.onchange("country_id")
+    def _onchange_country_id(self):
+        super()._onchange_country_id()
+        if self.country_id.lang:
+            self.lang = self.country_id.lang
```
rousseldenis, line 14: I would have tested if lang is not yet defined as partner lang may be different than the one defined in the country

victoralmau: Sorry but I am not sure about the change you suggest.
By default, a language is assigned (default of `lang` field), that's why in the onchange you are forced to change the language.

What approach do you suggest?

pedrobaeza: The idea is to always overwrites the lang, as the country lang has more priority than previously selected lang. If not, imagine this odd case:

- Associate "Spanish" to "Spain" and "French" to "France".
- Create partner.
- Select "France" as country > "French" is selected.
- Now select "Spain" as previous one was a mistake > keeping "French" is incorrect.previously.

partner_country_lang/__manifest__.py
```diff
@@ -0,0 +1,14 @@
+# Copyright 2022 Tecnativa - Víctor Martínez
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+{
+    "name": "Partner country lang",
```
pedrobaeza:
```suggestion
    "name": "Partner language according country",
```

partner_country_lang/models/res_country.py
```diff
@@ -0,0 +1,15 @@
+# Copyright 2022 Tecnativa - Víctor Martínez
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import api, fields, models
+
+
+@api.model
+def _lang_get(self):
+    return self.env["res.lang"].get_installed()
```
pedrobaeza:
```suggestion
from odoo.addons.base.models.res_partner import _lang_get
```

partner_country_lang/models/res_partner.py
```diff
@@ -0,0 +1,30 @@
...
+                if country.lang:
+                    vals["lang"] = country.lang
+        return super().create(vals_list)
+
+    def write(self, vals):
+        if vals.get("country_id") and "lang" not in vals:
+            country = self.env["res.country"].browse(vals["country_id"])
+            if country.lang:
+                vals["lang"] = country.lang
+        return super().write(vals)
```
pedrobaeza: Refactor for not repeating code and add docstrings:

```suggestion
    def _adjust_lang_by_country(self, vals):
        """Adjust vals dictionary for adding the language of the country
        if no one is included in it, but country is.
        """
        if vals.get("country_id") and "lang" not in vals:
            country = self.env["res.country"].browse(vals["country_id"])
            if country.lang:
                vals["lang"] = country.lang

    @api.model_create_multi
    def create(self, vals_list):
        """Add lang if the partner is created with a country through code
        and no language is specified.
        """
        for vals in vals_list:
            self._adjust_lang_by_country(vals)
        return super().create(vals_list)

    def write(self, vals):
        """Change language if a country is written through code and no 
        one is modified as well.
        """
        self._adjust_lang_by_country(vals)
        return super().write(vals)
```

pedrobaeza (changes requested):

partner_country_lang/models/res_country.py
```diff
@@ -0,0 +1,15 @@
...
+
+
+class ResCountry(models.Model):
+    _inherit = "res.country"
+
+    lang = fields.Selection(_lang_get, string="Language")
```
chienandalu:
```suggestion
    lang = fields.Selection(_lang_get, string="Enforce language", help="This country partners' language will be set to this value by default")
```

partner_country_lang/models/res_partner.py
chienandalu: I guess what Dennis has in mind is to be able to keep exceptions (it can go in the roadmap). For example: Switzerland (french and german) or Belgium (french and dutch) language depend on factors that sometimes can't be isolated to a country area. To keep those exceptions it would be necessary a language lock mechanism or something like that.

pedrobaeza: But I don't see it as a problem, as the sequence in such case would be:

- Set the country.
- A default language is set (or none).
- You select the correct language.

victoralmau:
Changes done.

partner_country_lang/models/res_partner.py
chienandalu: Yes, but sometimes a simple address change can trigger the country onchange, and then the default lang value for that country would be applied.

pedrobaeza: Are you sure about this? I think this was improved in the web client and the onchange is not triggered if current value = previous value.

chienandalu: Oh, you're right. The onchange isn't triggered anymore :+1:

chienandalu (approved):
Thanks @victoralmau :+1:

partner_country_lang/models/res_partner.py
pedrobaeza: @rousseldenis is it OK for you then?

pedrobaeza (approved):
/ocabot merge nobump

PR #1320, opened 2022-08-16
Add module for restricting duplicate reference for contact

AungKoKoLin1997:
This module will prevent duplicating reference for partner.


pedrobaeza:
Isn't this already done in https://github.com/OCA/partner-contact/tree/14.0/partner_ref_unique ?

AungKoKoLin1997:
> Isn't this already done in https://github.com/OCA/partner-contact/tree/14.0/partner_ref_unique ?

Yes. I didn't check that is done or not. After I checked, this is the same.

PR #1323, opened 2022-08-21
[15.0][IMP] partner_ref_unique: improve corporative partners

i-vyshnevska:
allow non unique references for partners under one company

customer request: partners under the company should operate under the same reference they propagate it to documents so it is possible to select them just by reference
we add one more option which should provide expected behaviour

### DIFF
--- a/partner_ref_unique/models/res_company.py
+++ b/partner_ref_unique/models/res_company.py
@@ -12,6 +12,7 @@ class ResCompany(models.Model):
         selection=[
             ("none", "None"),
             ("companies", "Only companies"),
+            ("exclude_corporative", "All, except contacts of a partner"),
             ("all", "All partners"),
         ],
         string="Unique partner reference for",
--- a/partner_ref_unique/models/res_partner.py
+++ b/partner_ref_unique/models/res_partner.py
@@ -9,23 +9,32 @@ from odoo.exceptions import ValidationError
 class ResPartner(models.Model):
     _inherit = "res.partner"
 
-    @api.constrains("ref", "is_company", "company_id")
+    @api.constrains("ref", "is_company", "company_id", "parent_id")
     def _check_ref(self):
         for partner in self.filtered("ref"):
             # If the company is not defined in the partner, take current user company
             company = partner.company_id or self.env.company
             mode = company.partner_ref_unique
-            if mode == "all" or (mode == "companies" and partner.is_company):
-                domain = [
-                    ("id", "!=", partner.id),
-                    ("ref", "=", partner.ref),
-                ]
-                if mode == "companies":
+            if mode == "none":
+                continue
+            domain = [
+                ("id", "!=", partner.id),
+                ("ref", "=", partner.ref),
+            ]
+            if mode == "exclude_corporative":
+                if partner.parent_id:
+                    # if reference same as the main on the company skip the check
+                    # reference will be checked on company level
+                    if partner.ref == partner.parent_id.ref:
+                        continue
+            if mode == "companies":
+                if partner.is_company:
                     domain.append(("is_company", "=", True))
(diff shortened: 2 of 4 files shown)


partner_ref_unique/models/res_partner.py
```diff
@@ -9,23 +9,34 @@
...
                     domain.append(("is_company", "=", True))
-                other = self.search(domain)
-                # Don't raise when coming from contact merge wizard or no duplicates
-                if other and not self.env.context.get("partner_ref_unique_merging"):
-                    raise ValidationError(
-                        _("This reference is equal to partner '%s'")
-                        % other[0].display_name
-                    )
+                else:
+                    domain.append(("is_company", "=", False))
```
TDu: Is this needed ? As it worked before and nothing has been changed for this mode.

partner_ref_unique/models/res_company.py
```diff
@@ -12,6 +12,7 @@ class ResCompany(models.Model):
         selection=[
             ("none", "None"),
             ("companies", "Only companies"),
+            ("exclude_corporative", "All, except partners in company"),
```
TDu: does it not make more sense like this ?
```suggestion
            ("exclude_corporative", "All, except contacts of a partner"),
```

partner_ref_unique/models/res_partner.py
```diff
@@ -9,23 +9,34 @@
...
+            if mode == "exclude_corporative":
+                if partner.is_company:
+                    domain.append(("is_company", "=", True))
```
TDu: With this we can have a company and a partner with the same ref ?

TDu (commented):
Also the readme needs to be updated with the new mode explained.

partner_ref_unique/models/res_partner.py
i-vyshnevska: I need to modify the domain as I refactored the first "if" condition, the behavior for this mode didn't change, and tests stayed with the same conditions

partner_ref_unique/models/res_partner.py
i-vyshnevska: refactored, I introduced all cases in the test, if the partner has a parent he can have the same ref no matter whether is it a company or individual

TDu:
Thanks for the changes, it would still be good to update this file https://github.com/OCA/partner-contact/blob/14.0/partner_ref_unique/readme/USAGE.rst

TDu (approved):

partner_ref_unique/models/res_partner.py
```diff
@@ -9,23 +9,32 @@
...
+            ]
+            if mode == "exclude_corporative":
+                if partner.parent_id:
+                    # if reference same as the main on the company skip the check
+                    # reference will be checked on company level
+                    if partner.ref == partner.parent_id.ref:
+                        continue
+            if mode == "companies":
+                if partner.is_company:
                     domain.append(("is_company", "=", True))
```
simahawk:
```suggestion
                domain.append(("is_company", "=", partner.is_company))
```

partner_ref_unique/models/res_partner.py
```diff
@@ -9,23 +9,32 @@
...
-                other = self.search(domain)
-                # Don't raise when coming from contact merge wizard or no duplicates
-                if other and not self.env.context.get("partner_ref_unique_merging"):
-                    raise ValidationError(
-                        _("This reference is equal to partner '%s'")
-                        % other[0].display_name
-                    )
+                else:
+                    domain.append(("is_company", "=", False))
```
simahawk:
```suggestion
```

simahawk (approved):
LG, minor remarks

pedrobaeza:
Tests are red...

PR #1327, opened 2022-08-24
[15.0][ADD] base_municipality

espo-tony:
This module introduces a new "Municipality" field on company and partner records, to be used as complement to all other address fields (e.g.: country, state, zip, city, street, ...).

### DIFF
--- a/base_municipality/__init__.py
+++ b/base_municipality/__init__.py
@@ -1,0 +1,1 @@
+from . import models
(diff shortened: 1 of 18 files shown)


pedrobaeza:
I think this is a localization concept that should be added in the corresponding localization repository, don't you think?

espo-tony:
Hi Pedro,
While it's true that municipalities are not often used, and their legal meaning may change from country to country, they also aren't unique to a single country, so I wouldn't know what localization repository should host this module.

Actually, they represent a second-level administrative division which, is a concept in many countries (e.g.: French departments, Dutch gemeenten, etc.) yet to the best of my knowledge they're not mapped anywhere in Odoo.

pedrobaeza:
We have `base_location_nuts` for a normalized administrative territory divisions.

espo-tony:
Thank you for referencing this module. I looked into it and indeed I agree that it addresses a similar scenario than the module I proposed. For this reason I will cancel this PR.

PR #1334, opened 2022-09-09
[FIX] filter on sector searching in name

gfcapalbo:
Old domain would create an error in search. 
By searching in sector_id name we have expected behaviour.


NL66278 (approved):
:+1: LGTM. Known to be working in live. Travis failed, but only in one job out of four, so seems some temporary problem.

ntsirintanis (approved):

PR #1335, opened 2022-09-10
[Fix] Opening Deduplicate Contacts when clicking on main Contacts men…

nikunjantala:
[FIX] partner_deduplicate_acl: Put menu on lower sequence

So that when clicking on Contacts main menu, it doesn't open the
wizard by default.

Bildschirmfoto_vom_2022-09-01_16-28-13

### DIFF
--- a/partner_deduplicate_acl/README.rst
+++ b/partner_deduplicate_acl/README.rst
@@ -66,7 +66,7 @@ Usage
 To use this module, you need to:
 
 #. Ask your admin to give you the new rights.
-#. Go to *CRM > Tools > Deduplicate Contacts* as usual.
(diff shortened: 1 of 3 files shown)


pedrobaeza:
Duplicate of #1310 . Can you check this one instead?

pedrobaeza:
I see yours more complete, but please change commit message to

```
[FIX] partner_deduplicate_acl: Put menu on lower sequence

So that when clicking on Contacts main menu, it doesn't open the
wizard by default.
```

nileshsheliya:
Commit message changed.

pedrobaeza (approved):

PR #1339, opened 2022-09-14
[15.0][MIG] partner_stage: Migration to 15.0

BT-dmontull:

PR #1216, opened 2022-02-01
[14.0][REF] partner_tier_validation: do not archive, depend on partner_stage

dreispt:
Major change.
Supersedes #1142 

- [x] Depends (bundles) #1192

PR #1340, opened 2022-09-14
[15.0][MIG] partner_tier_validation: Migration to 15.0

BT-dmontull:
dependency
[Partner_stage](https://github.com/OCA/partner-contact/pull/1339)


rousseldenis:
@BT-dmontull Thanks for this. For a better review, could you add the link with the PR needed to run this (partner_stage)?

BT-dmontull:
thanks @rousseldenis
I have also tried to put the dependency in the file test-requirements.txt but it seems that the tests don't detect it.

partner_tier_validation/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,12 @@
...
+before they can be used.
+
+The rule can be extended to new non-company contact,
+but beware that may cause issues with automatically created new contacts,
+such as the ones generated when processing incoming emails.
+
+If the 'Is Company' or 'Parent' field changes then the contact is Request
+for approval.
+
+For this, the new Contact record is kept as "Archived" until it is approved.
```
yostashiro, line 12: I suppose this description is no longer valid after https://github.com/OCA/partner-contact/pull/1216.

According to my limited understanding with no experience of using the module, it should be some thing like:

```
Note that there is nothing preventing the use of the newly created (unapproved) contact,
unless such function is implemented by another module depending on the partner stage.
```

Could @dreispt please confirm this?

yostashiro (changes requested):

AungKoKoLin1997 (approved):
Test functional in local environment and work as expected

partner_tier_validation/readme/DESCRIPTION.rst
dreispt: I believe so. It sounds like that, as is, this module might not make much sense?
Maybe it could enforce that only approved records are approved, encapsulating that decision in a computed field.
The Partner Stage related module could then extend that for richer state-based decisions.

PR #1350, opened 2022-09-30
[15.0][MIG] partner_pricelist_search: Migration to 15.0

victoralmau:
Superseed: https://github.com/OCA/partner-contact/pull/1254

Migration to 15.0

Please @pedrobaeza and @carlosdauden can you review it?

@Tecnativa TT37297

PR #1351, opened 2022-10-03
[13.0][FIX] partner_pricelist_search: speedup search

chienandalu:
We try to optimize the latter pricelist filter over the minimum set of records possible.

TT39742

ping @pedrobaeza @carlosdauden

### DIFF
--- a/partner_pricelist_search/models/res_partner.py
+++ b/partner_pricelist_search/models/res_partner.py
@@ -1,8 +1,7 @@
 # Copyright 2021 Tecnativa - Carlos Dauden
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
-from odoo import _, api, fields, models
-from odoo.exceptions import UserError
+from odoo import api, fields, models
 
 
 class Pricelist(models.Model):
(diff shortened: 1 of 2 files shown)


partner_pricelist_search/models/res_partner.py
```diff
@@ -39,5 +53,6 @@ def filter_func(partner):
                 _("Pricelist field do not support search with the operator '%s'.")
                 % operator
             )
-        partners = self.with_context(prefetch_fields=False).search([])
+        domain = self.env.context.get("search_partner_domain", [])
+        partners = self.with_context(prefetch_fields=False).search(domain)
         return [("id", "in", partners.filtered(filter_func).ids)]
```
carlosdauden: Can you add this changes to simplify code:
https://github.com/OCA/partner-contact/pull/1350/commits/60571d4406f887cc731150deb4eb0ca95981dbf4#diff-d7fde32bb36763c53ba086144a1584254003162fa6c604c59b7dbe68d7c6767aL17-R19

partner_pricelist_search/models/res_partner.py
```diff
@@ -12,6 +12,20 @@ class Pricelist(models.Model):
...
+    @api.model
+    def search(self, args, offset=0, limit=None, order=None, count=False):
+        # Substitute pricelist tuple
+        partner_domain = [
+            (1, "=", 1)
+            if (isinstance(x, tuple) and x[0] == "property_product_pricelist")
+            else x
+            for x in args
+        ]
+        self.env.context = dict(self.env.context, search_partner_domain=partner_domain)
```
carlosdauden: I understand that if you do this it's because you can't use with_context, right?
I don't know why I thought the context was "frozen" in this version.

carlosdauden (approved):

partner_pricelist_search/models/res_partner.py
chienandalu: Well, indeed that would be elegant way. I'll be changing it.

Anyway, what is immutable is the [`frozendict`](https://github.com/odoo/odoo/blob/745b5dad8c7bf817a7d60076c1d12307132e71b1/odoo/tools/misc.py#L997-L1014) object itself. But the `env` class property isn't so you can dump a new `dict` into it.

Anyway, internally it makes a difference to do it with `with_context`:
```python
>>> type(self.env.context)
<class 'odoo.tools.misc.frozendict'>

>>> self.env.context = dict(self.env.context, test1="")
>>> type(self.env.context)
<class 'dict'>

>>> self = self.with_context(test2="")
>>> type(self.env.context)
<class 'odoo.tools.misc.frozendict'>
```

partner_pricelist_search/models/res_partner.py
chienandalu: Sure I will

chienandalu:
Changes done

pedrobaeza (approved):

PR #1383, opened 2022-10-14
[13.0][FIX] partner_pricelist_search: Remove list type also from search arguments

sergio-teruel:
cc @Tecnativa

ping @chienandalu @victoralmau 

wiget domain set domain with lists instead tuples with this format [["field_name", "operator", value]]. If not take into account lists this cause "Maximum deep recursion"


carlosdauden (approved):

NL66278 (approved):
:+1: Change itself LGTM. Maybe also change the class to Partner instead of Pricelist.

And will you are at  it, change `http` to `https`.

pedrobaeza:
Changed Ronald's suggestions as well, so merging:

/ocabot merge patch

PR #1386, opened 2022-10-20
15.0 mig partner bank code

KM-joel:
This module adds a field Bank code and Bank branch code on Banks.

I just changed the version and checked if all the functionality worked because I needed this module, which is why I migrated it


rousseldenis:
This has been already migrated

PR #1387, opened 2022-10-25
[FIX] Corrected German Street translation to be uniform with odoo core

dhx:
Don't have the permissions to do it in Weblate myself, the links would be:

https://translation.odoo-community.org/translate/partner-contact-15-0/partner-contact-15-0-partner_address_street3/de/?checksum=3cda62386e1bd105

https://translation.odoo-community.org/translate/partner-contact-15-0/partner-contact-15-0-partner_address_street3/de/?checksum=1e878b4bc77bb9c7


dhx:
@psc as you did the last german translation, what do you think?

PR #1394, opened 2022-11-08
[FIX] pre-commit py-lint error

CarlosRoca13:


pedrobaeza:
Why not including #1395 here for solving everything together?

PR #1403, opened 2022-11-17
[12.0] Switch to GH actions

simahawk:

PR #1404, opened 2022-11-17
Fix manifests and pylint

SilvioC2C:
Fix needed for #1403


partner_multi_relation/views/res_partner_relation_all.xml
```diff
@@ -123,27 +123,4 @@
         </field>
     </record>
 
-    <record
```
NL66278, line 126: What is the reason for deleting this action?

SilvioC2C: It's a duplicate:
https://github.com/OCA/partner-contact/blob/12.0/partner_multi_relation/views/menu.xml#L6

NL66278 (approved):
:+1: LGTM

PR #1405, opened 2022-11-19
[12.0][ADD] partner_address_city_district

hbrunn:


quannguyen1560:
Hi based on the discussion in this https://github.com/OCA/partner-contact/pull/1460, I think you can check if this module https://github.com/OCA/partner-contact/tree/12.0/base_location_nuts is good enough for you or not.

hbrunn:
thanks for pointing this out, but the customer didn't want the baggage that comes with nuts and also have it integrate nicely with `base_address_city`. I guess a very similar rationale lead @PierrickBrun to the PR

PR #1409, opened 2022-11-22
[14.0] [ADD] `partner_duns`

ivantodorovich:
This module let's you set a DUNS number on your contacts.

**What's a DUNS Number?**

The **The Data Universal Numbering System**, abbreviated as **DUNS** or **D-U-N-S**, is a proprietary system developed and managed by [Dun & Bradstreet (D&B)](https://www.dnb.com/duns-number.html) that assigns a unique numeric identifier to a single business entity.

It was introduced in 1963 to support D&B's credit reporting practice. It is standard worldwide.


partner_duns/models/res_partner.py
```diff
@@ -0,0 +1,66 @@
...
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import api, fields, models
+
+
+class ResPartner(models.Model):
+    _inherit = "res.partner"
+
+    duns = fields.Char(
+        help="Data Universal Numbering System",
```
bodedra:
```suggestion
    duns = fields.Char(
        string="DUNS",
        help="Data Universal Numbering System",
```

bodedra (approved):
Functional Review 👍 

Suggest minor correction.

partner_duns/models/res_partner.py
ivantodorovich: Totally agree, thanks for the suggestion!
I've just changed it

bodedra (approved):
LGTM

jcoux (approved):

PR #1412, opened 2022-11-24
[16.0][MIG] partner_helper

florian-dacosta:
standard migraton, works without changes

@bealdav


bealdav (approved):
LGTM code review, it seems no changes in Odoo can break things

rousseldenis:
@florian-dacosta Could you maybe add tests for both uncovered code cases ?

image

bealdav:
@rousseldenis might we consider to only add tests request in the roadmap section here for the next migration considering that this piece of code is not so complex, then no so critical, because completely independent of odoo changes (except number of street fields ;-) )

It's also used in production since several years.

It's important that makes things happen with trivial PR and make security barriers proportional to risks, I suppose.

@florian-dacosta are you ok with that ?

florian-dacosta:
Adding test on this module is on my todolist but it is not in the top of it, that's why I did not do it yet !

I agree, it is a pity to block the migration in that case but also if we just add it in the roadmap, there are chance we never add tests ... 
So I have no strong opinion on this
I'll reopen it though, since I will eventually add the test one day, hopefully before the next 4 months!
@rousseldenis I can't reopen it, is it normal? Should I create a new one ?

rousseldenis:
> Adding test on this module is on my todolist but it is not in the top of it, that's why I did not do it yet !
> 
> I agree, it is a pity to block the migration in that case but also if we just add it in the roadmap, there are chance we never add tests ... So I have no strong opinion on this I'll reopen it though, since I will eventually add the test one day, hopefully before the next 4 months! @rousseldenis I can't reopen it, is it normal? Should I create a new one ?

Done

rousseldenis:
@florian-dacosta @bealdav I'm wondering why partner_helper ?
 
 As this module implements an adress splitting, it should be renamed by partner_address_split or something like this.
IMHO, this is not a good idea to have modules that have such names as they don't represent what module does in reality. Those 'fourre-tout' modules should be avoided. Moreover, it should help also people finding it

bealdav:
nice, partner_address_split seems a really good name. in that case, we could add a log warning to inform of the new name here and make a new PR with the new name.

florian-dacosta:
Replaced by https://github.com/OCA/partner-contact/pull/1543

PR #1415, opened 2022-11-24
[15.0][IMP] partner_ref_unique: Check also archived partner reference

ghost:


partner_ref_unique/tests/test_res_partner_ref.py
```diff
@@ -64,6 +64,16 @@ def test_partner1_w_company_new_partner_wo_company(self):
...
 
+    def test_partner1_archived_new_partner_same_ref(self):
+        self.company.partner_ref_unique = "all"
+        self.partner1.write({"company_id": False, "ref": "same_ref"})
+        self.partner1.active = False
+        with self.assertRaises(ValidationError):
+            self.partner_obj.create(
+                {"name": "other", "ref": "same_ref", "company_id": self.company.id}
+            )
+        self.partner1.unlink()
```
jcoux:
```suggestion
        with self.assertRaises(ValidationError) as error:
            self.partner_obj.create(
                {"name": "other", "ref": "same_ref", "company_id": self.company.id}
            )
        self.assertEqual(f"This reference is equal to partner '{self.partner1.display_name}'", error.exception.name)
        # No duplicate: no error
        self.partner1.unlink()
        self.partner_obj.create(
            {"name": "other", "ref": "same_ref", "company_id": self.company.id}
        )
```

jcoux (approved):

damdam-s (approved):

partner_ref_unique/models/res_partner.py
```diff
@@ -22,7 +22,7 @@ def _check_ref(self):
                 ]
                 if mode == "companies":
                     domain.append(("is_company", "=", True))
-                other = self.search(domain)
+                other = self.with_context(active_test=False).search(domain)
```
rousseldenis, line 25: I think the original behaviour is intended in order to archive partner without deleting it (to keep history).

Your test shows that you should unlink() the partner in order to create a new one with the same ref

damdam-s: @rousseldenis original behavior allows to have 2 partners w/ the same ref as soon as one of them is archived. With this code, we test that all the unicity of the ref is among all partners (archived or not).

About the test, we mimic the same 'unlink' than on the original tests. 
In the `test_partner1_archived_new_partner_same_ref` test, we archive the partner1 (line 70) and then try to create a new one w/ the same ref. 

Maybe/probably I misunderstood your comment :smile:

rousseldenis: > @rousseldenis original behavior allows to have 2 partners w/ the same ref as soon as one of them is archived. With this code, we test that all the unicity of the ref is among all partners (archived or not).

Yes I understood the purpose of this.

My question is about the original behaviour. Maybe was it at purpose to have archived partners with same reference than the active one.

Your change is not a fix but a module behavior change

NL66278: @damdam-s I have to agree with @rousseldenis here: this is a behavior change. You could give users the option to also test inactive partners by creating a new module, depending on this one, let us call it partner_ref_unique_total, building on this one. Module just needs to override _check_ref and call super with active_test=False. So users / admins have a change how they want the check to behave.

PR #1420, opened 2022-11-30
[16.0[]MIG] partner_phone_secondary

emagdalenaC2i:


partner_phone_secondary/models/res_partner.py
```diff
@@ -0,0 +1,17 @@
...
+from odoo import api, fields, models
+
+
+class ResPartner(models.Model):
+    _inherit = "res.partner"
+
+    phone2 = fields.Char("Phone (Secondary)")
+
+    @api.onchange("phone2", "country_id", "company_id")
+    def _onchange_phone2_validation(self):
```
rousseldenis, line 13: I think the filed should be turned into computed readonly False one.

@emagdalenaC2i

rousseldenis (changes requested):

partner_phone_secondary/models/res_partner.py
emagdalenaC2i: @rousseldenis Isn't readonly false by default?

rousseldenis: Computed fields are readonly=True by default.

But you can declare a field as computed, readonly=False. That was developed in order to replace completely the onchanges.

But, to be tested if it allows the same behaviour (format the field that triggered the compute)

partner_phone_secondary/tests/test_partner_phone_secondary.py
```diff
@@ -0,0 +1,19 @@
...
+
+class TestPartnerPhoneSecondary(TransactionCase):
+    def test_01_phone_validation_compatibility(self):
+        # This is here only to get code coverage.
+        with Form(self.env["res.partner"]) as form:
+            form.name = "French Partner"
+            form.country_id = self.env.ref("base.fr")
+            form.phone = "766666666"
+            form.phone2 = "766666666"
+            # It's not the purpose of this module to test phone_validation,
```
rousseldenis, line 15: I would say on this, that this is partly true.

As the code includes the call to `phone_format()`, we should test it is well called (e.g.: with mock use or ...)

Bilbonet:
Hi @emagdalenaC2i 

Thank you for your work on this migration. Would you be able to address the pending review comment by improving the test to verify that phone_format() is correctly applied?

For reference, you could take a look at the test in the 18.0 branch, where this module has already been merged:
https://github.com/OCA/partner-contact/blob/18.0/partner_phone_secondary/tests/test_partner_phone_secondary.py

It would be great to get this resolved so the 16.0 PR can move forward with the review and hopefully be approved and merged.

Thanks again!

Issue #1353, opened 2022-10-03
Migration to version 16.0


sbidoul:
# Todo

https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-16.0

# Modules to migrate

- [x] account_partner_company_group - By @Highcooley - #1521
- [x] animal - By @Deriman-Alonso - #1736
- [x] base_country_state_translatable - By @AungKoKoLin1997 - #1411
- [x] base_location - By @pedrobaeza - #1361
- [x] base_location_geonames_import - By @pedrobaeza - #1385
- [x] base_location_nuts - By @edlopen - #1768
- [x] base_partner_company_group - By @Highcooley - #1504
- [x] base_partner_sequence - By @Nikul-OSI - #1453
- [x] crm_partner_company_group - By @Highcooley - #1510
- [x] partner_address_street3 - By @SodexisTeam - #1433
- [x] partner_address_two_lines - By @yankinmax - #1470
- [ ] partner_affiliate
- [x] partner_bank_code - By @hbrunn - #1550
- [x] partner_capital - By @EmilioPascual - #1547
- [x] partner_category_security - By @victoralmau - #1669
- [x] partner_category_security_crm - By @victoralmau - #1676
- [ ] partner_company_group
- [x] partner_company_type - By @apineux - #1355
- [x] partner_contact_access_link - By @vib-adhoc - #1390
- [x] partner_contact_address_default - By @Reyes4711-S73 - #1449
- [x] partner_contact_age_range - By @Highcooley - #1501
- [x] partner_contact_birthdate - By @carmenbianca - #1368
- [x] partner_contact_birthplace - By @Borruso - #1734
- [x] partner_contact_department - By @aiendry-aktivsoftware - #1556
- [x] partner_contact_gender - By @carmenbianca - #1369
- [x] partner_contact_in_several_companies - By @stephansainleger - #1852
- [x] partner_contact_job_position - By @carmenbianca - #1381
- [x] partner_contact_lang - By @pegonzalezspesol - #1370
- [x] partner_contact_nationality - By @marielejeune - #1380
- [x] partner_contact_personal_information_page - By @marielejeune - #1377
- [x] partner_contact_type_end_user - By @acsonefho - #1561
- [ ] partner_data_vies_populator - By @dreispt - #1642
- [x] partner_deduplicate_acl - By @SoporteOperu - #1524
- [x] partner_deduplicate_by_ref - By @SoporteOperu - #1525
[...]

PR #1432, opened 2022-12-16
[16.0][MIG] partner_affiliate: Migration to 16.0

FernandoRomera:


pedrobaeza:
Why are you spamming all your PRs threads if you don't have the permission to mark an addon to be migrated in any of them?

FernandoRomera:
I thought that an action of simply notifying the rest of the community could be done by any contributor. I will not do it again.

pedrobaeza:
No, it's not, as you may overwrite another legitimate pull request with the same migration. That has to be decided by a PSC or the module maintainer. Become maintainer of some modules.

FernandoRomera:
But that is impossible to happen the first time you try to migrate the module. Therefore it would be legitimate, to inform the community of your action, to avoid duplication of work.

pedrobaeza:
Not if you are maintainer on the previous version, but the way to inform that the community is not here, but on the migration issue as centralized place to do that, as for example: https://github.com/OCA/partner-contact/issues/1353#issuecomment-1293876384

FernandoRomera:
#1353

AntoniRomera (approved):

MRomeera (approved):
ok

PR #1437, opened 2022-12-28
[14.0][ADD] partner_child_ids_confirm_remove

FrancescoBallerini:
I think the 'remove' button on child id partner form is dangerous because it doesn't ask for any confirmation and it directly remove record from database. 

In this case I think the rollback mechanism of 'discard' button might not be the way because the removal might be done accidentally.

I asked for this functionality and there doesn't seems to be any module yet so I made this small module that add the confirm dialog before record remove and might be good to prevent accidentally record deletion.


FrancescoBallerini:
Closed because of bad repository setup, I'm not sure how I can fix so I opened a new one: https://github.com/OCA/partner-contact/pull/1499

PR #1441, opened 2023-01-02
[14.0][IMP] base-location-geonames-import

lbelorgey:
Required for the module Base Location Geonames Import in l10n-france.
Zip column contains "CEDEX ..." in geonames file of France. "CEDEX ..." must be moved to city name.

Case :
Zip: `13626 CEDEX 1` => `13626`
City name: `Aix-en-Provence` => `Aix-en-Provence CEDEX 1`

### DIFF
--- a/base_location_geonames_import/__manifest__.py
+++ b/base_location_geonames_import/__manifest__.py
@@ -7,7 +7,7 @@
 
 {
     "name": "Base Location Geonames Import",
-    "version": "14.0.1.0.2",
+    "version": "14.0.1.0.3",
     "category": "Partner Management",
     "license": "AGPL-3",
     "summary": "Import zip entries from Geonames",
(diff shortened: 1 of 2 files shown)


pedrobaeza:
This is something that must be done in a localization module, like we have in l10n_es_toponyms.

lbelorgey:
> This is something that must be done in a localization module, like we have in l10n_es_toponyms.

I agree with you.
I supersede `transform_city_name` function in l10n-france but I need the zip field to transform the city name.

pedrobaeza:
But unfortunately this change makes any other override incompatible, so it can't be done on a stable branch. You should check another way (maybe through context). There's also the possibility to change the data source column by country. Have you checked it?

lbelorgey:
Ok. I understand. No solution found with context and no other column contains zip in data source but I found a solution not so good but it does the job.
I keep in mind for version 17.0.

PR #1453, opened 2023-01-18
[MIG] [16.0] base_partner_sequence

Nikul-OSI:


Nikul-OSI:
@dreispt Can you please take a look?

base_partner_sequence/models/partner.py
```diff
@@ -0,0 +1,65 @@
...
+
+    def _needs_ref(self, vals=None):
+        """
+        Checks whether a sequence value should be assigned to a partner's 'ref'
+
+        :param vals: known field values of the partner object
+        :return: true iff a sequence value should be assigned to the\
+                      partner's 'ref'
+        """
+        if not vals and not self:  # pragma: no cover
```
quannguyen1560, line 51: Can we use `self.ensure_one()` in this case?
so you don't need to `if not self` and `if self`. You just need to check if vals have value or not.

base_partner_sequence/models/partner.py
```diff
@@ -0,0 +1,65 @@
...
+
+from odoo import _, api, exceptions, models
+
+
+class ResPartner(models.Model):
+    """Assigns 'ref' from a sequence on creation and copying"""
+
+    _inherit = "res.partner"
+
+    def _get_next_ref(self, vals=None):
```
quannguyen1560, line 15: Why do we need `vals=None` in here? Because I don't see that you are using `vals` in function
Is there any module that depends to this module and need `vals`? In that case, Please add comment for this.
By the way, should we need `@api.model` for this function?

quannguyen1560 (changes requested):

base_partner_sequence/models/partner.py
Nikul-OSI: @quannguyen1560 code updated base on your comments.

base_partner_sequence/models/partner.py
Nikul-OSI: @quannguyen1560  code updated base on your comments.

base_partner_sequence/models/partner.py
Nikul-OSI: @quannguyen1560 This method used before record created, so here we haven't **self**.

**ensure_one()** This method check that the current recordset holds a single record or not and if not then raise singleton.

imlopes (approved):

lmignon (approved):
Thank you @Nikul-OSI LGTM (Code review + functional tests)

base_partner_sequence/models/partner.py
```diff
@@ -0,0 +1,65 @@
...
+    _inherit = "res.partner"
+
+    def _get_next_ref(self):
+        return self.env["ir.sequence"].next_by_code("res.partner")
+
+    @api.model
+    def create(self, vals):
+        if not vals.get("ref") and self._needs_ref(vals=vals):
+            vals["ref"] = self._get_next_ref()
+        return super(ResPartner, self).create(vals)
```
lmignon: @Nikul-OSI The model odoo.addons.base_partner_sequence.models.partner is not overriding the create method in batch 
```suggestion
    @api.model_create_multi
    def create(self, vals_list):
        for vals in vals_list:
            if not vals.get("ref") and self._needs_ref(vals=vals):
                vals["ref"] = self._get_next_ref()
        return super(ResPartner, self).create(vals_list)
```

lmignon (changes requested):

lmignon:
@Nikul-OSI Do you plan to work on this one to finalize the review or do you prefer I open a new pr based on your to include my proposal?

Regards,

lmi

Nikul-OSI:
I will check very soon 🙂

dreispt:
Ping @Nikul-OSI

Nikul-OSI:
@rousseldenis @lmignon can you please take a look?

dreispt (approved):

base_partner_sequence/models/partner.py
```diff
@@ -35,7 +36,7 @@ def write(self, vals):
                 and partner._needs_ref(vals=partner_vals)
                 and not partner.ref
             ):
-                partner_vals["ref"] = partner._get_next_ref(vals=partner_vals)
+                partner_vals["ref"] = partner._get_next_ref()
```
lmignon: @Nikul-OSI Why do you remove the `vals` paramter from `_get_next_ref` method? Even id it's useless in the current implementation, the `get_next_ref` method could be overridden into an other addon and build the sequence according to some infos into the vals dict. Here your change the existing api without benefits.

lmignon (changes requested):

base_partner_sequence/models/partner.py
Nikul-OSI: Because `_get_next_ref` method, not used **vals**.  Please check this method. https://github.com/OCA/partner-contact/pull/1453/files#diff-cec1f3044f9be2641c8e43fe9d3fb15bf0f9b957a37ee8fe908ef5b62b0fc84aR15-R16

lmignon: @Nikul-OSI I konw it doesn't use it in this addon but others addons could extend (override) the `get_next_ref` method and uses the information from `vals` (For exemple use the country code into the sequence). It's important to keep the same signature to still be able to extend `get_next_ref`  and use the the information from `vals`....

Nikul-OSI: @lmignon now done

lmignon (approved):
Thank you @Nikul-OSI LGTM (Code review + functional)

PR #1460, opened 2023-02-01
[ADD] 16.0: base_location_district module

PierrickBrun:
This module adds a district field on the partner.

It is inspired by the base_address_extended standard module.


rvalyi:
seems there is some overlap with https://github.com/OCA/partner-contact/tree/15.0/base_location_nuts but not sure if it  may still be useful, may be it should be an extension of base_location_nuts? Or some module may glue them later also? On the other hand having a such very small module only for the districts seems useful as well...
Any idea @pedrobaeza @antespi @SimoRubi @gaso ?

pedrobaeza:
I don't know the concept of district, but it can be one of the levels of NUTS. It's true however that the NUTS is very heavy and rigid classification.

PierrickBrun:
I did not know about base_location_nuts.

My use case is to select french "priority districts" (https://www.data.gouv.fr/fr/datasets/liste-des-quartiers-prioritaires-de-la-ville/ )

The goal is to create statistics about these districts/neighborhoods.

From what I understand about NUTS, this level is not covered by it.


Maybe the naming is wrong for what I intend to do, I want to represent a portion of a city (more precise than a borough)

pedrobaeza:
In NUTS, there are four standard levels, and one of them may be the district. If it's somehow official, it would be covered by NUTS, as it's an European initiative to put this information in a standard way.

rvalyi:
that would be a kind of level 4 NUTS in France, but indeed not officially. We may be interested with this district support for Brazil as we implemented it our own way. https://github.com/OCA/l10n-brazil/blob/16.0/l10n_br_base/models/party_mixin.py#L66
However in our case it's a simple Char field and not a many2one. I think they have districts at least in the USA, but apparently they can be of several kinds.

In SAP, the district field is a Char field https://help.sap.com/docs/SAP_NETWEAVER_702/fe2425b86c55101499da8b52f8f96122/4e43e382eb811552e10000000a421937.html#length-of-the-street%2C-city%2C-and-district-fields

Any chance we can play it nice with localization using Char fields for districts?

PierrickBrun:
> Any chance we can play it nice with localization using Char fields for districts?

Since my use case is statistics I want to have a defined list to make sure there is no duplicates so I'd like to have it like this (tbh this is not very different than a char field since only the name is used)

The link with the city may not work everywhere though

PierrickBrun:
In geonames that would be a PPLX (section of populated place) (but as @rvalyi said, in my case this is not the official list, this is a different one with a special use).

I do think that could be useful in other cases (for instance we could have a group of shops wanting to have statistics about customers depending on districts with their own definition of the districts)

Here is the geoname link for one of those [PPLX](https://www.geonames.org/12047796/quartier-charpennes-tonkin.html#)

rvalyi:
@hbrunn @PierrickBrun aren't your two modules pretty much the same thing/ https://github.com/OCA/partner-contact/pull/1405 ?

hbrunn:
looks like to me

PierrickBrun:
I think this is the same, yes.

The only difference I can point is that in my module a district can be split between several cities (sometimes it happens in urban areas)

PR #1462, opened 2023-02-07
[13.0][FIX] base_location: Force saving city_id

pedrobaeza:
Steps to reproduce the problem:

- Create a parent contact with a full location with city.
- Create a child contact of type "Contact".
- Try to save.

You get the error "The city of partner ... differs from that in location ...".

That's because the field `city_id` is readonly when type = 'contact' and no force_save is indicated.

We therefore include the force_save flag for avoiding the problem.

@Tecnativa TT41570


sergio-teruel (approved):

victoralmau (approved):

pedrobaeza:
The problem seems to not be present in v14 nor v15.

PR #1463, opened 2023-02-07
[ADD] 14.0 partner_category_unique_per_organization

PierrickBrun:
This module allows to mark a partner category as "unique per organization".

This means only 1 contact of this organization can have this category

This is useful to make sure the contact for a specific communication is up to date.

### DIFF
--- a/partner_category_unique_per_organization/__manifest__.py
+++ b/partner_category_unique_per_organization/__manifest__.py
@@ -1,0 +1,16 @@
+{
+    "name": "Partner Contact Unique per Organization",
+    "version": "14.0.1.0.0",
+    "category": "Partner",
+    "website": "https://github.com/OCA/partner-contact",
+    "author": "Odoo Community Association (OCA), Akretion",
+    "license": "AGPL-3",
+    "application": False,
+    "installable": True,
+    "depends": [
+        "base",
+    ],
+    "data": [
+        "views/res_partner_views.xml",
+    ],
+}
--- a/partner_category_unique_per_organization/models/partner.py
+++ b/partner_category_unique_per_organization/models/partner.py
@@ -1,0 +1,31 @@
+# Copyright 2023 Akretion (https://www.akretion.com).
+# @author Pierrick Brun <pierrick.brun@akretion.com>
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import _, api, models
+from odoo.exceptions import UserError
+
+
+class ResPartner(models.Model):
+    _inherit = "res.partner"
+
+    @api.constrains("category_id")
+    def _check_category_unique_per_organization(self):
+        for record in self:
+            for categ in record.category_id.filtered("unique_per_organization"):
+                already_existing_contacts = (
+                    record.commercial_partner_id.child_ids.filtered(
+                        lambda c: c.id != record.id and categ.id in c.category_id.ids
+                    )
+                )
+                if already_existing_contacts:
+                    raise UserError(
+                        _(
+                            "You can only have 1 contact with the category"
+                            " {} in the organization {} ({})"
+                        ).format(
+                            categ.name,
+                            record.commercial_partner_id.name,
+                            already_existing_contacts[0].name,
+                        )
+                    )
--- a/partner_category_unique_per_organization/__init__.py
+++ b/partner_category_unique_per_organization/__init__.py
@@ -1,0 +1,1 @@
+from . import models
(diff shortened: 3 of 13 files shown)


clementmbr (approved):
Perfect! LGTM.

partner_category_unique_per_organization/models/partner.py
```diff
@@ -0,0 +1,33 @@
...
+
+class ResPartner(models.Model):
+    _inherit = "res.partner"
+
+    @api.constrains("category_id")
+    def _check_category_unique_per_organization(self):
+        for record in self:
+            for categ in record.category_id.filtered(
+                lambda categ: categ.unique_per_organization
+            ):
```
fcayre: isn't `record.category_id.filtered("unique_per_organization")` enough and more readable than a lambda?

partner_category_unique_per_organization/models/partner.py
```diff
@@ -0,0 +1,33 @@
...
+                already_existing_contacts = (
+                    record.commercial_partner_id.child_ids.filtered(
+                        lambda c: c.id != record.id and categ.id in c.category_id.ids
+                    )
+                )
+                if already_existing_contacts:
+                    raise UserError(
+                        _(
+                            "You can only have 1 contact with the category"
+                            " {} for the organization {} ({})"
```
fcayre: s/for/in/ ?

fcayre (approved):
LGTM

partner_category_unique_per_organization/models/partner.py
```diff
@@ -0,0 +1,33 @@
...
+
+    @api.constrains("category_id")
+    def _check_category_unique_per_organization(self):
+        for record in self:
+            for categ in record.category_id.filtered(
+                lambda categ: categ.unique_per_organization
+            ):
+                already_existing_contacts = (
+                    record.commercial_partner_id.child_ids.filtered(
+                        lambda c: c.id != record.id and categ.id in c.category_id.ids
```
mourad-ehm, line 18: Why you use categ.id in c.category_id.ids rather than categ == c.category_id ?
Is'nt test without "for loop" possible ?
all_child_with_unique_categ = (
                    record.commercial_partner_id.child_ids.filtered(
                        lambda c: c.category_id and  c.category_id.unique_per_organization))
all_used_unique_categ =  record.commercial_partner_id.child_ids.mapped("category_id").filtered("unique_per_organization")
if (len(set(all_child_with_unique_categ.ids)) != len(set(all_used_unique_categ.ids)): --> error msg

clementmbr: > Why you use categ.id in c.category_id.ids rather than categ == c.category_id ?

`category_id`is (sadly) a [m2m](https://github.com/odoo/odoo/blob/14.0/odoo/addons/base/models/res_partner.py#L180) so need to use "in".

What is not useful is the use of the ".id"... and it might be better to use the letter "p" as "partner" rather than "c" which can be slightly confusing IMO :
`lambda p: p != record and categ in p.category_id`
... but not really a blocking question.

> Is'nt test without "for loop" possible ?

Your proposition doesn't work because of category_id being a m2m field.
But I think the for loop is necessary as you can have many categories "unique per organization" for the same organization and you want to catch which one is already in another partner, and what is the name of this partner.

rvalyi (approved):
LGTM. It would be better to attend the two comments by @fcayre

partner_category_unique_per_organization/models/partner.py
PierrickBrun: Thanks for anwering @clementmbr

PierrickBrun:
Changes done

PierrickBrun:
pre-commit failed seems unrelated (it runs well on my end)

PierrickBrun:
Is there a way to re-run the tests ?

After the rebase pre-commit is OK but now a test failed because of a query limit on the VIES API (MS_MAX_CONCURRENT_REQ)

partner_category_unique_per_organization/__manifest__.py
```diff
@@ -0,0 +1,18 @@
...
+    "license": "AGPL-3",
+    "application": False,
+    "installable": True,
+    "depends": [
+        "base",
+    ],
+    "data": [
+        "views/res_partner_views.xml",
+    ],
+    "demo": [],
```
NL66278: Should leave out empty keys.

NL66278 (approved):
:+1: LGTM. One nag, that should be nice to correct, but not blocking.

PierrickBrun:
> +1 LGTM. One nag, that should be nice to correct, but not blocking.

Thanks for the review

I don't really have an opinion on the empty keys in the manifest but maybe this should be added to the OCA pre-commit ?

clementmbr (approved):

klodr (approved):

PR #1472, opened 2023-02-22
[14.0][FIX] base_location: City/locations completion object display_name is not updated when changing the city name

eantones:
After changing a city name (adding an accent for instance), the display_name of res.city.zip is not updated and the address on the company form keeps displaying the old wrong value.

### DIFF
--- a/base_location/models/res_city_zip.py
+++ b/base_location/models/res_city_zip.py
@@ -37,7 +37,15 @@ class ResCityZip(models.Model):
         )
     ]
 
-    @api.depends("name", "city_id", "city_id.state_id", "city_id.country_id")
+    @api.depends(
+        "name",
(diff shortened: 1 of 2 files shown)


pedrobaeza (approved):

quannguyen1560:
should we update the display name when the state name and country name change as well ??

eantones:
> should we update the display name when the state name and country name change as well ??

Yes, we should! I just changed, thanks! :-)

pedrobaeza:
/ocabot merge patch

Please fw-port it to upper versions

eantones:
> /ocabot merge patch
> 
> Please fw-port it to upper versions

Done in v15!

In v16 is not necessary to do anything because the code is different and display_name is not computed

PR #1474, opened 2023-02-25
[16.0][ADD] partner_bank_acc_type_constraint

alexis-via:
This module aims at restoring the behavior of bank account types of Odoo 8.0 and previous versions : the user manually selects the bank account type and, if the checkum doesn't match the selected bank account type, Odoo raises a blocking error message.

With Odoo 9.0 and upper versions, when a user enters an invalid IBAN in Odoo, he often don't realize that he has entered an invalid IBAN because he doesn't have any explicit error message.

With this module, when a user enters an invalid IBAN in Odoo, he will get an explicit error message that will prevent him from saving the account number.

partner_bank_acc_type_raise


alexis-via:
I wanted to add tests, but as this module doesn't depend on base_iban (or another module that adds an acc_type with a checksum), I don't see how to add interesting tests.

PaulGoubert (commented):
Useful to avoid stupid typo mistake. Thank you @alexis-via

alexis-via:
@sebastienbeau here is the module I told you about

alexis-via:
FYI, this module works without modification on odoo v14

mourad-ehm (approved):

houzefa-abba (approved):

PR #1482, opened 2023-03-07
[ADD] 14.0 partner_category_restrict module

PierrickBrun:
This module allows to restrict the usage of a category/tag on a partner via domains.


partner_category_restrict/models/partner.py
```diff
@@ -0,0 +1,31 @@
...
+
+class ResPartner(models.Model):
+    _inherit = "res.partner"
+
+    @api.constrains("category_id")
+    def _check_category_restrict_filter(self):
+        for record in self:
+            for categ in record.category_id.filtered("restrict_filter_id"):
+                restrict_domain = categ.restrict_filter_id.sudo()._get_eval_domain()
+                if not self.env["res.partner"].search_count(
```
CoraVanga, line 17:
```suggestion
                if not self.search_count(
```

PierrickBrun: I prefer using the explicit way tbh.

Also I've already had problems with calling this kind of function with self being a non empty recordset. The function used self as if it was empty and it created weird issues

clementmbr (approved):
Perfect! Super concise. 
LGTM

CoraVanga (approved):

klodr (approved):

PR #1486, opened 2023-03-15
[15.0][FIX] base_location: City/locations completion object display_name is…

eantones:
… not updated when changing the city name

fw-port of #1472

ping @pedrobaeza

### DIFF
--- a/base_location/models/res_city_zip.py
+++ b/base_location/models/res_city_zip.py
@@ -37,7 +37,15 @@ class ResCityZip(models.Model):
         )
     ]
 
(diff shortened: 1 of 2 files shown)


pedrobaeza (approved):
Fast tracking this

/ocabot merge patch

And now it only rests v16

eantones:
> Fast tracking this
> 
> /ocabot merge patch
> 
> And now it only rests v16

Thanks!

Regarding v16 port, as I said in https://github.com/OCA/partner-contact/pull/1472#issuecomment-1470492591, it's not necessary to do anything because the code is different, display_name is not computed. I tested it and it already works.

PR #1496, opened 2023-04-02
[14.0] [IMP] partner_company_group: Change field position

nimarosa:
This PR changes the field position of the company_group_id to be after the name of the company.

I think that having the field between address fields and vat is a bit rare, i think the user would not expect to find it there. I propose to move it after the "name" when the parent_id field appears when the partner is not a company. So, if the partner is a company we show company_group there, and if not, original partner_id is shown. 

If you think we should move it to other place, please suggest it.

### DIFF
--- a/partner_company_group/views/contact_view.xml
+++ b/partner_company_group/views/contact_view.xml
@@ -5,10 +5,11 @@
         <field name="model">res.partner</field>
         <field name="inherit_id" ref="base.view_partner_form" />
         <field name="arch" type="xml">
-            <field name="vat" position="before">
+            <field name="parent_id" position="after">
                 <field
                     name="company_group_id"
                     attrs="{'invisible': [('is_company','=', False)]}"
+                    placeholder="Company Group"
                 />
             </field>
         </field>


nimarosa:
@pedrobaeza What do you think about this PR? I need some approvals. Do you think we can do it? It's a very simple customization but i think in terms of UI it's better to move this field there.

pedrobaeza:
I don't see why is better the new position, but one thing is clear: people using this module is already used to that position, so I don't think changing it is a good idea.

nimarosa:
The field in the original position is inside the address and vat group, before the vat, which for me don't make sense since it's not related to the fields inside that group in any way, and also might be confusing for a user creating a new partner. 

Since legacy parent_id field of odoo appears after the name of the partner (when the partner is a person) i think it makes sense to company_group field appear there when the partner is a company (since you can only link companies with companies). 

Maybe it's only my opinion but i think it makes more sense that having this field after the vat in a group of fields that are not really related to this. 

Let's see if anyone has any other opinions about this. Anyways I can make this change in a private module for my instances but it makes more sense to me to suggest it here first, to see if anyone thinks the same.

pedrobaeza:
What do you mean by "address group"? Any screenshot?

nimarosa:
@pedrobaeza Yes, i think it will be better with screenshots: 

**Original position:** 
Screenshot 2023-04-18 at 09 08 57

**My proposal:** 
Screenshot 2023-04-18 at 09 10 02

--- 

In screenshot 1 you see the company_group field is before "VAT", that is the original position of the field that for me it don't make sense concerning the UX. Groups of fields usually are related. In screenshot 2 is my proposal with this PR.

pedrobaeza:
Thanks for them. I reaffirm that the position is better as is. Where you want to put us already used for other things, like the parent company, and it's confusing.

nimarosa:
Yes but parent_id only appears when the partner is a "person" type and "company_group" appears only when the parent is "company" type, so for me was a good idea to have it in the same place. 

But no problem, I can make the changes in my instances but wanted to make the suggestion here in case anyone thought the same. Closing, thanks for your input, Pedro.

PR #1244, opened 2022-03-23
[15.0][ADD] base_partner_company_group

yankinmax:
The idea is to split `partner_company_group` to several modules depending on each of core modules. This is needed to avoid installing unwanted dependencies on the project, for example we need `partner_company_group`, but without additional installed `account`, `sale` or `crm`.
So, finally we want it to be:

1. `base_partner_company_group` -> https://github.com/OCA/partner-contact/pull/1244
2. `account_partner_company_group` depending on `base_partner_company_group` and `account` -> https://github.com/OCA/partner-contact/pull/1246
3. `crm_partner_company_group` depending on `base_partner_company_group` and `crm` -> https://github.com/OCA/partner-contact/pull/1245
4. `sale_partner_company_group` depending on `base_partner_company_group` and `sale` -> https://github.com/OCA/partner-contact/pull/1247

PR #1504, opened 2023-04-19
[16.0][MIG] base_partner_company_group

Highcooley:


sebalix (approved):

cyrilmanuel (approved):
Not an expert but LGTM

grindtildeath (approved):
@OCA/crm-sales-marketing-maintainers can we please merge? :pray:

rafaelbn:
It looks like we are missing all the commit history. Please could you review?

We should preserve the commit history like in https://github.com/OCA/partner-contact/pull/1244/commits

Isn't it?

Thank you!

rafaelbn (commented):
Please @Highcooley review: https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-16.0 to preserve history, here we have a long an important one!!! really needed

sebalix:
@rafaelbn it's because the module has been renamed, and the migration guide only retrieve the history of a given folder, not the previous one. I don't say we should not keep the previous history, but I think there is no ready-made solution in such case?

We could first retrieve the history of the original module name, then the new one (2 `git format-patch | git am` to run, and this needs to check a bit what were the previous module names).

And if a module has been split in multiple ones (I think it's the case here) it really starts to be complicated for a contributor to provide a clean history so we don't merge useless stuff from the previous module.

simahawk:
If the module was only renamed, you could rewrite the history (I did it in the past), BUT, what I see in the migration PR is:

* a new module has been added (w/o history)
* the old one was deleted (w/ its history)

I tend to say that, since we (c2c) are the authors of the module and there are no big changes in the past... we just don't care :wink:

base_partner_company_group/__manifest__.py
```diff
@@ -0,0 +1,16 @@
+# Copyright 2019 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+{
+    "name": "Base Partner Company Group",
+    "summary": "Adds the possibility to add a company group to a company",
+    "version": "16.0.1.0.0",
+    "category": "Sales",
+    "author": "Camptocamp SA, Odoo Community Association (OCA)",
```
simahawk:
```suggestion
    "author": "Camptocamp, Odoo Community Association (OCA)",
```
@Highcooley pls fix this :pray:

simahawk:
> If the module was only renamed, you could rewrite the history (I did it in the past), BUT, what I see in the migration PR is:
> 
>     * a new module has been added (w/o history)
> 
>     * the old one was deleted (w/ its history)
> 
> 
> I tend to say that, since we (c2c) are the authors of the module and there are no big changes in the past... we just don't care wink

Just one more point (mostly for our future selves): as I reported on #1244 `git move` should have been used. In fact, we should have blocked that PR, not this one. The authorship of some commits was not preserved. however, since there are no big changes from other authors and since the contributors added themselves to the contributors' list (I've just checked) it's kind of fine.

simahawk:
flappy tests due to GH issues https://www.githubstatus.com/

PR #1509, opened 2023-04-24
[16.0] [FIX] partner_identification partner form view: partner_id field on res.partner.id_number inside res.partner view

marielejeune:
`partner_id` field is currently visible in ID numbers form and tree view, even in the views of the o2m field id_numbers on res.partner. It is filled with active_id. 

The problem is that when creating a new partner from the interface and directly creating new ID numbers before saving (and thus creating the partner in DB), the active ID doesn't exist yet and the partner_id couldn't be filled. This is now necessary to save to create the partner before adding the ID numbers.

### DIFF
--- a/partner_identification/views/res_partner_view.xml
+++ b/partner_identification/views/res_partner_view.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8" ?>
 <odoo>
-    <!-- Modification of Partner - Adding Tab for Idenification Numbers -->
+    <!-- Modification of Partner - Adding Tab for Identification Numbers -->
     <record model="ir.ui.view" id="view_partner_form">
         <field name="name">res.partner.form.id_number</field>
         <field name="model">res.partner</field>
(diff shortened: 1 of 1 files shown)


partner_identification/views/res_partner_view.xml
```diff
@@ -14,8 +14,31 @@
...
+                            <field name="valid_until" />
+                            <field name="status" />
+                        </tree>
+                        <form>
+                            <group>
+                                <field name="category_id" />
+                                <field name="name" />
+                                <field name="partner_issued_id" />
+                                <field name="date_issued" />
+                                <field name="place_issuance" />
```
FrancoMaxime, line 33: Is it wanted that place_issuance is not listed in the tree part?

marielejeune: I took the same view as the one defined in `views/res_partner_id_number_view.xml`

FrancoMaxime (approved):
LGTM

PR #1513, opened 2023-05-01
[FIX] fixes #1389 Add display_name update upon name recalculation

### LINKED ISSUE #1389: [14.0] partner_firstname action_recalculate_partners_name does not update display_name
The action to recalculate partner names (action_recalculate_partners_name) is correctly updating name but not display_name on partners.

arpol-dev:


njeudy:
will test this fixes #1389

PR #1532, opened 2023-06-01
[14.0][ADD] partner_tag_smart_assignation

ecino:
New module for tagging partners automatically based on search filters or SQL queries.


ecino:
Anybody willing to review this new module? I think it brings a nice feature addition.

NL66278 (approved):
:+1: LGTM

NL66278:
@ecino please rebase and force push again to rerun tests

partner_tag_smart_assignation/cron/update_cron.xml
```diff
@@ -0,0 +1,14 @@
+<odoo>
```
NL66278, line 1: On cron data added this way it is best to use the noupdate flag, so changed to the cronfiguration (runtime, active or not, etc.) are not undone by an update.

partner_tag_smart_assignation/tests/test_smart_tagger.py
```diff
@@ -0,0 +1,161 @@
...
+    def test_check_validity_dates(self):
+        """
+        Test if the valid_until functionality works correctly
+        """
+
+        # Create a new tag with a 'valid_until' date set to yesterday
+        yesterday = fields.Date.to_string(fields.Date.today() - timedelta(days=1))
+
+        expired_tag = self.create_tag()
+        expired_tag.update({"valid_until": yesterday})
```
NL66278, line 138: You must use `write` here and then call the _check_validity_dates method manually, as the scheduler will not run immediately. `update`  only updates the in memory record.

partner_tag_smart_assignation/tests/test_smart_tagger.py
```diff
@@ -0,0 +1,161 @@
...
+        expired_tag.update_partner_tags()
+
+        # Reload the tags from the database
```
NL66278, line 141: This is not needed. Actually updating the database through write, will update the database (and the cache). invalidate_cache can be useful after changing stuff through SQL.

partner_tag_smart_assignation/models/hr_department.py
```diff
@@ -0,0 +1,9 @@
+# Copyright (C) 2019 Compassion CH (http://www.compassion.ch)
```
NL66278, line 1: Your module could be useful for many organisations that do not have `hr`  installed. So better not to have `hr`  as a dependency. Put the `hr` modification in a separate module (partner_tag_smart_assignation_hr), that depends on both this module and hr.

NL66278 (changes requested):
:+1: Thanks for your contribution. Please take up requested changed so we can approve this.

PR #1542, opened 2023-06-14
[15.0][FIX] partner_company_group: use rename_xmlids

ernestotejeda:
Cc @Tecnativa TT43676

The call to the `_update_xmlids` function of the `ir.model.data` model from the XML load data in `partner_company_group` module, causes duplication of views and duplication of their corresponding records in `ir.model.data`.

Main changes:

- Remove occurrences of `<function model="ir.model.data" name="_update_xmlids">` in `partner_company_group` module that were made to update the XML_IDs of the views.
- Add a call to `rename_xmlids` in the `pre_tinit_hook` of each of the other modules (base_partner_company_group, account_partner_company_group, crm_partner_company_group, sale_partner_company_group).


NL66278:
Why is this still draft? And please push again to run tests again.

PR #1544, opened 2023-06-21
[15.0][FIX] partner_company_default: multi company creation

AungKoKoLin1997:
Prior to this commit, the system would generate a partner before the creation of a new company and subsequently assign that partner to the company. As per `partner_company_default` module's behavior, the system would automatically assign the current company to the partner during creation, leading to a mismatch between the partner's company_id and its company.

In this Pull Request (PR), when a new company is created, the company_id of the associated partner will no longer be assigned during the partner's creation.
@qrtl

### DIFF
--- a/partner_company_default/models/res_company.py
+++ b/partner_company_default/models/res_company.py
@@ -1,0 +1,13 @@
+# Copyright 2023 Quartile Limited
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import api, models
+
+
+class ResCompany(models.Model):
+    _inherit = "res.company"
+
+    @api.model
+    def create(self, vals):
+        self = self.with_context(creating_from_company=True)
+        return super().create(vals)
--- a/partner_company_default/models/res_partner.py
+++ b/partner_company_default/models/res_partner.py
@@ -1,10 +1,23 @@
 # Copyright 2023 Quartile Limited
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 
-from odoo import fields, models
+from odoo import api, fields, models
+from odoo.tools import config
 
 
 class ResPartner(models.Model):
     _inherit = "res.partner"
 
-    company_id = fields.Many2one(default=lambda self: self.env.company)
+    company_id = fields.Many2one(default=lambda self: self._default_company_id())
+
+    @api.model
+    def _default_company_id(self):
+        """Return False for other tests or if creating a company."""
+        context = self.env.context
+        if (
+            context.get("creating_from_company")
+            or config["test_enable"]
+            and not context.get("test_partner_company_default")
+        ):
+            return False
+        return self.env.company
--- a/partner_company_default/tests/test_partner_company_default.py
+++ b/partner_company_default/tests/test_partner_company_default.py
@@ -1,0 +1,46 @@
+# Copyright 2023 Quartile Limited
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+import odoo.tests.common as common
+
+
+class TestPartnerCompanyDefault(common.TransactionCase):
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        cls.user = cls.env.ref("base.user_admin")
+
+    def test_partner_company_default(self):
+        # Check company of newly created partner
+        partner = (
+            self.env["res.partner"]
+            .with_user(self.user.id)
+            .with_context(test_partner_company_default=True)
+            .create({"name": "Test Partner 1"})
+        )
+        self.assertEqual(partner.company_id, self.user.company_id)
+
+        # Check company of the partner of newly created company
+        company_fr = (
+            self.env["res.company"]
+            .with_user(self.user.id)
+            .create(
+                {
+                    "name": "French company",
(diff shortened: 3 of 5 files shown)


AungKoKoLin1997:
One tests from base_location_nuts is failed. I think this should be handled by another PR and I will let this PR alone.

partner_company_default/models/res_company.py
```diff
@@ -0,0 +1,16 @@
...
+
+class ResCompany(models.Model):
+    _inherit = "res.company"
+
+    @api.model
+    def create(self, vals):
+        self = self.with_context(creating_from_company=True)
+        company = super(ResCompany, self).create(vals)
+        # Add partner's company_id to new company
+        company.partner_id.sudo().write({"company_id": company.id})
```
pedrobaeza: This is a potential problem. Companies' partner should have company_id = False.

AungKoKoLin1997: Yes. I agree.

partner_company_default/models/res_company.py
```diff
@@ -0,0 +1,14 @@
...
+
+
+class ResCompany(models.Model):
+    _inherit = "res.company"
+
+    @api.model
+    def create(self, vals):
+        self = self.with_context(creating_from_company=True)
+        company = super(ResCompany, self).create(vals)
+        return company
```
yostashiro:
```suggestion
        return super().create(vals)
```

partner_company_default/tests/test_partner_company_default.py
```diff
@@ -0,0 +1,60 @@
...
+                        6,
+                        0,
+                        [
+                            self.env.ref("base.group_system").id,
+                            self.env.ref("base.group_partner_manager").id,
+                        ],
+                    )
+                ],
+            }
+        )
```
yostashiro: As talked, let's use the admin user from the demo data.

partner_company_default/tests/test_partner_company_default.py
```diff
@@ -0,0 +1,44 @@
...
+    def test_partner_company_default(self):
+        # Test created partner company
+        partner = (
+            self.env["res.partner"]
+            .with_user(self.user.id)
+            .create({"name": "Test Partner 1"})
+        )
+        self.assertEqual(partner.company_id, self.user.company_id)
+
+        # Test partners of multi company
```
yostashiro:
```suggestion
        # Check company of the partner of newly created company
```

partner_company_default/tests/test_partner_company_default.py
```diff
@@ -0,0 +1,44 @@
...
+                {
+                    "name": "French company",
+                    "currency_id": self.env.ref("base.EUR").id,
+                    "country_id": self.env.ref("base.fr").id,
+                }
+            )
+        )
+        self.assertFalse(company_fr.partner_id.company_id)
+
+        self.user.company_ids = [(4, company_fr.id)]
```
yostashiro, line 38:
```suggestion
        # Switch user's company and create a partner
        self.user.company_ids = [(4, company_fr.id)]
```

partner_company_default/tests/test_partner_company_default.py
```diff
@@ -0,0 +1,44 @@
...
+        self.user.company_id = company_fr.id
+        partner = (
+            self.env["res.partner"]
+            .with_user(self.user.id)
+            .create({"name": "Test Partner 2"})
+        )
+
```
yostashiro:
```suggestion
```

partner_company_default/tests/test_partner_company_default.py
```diff
@@ -0,0 +1,44 @@
...
+
+class TestPartnerCompanyDefault(common.TransactionCase):
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        cls.user = cls.env.ref("base.user_admin")
+
+    def test_partner_company_default(self):
+        # Test created partner company
```
yostashiro:
```suggestion
        # Check company of newly created partner
```

yostashiro (approved):
LGTM. Thanks for the updates. 👍

NL66278 (approved):
:+1: LGTM

partner_company_default/models/res_partner.py
```diff
@@ -1,10 +1,17 @@
...
     _inherit = "res.partner"
 
     company_id = fields.Many2one(default=lambda self: self.env.company)
+
+    @api.model
+    def create(self, vals):
+        # The context value is set in the create method of res.company
+        if self.env.context.get("creating_from_company"):
+            vals["company_id"] = False
+        return super(ResPartner, self).create(vals)
```
pedrobaeza: Previous code is not multi-record creation aware, but I have tried and it's not always correct depending on when the create override in the stack is executed, so better to put company_id=False from the beginning this way:

```suggestion
    company_id = fields.Many2one(default=lambda self: self._default_company_id())

    @api.model
    def _default_company_id(self):
        """Return False for other tests or if creating a company."""
        context = self.env.context
        if (
            context.get("creating_from_company")
            or config["test_enable"]
            and not context.get("test_partner_company_default")
        ):
            return False
        return self.env.company
```

and on tests, pass the corresponding context.

pedrobaeza (changes requested):

pedrobaeza (approved):
/ocabot merge patch

PR #1546, opened 2023-06-22
[16.0] Replace partner_address_two_lines by partner_display_name_line_break

JuMiSanAr:
Fix the separation in two lines, and add tests.

Depends on: https://github.com/odoo/odoo/pull/126451


partner_address_two_lines/models/res_partner.py
```diff
@@ -9,8 +9,10 @@ class ResPartner(models.Model):
...
-            return "{}\n {}".format(
-                partner.commercial_company_name or partner.sudo().parent_id.name, name
-            )
-        else:
-            return super()._get_contact_name(partner, name)
+            company_name = partner.commercial_company_name
+            partner_name = partner.name
+            if company_name and partner_name:
+                # Only display two lines if both values are found,
+                # otherwise revert to the standard behavior.
```
grindtildeath: AFAIU if there's no name on the contact the "standard behaviour" will still display it as:
> Company name, address name

instead of 

> Company name
> address name

Don't we want do display only "Company name" on a single line here?

JuMiSanAr: If partner.name is not found, only the company name / name of the parent will be displayed:
```    
def _get_contact_name(self, partner, name):
        return "%s, %s" % (partner.commercial_company_name or partner.sudo().parent_id.name, name)
```

Where parameter `name` corresponds to partner.name, which will be empty (the comma is stripped afterwards).

So basically, if we find both a company name and a partner name, we display it in two rows. If there's no partner name, then we keep the one-liner as it is now.

If you can come up with a more elegant solution, I'll be glad to test it!

grindtildeath: It might be ok, but please add a test to ensure it is the correct behaviour. :pray:

JuMiSanAr: Of course, tests added.

partner_address_two_lines/models/res_partner.py
```diff
@@ -9,8 +9,10 @@ class ResPartner(models.Model):
 
     def _get_contact_name(self, partner, name):
         if self.env.context.get("_two_lines_partner_address"):
-            return "{}\n {}".format(
-                partner.commercial_company_name or partner.sudo().parent_id.name, name
-            )
-        else:
-            return super()._get_contact_name(partner, name)
+            company_name = partner.commercial_company_name
```
grindtildeath:
```suggestion
            company_name = partner.commercial_company_name or partner.sudo().parent_id.name
```

partner_address_two_lines/__manifest__.py
```diff
@@ -8,7 +8,7 @@
     "license": "AGPL-3",
     "category": "Partner",
     "website": "https://github.com/OCA/partner-contact",
-    "depends": ["base"],
+    "depends": ["base", "account"],
```
grindtildeath: Please no, there's no reason to add `account` as dependency here.

partner_address_two_lines/models/res_partner.py
grindtildeath: I'm wondering, can't we strip the `\n` afterwards as it's done with the `,`?

grindtildeath (changes requested):
For the test, please focus on writing a "unit" test by calling `name_get` on different partners structures. We just want to make sure we get a `\n` instead of a `,` and that we don't get the separator and the contact type in case there's no contact name.

partner_address_two_lines/models/res_partner.py
```diff
@@ -9,8 +9,11 @@ class ResPartner(models.Model):
...
         if self.env.context.get("_two_lines_partner_address"):
-            return "{}\n {}".format(
-                partner.commercial_company_name or partner.sudo().parent_id.name, name
+            company_name = (
+                partner.commercial_company_name or partner.sudo().parent_id.name
             )
-        else:
-            return super()._get_contact_name(partner, name)
+            partner_name = partner.name
+            if company_name and partner_name:
```
grindtildeath:
```suggestion
            if company_name and name:
```

partner_address_two_lines/models/res_partner.py
```diff
@@ -9,8 +9,12 @@ class ResPartner(models.Model):
...
+            company_name = (
+                partner.commercial_company_name or partner.sudo().parent_id.name
             )
-        else:
-            return super()._get_contact_name(partner, name)
+            # Depends on https://github.com/odoo/odoo/pull/126451
+            # This change can only be merged after this Odoo PR is merged as well.
+            if company_name and name:
+                # Only display two lines if both values are found.
+                return "{}\n {}".format(company_name, name)
```
mmequignon:
```suggestion
                return "{}\n {}".format(company_name, name)
```
Why this extra space before the name?

mmequignon (approved):
would be nice to have a screenshot or smth…
LG tough.

ajaniszewska-dev (approved):

mmequignon (approved):

rafaelbn (commented):
We have to change the module name please! 🙏🏼 😄 ❤️ 

It's really confusing :-(

The module make this:

imagen

rafaelbn:
@rousseldenis which name in your head?

partner_address_two_lines/tests/test_res_partner.py
```diff
@@ -0,0 +1,40 @@
...
+                "parent_id": partner.id,
+            }
+        )
+
+    def test_get_name(self):
+        # Partner with name.
+        name = self.child_partner_name.with_context(
+            _two_lines_partner_address=True
+        )._get_name()
+        self.assertEqual(name, "Test Company Name\n Test Partner Name")
```
NL66278: Leave out the blank after ` \n` . Same goes for the invoice address.

The improved code does not add a space at the beginning of the second line, which in my opinion is correct. The old code did however.

trisdoan:
Hi @rafaelbn, do you still want to rename the module

I am migrating it to 17.0. I can rename it as a part of the migration

rafaelbn:
@NL66278 @JuMiSanAr is this still need fixing PR?

Any idea with module name @hbrunn ?

partner_display_name_break_line ?

Thank you! 😄 

Remember @trisdoan , this PR depends in a Odoo/odoo PR too

NL66278:
@rafaelbn My comment on how to fix the failing test has not been taken up after half a year. The prerequisite Odoo PR has a merge conflict that is also not resolved. If nothing is done on these issues I would propose closing the PR.

JuMiSanAr:
Test updated. Regarding the module name, I'm open to suggestions.

rafaelbn:
@NL66278 right! Test are fixed now! 🥳

NL66278:
@JuMiSanAr I have two more request for approval and merge:
- Squash your commits, so we are left with one commit;
- Rename the module to partner_display_name_line_break;

Nearly the same as the module name proposed by @rafaelbn , but I think "line break" is the recognized term for this.

rafaelbn:
I agree

JuMiSanAr:
> @JuMiSanAr I have two more request for approval and merge:
> 
> * Squash your commits, so we are left with one commit;
> * Rename the module to partner_display_name_line_break;
> 
> Nearly the same as the module name proposed by @rafaelbn , but I think "line break" is the recognized term for this.

There you go :relaxed:

NL66278 (approved):
:+1: LGTM. Thanks for your contribution. Two questions, not breaking::
- Why the dependency on context key? If this module is installed, apparently the instance administrators want to display the name and the company on two lines.
- Does this really depend on the Odoo PR? If the Odoo PR is not accepted or merged, you will also get possibly the word 'Invoice' or one of the other type descriptions on a separate line. To me this seems to be consistent with the partner name on a second line.

rafaelbn (approved):
Thank you! 👍🏼 


> [NL66278](https://github.com/NL66278) approved these changes
> 
> 👍 LGTM. Thanks for your contribution. Two questions, not breaking::
> 
> - Why the dependency on context key? If this module is installed, apparently the instance administrators want to display the name and the company on two lines.
> - Does this really depend on the Odoo PR? If the Odoo PR is not accepted or merged, you will also get possibly the word 'Invoice' or one of the other type descriptions on a separate line. To me this seems to be consistent with the partner name on a second line.

trisdoan (approved):

trisdoan:
Hello @@JuMiSanAr, LGTM.

I agree with NL66278 that it does not need to depend on that PR

hbrunn:
please also add a PR to https://github.com/OCA/OpenUpgrade/blob/16.0/openupgrade_scripts/apriori.py#L6 to document the renaming for openupgrade.

/ocabot merge minor

PR #1550, opened 2023-06-27
[16.0][MIG] partner_bank_code

hbrunn:
I took the liberty to add overrides for name_search, name_get using those fields here


partner_bank_code/models/res_bank.py
```diff
@@ -0,0 +1,75 @@
...
+                            if self.browse(_id).bank_branch_code
+                            else "",
+                        )
+                    )
+                    if self.browse(_id).bank_code
+                    else ""
+                ),
+            )
+            for _id, name in result
+        ]
```
NL66278, line 44: I just wondered whether the repeated browse could be prevented with a little helper function, something along the lines off:
```
def make_name(id, name):
    bank = self.browse(id)
    return " ".join(
        [
            name,
            "/".join(
                [
                    part
                    for part in [bank.bank_code, bank.bank_branch_code]
                    if part
                ]
            )
        ]
    )
result = super().name_get()
return [(id, make_name(id, name)) for id, name in result]
```

NL66278 (approved):
:+1: LGTM Nice addition, one suggestion/question.

partner_bank_code/models/res_bank.py
hbrunn: we could, but nowadays this just looks at the same cache, so that's basically syntactic sugar

NL66278: It was  a suggestion. I might be old fashioned in not wanting to retrieve the same data repeatedly, even though I am aware of the caching. But I approved the PR anyway already.

rafaelbn (approved):
/ocabot merge patch

PR #1554, opened 2023-07-15
[16.0][MIG] partner_company_group

Highcooley:
This is a legacy port of partner_company_group to migrate projects hosted on SH. Use base_partner_company_group for a fresh install or if you have better migration tools on hand.


pedrobaeza (changes requested):
This module has been split in parts, so even on Odoo.sh, you have to transition to the new modules. Use openupgradelib with `update_module_names` method or any other SQL thing to get the module renaming.

PR #1565, opened 2023-08-22
tr_TR.po file updated

yemrekocamese:
tr_TR.po file updated


pedrobaeza:
Please use Weblate for that:

https://odoo-community.org/page/translate

PR #1566, opened 2023-08-22
Translation

sel1mErdogan:
tr_TR.po translation


pedrobaeza:
Please use Weblate for that:

https://odoo-community.org/page/translate

PR #1577, opened 2023-10-06
[16.0][ADD] partner_lastname_uppercase

robinkeunen:
This module uppercases last names on create and write.

A server action (Uppercase all lastnames) is provided to initialize the data.
To launch it :

- activate debug mode
- find the action in Technical > Server Actions
- click "add contextual action"
- on any partner, from the action menu, run "Uppercase all lastnames"
- return to the server action
- click "remove the contextual action"


partner_lastname_uppercase/__manifest__.py
```diff
@@ -0,0 +1,19 @@
+# Copyright 2023 Coop IT Easy SC
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+{
+    "name": "Partner last name uppercase",
+    "summary": "Uppercases the the last names of partners",
+    "version": "16.0.1.0.1",
+    "author": "Coop IT Easy SC, " "Odoo Community Association (OCA)",
```
huguesdk:
```suggestion
    "author": "Coop IT Easy SC, Odoo Community Association (OCA)",
```

partner_lastname_uppercase/data/actions.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <record id="res_partner_uppercase_lastname_action" model="ir.actions.server">
```
huguesdk: shouldn’t `uppercase` and `lastname` follow the same order as in the module name?

partner_lastname_uppercase/readme/DESCRIPTION.rst
huguesdk, line 1: i think it should be more obvious that this causes data loss.

partner_lastname_uppercase/models/res_partner.py
huguesdk, line 1: the upper-casing is done in 3 different places. it should be done in only one place, in a function used by the other functions.

partner_lastname_uppercase/models/res_partner.py
```diff
@@ -0,0 +1,35 @@
...
+    @api.model
+    def uppercase_all_lastnames(self):
+        individuals = self.search(
+            [
+                ("is_company", "=", False),
+                ("lastname", "!=", False),
+            ]
+        )
+        for partner in individuals:
+            partner.lastname = partner.lastname.upper()
```
huguesdk, line 29: note that `str.upper()` [uses the default unicode case folding](https://docs.python.org/3/library/stdtypes.html#str.upper). this works for many languages, but not for all, like turkish where ı/I i/İ pairs work differently (see [cpython issue 17252](https://bugs.python.org/issue17252)). this limitation should be explained in the readme.

partner_lastname_uppercase/models/res_partner.py
```diff
@@ -0,0 +1,35 @@
...
+            if lastname and not is_company:
+                vals["lastname"] = lastname.upper()
+        return super().create(vals_list)
+
+    def write(self, vals):
+        res = super().write(vals)
+        individuals = self.filtered(lambda rp: rp.lastname and not rp.is_company)
+        for partner in individuals:
+            super(ResPartner, partner).write({"lastname": partner.lastname.upper()})
+        return res
```
huguesdk: this upper-cases the last name at each write of a partner instead of only when the last name is changed. there could be common code between `create()` and `write()` in handling `vals` (maybe not that feasible because of the `is_company` test).

partner_lastname_uppercase/__manifest__.py
```diff
@@ -0,0 +1,19 @@
...
+    "license": "AGPL-3",
+    "maintainer": "Coop IT Easy SC",
+    "category": "Extra Tools",
+    "website": "https://github.com/OCA/partner-contact",
+    "depends": ["partner_firstname"],
+    "data": [
+        "data/actions.xml",
+    ],
+    "auto_install": False,
+    "installable": True,
```
huguesdk: default values should be omitted.

partner_lastname_uppercase/models/res_partner.py
carmenbianca: Indeed. `if "lastname" in vals` would be a good check to add.

I'm not certain it's valuable to spend time creating common code between `write` and `create`. The module's too small for there to be any benefits.

partner_lastname_uppercase/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,11 @@
...
+
+A server action (Uppercase all lastnames) is provided to initialize the data.
+To launch it :
+
+- activate debug mode
+- find the action in Technical > Server Actions
+- click "add contextual action"
+- on any partner, from the action menu, run "Uppercase all lastnames"
+- return to the server action
+- click "remove the contextual action"
```
carmenbianca: This should go to USAGE.rst.

carmenbianca (changes requested):

partner_lastname_uppercase/models/res_partner.py
robinkeunen: I tend to agree with Carmen, I started doing common code which pretty much reduced to calling `upper` so I left it as is.

robinkeunen:
@carmenbianca @huguesdk I fixed following your comments.

partner_lastname_uppercase/models/res_partner.py
```diff
@@ -0,0 +1,36 @@
...
+            if lastname and not is_company:
+                vals["lastname"] = lastname.upper()
+        return super().create(vals_list)
+
+    def write(self, vals):
+        res = super().write(vals)
+        if "lastname" in vals:
+            individuals = self.filtered(lambda rp: rp.lastname and not rp.is_company)
+            for partner in individuals:
+                super(ResPartner, partner).write({"lastname": partner.lastname.upper()})
```
huguesdk: at first glance, this way of doing seemed a little confusing to me (writing first then upper-casing). why not simply upper-casing the value in `vals`? i understand that it is because companies should be excluded, but a comment would be useful.

also, is it useful to test for `rp.lastname` while we know that we just wrote to it? oh ok, it is to allow to set it to `False`? a comment would be useful also. ☺️

huguesdk (approved):

carmenbianca (approved):

victor-champonnois (approved):
Functional test OK :)
-> set last name in uppercase when creating a contact : OK (tried on contact form, from a SO, or with a contact created with a cooperator subscription request)
-> server action : OK

robinkeunen:
Hi @pedrobaeza this module is ready to be merged, although I have a question : I updated the copier template to include the module in the rebel module list. It means we know have 11 checks running on each PR. Isn't that a bit much ? Should we remove the tests against Odoo and only keep OCB ?

pedrobaeza:
Why do you need to put it on rebel modules? The ideal is to not activate by default conflicting behaviors, or have some code testing if `config["test_enable"]` for inhibiting such behavior.

robinkeunen:
Thanks for the recommendation, I added a parameter and removed the module from the rebel module list.

Can some of you take a look again ? @victor-champonnois @huguesdk @carmenbianca

partner_lastname_uppercase/models/res_config_settings.py
```diff
@@ -0,0 +1,13 @@
...
+from odoo import fields, models
+
+
+class ResConfigSettings(models.TransientModel):
+    _inherit = "res.config.settings"
+
+    uppercase_lastname_activated = fields.Boolean(
+        default=False,
+        config_parameter="uppercase_lastname_activated",
+    )
```
huguesdk: i don’t like the name so much. i would prefer something like `convert_lastnames_to_uppercase`. look at the other boolean settings in odoo.

moreover, most prefix the `config_parameter` with the module name.

partner_lastname_uppercase/readme/CONFIGURATION.rst
huguesdk, line 1: i think the correct name for this file is `CONFIGURE.rst` (see [here](https://github.com/OCA/maintainer-tools/blob/master/tools/gen_addon_readme.py#L42)).

huguesdk (changes requested):

robinkeunen:
@huguesdk @pedrobaeza

pedrobaeza:
I personally don't like to put everything in upper. It's a very bad technique inherited from old systems where the search doesn't allow to insensitively search strings. Uppercase is a language tool for highlighting elements in the words. Putting everything in upper is losing this feature, take more space in screen, and making things less readable.

robinkeunen:
I understand your POV. The customer wants to make the difference more clear on the reports etc. I was expecting comments on the parameter I introduced based on your suggestion : what do you think about that ?

robinkeunen:
@huguesdk @rousseldenis can you take a look at this ?

robinkeunen:
@huguesdk I fixed and added comments based on your suggestion. Can you check ?

robinkeunen:
@huguesdk i need your review here please 🙏

robinkeunen:
@gurneyalex @rousseldenis i think this is ready to be merged.

huguesdk (approved):

robinkeunen:
@tafaRU @HaraldPanten  I think this is ready to be merged

robinkeunen:
@gurneyalex @rousseldenis i think this is ready to be merged.

NL66278 (approved):
:+1:  I would not recommend this to my customers, as indeed uppercase names IMHO are not to be recommended. But technically this looks OK and if customers want it, they can use it, and if not, they can just ignore the module.

PR #1580, opened 2023-10-10
[16.0][IMP] partner_manual_rank: hide base customer and supplier filters

JordiMForgeFlow:
The base filters added by Odoo create confusion with the new ones added by the OCA module

### DIFF
--- a/partner_manual_rank/views/res_partner.xml
+++ b/partner_manual_rank/views/res_partner.xml
@@ -5,7 +5,7 @@
         <field name="inherit_id" ref="base.view_res_partner_filter" />
         <field name="model">res.partner</field>
         <field type="xml" name="arch">
-            <xpath expr="//filter[@name='type_person']" position="before">
+            <xpath expr="//filter[@name='inactive']" position="before">
                 <filter
                     string="Customers"
                     name="customers"
@@ -17,6 +17,19 @@
                     domain="[('is_supplier', '=', True)]"
                 />
             </xpath>
+        </field>
+    </record>
+    <record id="view_res_partner_filter_account" model="ir.ui.view">
+        <field name="name">partner_manual_rank.view_partner_filter.account</field>
+        <field name="inherit_id" ref="account.res_partner_view_search" />
+        <field name="model">res.partner</field>
+        <field type="xml" name="arch">
+            <filter name="customer" position="attributes">
+                <attribute name="invisible">1</attribute>
+            </filter>
+            <filter name="supplier" position="attributes">
+                <attribute name="invisible">1</attribute>
+            </filter>
         </field>
     </record>
     <record id="view_partner_form" model="ir.ui.view">


AaronHForgeFlow (approved):

MiquelRForgeFlow (approved):

DavidJForgeFlow (approved):
LGTM!

partner_manual_rank/views/res_partner.xml
```diff
@@ -2,10 +2,16 @@
 <odoo>
     <record id="view_res_partner_filter" model="ir.ui.view">
         <field name="name">partner_manual_rank.view_partner_filter</field>
-        <field name="inherit_id" ref="base.view_res_partner_filter" />
+        <field name="inherit_id" ref="account.res_partner_view_search" />
```
luisg123v: Inherit the base view instead, just increase priority if necessary. Otherwise, this will produce a more complex inheritance tree. The default priority is 16, so a priority of 20 should be enough.

luisg123v (changes requested):
Missing commit description, description is only present in PR. Commit description is even even more important than PR description, as it cannot be modified once the change is merged.

partner_manual_rank/views/res_partner.xml
JordiMForgeFlow: Hi @luisg123v 

I am inheriting from account because the filters I am hiding are added in the account module.

luisg123v: Yes, I know. But you don't need to, just increase priority and it will work.

Take into account when you inherit a view that is not the base one, internally the base one is looked for anyway, the only difference is the order on which inheritances are applied, which is the same as changing priority.

It's recommended to always inherit the base view, to avoid inheriting the same view on different places (e.g. one for fields available in account and other ones available in product)

JordiMForgeFlow: Hi @luisg123v 

Sorry for the late reply, I was on holidays. I am keeping the base and only using the account one to hide the filters added in account :)

luisg123v: As I said, you don't need two inheritances, you just need to increase view's priority to something like 20 and it will work inheriting the base view.

MiquelRForgeFlow: > It's recommended to always inherit the base view, to avoid inheriting the same view on different places (e.g. one for fields available in account and other ones available in product)

Not always. @luisg123, I think you are being a bit picky here. We are not increasing complexity too much just by inheriting two views instead of one. We understand your logic and your approach. It's nice sometimes to implement this priority approach, but we can agree that it's not required. Furthermore, the priority approach is more error prone. Hypothetically, we may have an implementation that customize some view priorities due to some other inheriting views that change or move some part of the view (and that contain priority changes too), then your approach of putting here priority 20 could fail. So, in summary, we prefer to be overly cautious and edit the two views instead of adding the priority approach here. And I think you should tolerate it instead of blocking it, thanks.

luisg123v: Thanks for the detailed reply.

luisg123v (approved):
LGTM :+1: 

/ocabot merge patch

Issue #1578, opened 2023-10-09
[16.0] base_location: res.city.zip search


odooNextev:
## Module
base_location

## Describe the bug
Searching in `res.city.zip` field without use `search more` can be difficult in Odoo 16.0 because it doesn't allow you to filter based on more than one parameter at a time.
In older versions (for example 12.0), it was possible to search exactly the string you see in the dropdown (space and dot included) and get the results starts with characters in the input.

## To Reproduce
**Affected versions**:

- [ ] 16.0 https://github.com/OCA/partner-contact/pull/1581

1. Create or edit a contact
2. Search a city in `res.city.zip` field (with "Location completion" placeholder) by "zip code, city name"
![image](https://github.com/OCA/partner-contact/assets/93990941/751ab821-1b81-4389-b718-e416a9efa282)

3. You won't get results
![image](https://github.com/OCA/partner-contact/assets/93990941/8632f2de-3c23-49f8-b074-389de18b980c)

**Expected behavior**
Exactly as it was in 12.0 version.
![image](https://github.com/OCA/partner-contact/assets/93990941/e138a08f-a42e-4f8b-87fa-3b27ed9f8079)

pedrobaeza:
It should be a consequence of the several refactorings in the framework and the module.

Pull requests are accepted improving this.

PR #1581, opened 2023-10-10
[16.0] [FIX] base_location search by display_name as in older versions

odooNextev:
I'd propose this PR to fix this issue: https://github.com/OCA/partner-contact/issues/1578

I restored the old behaviour searching by `display_name` instead of separated fields.

### DIFF
--- a/base_location/migrations/16.0.1.0.1/post-migration.py
+++ b/base_location/migrations/16.0.1.0.1/post-migration.py
@@ -1,0 +1,10 @@
+# Copyright 2018 Tecnativa - Pedro M. Baeza
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    # Recompute display name for entries inserted by SQL
+    env["res.city.zip"].search([])._compute_display_name()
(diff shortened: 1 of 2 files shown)


odooNextev:
Hi @pedrobaeza,
what do you think about this PR?
It downgrades search and `display_name` computation to the older versions (I copied from 12.0).
Why were these changed in the newer versions?

PS. If you think it's fine, how can I trigger migration script to recompute `display_name` on module update/upgrade?
I noticed it doesn't be triggered when you update module.
Should I rename the version folder to "16.0.1.0.2"? I guess it will be the new module version if the PR will be merged.

pedrobaeza:
I don't think that change is sane for the DB size, as it requires to store the composed name for all the things, so I prefer to keep as it was and don't have the feature. One thing you can do is that you put this in an extra module, but I let also other's to give their opinion.

odooNextev:
@pedrobaeza I understand your doubts.
If there aren't supporters for this PR I'll make a private extra module.
I think the only solution to keep multiple `_rec_names_search` and search the keywords in the search input in all the fields in the list at the same time is changing the framework core.
The best should be writing "**Brussels 1000**" and get "**1000, Brussels, Belgium**".
This can be done separating the keywords by spaces and search every of these in `res.city.zip` name and zip.

What do you think?
Ain't anything similar already implemented?

pedrobaeza:
I don't know anything similar to that, sorry.

pedrobaeza:
Other contributor has put in my radar this module: https://github.com/OCA/server-tools/tree/16.0/base_name_search_improved

Is that what you intends?

odooNextev:
> Other contributor has put in my radar this module: https://github.com/OCA/server-tools/tree/16.0/base_name_search_improved
> 
> Is that what you intends?

Thank you so much Pedro, it can be very useful!
I'll try it now

PR #1583, opened 2023-10-17
[16.0][ADD] partner_identification_eori

JordiMForgeFlow:
Module to extend partner identification with new category for EORI numbers.

Also moving a helper method used already in partner_identification_gln to get duplicate numbers.


JordiMForgeFlow:
Hi @rousseldenis @lmignon @simahawk 

Could I get your input here? :)

lmignon (approved):
LGTM (Code review only)

partner_identification_eori/models/res_partner_id_category.py
```diff
@@ -0,0 +1,30 @@
...
+
+        if (
+            id_number.partner_id
+            and id_number.partner_id.country_id
+            and not id_number.name.startswith(id_number.partner_id.country_id.code)
+        ):
+            return True
+
+        cat = self.env.ref(
+            "partner_identification_eori." "partner_identification_eori_number_category"
```
simahawk:
```suggestion
            "partner_identification_eori.partner_identification_eori_number_category"
```

partner_identification/models/res_partner_id_category.py
```diff
@@ -40,6 +40,21 @@ def _get_default_color(self):
         "Python validation code", help="Python code called to validate an id number."
     )
 
+    @api.model
+    def _search_duplicate(self, category_id, id_number, force_active=False):
+        """Helper method to use in validation code in order to get an existing
+        duplicate for the specified number."""
```
simahawk:
```suggestion
        """Find duplicates for the given category and number."""
```

simahawk (changes requested):
LG. Just 2 remarks:
1. blocker: either you change the version of the old modules manually or you move out the changes to another PR because otherwise we won't get a version bump for them
2. nice to have: last 2 commits can be squashed and shortened by rewording like `partner_identification*: move _search_duplicate`

lmignon:
> blocker: either you change the version of the old modules manually or you move out the changes to another PR because otherwise we won't get a version bump for them

@simahawk  What do you mean? The bump command from the ocabot will bump the version for all the modified addons. Even the nobump will publish a new version on pypi as 16.0.X.Y.Z.W in place of 16.0.X.Y.Z AFAIK.

simahawk:
I mean that when you add a new module we normally merge w/ nobump which will leave the other modules' version unchanged. You get a new version on pypi but the module stays on the same version. Not the same game :wink: 

Now, it might sound nitpicking but I think is just a good practice to take of getting the right version for each module. 
It costs nothing...

JordiMForgeFlow:
Hi @simahawk 

I have applied the changes. I didn't thought about the version bump in this scenario, thank you for pointing it out! :)

simahawk (approved):
Cool, thanks for you understanding and collaboration! :raised_hands:

simahawk:
Worth backporting?

/ocabot merge nobump

PR #1586, opened 2023-10-20
[16.0][FIX] base_partner_sequence: consider written values in _needs_ref check

JordiMForgeFlow:
When evaluating the `_needs_ref` on a partner write, the method is not taking into consideration the values that are going to be written.

For example, imagine a scenario where you are converting a child partner into a parent partner by removing the parent_id. When the evaluation is done, it is done before the write, so the method considers that the partner is having a parent (although the write will remove it).

The fix is aimed at making sure the written values are considered in the evaluation. Due to having to modify the passed argument `vals` we need to make a copy in order to not mess up with the dict that is being passed from the write method.

### DIFF
--- a/base_partner_sequence/models/partner.py
+++ b/base_partner_sequence/models/partner.py
@@ -53,9 +53,17 @@ class ResPartner(models.Model):
                 _("Either field values or an id must be provided.")
             )
         # only assign a 'ref' to commercial partners
+        fields_for_check = ["is_company", "parent_id"]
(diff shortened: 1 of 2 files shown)


MiquelRForgeFlow (approved):

LoisRForgeFlow (approved):
LGTM, thanks! 

A test case would be great, but not blocking.

JordiMForgeFlow:
test case added cc @LoisRForgeFlow @MiquelRForgeFlow

PR #1597, opened 2023-11-06
[16.0][MIG] Migrate partner_coc

ddejong-therp:
When upgrading odoo from 15.0 to 16.0, there is a new field on `res.partner`: `company_registry`. This gets copied from the partner's company here: https://github.com/OCA/OpenUpgrade/blob/16.0/openupgrade_scripts/scripts/base/16.0.1.3/post-migration.py
However, it is basically a more basic implementation of what the `partner_coc` module did in previous versions. So there is no need for `partner_coc` in 16.0 anymore, but its data doesn't get properly migrated to the new field.

This module adds nothing, but is only here to migrate the company registry number.


pedrobaeza:
You don't need this dummy module if everything is in OCA (as it is) and you use OpenUpgrade. Just include the module in the apriori.py in merged modules, and then on `base` module for example (or wherever the new field is added, add a script renaming the column if found).

ddejong-therp:
@pedrobaeza 
Sorry for letting this MR go stale some years ago,
but this pseudo module was also there to be able to run some sql queries to migrate the field, because it is not just a renamed field or something.
What would you suggest is the best place to put these migration queries, if it is not to be put in a pseudo module?
Thank you.

pedrobaeza:
In OpenUpgrade itself, in `base` module, triggering them by the presence of a column or a table.

PR #1604, opened 2023-11-16
[16.0][IMP] partner_company_type: allow to set up countries and states, display shortcut in name

OriolMForgeFlow:


partner_company_type/models/res_partner.py
```diff
@@ -10,3 +11,33 @@ class ResPartner(models.Model):
...
     )
+
+    @api.constrains(
+        "country_id",
+        "state_id",
+        "partner_company_type_id",
+    )
+    def _check_partner_company_type_country_state(self):
+        for rec in self:
+            if not rec.partner_company_type_id:
```
JordiMForgeFlow, line 22:
```suggestion
            if not rec.partner_company_type_id or not rec.partner_company_type_id.country_ids or not rec.partner_company_type_id.state_ids:
```
with this change you can then remove the first check in the other two 'if' statements

JordiMForgeFlow (approved):
Code review LGTM

AaronHForgeFlow (approved):
Code review LG

LoisRForgeFlow (approved):
Code review :+1:

JordiMForgeFlow:
@rousseldenis could we get a review here? :)

partner_company_type/models/res_partner.py
```diff
@@ -10,3 +11,33 @@ class ResPartner(models.Model):
...
+                return
+            if (
+                rec.partner_company_type_id.country_ids
+                and rec.country_id not in rec.partner_company_type_id.country_ids
+            ):
+                raise UserError(
```
rousseldenis: You should raise `ValidationError`in a contraint.

rousseldenis:
@OriolMForgeFlow @len-foss Isn't it the same problematic ?

https://github.com/OCA/partner-contact/pull/1917

I would say I'd prefer an EXCLUDE

len-foss:
@rousseldenis 
#1917 is to relax the name check to allow the same name to be used for 2 different country (which I assume would speak the same language, which isn't so rare in the world), which isn't actually done here. 

Here there's a check to choose a company type to match the partner's country which is nice though.

Since it seems that different use-cases want for different constraint, it would make sense to have a configuration option to choose what is allowed. Therefore entirely drop the sql constaint for the python check.

partner_company_type/models/res_partner.py
JordiMForgeFlow: @rousseldenis corrected :)

JordiMForgeFlow:
@len-foss @rousseldenis could we get this one merged?

JordiMForgeFlow:
@rousseldenis could we move this one forward?

partner_company_type/models/res_partner.py
```diff
@@ -10,3 +11,33 @@ class ResPartner(models.Model):
...
+
+    @api.constrains(
+        "country_id",
+        "state_id",
+        "partner_company_type_id",
+    )
+    def _check_partner_company_type_country_state(self):
+        for rec in self:
+            if not rec.partner_company_type_id:
+                return
```
NL66278: Should this not be continue?

NL66278 (changes requested):
This are very extensive changes without any reflection in the README. To me it is completely unclear what problem this PR is supposed to solve.

JordiMForgeFlow:
@NL66278 you are right, README is now updated to reflect the purpose.

We don't have a problem with the module, we are simply proposing an improvement in order to manage the relation of Legal Forms with countries and their states. Legal Forms are very structured and predefined in most countries so it is a way to ensure and validate that the data is correct.

MiquelRForgeFlow (approved):

NL66278:
@JordiMForgeFlow It is becoming more clear now. Would it be possible to consistently use the vocabulary of "Legal Entity Type" instead of Legal Form. The word "Form" is confusing at it suggest some kind of document or form that must be filled.

JordiMForgeFlow:
@NL66278 done :)

partner_company_type/migrations/16.0.1.0.1/post-migration.py
```diff
@@ -0,0 +1,11 @@
+# Copyright 2023 ForgeFlow S.L. <http://www.forgeflow.com>
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
```
NL66278: The module itself has an AGPL license.

partner_company_type/views/res_partner.xml
```diff
@@ -12,6 +12,7 @@
                     name="partner_company_type_id"
                     options='{"no_open": True}'
                     attrs="{'invisible': [('is_company', '=', False)]}"
+                    domain="[ '|', ('country_ids', '=', False), ('country_ids', 'in', [country_id]), '|', ('state_ids', '=', False), ('state_ids', 'in', [state_id])]"
```
NL66278: This can actually be split over multiple lines, would be nicer.

NL66278:
Two last small remarks and we should be good to go.

JordiMForgeFlow:
@NL66278 thank you for the reviews, should be now ready :)

NL66278 (approved):
:+1: Thanks for your contribution! LGTM

PR #1611, opened 2023-11-20
[16.0][IMP] base_country_state_translatable: make state code also translatable

JordiMForgeFlow:


GuillemCForgeFlow (approved):
👍🏽

AaronHForgeFlow (approved):

pedrobaeza:
Please explain the rationale behind this

JordiMForgeFlow:
> Please explain the rationale behind this

Odoo loads chinese states using the chinese characters. We want to add a translation for them based on https://en.wikipedia.org/wiki/ISO_3166-2:CN

pedrobaeza:
But this provokes hard side effects: if a code is incorrect, next time it's changed, translations will have an incorrect value. This should be done on your custom, not here in the module.

JordiMForgeFlow:
> But this provokes hard side effects: if a code is incorrect, next time it's changed, translations will have an incorrect value. This should be done on your custom, not here in the module.

You are right, the side effect is not worth adding it in OCA. Thank you for the review @pedrobaeza :)

PR #1613, opened 2023-11-22
[16.0][IMP] base_partner_sequence: add setting to mark reference as readonly

JordiMForgeFlow:
New company setting to allow marking the reference as readonly. Useful if you want to rely 100% on the partner sequence, not allowing users to modify it.

### DIFF
--- a/base_partner_sequence/views/partner_view.xml
+++ b/base_partner_sequence/views/partner_view.xml
@@ -5,10 +5,13 @@
         <field name="model">res.partner</field>
         <field name="inherit_id" ref="base.view_partner_form" />
         <field name="arch" type="xml">
+            <field name="ref" position="before">
+                <field name="ref_readonly" invisible="1" />
+            </field>
             <field name="ref" position="attributes">
                 <attribute
                     name="attrs"
-                >{'readonly': [('is_company', '=', False), ('parent_id', '!=', False)]}</attribute>
+                >{'readonly': ['|', ('ref_readonly', '=', True), '&amp;', ('is_company', '=', False), ('parent_id', '!=', False)]}</attribute>
             </field>
         </field>
     </record>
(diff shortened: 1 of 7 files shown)


aliciagaarzo (approved):
LGTM!

amarcosg (approved):

dalonsofl (approved):
LGTM!

amarcosg:
Hi @NL66278, @JordiMForgeFlow proposes some changes that we have validated and find very useful. It is possible to merge them? 

Thanks!

base_partner_sequence/views/partner_view.xml
```diff
@@ -5,10 +5,13 @@
...
         <field name="inherit_id" ref="base.view_partner_form" />
         <field name="arch" type="xml">
+            <field name="ref" position="before">
+                <field name="ref_readonly" invisible="1" />
+            </field>
             <field name="ref" position="attributes">
                 <attribute
                     name="attrs"
-                >{'readonly': [('is_company', '=', False), ('parent_id', '!=', False)]}</attribute>
+                >{'readonly': ['|', ('ref_readonly', '=', True), '&amp;', ('is_company', '=', False), ('parent_id', '!=', False)]}</attribute>
```
NL66278: Actually there are now two distinct places that determine whether `ref` is readonly. Even if `ref_readonly`  is set to False in the compute method, there is still a possibility that `ref` can not be updated, that is, when the partner is a person that belongs to a company.

I think it is better to have the logic for readonly in only one place, which should be the compute method. Then the readonly condition in the view can be simply `[('ref_readonly', '=', True)]`.

This also would have the advantage that if some customization is needed, the compute method can be easily inherited and extended  in custom modules.

NL66278 (changes requested):
Thank you for your Pull Request. I think it is a very useful addition, but I have one suggestion that I hope you will take up.

base_partner_sequence/views/partner_view.xml
JordiMForgeFlow: @NL66278 totally agree! I have added the changes, thank you :)

NL66278 (approved):
:+1: LGTM

PR #1628, opened 2023-12-13
[16.0] partner_industry_secondary: fix name_get w/o name

simahawk:
Surprisingly you can have industry records w/o a name. The field is not required in odoo core so...
if by chance you get an empty name, name_get is broken w/o this fix.

### DIFF
--- a/partner_industry_secondary/models/res_partner_industry.py
+++ b/partner_industry_secondary/models/res_partner_industry.py
@@ -28,8 +28,7 @@ class ResPartnerIndustry(models.Model):
             """Return the list [cat.name, cat.parent_id.name, ...]"""
             res = []
             while cat:
-                res.insert(0, cat.name)
-                cat = cat.parent_id
+                res.insert(0, cat.name or f"#{cat.id}")
             return res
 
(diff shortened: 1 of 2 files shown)


santostelmo:
@simahawk that will generate a infinite loop.

simahawk:
> @simahawk that will generate a infinite loop.

sorry, deleted 1 line by errror

santostelmo (approved):

simahawk:
hmmm it might be that something has changed upstream because now there's a constraint on the name
```
2024-01-23 15:05:17,670 427 ERROR odoo odoo.sql_db: bad query: UPDATE "res_partner_industry" SET "name" = NULL, "write_date" = '2024-01-23T15:05:17.600787'::timestamp, "write_uid" = 1 WHERE id IN (23)
ERROR: null value in column "name" violates not-null constraint
DETAIL:  Failing row contains (23, 1, 1, null, null, t, 2024-01-23 15:05:17.600787, 2024-01-23 15:05:17.600787, 22/23/, 22).
```

PR #1629, opened 2023-12-18
[17.0][MIG] partner_stage: Migration to 17.0

fasilwdr:
Standard migration to 17.0


dreispt (changes requested):
Commit history is missing.
Can you check the migration instructions?

PR #1639, opened 2024-01-09
[16.0][FIX] partner_company_default

AungKoKoLin1997:
Before this PR, multi-record creation is not awared and not ways correct depending on when the create override in the stack is executed.

With current PR, this will be solved.

@qrtl


pedrobaeza (approved):
Patch already merged in 15.0 according my suggestion in https://github.com/OCA/partner-contact/pull/1544#pullrequestreview-1796865084, so fast-tracking:

/ocabot merge patch

PR #1643, opened 2024-01-13
[IMP] partner_email_check: option to ignore Users emails

dreispt:


partner_email_check/models/res_partner.py
```diff
@@ -84,7 +83,11 @@ def _should_check_syntax(self):
         return self.env.company.partner_email_check_syntax
 
     def _should_filter_duplicates(self):
-        return self.env.company.partner_email_check_filter_duplicates
+        check_duplicates = self.env.company.partner_email_check_filter_duplicates
+        ignore_users = self.env.company.partner_email_duplicates_ignore_users
+        # The Users Window Action has some context values to know we are there
+        on_user_form = "search_default_filter_no_share" in self.env.context
```
qgroulard, line 89: I don't think this is a good idea to be based on the key `search_default_filter_no_share`.
This key has nothing to do with this feature, we don't not how it will evolve and it makes the code difficult to understand.

IMO we should work with our own key or add this check to the search domain when we are looking for duplicates.

dreispt: It is explicitly explained in a comment, so it shouldn't be difficult to understand, of change in the future in case of need.
I hesitate in adding unneeded complexity to the change.

PR #1126, opened 2021-08-04
[14.0][REF] partner_firstname: Support duplicate multiple names

fernandahf:
Description of the issue/feature this PR addresses:

Considering following name:

firstname = Francisco Javier
lastname = Garcia
second_lastname = Cabeza de Vaca

When duplicate:

firstname = de Vaca Francisco Javier (copy)
lastname = Garcia
second_lastname = Cabeza

It is inverted based on the 'name' and changed the order of the fields

But when we are duplicating a contact we don't need to recompute the names since that it should be duplicated as it

Video:

https://youtu.be/CEFik5COK18

Current behavior before PR:

Duplicate partner with multiple last name or second last name, name is inverted based on the 'name'

Desired behavior after PR is merged:

Duplicate partner with multiple last name or second last name, name is not inverted.

-- I confirm I have signed the CLA and read the PR guidelines at www.odoo.com/submit-pr

PR #1647, opened 2024-01-17
[16.0][FIX] partner_firstname,partner_second_lastname: Avoid name fields rewrite

payen000:
Previously, creating a non-company partner by passing the `name` field
would trigger the `_inverse_name_after_cleaning_whitespace` which, for
values of `firstname`, `lastname` and `lastname2` with multiple words,
would result in most of those words being set to the `lastname2`, even if the
correct values were explicitly passed to the `create` method. For example,
duplicating a non-company partner with the following field values:
    
firstname = Francisco Javier
lastname = Garcia
lastname2 = Cabeza de Vaca
    
Would result in a partner with these values:
    
firstname = Francisco
lastname = Javier
lastname2 = Garcia Cabeza de Vaca (copy)

because the full name (Francisco Javier Garcia Cabeza de Vaca) is passed to the `create` method, and the `_inverse_name_after_cleaning_whitespace` method is triggered afterwards.
    
This PR solves the issue and also prevents the `_inverse_name_after_cleaning_whitespace`
method from triggering when recomputing names from settings, which causes the same
effect described above.

This PR is a forward port of the following 14.0 PRs:

* https://github.com/OCA/partner-contact/pull/1126
* https://github.com/OCA/partner-contact/pull/1276

### DIFF
--- a/partner_firstname/models/base_config_settings.py
+++ b/partner_firstname/models/base_config_settings.py
@@ -61,7 +61,10 @@ class ResConfigSettings(models.TransientModel):
         )
         partners = self._partners_for_recalculating()
         _logger.info("Recalculating names for %d partners.", len(partners))
-        partners._compute_name()
+        # Use add_to_compute instead of _compute_name to avoid triggering
+        # _inverse_name_after_cleaning_whitespace, which can
+        # modify a partner's firstname, lastname and lastname2
+        self.env.add_to_compute(self.env["res.partner"]._fields["name"], partners)
         self.partner_names_order_changed = False
         self.execute()
         _logger.info("%d partners updated.", len(partners))
--- a/partner_firstname/models/res_partner.py
+++ b/partner_firstname/models/res_partner.py
@@ -99,8 +99,7 @@ class ResPartner(models.Model):
         ignored in :meth:`~.create` because it also copies explicitly firstname
         and lastname fields.
         """
-        if default is None:
-            default = {}
+        default = default or {}
         if not self.is_company:
             order = self._get_names_order()
             extra_default_values = self.get_extra_default_copy_values(order)
--- a/partner_firstname/models/res_partner.py
+++ b/partner_firstname/models/res_partner.py
@@ -4,7 +4,7 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 import logging
 
-from odoo import api, fields, models
+from odoo import _, api, fields, models
 
 from .. import exceptions
 
@@ -26,32 +26,71 @@ class ResPartner(models.Model):
         readonly=False,
     )
 
+    @api.model
+    def name_fields_in_vals(self, vals):
+        """Method to check if any name fields are in `vals`."""
+        return vals.get("firstname") or vals.get("lastname")
+
     @api.model_create_multi
     def create(self, vals_list):
-        """Add inverted names at creation if unavailable."""
-        context = dict(self.env.context)
+        """Add inverted names at creation if unavailable. Also, remove the full name
+        from `vals` and context if the partner is an individual and is being created
+        with any name fields, as the name must be computed from the provided name parts;
+        otherwise, the name fields will be computed from the `name` again, when calling
+        its inverse method.
+
+        Note that, to avoid deleting the 'default_name' context for all partners when it's
+        not appropriate, we must call `create` for each partner individually with the correct
+        context.
+        """
+        created_partners = self.browse()
         for vals in vals_list:
-            name = vals.get("name", context.get("default_name"))
+            partner_context = dict(self.env.context)
+            if (
+                not vals.get("is_company")
+                and self.name_fields_in_vals(vals)
+                and "name" in vals
+            ):
+                del vals["name"]
+                partner_context.pop("default_name", None)
+            else:
+                name = vals.get("name", partner_context.get("default_name"))
+                if name is not None:
+                    # Calculate the split fields
+                    inverted = self._get_inverse_name(
+                        self._get_whitespace_cleaned_name(name),
+                        vals.get(
+                            "is_company", self.default_get(["is_company"])["is_company"]
+                        ),
+                    )
+                    for key, value in inverted.items():
+                        if not vals.get(key) or partner_context.get("copy"):
+                            vals[key] = value
 
-            if name is not None:
-                # Calculate the splitted fields
-                inverted = self._get_inverse_name(
-                    self._get_whitespace_cleaned_name(name),
-                    vals.get(
-                        "is_company", self.default_get(["is_company"])["is_company"]
-                    ),
-                )
-                for key, value in inverted.items():
-                    if not vals.get(key) or context.get("copy"):
-                        vals[key] = value
+                    # Remove the combined fields
+                    vals.pop("name", None)
+                    partner_context.pop("default_name", None)
+            # pylint: disable=W8121
+            created_partners |= super(
+                ResPartner, self.with_context(partner_context)
+            ).create(vals_list)
+        return created_partners
 
-                # Remove the combined fields
-                if "name" in vals:
-                    del vals["name"]
-                if "default_name" in context:
-                    del context["default_name"]
-        # pylint: disable=W8121
-        return super(ResPartner, self.with_context(context)).create(vals_list)
+    def get_extra_default_copy_values(self, order):
+        """Method to add '(copy)' suffix to lastname or firstname, depending on name
+        order configuration.
+        """
+        if order == "first_last":
+            return {
+                "lastname": _("%s (copy)", self.lastname)
+                if self.lastname
+                else _("(copy)")
+            }
+        return {
+            "firstname": _("%s (copy)", self.firstname)
+            if self.firstname
+            else _("(copy)")
+        }
 
     def copy(self, default=None):
         """Ensure partners are copied right.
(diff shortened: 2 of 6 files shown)


payen000:
Hello @luisg123v, @hugho-ad, could you review this, please?

partner_firstname/models/base_config_settings.py
```diff
@@ -61,7 +61,10 @@ def action_recalculate_partners_name(self):
         )
         partners = self._partners_for_recalculating()
         _logger.info("Recalculating names for %d partners.", len(partners))
-        partners._compute_name()
+        # Use add_to_compute instead of _compute_name to avoid triggering
+        # _inverse_name_after_cleaning_whitespace, which can
```
hugho-ad, line 65: why the inverse it's triggered?

partner_firstname/models/res_partner.py
```diff
@@ -26,11 +26,28 @@ class ResPartner(models.Model):
...
+    def name_fields_in_vals(self, vals):
+        """Method to check if any name fields are in `vals`."""
+        return vals.get("firstname") or vals.get("lastname")
+
     @api.model_create_multi
     def create(self, vals_list):
         """Add inverted names at creation if unavailable."""
         context = dict(self.env.context)
         for vals in vals_list:
+            # Remove the full name from `vals` and context if the partner is an
```
hugho-ad: Move it to the method dosctrings

partner_firstname/models/res_partner.py
```diff
@@ -53,13 +70,35 @@ def create(self, vals_list):
...
+        }
+
     def copy(self, default=None):
         """Ensure partners are copied right.
 
         Odoo adds ``(copy)`` to the end of :attr:`~.name`, but that would get
         ignored in :meth:`~.create` because it also copies explicitly firstname
         and lastname fields.
         """
+        if default is None:
```
hugho-ad: default = default or {}

partner_firstname/models/base_config_settings.py
payen000: This is because the `name` field (which is now a stored field) has as its inverse the `inverse_name_after_cleaning_whitespace` method:

https://github.com/OCA/partner-contact/blob/23c2d101e08c36f3909091c5f488110ce45e14cc/partner_firstname/models/res_partner.py#L23

so when calling the `_compute_name` method directly, its inverse is called afterwards:

https://github.com/odoo/odoo/blob/72a5da62d00feb247e4dcc5e6be52955de1c83f8/odoo/models.py#L3808

When using `add_to_compute` instead, the `name` is recomputed during `flush` here:

https://github.com/odoo/odoo/blob/72a5da62d00feb247e4dcc5e6be52955de1c83f8/odoo/fields.py#L1370

 and the `write` method is not called*, so the inverse is not triggered.
 
 *Note, I am unsure of why the write method is not called, I believe it has to do with the `_compute_name` method being [called as a function](https://github.com/odoo/odoo/blob/72a5da62d00feb247e4dcc5e6be52955de1c83f8/odoo/fields.py#L1343) with arguments instead of as a method belonging to the `Field` class, which might prevent other methods from `self` from triggering.

payen000:
Hello @hugho-ad, thanks for reviewing, I've applied your suggested changes, could you check again please?

hugho-ad (approved):

payen000:
Hello @luisg123v, could you review this please?

partner_firstname/models/res_partner.py
```diff
@@ -26,11 +26,30 @@ class ResPartner(models.Model):
...
+        """
         context = dict(self.env.context)
         for vals in vals_list:
+            if (
+                not vals.get("is_company")
+                and self.name_fields_in_vals(vals)
+                and "name" in vals
+            ):
+                vals.pop("name", None)
+                context.pop("default_name", None)
```
luisg123v: This will affect context for all records. To avoid that, you should copy context on each iteration and call super individually.

partner_firstname/tests/test_copy.py
```diff
@@ -22,6 +23,16 @@ def create_original(self):
             }
         )
 
+    def create_original_multinames(self):
+        self.original_multinames = self.env["res.users"].create(
```
luisg123v: Add context to avoid sending emails.

partner_firstname/tests/test_copy.py
```diff
@@ -50,3 +61,19 @@ def test_copy_firstname_lastname_name(self):
             }
         )
         self.compare(copy)
+
+    def test_copy_multiple_names(self):
+        copy = self.original_multinames.partner_id.copy()
+        self.assertEqual(copy.firstname, "Firstname1 Firstname2")
+        self.assertEqual(copy.lastname, "Lastname1 Lastname2 (copy)")
+        self.assertEqual(copy.name, "Firstname1 Firstname2 Lastname1 Lastname2 (copy)")
```
luisg123v: Instead of several assertions, use:

`self.assertRecordValues`

Same below.

partner_firstname/models/res_partner.py
```diff
@@ -26,11 +26,30 @@ class ResPartner(models.Model):
         readonly=False,
     )
 
+    @api.model
+    def name_fields_in_vals(self, vals):
+        """Method to check if any name fields are in `vals`."""
+        return vals.get("firstname") or vals.get("lastname")
```
luisg123v, line 32: I think this method should instead return a list of fields that you could check using:

`set(vals) & self._get_name_fields()`

Or are you doing it this way to consider provided values that are falsy?

partner_second_lastname/tests/test_multiple_name.py
```diff
@@ -0,0 +1,31 @@
+from odoo.tests.common import TransactionCase
+
+
+class TestRecalculateNames(TransactionCase):
```
luisg123v: TestMultipleName

partner_second_lastname/tests/__init__.py
```diff
@@ -2,4 +2,5 @@
 
 from . import test_name
 from . import test_config
+from . import test_multiple_name
```
luisg123v: Did you mean names in plural?

partner_second_lastname/tests/test_multiple_name.py
```diff
@@ -0,0 +1,31 @@
...
+                "lastname2": lastname2,
+            }
+        )
+        config_settings = self.env["res.config.settings"].create({})
+        order_names_dict = {
+            "first_last": f"{firstname} {lastname} {lastname2}",
+            "last_first": f"{lastname} {lastname2} {firstname}",
+            "last_first_comma": f"{lastname} {lastname2}, {firstname}",
+        }
+        for order, expected_name in order_names_dict.items():
```
luisg123v: We should avoid loops inside tests, because tests should be as deterministic as possible.

If you move code inside loop to a different method and you call that method three times, you will end up with even less lines than currently.

luisg123v (changes requested):
Bot commit messages say: "This PR". But not a PR but a commit.

partner_firstname/tests/test_copy.py
payen000: What context are you referring to?

I don't see any `mail.mail` records created after this `create` call, only `mail.message` records  - which can be avoided with the `mail_create_nolog` context, is that what you meant?

partner_firstname/models/res_partner.py
payen000: > Or are you doing it this way to consider provided values that are falsy?

Yes, if neither `firstname` nor `lastname` are set, then they should be computed from the `name` field; otherwise, if any of them are set, `name` should be computed from them instead; deciding this is what the method is made for.

partner_firstname/tests/test_copy.py
luisg123v: Whenever creating a user, it's recommended to use the context `no_reset_password=True`

payen000: Got it, thanks!

payen000:
@luisg123v I've applied your suggestions, could you check again please?

partner_firstname/models/res_partner.py
```diff
@@ -26,32 +26,71 @@ class ResPartner(models.Model):
...
+        from `vals` and context if the partner is an individual and is being created
+        with any name fields, as the name must be computed from the provided name parts;
+        otherwise, the name fields will be computed from the `name` again, when calling
+        its inverse method.
+
+        Note that, to avoid deleting the 'default_name' context for all partners when it's
+        not appropriate, we must call `create` for each partner individually with the correct
+        context.
+        """
+        created_partners = self.env["res.partner"]
```
luisg123v:
```suggestion
        created_partners = self.browse()
```

partner_firstname/models/res_partner.py
```diff
@@ -26,32 +26,71 @@ class ResPartner(models.Model):
...
+                not vals.get("is_company")
+                and self.name_fields_in_vals(vals)
+                and "name" in vals
+            ):
+                del vals["name"]
+                partner_context.pop("default_name", None)
+            else:
+                name = vals.get("name", partner_context.get("default_name"))
+                if name is not None:
+                    # Calculate the splitted fields
```
luisg123v:
```suggestion
                    # Calculate the split fields
```

partner_second_lastname/tests/test_multiple_names.py
```diff
@@ -0,0 +1,28 @@
+from odoo.tests.common import TransactionCase
+
+
+class TestMultipleNames(TransactionCase):
+    def test_recalculate_names(self):
+        def assert_name(partner, vals, order, expected_name):
```
luisg123v: Avoid nesting functions inside methods, please.

luisg123v (changes requested):

payen000:
Hello @luisg123v, I've applied your suggestions, could you review again please?

luisg123v (approved):
LGTM :+1:

luisg123v:
@moylop260 I can't merge here. Could you help me with that, please?

payen000:
Hello @moylop260, could you help us merge this please?

PR #1651, opened 2024-01-30
[17.0][MIG] base_location: Migration to 17.0

beagle-cloud:


base_location/views/res_company_view.xml
```diff
@@ -16,14 +16,11 @@
...
-                    name='city_id'
-                    attrs="{'invisible': [('country_enforce_cities', '=', False)]}"
-                />
+                <field name='city_id' invisible="country_enforce_cities == False" />
             </field>
             <field name="city" position="attributes">
-                <attribute name="attrs">
-                    {'invisible': [('country_enforce_cities', '=', False)]}
+                <attribute name="invisible">
+                    country_enforce_cities == False
```
pedrobaeza:
```suggestion
                <attribute name="invisible">
                    not country_enforce_cities
```

base_location/views/res_country_view.xml
```diff
@@ -20,7 +19,7 @@
                     class="oe_stat_button"
                     icon="fa-globe"
                     type="action"
-                    context="{'default_country_id': active_id, 'search_default_country_id': active_id}"
+                    context="{'default_country_id': id, 'search_default_country_id': id}"
```
pedrobaeza, line 22: This is not working properly: all ZIPs are visible from all countries.

base_location/views/res_partner_view.xml
```diff
@@ -10,7 +10,6 @@
                     name="zip_id"
                     options="{'create_name_field': 'city', 'no_open': True, 'no_create': True}"
                     placeholder="Location completion"
-                    attrs="{'readonly': [('type', '=', 'contact'),('parent_id', '!=', False)]}"
```
pedrobaeza: Why removing this?

pedrobaeza (changes requested):
There are 2 "Cities" menus inside Contacts > Configuration

base_location/views/res_country_view.xml
beagle-cloud: I found the reason, it is not caused by the migration, it is not working on V16 runboat either. It was because country_id was not in the search, I added it.

base_location/views/res_partner_view.xml
beagle-cloud: restored

pedrobaeza:
Please don't add new commits, but squash it onto the existing migration one, and pass pre-commit. You can't remove the space at the end of ` />` entries.

pedrobaeza:
But please don't add extra commits. Squash them in the migration script.

HaraldPanten (approved):
Functional review in runboat. Seems to be OK.

pedrobaeza (approved):
/ocabot merge nobump

PR #1452, opened 2023-01-16
[15.0][FW][IMP] base_partner_sequence: improve write performance

JordiMForgeFlow:
FW of https://github.com/OCA/partner-contact/pull/1443

PR #1655, opened 2024-02-01
[16.0][IMP] base_partner_sequence: improve write performance

aliciagaarzo:
Prevent forcing to do record by record write always and for all the records, and only do it when needed.

Implement improvement https://github.com/OCA/partner-contact/pull/1452

### DIFF
--- a/base_partner_sequence/models/partner.py
+++ b/base_partner_sequence/models/partner.py
@@ -29,16 +29,16 @@ class ResPartner(models.Model):
         return super(ResPartner, self).copy(default=default)
 
     def write(self, vals):
-        for partner in self:
-            partner_vals = vals.copy()
-            if (
-                not partner_vals.get("ref")
-                and partner._needs_ref(vals=partner_vals)
-                and not partner.ref
-            ):
+        partners_needing_ref = self.env["res.partner"]
+        if not vals.get("ref"):
+            partners_needing_ref = self.filtered(
+                lambda p: p._needs_ref(vals=vals) and not p.ref
+            )
+            for partner in partners_needing_ref:
+                partner_vals = vals.copy()
                 partner_vals["ref"] = partner._get_next_ref(vals=partner_vals)
-            super(ResPartner, partner).write(partner_vals)
-        return True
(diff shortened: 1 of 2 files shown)


amarcosg (approved):
LGTM!

dalonsofl (approved):
LGTM!

imoralesc:
Hello @pedrobaeza, we have an improvement to avoid forcing the write record by record always, and do it only when necessary. Could you take a look at it and merge it?

pedrobaeza:
OK. I think this module is over-engineered, checking on each write for the sequence, while it should be only on create, but this performance improvement is welcome not touching the main reason of the bottleneck.

Anyway, I don't use this module in any of my customers, as the partner reference is something anachronistic, and I recommend you to recheck its need. It's hard at the beginning to convince old users about its lack of need, but they won't miss it in few weeks.

/ocabot merge patch

pedrobaeza:
Can you please rebase to see if the CI is OK?

amarcosg:
Thanks for the feedback @pedrobaeza, I just did the rebase and the error has to do with being February 29th in the test of partner_contact_age_range.

Bad luck i guess, i will try again tomorrow :disappointed:

pedrobaeza:
Well, the test should be designed for not having that problems...

Usually, it's a problem of the test having "variable dates" based on today, and it's bypassed freezing the date with `freezegun`, but if it's showcasing a problem in the feature itself that won't work on this day, then it should be correctly programmed.

I know it's not your module, but if you want to explore this as exercise, go ahead.

If not, ping me tomorrow to launch again the merge command.

amarcosg:
Sure no problem, I've already made the change so we can make sure it doesn't happen again.

pedrobaeza:
Great, thanks, but you have forgotten to put the module name in the commit message. And have you checked if the problem is in the tests or in the inner feature?

amarcosg:
Right, I have already added the module name. I have not entered to validate the functionality of the module to check if the problem is in the inner feature, I have limited myself only to the test

NL66278 (approved):
:+1: LGTM

PR #1668, opened 2024-02-20
[16.0][ADD] partner_auto_archive

MarinaAForgeFlow:


partner_auto_archive/models/res_partner.py
```diff
@@ -0,0 +1,15 @@
...
+
+class Partner(models.Model):
+    _inherit = "res.partner"
+
+    auto_archive = fields.Boolean(help="Archive contact automatically after a period of time")
+
+    @api.model
+    def _auto_archive_contacts(self):
+        contacts_to_archive = self.env["res.partner"].search([("auto_archive", "=", True)])
+        contacts_to_archive.write({"active": False})
```
JordiMForgeFlow:
```suggestion
        contacts_to_archive.action_archive()
```

JordiMForgeFlow:
@MarinaAForgeFlow could you add a simple test and a USAGE section in the README?

partner_auto_archive/views/res_partner_views.xml
```diff
@@ -0,0 +1,17 @@
...
+        <field name="model">res.partner</field>
+        <field name="inherit_id" ref="base.view_partner_form" />
+        <field name="arch" type="xml">
+            <!-- Add checkbox to set as auto-archive -->
+            <xpath expr="//span[@name='address_name']" position="before">
+                <field name="auto_archive" />
+            </xpath>
+            <xpath expr="//label[@for='street']" position="before">
+                <field name="auto_archive" />
+            </xpath>
```
JordiMForgeFlow: maybe it would be better to put the field after "category_id" so that it is shown in the right column of field information

partner_auto_archive/models/res_partner.py
```diff
@@ -0,0 +1,19 @@
...
+    auto_archive = fields.Boolean(
+        help="Archive contact automatically after a period of time"
+    )
+
+    @api.model
+    def _auto_archive_contacts(self):
+        contacts_to_archive = self.env["res.partner"].search(
+            [("auto_archive", "=", True)]
+        )
+        contacts_to_archive.action_archive()
```
GuillemCForgeFlow, line 19: Perhaps a good improvement could be unsetting the `auto_archive` after running the `action_archive()` just so that if the partner gets unarchived in the future, the user doesn't have manually unmark the `auto_archive`.
Just as an nice to have, not needed for the funcionality of the module.

MarinaAForgeFlow: I think it is a good idea. I will add it. Thanks!

GuillemCForgeFlow (approved):
I guess that you will need to squash all commits to just leave the [ADD] one, but for the rest, LGTM! 👍🏽

partner_auto_archive/tests/test_partner_auto_archive.py
```diff
@@ -0,0 +1,33 @@
...
+                "auto_archive": False,
+            }
+        )
+
+    def test_partner_auto_archive(self):
+        self.cron.method_direct_trigger()
+        archived_contacts = self.env["res.partner"].search(
+            [("active", "=", False), ("name", "ilike", "Partner Archive")]
+        )
+        self.assertEqual(len(archived_contacts), 2)
```
JordiMForgeFlow, line 33: you could also add a check that the auto_archive has been set to False when they have been archived

JordiMForgeFlow (approved):
Code + functional review 👍🏼

PR #1673, opened 2024-02-22
[16.0] [IMP] partner_industry_secondary: add unaccent=False on parent_path field

marielejeune:
As asked by Odoo in a warning in `_check_parent_path` method.

See https://github.com/odoo/odoo/blob/16.0/odoo/models.py#L2615


pedrobaeza (approved):
Straight-forward, so fast-tracking:

/ocabot merge patch

PR #1675, opened 2024-02-27
[FIX] partner_vat_unique: VAT uniqueness constraint for XMLRPC contact creation

randall-vx:
Modifies the `res.partner` model in Odoo to apply the VAT uniqueness constraint specifically when a contact is created via XMLRPC.

### DIFF
--- a/partner_vat_unique/models/res_partner.py
+++ b/partner_vat_unique/models/res_partner.py
@@ -26,3 +26,11 @@ class ResPartner(models.Model):
                 raise ValidationError(
                     _("The VAT %s already exists in another partner.") % record.vat
                 )
+
+    @api.model
+    def create(self, vals_list):
+        new_records = super(ResPartner, self).create(vals_list)
+        if "xmlrpc" in self.env.context:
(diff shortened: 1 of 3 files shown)


randall-vx:
Tested on [runboat](http://oca-partner-contact-16-0-pr1675-a32c38a00de6.runboat.odoo-community.org)

Screenshot 2024-02-27 at 1 21 10 PM

Python script to test create method from XML-RPC

https://gist.github.com/randall-vx/4fb7959e216d42f6d6f384be62c511ad

cc: @moylop260

partner_vat_unique/models/res_partner.py
```diff
@@ -26,3 +26,10 @@ def _check_vat_unique(self):
                 raise ValidationError(
                     _("The VAT %s already exists in another partner.") % record.vat
                 )
+
+    @api.model
+    def create(self, vals):
```
luisg123v: Support multiple records, otherwise this will raise warnings

luisg123v (changes requested):

Issue #1677, opened 2024-02-28
[16.0] partner_firstname cannot be installed, if the system contains a contact with an empty name


mohs8421:
When trying to install this module, I already get the following error:
> RPC_ERROR
> Odoo Server Error
> Traceback (most recent call last):
  …
  File "/usr/lib/python3/dist-packages/odoo/modules/registry.py", line 90, in new
    odoo.modules.load_modules(registry, force_demo, status, update_module)
  File "/usr/lib/python3/dist-packages/odoo/modules/loading.py", line 488, in load_modules
    processed_modules += load_marked_modules(cr, graph,
  File "/usr/lib/python3/dist-packages/odoo/modules/loading.py", line 372, in load_marked_modules
    loaded, processed = load_module_graph(
  File "/usr/lib/python3/dist-packages/odoo/modules/loading.py", line 249, in load_module_graph
    getattr(py_module, post_init)(cr, registry)
  File "partner_firstname/hooks.py", line 9, in post_init_hook
    env["res.partner"]._install_partner_firstname()
  File "partner_firstname/models/res_partner.py", line 220, in _install_partner_firstname
    records._inverse_name()
  File "partner_firstname/models/res_partner.py", line 193, in _inverse_name
    record.lastname = parts["lastname"]
  File "/usr/lib/python3/dist-packages/odoo/fields.py", line 1325, in __set__
    records.write({self.name: write_value})
    …
  File "/usr/lib/python3/dist-packages/odoo/models.py", line 3798, in write
    real_recs._validate_fields(vals, inverse_fields)
  File "/usr/lib/python3/dist-packages/odoo/models.py", line 1359, in _validate_fields
    check(self)
  File "partner_firstname/models/res_partner.py", line 206, in _check_name
    raise exceptions.EmptyNamesError(record)
odoo.addons.partner_firstname.exceptions.EmptyNamesError: ('Fehler beim Namen des Partners 1270 .', 'Es ist kein Name eingetragen.')

It might be an issue referring to the general [misconceptions about names](https://www.kalzumeus.com/2010/06/17/falsehoods-programmers-believe-about-names/) [also](https://shinesolutions.com/2018/01/08/falsehoods-programmers-believe-about-names-with-examples/)
> 40. People have names.

Besides the edgecases, it might also be relevant for the system to be aware that the information about the firstname might simply be missing and only other details might be available.

## Module
partner_firstname

## To Reproduce
**Affected versions**: 16.0

[...]

PR #1678, opened 2024-02-28
[16.0][FIX] partner_firstname: fix installation w/ missing names

mohs8421:
This change fixes this bug: https://github.com/OCA/partner-contact/issues/1677

With that the module can be installed as only partners with names will be processed

### DIFF
--- a/partner_firstname/models/res_partner.py
+++ b/partner_firstname/models/res_partner.py
@@ -259,7 +259,9 @@ class ResPartner(models.Model):
         correctly into the database. This can be called later too if needed.
         """
         # Find records with empty firstname and lastname
-        records = self.search([("firstname", "=", False), ("lastname", "=", False)])
+        records = self.search(
+            [("firstname", "=", False), ("lastname", "=", False), ("name", "!=", "")]
+        )
 
         # Force calculations there
         records._inverse_name()
--- a/partner_firstname/tests/test_create.py
+++ b/partner_firstname/tests/test_create.py
@@ -33,6 +33,18 @@ class PersonCase(TransactionCase):
 
         super(PersonCase, self).tearDown()
 
+    def test_module_installation(self):
+        model = self.env[self.model].with_context(**self.context)
+        unnamed = model.create(
+            {"name": "", "email": "anon@ymous.org", "type": "invoice"}
+        )
+        named = model.create(
+            {"name": "firstname lastname", "email": "firstname@lastname.org"}
+        )
+        model._install_partner_firstname()
+        self.assertEqual(unnamed.name, "")
+        self.assertEqual((named.firstname, named.lastname), ("firstname", "lastname"))
+
     def test_no_name(self):
         """Name is calculated."""
         del self.values["name"]


simahawk:
@mohs8421 could you please add a test for this change?
When you are done, could you please:
1. squash your commits
2. rewrite the commit message according to our guidelines https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#commit-message

Thanks!

mohs8421:
I tried to write a test for it, but it is not possible with the module in place to test this, as the module itself prevents that.
So when trying to create a test record to run `_install_partner_firstname()` afterwards, it already exits with the empty name error.
Yes, that's partly the point of this module, but it should not be that strict during installation.

For reference, here is the testcode:
```python
    def test_module_installation(self):
        model = self.env[self.model].with_context(**self.context)
        unnamed = model.create({'name': '', 'email': 'anon@ymous.org'})
        named = model.create({'name': 'firstname lastname', 'email': 'firstname@lastname.org'})
        model._install_partner_firstname()
        self.assertEqual(unnamed.name, '')
        self.assertEqual((named.firstname, named.lastname), ('firstname', 'lastname'))
```

yvaucher (approved):
LGTM, seems safer that way.

SirAionTech:
> I tried to write a test for it, but it is not possible with the module in place to test this, as the module itself prevents that.
So when trying to create a test record to run `_install_partner_firstname()` afterwards, it already exits with the empty name error.
Yes, that's partly the point of this module, but it should not be that strict during installation.
> 
> For reference, here is the testcode:
> 
> ```python
>     def test_module_installation(self):
>         model = self.env[self.model].with_context(**self.context)
>         unnamed = model.create({'name': '', 'email': 'anon@ymous.org'})
>         named = model.create({'name': 'firstname lastname', 'email': 'firstname@lastname.org'})
>         model._install_partner_firstname()
>         self.assertEqual(unnamed.name, '')
>         self.assertEqual((named.firstname, named.lastname), ('firstname', 'lastname'))
> ```

You could `mock` whatever method is preventing you from creating `unnamed`, just while creating it, so that it's like you are creating the partner while the module is not installed yet.
You can see an example in https://github.com/OCA/account-invoicing/blob/86b9060ce96d91251f6a03bdc8021176880d3586/partner_invoicing_mode/tests/test_invoice_mode.py#L112:
```py
with mock.patch.object(ResPartner, "_update_next_invoice_date") as mock_update:
    ...
```

CRogos (commented):
@mohs8421 I think your test is nearly working. I've added your test to the test_empty.py file and just added the contact **"type": "invoice"** because empty name is not allowed on all contact types. Than the tests run successfully.


```
class AddressCase(TransactionCase):

...

    def test_module_installation(self):
        """Create an invoice patner without name."""
        model = self.env["res.partner"]
        unnamed = model.create({'name': '', 'email': 'anon@ymous.org', "type": "invoice"})
        named = model.create({'name': 'firstname lastname', 'email': 'firstname@lastname.org'})
        model._install_partner_firstname()
        self.assertEqual(unnamed.name, '')
        self.assertEqual((named.firstname, named.lastname), ('firstname', 'lastname'))
        
```

If you add your tests, and squash all commits, I will also approve this PR.

PR #1679, opened 2024-02-28
[17.0] [MIG] partner_industry_secondary

mmrondon:


partner_industry_secondary/views/res_partner_view.xml
```diff
@@ -0,0 +1,79 @@
...
+        <field name="name">Partner form with industry</field>
+        <field name="model">res.partner</field>
+        <field name="inherit_id" ref="base.view_partner_form" />
+        <field name="arch" type="xml">
+            <field name="industry_id" position="after">
+                <field name="show_partner_industry_for_person" invisible="1" />
+                <field
+                    name="secondary_industry_ids"
+                    widget="many2many_tags"
+                    invisible="is_company == False and show_partner_industry_for_person == False"
```
DucTruongKomit, line 16: I imagine that the user should set this field after they set the field industry_id 1st. Because the this field got domain [('id', '!=', industry_id)]

So, we should hide this field also if industry_id = False

For example:
I set Partner A with industry_id then, display this field, so the user may think about their options on secondary_industry_ids

Do you agree with me?

DucTruongKomit (commented):
Hello! Do you agree with me on my comments?

partner_industry_secondary/views/res_partner_view.xml
mmrondon: hello @DucTruongKomit 
Yes, your comment makes sense, but in this case the module already has a restriction in case the same industry is selected as main and secondary, only possible if a secondary industry is selected first.

validationError

DucTruongKomit (approved):
LGTM! All gud!

HaraldPanten:
@mmrondon You have too many commits for this migration. Don't you?

Could you clean the commits and leave only what we need for the migration?

THX.

partner_industry_secondary/views/res_config_settings.xml
```diff
@@ -0,0 +1,49 @@
...
+                                        icon="fa-arrow-right"
+                                        type="action"
+                                        string="Industries"
+                                        class="btn-link"
+                                    />
+                                </div>
+                            </div>
+                        </div>
+                    </div>
+                </div>
```
Tisho99, line 45:
```suggestion
                <block>
                    <setting string="Enable industries for individuals" help="Enable industries for individuals">
                        <field name="group_use_partner_industry_for_person" />
                    </setting>
                </block>
```

partner_industry_secondary/views/res_partner_view.xml
```diff
@@ -0,0 +1,81 @@
...
+        <field name="name">Partner tree with industry</field>
+        <field name="model">res.partner</field>
+        <field name="inherit_id" ref="base.view_partner_tree" />
+        <field name="arch" type="xml">
+            <field name="email" position="after">
+                <field
+                    name="industry_id"
+                    invisible="is_company == False and show_partner_industry_for_person == False"
+                />
+                <field name="show_partner_industry_for_person" invisible="1" />
```
Tisho99, line 54:
```suggestion
                <field name="show_partner_industry_for_person" column_invisible="1" />
```

Tisho99 (changes requested):
Hi @mmrondon 

I have made a functional and code review and this migration looks good.

However, i found that you haven't applied some steps of the [migration guide](https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-17.0)

Can you review and apply my suggestions in your last commit?

Then this will be ready for merge

Thank you!

HaraldPanten:
Superseded by https://github.com/OCA/partner-contact/pull/1761

PR #1684, opened 2024-03-04
[15.0][IMP] partner_pricelist_search: support searching subfields

chienandalu:
Previously we couldn't make an advanced search to filter pricelist subfields.

Currently, only possitive operators are working for advanced searches as negative ones seem to be encapsulating the queries and we can't use them in the point we're overriding. Further research would need to be done in order to cover it.

cc @Tecnativa TT47648

please check @carlosdauden 
please review @victoralmau


carlosdauden (approved):

pedrobaeza (approved):
/ocabot merge major

pedrobaeza:
Please fw-port it to v16

PR #1526, opened 2023-05-22
[16.0][MIG] partner_employee_quantity

SoporteOperu:

PR #225, opened 2015-12-17
[8.0][partner_employee_quantity] New module to know the quantity of employees per partner

yajo:
If you need to keep track of the size of your partners by employee number, here's your module. It allows to set a number and a classification.

@rafaelbn @antespi @JavierIniesta.

PR #1687, opened 2024-03-04
[16.0] [MIG] partner_employee_quantity

edlopen:
I only added the commits of [this stale PR](https://github.com/OCA/partner-contact/pull/1526) and changed the module icon to the OCA one.

@moduon MT-1749

Please @Shide @yajo @EmilioPascual @rafaelbn please review, if you can.

Closes #1526


yajo (changes requested):
I'm missing Git history before the rename (see 1st commit, which comes from https://github.com/OCA/partner-contact/pull/225). This module had several renames and we should respect all that history in the migration if possible.

Apart from that, it's OK.

/ocabot migration partner_employee_quantity

yajo (approved):

fcvalgar (approved):
Functional review.

LGTM, Thank you @edlopen 

image


image

image

PR #1697, opened 2024-03-12
[17.0][MIG] partner_contact_job_position

WorrapongOngsakul:
The XML attributes attrs and states are no longer used.


newtratip:
Please use module name like this on subject, it is easier to search by others, e.g., partner_contact_job_position

HaraldPanten:
@WorrapongOngsakul could you rebase, please?

THX

PR #1704, opened 2024-03-15
Revert "[IMP] partner_(ref,vat)_unique: enable name search by the VAT Identification Number and Reference"

CLaurelB:
Revert commits 569f43b9 and ef13b589. This is because more research has to be done in order to account for all possible cases where the search is involved. For example, when a domain needs to be considered with an AND operator.

### DIFF
--- a/partner_ref_unique/models/res_partner.py
+++ b/partner_ref_unique/models/res_partner.py
@@ -4,7 +4,6 @@
 
 from odoo import _, api, models
 from odoo.exceptions import ValidationError
-from odoo.osv import expression
 
 
 class ResPartner(models.Model):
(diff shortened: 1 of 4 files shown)


luisg123v (changes requested):
Explanation is missing from commit descriptions.

CLaurelB:
> Explanation is missing from commit descriptions.

Added.

luisg123v (approved):
LGTM :+1:

luisg123v:
It seems it's a flaky error on an unrelated module, let's try again:

/ocabot merge patch

sergio-teruel (approved):

PR #1724, opened 2024-04-03
[16.0][ADD] partner_shipping_policy

MarinaAForgeFlow:


partner_shipping_policy/models/res_partner.py
```diff
@@ -0,0 +1,17 @@
...
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class Partner(models.Model):
+    _inherit = "res.partner"
+
+    picking_policy = fields.Selection(
+        [("direct", "As soon as possible"), ("one", "When all products are ready")],
```
GuillemCForgeFlow: Perhaps it would be great to get the selection values from the `picking_policy` field from the `sale.order`? Then, if more values are added in custom modules for example, it will already be caught here.

partner_shipping_policy/models/res_partner.py
```diff
@@ -0,0 +1,17 @@
+# Copyright 2024 ForgeFlow S.L. (https://www.forgeflow.com)
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class Partner(models.Model):
+    _inherit = "res.partner"
+
+    picking_policy = fields.Selection(
```
GuillemCForgeFlow, line 10: I'm guessing the field could be added to the `_commercial_fields` method result? WDYT?

GuillemCForgeFlow (commented):
For the rest, functional review 👍🏿

partner_shipping_policy/models/res_partner.py
MarinaAForgeFlow: Yes, I did not think of that but it makes sense. Thanks!

partner_shipping_policy/__manifest__.py
```diff
@@ -0,0 +1,16 @@
+# Copyright 2024 ForgeFlow S.L. (https://www.forgeflow.com)
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+{
+    "name": "Partner Shipping Policy",
+    "summary": "Define shipping policy at partners level.",
+    "author": "ForgeFlow,Odoo Community Association (OCA)",
+    "website": "https://github.com/OCA/partner-contact",
+    "category": "Sales",
+    "version": "16.0.1.0.1",
```
GuillemCForgeFlow:
```suggestion
    "version": "16.0.1.0.0",
```

GuillemCForgeFlow (approved):
Apart from this small change, the rest looks great.
LGTM code and functional review 👍🏿

partner_shipping_policy/__manifest__.py
```diff
@@ -0,0 +1,16 @@
...
+
+{
+    "name": "Partner Shipping Policy",
+    "summary": "Define shipping policy at partners level.",
+    "author": "ForgeFlow,Odoo Community Association (OCA)",
+    "website": "https://github.com/OCA/partner-contact",
+    "category": "Sales",
+    "version": "16.0.1.0.0",
+    "license": "AGPL-3",
+    "depends": ["base", "sale_stock"],
```
JordiMForgeFlow:
```suggestion
    "depends": ["sale_stock"],
```

partner_shipping_policy/models/sale_order.py
```diff
@@ -0,0 +1,22 @@
...
+    )
+
+    @api.depends("partner_id")
+    def _compute_picking_policy(self):
+        for this in self:
+            if this.partner_id.picking_policy:
+                this.picking_policy = this.partner_id.picking_policy
+            else:
+                # Default to 'direct' if partner's policy is not set
+                this.picking_policy = "direct"
```
JordiMForgeFlow: the default may not be "direct", it is actually taken from a setting

partner_shipping_policy/models/res_partner.py
```diff
@@ -0,0 +1,32 @@
...
+
+    def _get_default_picking_policy(self):
+        return (
+            self.env["sale.order"].default_get(["picking_policy"]).get("picking_policy")
+        )
+
+    picking_policy = fields.Selection(
+        selection="_get_picking_policy_selection",
+        string="Shipping Policy",
+        default=_get_default_picking_policy,
```
JordiMForgeFlow: not sure if it is the best to put a default. specifying a picking policy in a partner should be to cover specific cases, I don't think we should fill it always

JordiMForgeFlow:
@GuillemCForgeFlow @MarinaAForgeFlow I think that the picking_policy should not be a commercial field, as you can have different delivery addresses for a customer and maybe the policy for the different addresses needs to be different.

What we could do is to not have it as a commercial field, but when deciding the picking policy in the SO we should evaluate in the following order: partner_shipping_id.picking_policy or partner_id.picking_policy

WDYT?

GuillemCForgeFlow:
> @GuillemCForgeFlow @MarinaAForgeFlow I think that the picking_policy should not be a commercial field, as you can have different delivery addresses for a customer and maybe the policy for the different addresses needs to be different.
> 
> What we could do is to not have it as a commercial field, but when deciding the picking policy in the SO we should evaluate in the following order: partner_shipping_id.picking_policy or partner_id.picking_policy
> 
> WDYT?

Makes sense to me to do like so. Then if we proceed with this proposal, I'd say it'd be great to add this to the Description file, just to be aligned.

GuillemCForgeFlow:
Great, thanks for the improvement 👍🏿

JordiMForgeFlow (approved):
Code review

PR #1725, opened 2024-04-03
[17.0][FIX] partner_firstname: fix error when creating a user from an employee

BurkhalterY:
Steps to reproduce:
- Install `hr` and `partner_firstname`
- Go to Employees > any employee > action menu > Create User > Save
- 💥 Traceback

Suspected cause: in `defaul_get()`, the field `name` is not present in `fields_list` (probably because it's computed). So `result.get("name", "")` will always return `""` and `firstname` and `lastname` are not computed.

### DIFF
--- a/partner_firstname/models/res_partner.py
+++ b/partner_firstname/models/res_partner.py
@@ -65,6 +65,8 @@ class ResPartner(models.Model):
     @api.model
     def default_get(self, fields_list):
         """Invert name when getting default values."""
+        if "firstname" in fields_list or "lastname" in fields_list:
+            fields_list.append("name")
         result = super().default_get(fields_list)
 
         inverted = self._get_inverse_name(


mymage (approved):
OK on RunBoat

luisDIXMIT (approved):
LGTM!

partner_firstname/models/res_partner.py
```diff
@@ -65,6 +65,7 @@ def copy(self, default=None):
     @api.model
     def default_get(self, fields_list):
         """Invert name when getting default values."""
+        fields_list.append("name")
```
pedrobaeza: IMO, the proper fix is to do the stuff only if `name` if `fields_list`, not to force always to include that field in the results.

pedrobaeza (changes requested):

partner_firstname/models/res_partner.py
BurkhalterY: Hi @pedrobaeza, I'm not sure to understand what do you propose, but I added a condition to add `name` only if `firstname` or `lastname` is in `fields_list`.
Let me know if there is a better solution

pedrobaeza: OK, that can work.

pedrobaeza (approved):
/ocabot merge patch

partner_firstname/models/res_partner.py
rpinset: other issue https://github.com/OCA/partner-contact/issues/1757

rpinset: Pr for FIX https://github.com/OCA/partner-contact/pull/1793

PR #1726, opened 2024-04-03
[16.0][FIX] partner_firstname: fix error wehn creating a user from an employee

BurkhalterY:
Steps to reproduce:
- Install `hr` and `partner_firstname`
- Go to Employees > any employee > action menu > Create User > Save
- 💥 Traceback

Suspected cause: in `defaul_get()`, the field `name` is not present in `fields_list` (probably because it's computed). So `result.get("name", "")` will always return `""` and `firstname` and `lastname` are not computed.


BurkhalterY:
This is actually not an issue in 16.0

PR #1736, opened 2024-04-11
[16.0][MIG] animal

Deriman-Alonso:
Standart migration


animal/models/animal.py
```diff
@@ -0,0 +1,41 @@
...
+    _inherit = ["mail.thread", "mail.activity.mixin"]
+    _order = "name"
+
+    name = fields.Char()
+    ref = fields.Char(string="Reference")
+    species_id = fields.Many2one("animal.species", string="Species", required=True)
+    breed_id = fields.Many2one("animal.breed", string="Breed", required=True)
+    color_id = fields.Many2one("animal.color", string="Color")
+    size = fields.Char()
+    weight = fields.Float(string="Weight (in kg)")
```
christian-ramos-tecnativa, line 18: Hi @Deriman-Alonso , I think we should take advantage of the opportunity to remove the specification regarding the pet's weight in kilograms. I believe the best approach would be to do something similar to how products/pickings handle weight based on the overall adjustment. 
image
image

christian-ramos-tecnativa (changes requested):

animal/models/animal.py
Deriman-Alonso: Makes sense to me, I'm working on it 👍.

Deriman-Alonso: Hi @Christian-RB , as far as my understanding goes, those UoM settings comes from the "product" module. Would it be properly to make the "animals" module depend on  "product", or rather create its own menu in the settings, even tho it means there could be different types of UoM for the weight?

christian-ramos-tecnativa: Oh You are right @Deriman-Alonso, in that case I assume is better to keep it the same way. Thanks!

christian-ramos-tecnativa (approved):

jado95 (approved):

MarcoCalcagni (approved):
Go 

we have in production 

@pedrobaeza

pedrobaeza (approved):
/ocabot migration animal
/ocabot merge nobump

PR #1740, opened 2024-04-18
[16.0][FIX] partner_multi_relation: fix name search on relation

NL66278:
Before this fix the following exception was logged (and the search not done):
```
2024-04-18 13:13:15,110 320 ERROR vwnvrs-testdatabase odoo.osv.expression: Non-stored field res.partner.relation.type.selection.display_name cannot be searched. 
NoneType: None
```

After fix no more errors in the log and the search works.

### DIFF
--- a/partner_multi_relation/models/res_partner_relation_all.py
+++ b/partner_multi_relation/models/res_partner_relation_all.py
@@ -54,7 +54,7 @@ class ResPartnerRelationAll(models.Model):
     _order = "this_partner_id, type_selection_id, date_end desc, date_start desc"
     _rec_names_search = [
         "this_partner_id.name",
-        "type_selection_id.display_name",
+        "type_selection_id.name",
         "other_partner_id.name",
     ]
 


SirAionTech:
Could you please explain how to obtain this error?
I tried to search for a relation in its `Many2one` field but the search is performed correctly and no ERROR is logged.

NL66278:
@SirAionTech The error occurs if you have a model with a Many2one field to res.partner.relation.all, and then in the form you enter a name to search on.

SirAionTech (changes requested):
Thanks for the PR!
I had a look at the code and tested this change as you suggested in https://github.com/OCA/partner-contact/pull/1740#issuecomment-2312116188:
> The error occurs if you have a model with a Many2one field to res.partner.relation.all, and then in the form you enter a name to search on.

Everything looks fine.

Could you please add a test that fails without this fix? Searching for records is a feature I'd say is worth protecting from regression.

NL66278:
@OCA-git-bot merge minor

PR #1742, opened 2024-04-22
[FIX] partner_firstname: Multiple creation with "name"

SirAionTech:
This happens after https://github.com/OCA/partner-contact/pull/1647.

**Steps**:
1. Create 2 partners in one `create`, only assigning their `name` (also see added test):
   ```
   self.env["res.partner"].create(
       [
           {
               "name": "Test partner1",
           },
           {
               "name": "Test partner2",
           },
       ]
   )
   ```

**Actual behavior**:
Exception raised:
> odoo.addons.partner_firstname.exceptions.EmptyNamesError: ("Error(s) with partner 1425's name.", 'No name is set.')

**Expected behavior**:
2 partners created

### DIFF
--- a/partner_firstname/models/res_partner.py
+++ b/partner_firstname/models/res_partner.py
@@ -1,6 +1,7 @@
 # Copyright 2013 Nicolas Bessi (Camptocamp SA)
 # Copyright 2014 Agile Business Group (<http://www.agilebg.com>)
 # Copyright 2015 Grupo ESOC (<http://www.grupoesoc.es>)
+# Copyright 2024 Simone Rubino - Aion Tech
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 import logging
 
@@ -73,7 +74,7 @@ class ResPartner(models.Model):
             # pylint: disable=W8121
             created_partners |= super(
                 ResPartner, self.with_context(partner_context)
-            ).create(vals_list)
+            ).create([vals])
         return created_partners
 
     def get_extra_default_copy_values(self, order):
--- a/partner_firstname/readme/CONTRIBUTORS.rst
+++ b/partner_firstname/readme/CONTRIBUTORS.rst
@@ -18,3 +18,6 @@
 * Graeme Gellatly <graeme@o4sb.com>
 * Laurent Mignon <laurent.mignon@acsone.eu>
 * Bjorn Billen <bjorn.billen@dynapps.be>
+* `Aion Tech <https://aiontech.company/>`_:
+
+  * Simone Rubino <simone.rubino@aion-tech.it>
--- a/partner_firstname/static/description/index.html
+++ b/partner_firstname/static/description/index.html
@@ -1,3 +1,4 @@
+<?xml version="1.0" encoding="utf-8"?>
 <!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN" "http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd">
 <html xmlns="http://www.w3.org/1999/xhtml" xml:lang="en" lang="en">
 <head>
@@ -466,6 +467,10 @@ If you spotted it first, help us to smash it by providing a detailed and welcome
 <li>Graeme Gellatly &lt;<a class="reference external" href="mailto:graeme&#64;o4sb.com">graeme&#64;o4sb.com</a>&gt;</li>
 <li>Laurent Mignon &lt;<a class="reference external" href="mailto:laurent.mignon&#64;acsone.eu">laurent.mignon&#64;acsone.eu</a>&gt;</li>
 <li>Bjorn Billen &lt;<a class="reference external" href="mailto:bjorn.billen&#64;dynapps.be">bjorn.billen&#64;dynapps.be</a>&gt;</li>
+<li><a class="reference external" href="https://aiontech.company/">Aion Tech</a>:<ul>
+<li>Simone Rubino &lt;<a class="reference external" href="mailto:simone.rubino&#64;aion-tech.it">simone.rubino&#64;aion-tech.it</a>&gt;</li>
+</ul>
+</li>
 </ul>
 </div>
 <div class="section" id="maintainers">
--- a/partner_firstname/tests/test_name.py
+++ b/partner_firstname/tests/test_name.py
@@ -23,6 +23,9 @@
 # along with this program; if not, write to the Free Software
 # Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA  02111-1307, USA.
 
+# Copyright 2024 Simone Rubino - Aion Tech
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
 """Test naming logic.
 
 To have more accurate results, remove the ``mail`` module before testing.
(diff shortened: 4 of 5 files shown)


SirAionTech:
@payen000 @hugho-ad @luisg123v @moylop260 you were involved in https://github.com/OCA/partner-contact/pull/1647, could you please have a look?

partner_firstname/readme/CONTRIBUTORS.rst
```diff
@@ -18,3 +18,6 @@
 * Graeme Gellatly <graeme@o4sb.com>
 * Laurent Mignon <laurent.mignon@acsone.eu>
 * Bjorn Billen <bjorn.billen@dynapps.be>
+* `Aion Tech <https://aiontech.company/>`_:
```
luisg123v, line 21: Let's leave these automatic changes to the bot, so the commit contains only the actual changes.

partner_firstname/models/res_partner.py
```diff
@@ -73,7 +74,7 @@ def create(self, vals_list):
             # pylint: disable=W8121
             created_partners |= super(
                 ResPartner, self.with_context(partner_context)
-            ).create(vals_list)
+            ).create(vals)
```
luisg123v:
```suggestion
            ).create([vals])
```

partner_firstname/readme/CONTRIBUTORS.rst
luisg123v, line 21: This syntax seems to not be correct, I'd include only the contributor name, as in the already-existing list items.

partner_firstname/static/description/index.html
```diff
@@ -466,6 +467,10 @@ <h2><a class="toc-backref" href="#toc-entry-7">Contributors</a></h2>
 <li>Graeme Gellatly &lt;<a class="reference external" href="mailto:graeme&#64;o4sb.com">graeme&#64;o4sb.com</a>&gt;</li>
 <li>Laurent Mignon &lt;<a class="reference external" href="mailto:laurent.mignon&#64;acsone.eu">laurent.mignon&#64;acsone.eu</a>&gt;</li>
 <li>Bjorn Billen &lt;<a class="reference external" href="mailto:bjorn.billen&#64;dynapps.be">bjorn.billen&#64;dynapps.be</a>&gt;</li>
+<li><a class="reference external" href="https://aiontech.company/">Aion Tech</a>:<ul>
```
luisg123v, line 470: Same here, let's leave automatic changes to the bot.

partner_firstname/tests/test_name.py
```diff
@@ -50,6 +53,19 @@ def test_whitespace_cleanup(self):
...
+            [
+                {
+                    "name": "Test partner1",
+                },
+                {
+                    "name": "Test partner2",
+                },
+            ]
+        )
+
```
luisg123v, line 82: Provide atleast one assertion, e.g. `assertRecordValues`

luisg123v (changes requested):
Missing commit description.

payen000:
Hello @SirAionTech,

I would only add to the current review comments that the commit message `[FIX] partner_firstname: Multiple creation with "name"` could probably be more descriptive, like `[FIX] partner_firstname: Avoid duplicate partner creation with "name"` (or something different that points the attention to the issue being fixed). Then the commit description would clarify the details.

partner_firstname/models/res_partner.py
SirAionTech: Done!
Since it works either way I'm curious: is there any particular reason to prefer this syntax to the one I proposed?

partner_firstname/tests/test_name.py
SirAionTech: Assertion added; there was none before because I was already happy that no exception is raised, as it happens before these changes.

partner_firstname/readme/CONTRIBUTORS.rst
SirAionTech: This is not an automatic change, this is me adding myself as a contributor.

SirAionTech: Why is it not correct? It looks fine in the README: 
image

partner_firstname/static/description/index.html
SirAionTech: This is an automatic change done by **pre-commit** installed locally when I commit, since it ensures coding conventions I'd rather leave it installed; do I really have to change this in order to have the PR merged?

SirAionTech:
> Hello @SirAionTech,
> 
> I would only add to the current review comments that the commit message `[FIX] partner_firstname: Multiple creation with "name"` could probably be more descriptive, like `[FIX] partner_firstname: Avoid duplicate partner creation with "name"` (or something different that points the attention to the issue being fixed). Then the commit description would clarify the details.

Hi,
> [FIX] partner_firstname: Avoid duplicate partner creation with "name"

Not sure what I would be avoiding here, could you please clarify that?
Right now the commit message states that it is `[FIX]`ing `Multiple creation with "name"` because when trying to create multiple partners with only the `name` field, an exception is raised, so it needs `[FIX]`ing; I added the exception message being raised in the commit description, does it look better now?

SirAionTech (commented):
Thanks for the review!

SirAionTech:
> Missing commit description.

Added, see also https://github.com/OCA/partner-contact/pull/1742#issuecomment-2071660402.

partner_firstname/static/description/index.html
simahawk: @SirAionTech that's more for func ppl :wink: 
As a techie I would recommend setting this env var: `SKIP=oca-gen-addon-readme`

simahawk (approved):
LG

payen000:
Hello @SirAionTech,

> Not sure what I would be avoiding here, could you please clarify that?

Of course!

From reviewing the partner's create method, it seems that previous to your fix all partners would be created on each iteration (as opposed to only one partner per iteration, as your change correctly does). I mean, if I sent a list of 3 partner value dicts, I would end up with 9 partners instead of the intended 3, effectively duplicating the partners on each iteration, no?

I thought at first that your commit message would be focused on that idea, but I now see that the focus of your commit message is in preventing the error that you describe in the commit message, so I think you can ignore my suggestion :)

SirAionTech:
> From reviewing the partner's create method, it seems that previous to your fix all partners would be created on each iteration (as opposed to only one partner per iteration, as your change correctly does). I mean, if I sent a list of 3 partner value dicts, I would end up with 9 partners instead of the intended 3, effectively duplicating the partners on each iteration, no?
>
> I thought at first that your commit message would be focused on that idea, but I now see that the focus of your commit message is in preventing the error that you describe in the commit message, so I think you can ignore my suggestion :)

I see, I have clarified what I am fixing in the PR description too.

partner_firstname/models/res_partner.py
luisg123v: Both ways work, but the old one is deprecated and it's internally "translated" to the new one, so better use the new one directly.

Saran440 (approved):
Functional Test :+1: 

This PR solved with case import user too.

Step to test:
1. Import `res.user` with 2 users (more than 1) with firstname, lastname, login
Selection_002

2. Check partner, it created partner duplicate. if import 10 users, it will create 1 user = 10 partner
Selection_003

moylop260 (approved):
👍

PR #1750, opened 2024-04-29
[17.0][MIG] partner_contact_type_end_user

Deriman-Alonso:
Standart Migration.


CRogos (commented):
Could you fix the commit naming and merge the yellow marked according to the migration guide?
image

image

sanderlienaerts (changes requested):
Hi there! Two comments. Pre-commit is failing. And also please rebase instead of adding a merge commit to prevent merge commits with base `17.0` branch. The two last commits should not be present.

PR #1548, opened 2023-06-27
[16.0] add partner_no_address_type

JuMiSanAr:
New module to remove the address type from the report address in case the partner name is not defined.

NOTE: This PR depends on this other one from Odoo:
- https://github.com/odoo/odoo/pull/126451

Do not merge until that one is merged as well.

PR #1752, opened 2024-05-06
[17.0][MIG] partner_no_address_type: Migration to 17.0

nguyenminhchien:


mmequignon (commented):
aren't we losing the git history here ?

nguyenminhchien:
> aren't we losing the git history here ?

i don't think so. this module was added to 16.0 by this https://github.com/OCA/partner-contact/pull/1548 which had only 1 commit. but it had not been merged yet.

mmequignon:
> > aren't we losing the git history here ?
> 
> i don't think so. this module was added to 16.0 by this #1548 which had only 1 commit. but it had not been merged yet.

ah perfect, thanks for the clarification

partner_no_address_type/models/res_partner.py
```diff
@@ -0,0 +1,35 @@
...
+        displayed_types = self._complete_name_displayed_types
+        type_description = dict(self._fields["type"]._description_selection(self.env))
+        # These filters extracted from def _get_complete_name(self)
+        partners = self.filtered(
+            lambda p: not p.name
+            and p.type in displayed_types
+            and (self.company_name or self.parent_id)
+        )
+        for partner in partners:
+            name_by_address_type = f", {type_description[self.type]}"
```
mmequignon: be careful here, self possibly contains multiple records.
```suggestion
            name_by_address_type = f", {type_description[partner.type]}"
```

partner_no_address_type/models/res_partner.py
```diff
@@ -0,0 +1,35 @@
...
+        if not self._context.get("no_address_type"):
+            return
+        displayed_types = self._complete_name_displayed_types
+        type_description = dict(self._fields["type"]._description_selection(self.env))
+        # These filters extracted from def _get_complete_name(self)
+        partners = self.filtered(
+            lambda p: not p.name
+            and p.type in displayed_types
+            and (self.company_name or self.parent_id)
+        )
```
mmequignon: same here, self is a recordset.
Also if I'm not mistaking, p is not a valid variable name anymore.
```suggestion
        partners = self.filtered(
            lambda partner: not partner.name
            and partner.type in displayed_types
            and (partner.company_name or partner.parent_id)
        )
```

partner_no_address_type/models/res_partner.py
```diff
@@ -0,0 +1,35 @@
...
+    def _compute_display_name(self):
+        super()._compute_display_name()
+        if not self._context.get("no_address_type"):
+            return
+        displayed_types = self._complete_name_displayed_types
+        type_description = dict(self._fields["type"]._description_selection(self.env))
+        # These filters extracted from def _get_complete_name(self)
+        partners = self.filtered(
+            lambda p: not p.name
+            and p.type in displayed_types
```
mmequignon: I guess this will only be remove partners where type isn't set
```suggestion
and not p.type
```

nguyenminhchien: Hello, These filters extracted from def _get_complete_name(self)
image

We don't need to re-compute the display_name if the record does not meet these filters.

partner_no_address_type/models/res_partner.py
nguyenminhchien: My bad, i just fixed it. please have a look.

mmequignon (approved):

partner_no_address_type/models/res_partner.py
```diff
@@ -0,0 +1,35 @@
...
+        partners = self.filtered(
+            lambda partner: not partner.name
+            and partner.type in displayed_types
+            and (partner.company_name or partner.parent_id)
+        )
+        for partner in partners:
+            name_by_address_type = f", {type_description[partner.type]}"
+            partner.display_name = (partner.display_name).replace(
+                name_by_address_type, ""
+            )
```
grindtildeath: To avoid having to rewrite everything, you should instead have migrated the extra hook that was not accepted in Odoo yet. I did it there: https://github.com/odoo/odoo/pull/185166

grindtildeath (commented):
Can you please merge https://github.com/nguyenminhchien/partner-contact/pull/4 ?

partner_no_address_type/tests/test_res_partner.py
```diff
@@ -0,0 +1,38 @@
...
+
+    def test_get_name(self):
+        # Partner with name.
+        name = self.child_partner_name.with_context(no_address_type=True).display_name
+        self.assertEqual(name, "Test Company Name, Test Partner Name")
+        # Partner without a name.
+        name = self.child_partner_no_name.with_context(
+            no_address_type=True
+        ).display_name
+        self.assertEqual(name, "Test Company Name")
```
grindtildeath, line 38: BTW: This is actually testing the context key we must have in Odoo, not the application of context key when rendering report, so I'm not sure it makes any sense to have that here.

nguyenminhchien: Let's keep the test until https://github.com/odoo/odoo/pull/185166 is merged . Otherwise, we don't know whether the module works or not.

nguyenminhchien:
Hi @grindtildeath , i merged your PR. On the other hand, i added your PR https://github.com/odoo/odoo/pull/185166 to test-requirement.txt. But it didn't work (test failed). Do you have any idea?

grindtildeath:
@nguyenminhchien Thank you. However, after reading answer from Odoo dev on my PR I did just test this behaviour on runbot and I cannot reproduce it.
So I guess this module is not needed anymore (although I could not even reproduce on 16.0).

grindtildeath:
@nguyenminhchien Actually we don't need this module, as it was another one that was breaking Odoo core code, cf https://github.com/OCA/partner-contact/pull/1882

PR #1763, opened 2024-05-21
[16.0][IMP] partner_fax: reformat fax numbers

alexis-via:
Remove placeholder='Fax ...' on partner form view
Add fax as optional="hide" in partner tree view
Add unaccent=False on fax field definition, as for mobile and phone in v16 On partner form view, position the fax field after the mobile field (it was its position until odoo remove the fax field in v11)


alexis-via:
All green. Ready to merge

rvalyi (approved):

PR #1765, opened 2024-05-22
[16.0][IMP] partner_deduplicate_acl: Test compatibility with website.visitor records

victoralmau:
Test compatibility with `website.visitor` records

`ERROR prod odoo.addons.partner_deduplicate_acl.tests.test_partner_deduplicate_acl: ERROR: TestPartnerDeduplicateAcl.test_same_email_restriction`
```
odoo.exceptions.AccessError: You are not allowed to access 'Website Visitor' (website.visitor) records.
This operation is allowed for the following groups:
	- Administration/Settings
	- Sales/User: Own Documents Only
	- Website/Editor and Designer

```
@Tecnativa

### DIFF
--- a/partner_deduplicate_acl/tests/test_partner_deduplicate_acl.py
+++ b/partner_deduplicate_acl/tests/test_partner_deduplicate_acl.py
@@ -2,13 +2,16 @@
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
 
 from odoo import exceptions
-from odoo.tests import common
+from odoo.tests import new_test_user
+
+from odoo.addons.base.tests.common import BaseCommon
 
 
-class TestPartnerDeduplicateAcl(common.TransactionCase):
-    def setUp(self):
-        super().setUp()
-        self.partner_1 = self.env["res.partner"].create(
+class TestPartnerDeduplicateAcl(BaseCommon):
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        cls.partner_1 = cls.env["res.partner"].create(
             {
                 "name": "Partner 1",
                 "email": "partner1@example.org",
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Please indicate at least the stack trace of the error.

victoralmau:
Updated in PR description

pedrobaeza:
Shouldn't this be done with sudo instead? In fact, one of the permissions of this module does this IIRC

partner_deduplicate_acl/wizards/partner_merge.py
```diff
@@ -13,8 +13,11 @@ def _merge(self, partner_ids, dst_partner=None, extra_checks=True):
         """Pass extra_checks=False if we have the extra group for avoiding
         the checks.
         """
+        _self = self
```
pedrobaeza: No need of the extra variable. You can store it in `self` and don't change `super`. Anyway, I commented about the sudo just because there should be a permission that does this, but I think it shouldn't be applied for everything.

victoralmau:
Finally I have applied a `sudo()` in the tests to avoid the error in some cases and added a comment to explain it.

Do you think this would be the right way to go in this case?

partner_deduplicate_acl/tests/test_partner_deduplicate_acl.py
```diff
@@ -40,7 +38,12 @@ def test_same_email_restriction(self):
         with self.assertRaises(exceptions.UserError):
             self.wizard.action_merge()
         self.user.groups_id = [
-            (4, self.env.ref("partner_deduplicate_acl.group_unrestricted").id)
+            (4, self.env.ref("partner_deduplicate_acl.group_unrestricted").id),
+            # Add extra group to avoid error with website.visitor records
```
pedrobaeza: You haven't removed this.

pedrobaeza (commented):
OK to put that workaround

pedrobaeza (approved):
/ocabot merge patch

pedrobaeza:
@victoralmau now it's red.

victoralmau:
The tests will be solved with https://github.com/OCA/partner-contact/pull/1770

PR #1768, opened 2024-05-27
[16.0][MIG] base_location_nuts

edlopen:
Migration to 16.0 using the new European ShowVoc service.

@moduon MT-1655

https://www.loom.com/share/d890f3b8cc4f4864a7c658685ca7b67d?sid=b1afe6dc-2d43-4aa6-b0ed-3f9eb1b17e41

@yajo @EmilioPascual @Shide @fcvalgar @rafaelbn review this PR when you can. Thank you!

The tests are failing but the errors came from another module (partner_category_security) which I'm currently looking for fixing it up.

PS: Apologies, English isn't my first language.


base_location_nuts/tests/__init__.py
```diff
@@ -1,3 +1,3 @@
 # FIXME: temporarily disable tests while issue below is not solved.
 #  https://github.com/OCA/partner-contact/issues/1557
```
yajo: Should these comments be removed?

base_location_nuts/wizard/nuts_import_view.xml
```diff
@@ -37,12 +37,12 @@
...
         action="nuts_import_action"
         id="nuts_import_menu"
-        name="Import NUTS 2013"
+        name="Import NUTS 2024"
         parent="contacts.menu_localisation"
         sequence="45"
     />
     <record id="config_wizard_nuts" model="ir.actions.todo">
-        <field name="name">Import NUTS 2013 from RAMON</field>
+        <field name="name">Import NUTS 2024 from </field>
```
yajo: issue: from... ShowVoc?

base_location_nuts/migrations/16.0.1.0.0/pre-migration.py
```diff
@@ -0,0 +1,21 @@
...
+from odoo import SUPERUSER_ID, api
+
+_logger = logging.getLogger(__name__)
+
+
+def migrate(cr, version):
+    # Create or update NUTs
+    env = api.Environment(cr, SUPERUSER_ID, {})
+    try:
+        env["nuts.import"].import_update_partner_nuts()
```
yajo: issue: AFAICS there are no changes related to our data model, right? Only to the way we import that from official sources.

Then, is it really necessary to force a NUTS update while upgrading?
- Not: remove migration script.
- Yes: you cannot load a model you're declaring in the pre-migration stage, only in post-migration.

yajo (changes requested):
I'll take a look at the failing module tests, but in the mean time I reviewed your code. Thanks!

base_location_nuts/migrations/16.0.1.0.0/pre-migration.py
edlopen: Well, the last RAMON file used in this module is from 2013. I came across several differences between the quantity of nuts loaded because UK leaving the EU for instance. This update action should be declared in the configuration steps instead?

base_location_nuts/wizard/nuts_import_view.xml
edlopen: Fixed!

base_location_nuts/tests/__init__.py
edlopen: Indeed!

base_location_nuts/migrations/16.0.1.0.0/pre-migration.py
yajo: That would be better.

In any case, when RAMON was working, data could still be very outdated since 2013 (it's been 11 years). So it doesn't seem like a migration problem, but something that can be done later.

yajo:
Fix for the flaky tests in https://github.com/OCA/partner-contact/pull/1770.

fcvalgar (approved):
Functinoal review, 

LGTM. thank you @edlopen 👏👏👏

image

PR #1771, opened 2024-05-29
[17.0][MIG] partner_contact_department: Migration to 17.0

manuelregidor:
Standard migration

T-6098


manuelregidor:
@ValentinVinagre @luis-ron

partner_contact_department/models/res_partner.py
```diff
@@ -0,0 +1,28 @@
...
+
+    department_id = fields.Many2one("res.partner.department", "Department")
+
+
+class ResPartnerDepartment(models.Model):
+    _name = "res.partner.department"
+    _order = "parent_path"
+    _parent_order = "name"
+    _parent_store = True
+    _description = "Department"
```
ValentinVinagre, line 19: If I'm not mistaken, the attribute "_parent_name = "parent_id"" would be missing

partner_contact_department/models/res_partner.py
```diff
@@ -0,0 +1,28 @@
...
+
+    name = fields.Char(required=True, translate=True)
+    parent_id = fields.Many2one(
+        "res.partner.department", "Parent department", ondelete="restrict"
+    )
+    child_ids = fields.One2many(
+        "res.partner.department", "parent_id", "Child departments"
+    )
+    parent_path = fields.Char(index=True, unaccent=False)
```
ValentinVinagre, line 28: it would be necessary to review the recursion
https://github.com/odoo/odoo/blob/2a08ba3ac193965e3394bb4e75d7e3abd39e68d3/addons/product/models/product_category.py#L46

ValentinVinagre (changes requested):
Some changes

partner_contact_department/models/res_partner.py
manuelregidor: I made functional tests and recursions seem to be under control.

ValentinVinagre: But the tests are only executed when Odoo is started so that it can execute them, but this verification is NEVER performed until this is the case and it can cause problems. If the system has a revision we can take advantage of it.

ValentinVinagre (approved):
LGTM 👍🏻

luis-ron (approved):
Functional review, LGTM 👍🏻

ValentinVinagre:
@HaraldPanten

HaraldPanten (approved):

PR #1694, opened 2024-03-10
[14.0][ADD] partner_address_ok

norlinhenrik:

PR #1782, opened 2024-06-16
[16.0][ADD] partner_address_ok

norlinhenrik:


legalsylvain (commented):
Hi @norlinhenrik. What is an adress "OK" ?

norlinhenrik:
> Hi @norlinhenrik. What is an adress "OK" ?

A valid postal address.
When we send physical mail to a mailing list (contact tag), we need to filter out those who don't have a valid address.

legalsylvain:
> A valid postal address.  When we send physical mail to a mailing list (contact tag), we need to filter out those who don't have a valid address.

Thanks ! 
- Maybe such text could be added in the "help" field ? 
- By default, it should be False, don't you think ?

Deriman-Alonso:
Hi, as @legalsylvain already said, it should be False by default.
Also, could you check those tests?.
The rest looks just fine 👍.

norlinhenrik:
On the PR for 14.0 I set [default=True](https://github.com/OCA/partner-contact/pull/1694/commits/0fe6b4109c2b73f7b8fde8bd8d7fed85c4b1fc10).
In this PR for 16.0 it is removed. Then by default the boolean value is False.

norlinhenrik:
I set default=False to be explicit.
I also set a help description.

partner_address_ok/tests/__init__.py
```diff
@@ -0,0 +1,2 @@
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+from . import test_partner_contact_id
```
Deriman-Alonso: Hi, I don't think this is the correct test you're trying to import, please fix this.

Deriman-Alonso (changes requested):

partner_address_ok/tests/__init__.py
norlinhenrik: Thank you for finding this error. I have fixed it now.

partner_address_ok/models/res_partner.py
```diff
@@ -0,0 +1,15 @@
+# Copyright 2024 Henrik Norlin
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import fields, models
+
+
+class ResPartner(models.Model):
+    _inherit = "res.partner"
+
+    address_ok = fields.Boolean(
```
leemannd, line 10: I'm a bit perplexed.

It adds a boolean fields without adding any logic such as testing the validity of the address. We have something that exists in -> https://github.com/odoo/odoo/blob/17.0/addons/snailmail/models/snailmail_letter.py#L453

It could be achieved with tags on partners.
I would have prefered a field naming like `is_valid_postal_address` which is self explanatory.

NB: I will not block this PR. Feel free to take into account these points

norlinhenrik: Thank you @leemannd for very good feedback!

leemannd: NB: another solution that was used in 12.0 was to add it into the type -> https://github.com/OCA/partner-contact/tree/12.0/partner_postal_address
But it may not fulfill your need.

PR #1785, opened 2024-06-19
[16.0][ADD] partner_property: New module

victoralmau:
New module

Please @pedrobaeza can you review it?

@Tecnativa TT49480


victoralmau:
Ping @pedrobaeza

partner_property/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1 @@
+This module add the use of properties in partners
```
pedrobaeza:
```suggestion
This module adds the use of properties in partners.
```

partner_property/readme/USAGE.rst
```diff
@@ -0,0 +1,4 @@
+#. Go to `Contacts` and create or edit any record.
+#. The “Add a Property” option is available (just above the tabs).
+
+The properties defined are per company.
```
pedrobaeza:
```suggestion
#. Properties are defined per company. If the contact has a company set, they 
   are fetched from it. If not, current environment company properties are
   displayed.
```

pedrobaeza (approved):
Tested on runboat and working. Just 2 minor comments.

carolinafernandez-tecnativa:
If i go to Contacts and open Azure i got this error message:

image

victoralmau:
Thanks, fixed, can you check it again?

partner_property/models/res_company.py
```diff
@@ -0,0 +1,29 @@
...
+
+    partner_properties_definition = fields.PropertiesDefinition(
+        string="Partner Properties",
+        compute="_compute_partner_properties_definition",
+        inverse="_inverse_partner_properties_definition",
+    )
+
+    @api.depends_context("company")
+    def _compute_partner_properties_definition(self):
+        ICP = self.env["ir.config_parameter"]
```
pedrobaeza, line 18: One thing, why don't you store this on the company itself? As it's right now, the properties are shared across all companies.

victoralmau: Sorry, I thought we had already discussed that internally.

The option to make the partner_properties_definition field of res.company store was discarded because:

1- Set properties and values for a record in company 1 > OK.
ejemplo-1

2- Set properties and values for a record in company 2 > KO
ejemplo-2

In point 2 the values have **not** been saved, and when changing to company 1 the values have **disappeared**!
Another discarded option is to use the field **company_dependet=True** (odoo core has not taken into account this use case).

Having clarified all this, the idea is that the properties are defined at a "_global_" level (of all companies) and can be used in the context of any company.

pedrobaeza: We have to keep current approach due to ORM limitations.

carolinafernandez-tecnativa (approved):

PR #1787, opened 2024-06-19
[17.0][MIG] animal: Migration to 17.0

santiagordz:

PR #1788, opened 2024-06-19
[17.0][MIG] animal_owner: Migration to 17.0

santiagordz:
@max3903 can you check this please?

Depends on #1787


animal_owner/static/description/icon.png
max3903, line 1: @santiagordz Please use the same icon from the `animal` module.

max3903 (changes requested):
Icon to change

PR #1790, opened 2024-06-22
[14.0] partner_helper: add strict mode that raise an error if address is too long

sebastienbeau:
Add an optional strict mode, so you can raise an error if the address can not be splitted correctly because it's too long

@bealdav @florian-dacosta

### DIFF
--- a/partner_helper/models/partner.py
+++ b/partner_helper/models/partner.py
@@ -2,11 +2,12 @@
 # Author: Sébastien BEAU <sebastien.beau@akretion.com>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
-from odoo import models
+from odoo import _, models
+from odoo.exceptions import UserError
 
 
-def split_char(char, output_number, size):
-    words = char.split(" ")
+def split_char(char, output_number, size, strict=False):
+    words = char.strip().split(" ")
     result = []
     word = words.pop(0)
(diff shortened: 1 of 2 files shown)


partner_helper/models/partner.py
```diff
@@ -2,11 +2,12 @@
...
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
-from odoo import models
+from odoo import _, models
+from odoo.exceptions import UserError
 
 
-def split_char(char, output_number, size):
-    words = char.split(" ")
+def split_char(char, output_number, size, strict):
```
florian-dacosta: Isn't it better to set strict=False in signature in case some modules use this method directly ?

florian-dacosta (approved):
LGTM (minor  comment)
Please port this to last version of the module

partner_helper/models/partner.py
bealdav: yes, it avoid to break things

bealdav:
The name of this module is now https://github.com/OCA/partner-contact/tree/16.0/partner_address_split

partner_helper/models/partner.py
sebastienbeau: done

bealdav (approved):

dreispt (approved):

PR #1791, opened 2024-06-24
partner_helper: add strict mode that raise an error if address is too long

sebastienbeau:
Port of PR: https://github.com/OCA/partner-contact/pull/1790
@bealdav @florian-dacosta


florian-dacosta:
@sebastienbeau Can you check the failing tests?

PR #1799, opened 2024-07-04
[16.0][Feat] Added store location type contact

wouitmil:
Screenshot 2024-07-04 at 11 20 13


partner_store/models/res_partner.py
```diff
@@ -0,0 +1,16 @@
+import logging
```
yvaucher, line 3: Missing copyright

yvaucher (changes requested):
Copyright is missing. Please fix this.

partner_store/models/res_partner.py
wouitmil: Done

yvaucher (approved):

StephaneMangin (approved):
Nice !

partner_store/README.rst
```diff
@@ -0,0 +1,64 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
```
gurneyalex: This file is supposed to e generated from documents in the readme/ subdirectory but this directory is missing. 

Additionally the content of the file is suspicious. I see no "geospatial support" in the module, for instance.

wouitmil: Better like this?

PR #1803, opened 2024-07-05
[17.0] [MIG] partner_identification_eori

Deriman-Alonso:
Standart migration


flotho (changes requested):
Hi,

Thanks for the PR.
Functionnal review on Runboat and I have some comments.
AFAIK, it shouldn't be possible to have 2 EORIs per partner
image

Duplicate EORI should raise a more explicit message :
image

Hope it could help

Deriman-Alonso:
Hi @flotho, TY for the review, but I feel that a change to only allow 1 EORI would be in a FIX or IMP, not in the migration itself, as this happens in other versions.

PR #1811, opened 2024-07-09
[16.0][IMP] partner_property: Define different properties for Individiaul or Company partners

victoralmau:
Define different properties for Individiaul or Company partners

Please @pedrobaeza and @carolinafernandez-tecnativa can you review it?

@Tecnativa TT49480

PR #1813, opened 2024-07-09
[16.0][OU-FIX] partner_property: Fix migration script (create properly partner_property.properties_definition_person)

victoralmau:
Fix migration script (create `properly partner_property.properties_definition_person`)

Related to https://github.com/OCA/partner-contact/pull/1811

@Tecnativa TT49480

### DIFF
--- a/partner_property/migrations/16.0.1.1.0/pre-migration.py
+++ b/partner_property/migrations/16.0.1.1.0/pre-migration.py
@@ -15,7 +15,7 @@ def migrate(env, version):
         WHERE key = 'partner_property.properties_definition'
         """,
     )
-    icp = env["ir.config_parameter"].get_param(
-        "partner_property.properties_definition_company"
+    icp = env["ir.config_parameter"].search(
+        [("key", "=", "partner_property.properties_definition_company")]
     )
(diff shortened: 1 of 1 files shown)


partner_property/migrations/16.0.1.1.0/pre-migration.py
```diff
@@ -15,7 +15,23 @@ def migrate(env, version):
...
+            write_uid,
+            'partner_property.properties_definition_person',
+            value,
+            create_date,
+            write_date
+            FROM ir_config_parameter
+            WHERE key = 'partner_property.properties_definition_company'
+        """,
     )
-    icp.copy({"key": "partner_property.properties_definition_person"})
```
pedrobaeza, line 21: Why not doing the copy before the renaming by SQL?

victoralmau: Because the `get_param()` method returns the value, not the record.

pedrobaeza: You can do a search as well.

pedrobaeza (approved):

PR #1829, opened 2024-08-01
[16.0][FIX] translate partner name order config values

benwillig:


partner_firstname/i18n/fr.po
SirAionTech, line 1: `.po` files should not be edited in PRs
> Pull requests should never directly modify ``.po`` files, because this could produce merge conflicts with pushes coming from weblate.

(from https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#11translations)

I think this also applies to `.pot` files.

partner_firstname/models/base_config_settings.py
```diff
@@ -25,9 +25,9 @@ class ResConfigSettings(models.TransientModel):
 
     def _partner_names_order_selection(self):
         return [
-            ("last_first", "Lastname Firstname"),
-            ("last_first_comma", "Lastname, Firstname"),
-            ("first_last", "Firstname Lastname"),
+            ("last_first", _("Lastname Firstname")),
+            ("last_first_comma", _("Lastname, Firstname")),
+            ("first_last", _("Firstname Lastname")),
```
SirAionTech, line 30: Just adding a bit of technical background: when the `selection` attribute of a field requires to call a method, the labels in its result are not translated:
```py
        if isinstance(selection, str) or callable(selection):
            return determine(selection, env[self.model_name])

        # translate selection labels
        if env.lang:
            return env['ir.model.fields'].get_field_selection(self.model_name, self.name)
        else:
            return selection
```
(from https://github.com/odoo/odoo/blob/38ec36e9b5dde0c02a4ff47714b1056e0db76812/odoo/fields.py#L2701-L2708)

SirAionTech (changes requested):
Thanks for the PR!
Please mention the changed module in the commit message so that it follows the guidelines
> commit messages should be self explanatory (long enough) including the name of the module that has been changed

(from https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#commit-message)

A test would be nice too

benwillig:
Removed translations and updated commit message.

> A test would be nice too

I don't think I should test this

yvaucher (commented):
LGTM

PR #1833, opened 2024-08-22
[16.0][ADD] partner_search_alias

AungKoKoLin1997:
This module allows users to set non-official names for partners in the search_alias field, improving search usability by enabling searches using alternative names.

@qrtl QT4793


partner_search_alias/views/res_partner_views.xml
```diff
@@ -0,0 +1,23 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <record model="ir.ui.view" id="view_partner_tree_alias">
+        <field name="name">res.partner.tree</field>
+        <field name="model">res.partner</field>
+        <field name="inherit_id" ref="base.view_partner_tree" />
+        <field name="arch" type="xml">
+            <field name="translated_display_name" position="after">
+                <field name="search_alias" />
```
yostashiro:
```suggestion
                <field name="search_alias" optional="hide" />
```

partner_search_alias/views/res_partner_views.xml
```diff
@@ -0,0 +1,23 @@
...
+            </field>
+        </field>
+    </record>
+    <record model="ir.ui.view" id="view_partner_form_alias">
+        <field name="name">res.partner.form</field>
+        <field name="model">res.partner</field>
+        <field name="inherit_id" ref="base.view_partner_form" />
+        <field name="arch" type="xml">
+            <field name="vat" position="after">
```
yostashiro: Let's move the field somewhere less obtrusive.
```suggestion
            <field name="ref" position="after">
```

yostashiro (commented):
@AungKoKoLin1997 Can you please also extend the partner search view to consider `search_alias` for the name field?

Please also make it clear in the readme file where in the form view users can find the added field.

partner_search_alias/views/res_partner_views.xml
AungKoKoLin1997: Done!

partner_search_alias/views/res_partner_views.xml
AungKoKoLin1997: I updated the code.

AungKoKoLin1997:
Hi community,
I am having trouble adding my new field `search_alias` to `_name_search()`. The challenging part is that I need to combine it with my new domain `('search_alias', operator, name)` using an `OR` condition along with Odoo's standard `_rec_names_search`.
https://github.com/odoo/odoo/blob/16.0/odoo/models.py#L1633-L1640

With my current design, there might be issues if another module also extends `_name_search()` and calls `super()`, as my new domain will not be combined with the `_rec_names_search` fields using an `OR` condition. I believe overriding `_rec_names_search` is not a good solution, and I haven't seen any modules that override it. I welcome any suggestions.

partner_search_alias/models/res_partner.py
```diff
@@ -0,0 +1,19 @@
...
+    search_alias = fields.Char(help="Enter the name that is also used for name search.")
+
+    @api.model
+    def _name_search(
+        self, name="", args=None, operator="ilike", limit=100, name_get_uid=None
+    ):
+        args = list(args or [])
+        if name:
+            args += ["|", ("search_alias", operator, name)]
+        return super()._name_search(name, args, operator, limit, name_get_uid)
```
yostashiro: @AungKoKoLin1997 Please try this.
```suggestion
    @property
    def _rec_names_search(self):
        return list(set(super()._rec_names_search + ["search_alias"]))
```

AungKoKoLin1997: It works. Thank you.

yostashiro (approved):
Code review.

kanda999 (approved):
Code and functional review.

partner_search_alias/models/res_partner.py
```diff
@@ -12,3 +14,19 @@ class ResPartner(models.Model):
...
+        res = super().get_view(view_id, view_type, **options)
+        if view_type == "search":
+            xml = etree.XML(res["arch"])
+            name_field = xml.xpath("//field[@name='name']")
+            if name_field:
+                name = name_field[0]
+                filter_domain = name.get("filter_domain", "[]")
+                filter_domain = filter_domain.replace(
+                    "[", "['|', ('search_alias', 'ilike', self), "
+                )
```
yostashiro, line 29: Shouldn't we be using `expression.OR()` instead?

AungKoKoLin1997: To use` expression.OR()`, the XML string must be converted to a Python list using `safe_eval`. However, `safe_eval` fails if the string includes unsupported expressions like `self`. Seem like we have to replace `self` with some values and I think we can't do it. So, I prefer using current approach. How do you think?

yostashiro: I think we can leave the code as is on second thought. Thanks for checking.

kanda999 (approved):

yostashiro (approved):
LGTM.

NL66278 (approved):
:+1: LGTM

PR #1835, opened 2024-08-26
[MIG][16.0] partner_industry_parent

len-foss:


partner_industry_parent/models/res_partner.py
```diff
@@ -0,0 +1,54 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import SUPERUSER_ID, _, api, models
+
+
+class Partner(models.Model):
+    _inherit = "res.partner"
+
+    @api.model
+    def _search(
```
gdgellatly, line 11: This is just a minor point, I know you didn't write yourself, and purely opinion. I much prefer these overrides where only args is amended to use the idiom 

```python
def _search(args, **kwargs):
   # modify args code
  return super()._search(args, **kwargs)
```
It is loads shorter, more resistant to api changes, and cleaner for people to read as they know only args is being affected.

Then super()

partner_industry_parent/models/res_partner.py
```diff
@@ -0,0 +1,54 @@
...
+            limit=limit,
+            order=order,
+            count=count,
+            access_rights_uid=access_rights_uid,
+        )
+
+    @api.model
+    def view_header_get(self, view_id, view_type):
+        if self._context.get("industry_id"):
+            return _(
```
gdgellatly, line 40: Is this right? I don't really know translations but I thought the rule was `_(<format_string>), args` not `_(<format_string>, args)`

gdgellatly (commented):
Just 1 concern to check really. The translation bit, but I could well be wrong not my specialty. Static review. Otherwise LGTM.

partner_industry_parent/models/res_partner.py
rousseldenis: @gdgellatly Yes, Odoo introduced args to translation function in v16 I think. So, this is the correct way of doing it.

partner_industry_parent/models/res_partner.py
```diff
@@ -0,0 +1,54 @@
...
+            offset=offset,
+            limit=limit,
+            order=order,
+            count=count,
+            access_rights_uid=access_rights_uid,
+        )
+
+    @api.model
+    def view_header_get(self, view_id, view_type):
+        if self._context.get("industry_id"):
```
rousseldenis: `_context`is deprecated. Use `self.env.context.get()`instead.

partner_industry_parent/models/res_partner.py
```diff
@@ -0,0 +1,54 @@
...
+    def _search(
+        self,
+        args,
+        offset=0,
+        limit=None,
+        order=None,
+        count=False,
+        access_rights_uid=None,
+    ):
+        if self._context.get("search_default_industry_id"):
```
rousseldenis: `_context` is deprecated. Use `self.env.context.get()` instead.

rousseldenis (changes requested):

partner_industry_parent/models/res_partner.py
len-foss: All right, fixed :-)

len-foss:
@rousseldenis any chances this can be merged before the release of 20.0?

rousseldenis:
> @rousseldenis any chances this can be merged before the release of 20.0?

@len-foss Thanks for your concern. As you know, OCA is made of contributors and their time. If you don't have answer from individuals, don't hesitate to ping PSC members with @OCA/product-maintainers

rousseldenis (approved):

len-foss:
@rousseldenis 

> OCA is made of contributors and their time.

Doesn't it go both ways? Migrations that are stuck in limbo make the whole community lose time, including the many people who just give up on trying to upstream their changes, as they'll probably have to merge into standard OCA once some bug inevitably arises. 

Anyway, good luck with your projects.

NL66278 (approved):
:+1: LGTM Altough would have been nice to indeed use the `*args **kwargs` override when nothing done with the arguments.

PR #1837, opened 2024-08-29
[FIX] [16.0] partner_contact_address_default child layout

gdgellatly:
Before this commit, strange English populated a single column on the child contact layout.

After this commit, we span the 4 columns and clean up the text.

PR #1838, opened 2024-09-01
[16.0][FIX] partner_contact_address_default: Tidy up view

### LINKED ISSUE #1659: partner_contact_address_default  odoo16
the module works but in the drop-down menu where you can choose the shipping address it is limited to 99 lines. If a company has more than 99 locations, those beyond 99 are not exposed

Vittorio

gdgellatly:
Before this commit the partner default text on children in partner form only filled one column causing a long vertical text. Also, the selection widget has a limit of 99 entries and is unnecessary, so switched to a many2one with no create and open options. Resolves #1659.

Refer closed PR #1837

### DIFF
--- a/partner_contact_address_default/views/res_partner_views.xml
+++ b/partner_contact_address_default/views/res_partner_views.xml
@@ -10,19 +10,19 @@
                         <field
                             name="partner_delivery_id"
                             domain="[('id', 'child_of', commercial_partner_id), ('type', '=', 'delivery')]"
-                            widget="selection"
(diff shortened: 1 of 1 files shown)


gdgellatly:
@pedrobaeza IRT to the commit message format. I was looking for some solutions so I stop doing this and you stop wasting time with this. I ran into https://commitizen-tools.github.io/commitizen/ which would seem to allow us to codify our standards. Might be worth a look.

pedrobaeza (approved):
Fast-tracking it

/ocabot merge minor

PR #1840, opened 2024-09-03
[FIX] base_location_geonames_import: avoid code duplicates in country states

Tisho99:
Currently, some countries like Bulgaria, Latvia and Sweden can't be imported with this module due to a duplicity in the code of their states on the data files.

With this fix, the states with the same code of another already parsed are discarded, and countries can be imported in spite of having an error of one of their state codes

T-5767

### DIFF
--- a/base_location_geonames_import/tests/test_base_location_geonames_import.py
+++ b/base_location_geonames_import/tests/test_base_location_geonames_import.py
@@ -177,6 +177,42 @@ class TestBaseLocationGeonamesImport(common.TransactionCase):
                 "Incorrect state for {} {}".format(state_name, zip_code),
             )
 
+    def test_import_duplicated_state_code(self):
+        country = self.env.ref("base.bg")
+        self.wizard.country_ids = [(6, 0, country.ids)]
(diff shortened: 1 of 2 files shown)


Tisho99:
@HaraldPanten @ValentinVinagre @luis-ron

luis-ron (approved):
Functional review: LGTM 👍🏻

pedrobaeza (changes requested):
I don't think we should simply put all the cities in the first state with that code. We should fix it in the source (Geonames): https://forum.geonames.org/, or patch the file on the fly for not having duplicated data for that countries.

manuelregidor (approved):
LGTM

Tisho99:
Thanks for the reviews @pedrobaeza 

I have tried to edit the files directly in geonames, but with no success.

Then i have patched the files on the fly and imported them with other method.

I have also updated the #1841 PR, but it still needs the adition of a unit test and more work.

Do you think it is worth merging any of these PRs?

pedrobaeza:
Well, if they are correctly done according the comments, why not?

PR #1843, opened 2024-09-10
[16.0][ADD] partner_contact_restrict_depth

AlexPForgeFlow:
This module adds the possibility of defining the maximum depth of the hierarchy of contacts


partner_contact_depth/data/system_parameter.xml
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo noupdate="1">
+    <record id="contacts_max_depth" model="ir.config_parameter">
```
GuillemCForgeFlow: IMHO, it should not be possible to update the configuration parameter if there is any partner which will then trigger the ValidationError.
WDYT?

partner_contact_depth/models/res_partner.py
```diff
@@ -0,0 +1,38 @@
...
+    def _compute_contact_depth(self):
+        for partner in self:
+            depth = 0
+            parent = partner.parent_id
+            while parent:
+                depth += 1
+                parent = parent.parent_id
+            partner.depth = depth
+
+    depth = fields.Integer(compute="_compute_contact_depth", store=True)
```
GuillemCForgeFlow: I believe this may cause inconsistencies in the long run for specific scenarios. Notice that the `depth` of a partner will only be updated if the `parent_id` changes.

Then, if we have a hierarchy relation such as:
- Contact **A** with no parent.
- Contact **B**, being **A** its parent.
- Contact **C**, being **B** its parent.

If for some reason, the Contact **A** gets assigned a Parent now, the hierarchy will change and the `depth` will only be changed for Contact **A**, when it should be done for **B** and **C** as well.

Therefore, I'd say it should perhaps be a non-computed field which would get updated every time the parent_id changes for itself, and for its childs. Or, perhaps adding something like:

```suggestion
    @api.depends("parent_id")
    def _compute_contact_depth(self):
        for partner in self:
            depth = 0
            parent = partner.parent_id
            while parent:
                depth += 1
                parent = parent.parent_id
            partner.depth = depth
            partner.child_ids._compute_contact_depth()

    depth = fields.Integer(compute="_compute_contact_depth", store=True)
```

I guess this would work fine, but of course I would add some tests to reinforce the desired functionality.

partner_contact_depth/models/res_partner.py
```diff
@@ -0,0 +1,38 @@
...
+                raise ValidationError(
+                    _(
+                        "You have reached the maximum depth of contacts. "
+                        "You cannot create childs for this contact."
+                    )
+                )
+
+    def write(self, vals):
+        self.check_maximum_contact_depth()
+        return super().write(vals)
```
GuillemCForgeFlow: Is this really needed? I guess that, as the `depth` will be a stored field, updated every time there is a relevant change, we should already be covered by that.
WDYT?

GuillemCForgeFlow (changes requested):

partner_contact_depth/models/res_partner.py
AlexPForgeFlow: I refactored the depth computation in order to handle with reparenting. Thanks for your example :)

partner_contact_depth/models/res_partner.py
AlexPForgeFlow: The aim is to avoid contacts modification with depth higher than the maximum defined.

partner_contact_depth/models/res_partner.py
```diff
@@ -0,0 +1,62 @@
+from odoo import _, api, fields, models
+from odoo.exceptions import ValidationError
+
+
+class ResPartner(models.Model):
+    _inherit = "res.partner"
+
+    def _get_partners_with_no_parent(self):
```
AaronHForgeFlow: does this code do the same as this other?

```suggestion
    def _get_partners_with_no_parent(self):
        return self.search([('parent_id', '=', False)])
```

partner_contact_depth/models/res_partner.py
```diff
@@ -0,0 +1,62 @@
...
+
+    @api.depends("parent_id")
+    def _compute_contact_depth(self):
+        partners_with_no_parent = self._get_partners_with_no_parent()
+        partners_depth_dict = {}
+        partners_with_no_parent._get_partners_depth_dict(partners_depth_dict)
+        for depth, partners in partners_depth_dict.items():
+            partners.depth = depth
+
+    depth = fields.Integer(compute="_compute_contact_depth", store=True)
```
AaronHForgeFlow: Perhaps we can add an index, WDYT?

GuillemCForgeFlow (approved):
LGTM 🤖 👍🏿

AaronHForgeFlow (commented):
It works ok, some minor questions.

I will also change the name, partner_contact_depth seems to be calculating the depth, but it is doing more. Maybe partner_contact_maximum_depth or partner_contact_restrict_depth WDYT?

partner_contact_depth/models/res_partner.py
AlexPForgeFlow: The aim it's just to recompute hierarchy related to partners in self, otherwise we will be recomputing depth for all partners.

partner_contact_depth/models/res_partner.py
AlexPForgeFlow: Yes, sounds a good idea

AlexPForgeFlow:
> It works ok, some minor questions.
> 
> I will also change the name, partner_contact_depth seems to be calculating the depth, but it is doing more. Maybe partner_contact_maximum_depth or partner_contact_restrict_depth WDYT?

@AaronHForgeFlow Thanks for the suggestion, I have renamed it as partner_contact_restrict_depth

AaronHForgeFlow (approved):
LGTM!

PR #1845, opened 2024-09-19
[17.0][IMP] base_partner_company_group: restrict choosing active partner itself on the form

yankinmax:
As for now, you can set yourself as a group company which doesn't make any sense.


gurneyalex (approved):

CRogos:
Why do you think that this does not make sense?

We have a company A and this company A owns Company B & C.

My expectation would be, that all companies (and contacts in these companies) have "Company A" as common Company Group.

vvrossem:
Hello @OCA/crm-sales-marketing-maintainers, could you please reopen this PR and set the "no stale" label?

PR #1846, opened 2024-09-24
[17.0][ADD]: partner_contact_block_invoice_address

imlopes:


partner_contact_block_invoice_address/models/res_partner.py
```diff
@@ -0,0 +1,71 @@
...
+
+from odoo import _, api, models
+from odoo.exceptions import UserError
+
+
+class ResPartner(models.Model):
+    _inherit = "res.partner"
+
+    def _check_setting_block_invoice_address(self):
+        partner_block_invoice_address = (
```
yankinmax:
```suggestion
        partner_block_invoice_address = bool(
```

partner_contact_block_invoice_address/models/res_partner.py
```diff
@@ -0,0 +1,71 @@
...
+            self.env["ir.config_parameter"]
+            .sudo()
+            .get_param("partner_block_invoice_address")
+        )
+        if partner_block_invoice_address == "True":
```
yankinmax:
```suggestion
        if not partner_block_invoice_address:
            return False
        else:
```

partner_contact_block_invoice_address/models/res_partner.py
```diff
@@ -0,0 +1,71 @@
...
+                .get_param("partner_block_invoice_address_default_type")
+            )
+            return partner_block_invoice_address_default_type
+        return False
+
+    @api.onchange("type")
+    def _onchange_contact_type(self):
+        default_type = self._check_setting_block_invoice_address()
+        if default_type and self.type == "invoice":
+            self.type = default_type
```
yankinmax: I would first check is `self.type == "invoice"` then call `_check_setting_block_invoice_address`

partner_contact_block_invoice_address/models/res_partner.py
```diff
@@ -0,0 +1,71 @@
...
+                    "message": _(
+                        "You cannot set a contact as an invoice address. "
+                        "Check the settings."
+                    ),
+                },
+            }
+        return {}
+
+    def write(self, vals):
+        if (
```
yankinmax: same here

gurneyalex (commented):
We don't need a module for this, a record rule is enough.

yankinmax:
Thanks for the work done @imlopes 
We'll handle it with a record rule

PR #1849, opened 2024-09-26
[16.0][FIX] partner_manual_rank: Compatibility with Odoo default

lmignon:
The introduction of default values for the fields 'is_customer' and 'is_supplier' breaks the compatibility when the create method on partner is called with once of the fields 'supplier_rank', 'customer_rank'. To restore this compatibility, we ovveride the create method to initialize the 'is_customer' and 'is_supplier' fields into the given dict if 'supplier_rank' or 'customer_rank' are passed to the method.

### DIFF
--- a/partner_manual_rank/models/res_partner.py
+++ b/partner_manual_rank/models/res_partner.py
@@ -72,3 +72,12 @@ class ResPartner(models.Model):
 
     def _default_is_supplier(self):
         return self.env.context.get("res_partner_search_mode") == "supplier"
+
+    @api.model_create_multi
+    def create(self, vals_list):
+        for vals in vals_list:
+            if "customer_rank" in vals and "is_customer" not in vals:
+                vals["is_customer"] = bool(vals["customer_rank"])
+            if "supplier_rank" in vals and "is_supplier" not in vals:
+                vals["is_supplier"] = bool(vals["supplier_rank"])
+        return super().create(vals_list)
--- a/partner_manual_rank/tests/test_res_partner.py
+++ b/partner_manual_rank/tests/test_res_partner.py
@@ -8,6 +8,30 @@ class TestResPartner(common.TransactionCase):
         self.partner = self.env["res.partner"].create({"name": "Microsoft Corporation"})
         self.partner_2 = self.env["res.partner"].create({"name": "Apple Inc."})
 
+    def test_create_with_rank(self):
+        p = self.env["res.partner"].create({"name": "customer", "customer_rank": 1})
+        self.assertTrue(p.is_customer)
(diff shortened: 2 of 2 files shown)


lmignon:
ping @ypapouin @sbejaoui @rousseldenis

sbejaoui (approved):

partner_manual_rank/models/res_partner.py
```diff
@@ -72,3 +72,12 @@ def _default_is_customer(self):
 
     def _default_is_supplier(self):
         return self.env.context.get("res_partner_search_mode") == "supplier"
+
+    @api.model_create_multi
+    def create(self, vals_list):
+        for vals in vals_list:
+            if "customer_rank" in vals and "is_customer" not in vals:
```
luisg123v, line 79: Shouldn't this be covered by the compute method?

luisg123v (changes requested):

partner_manual_rank/models/res_partner.py
lmignon: > Shouldn't this be covered by the compute method?

No because you defined a default value for the `is_customer` and `is_supplier` field. Therefore, if these fields are not part of the dict, the orm call the write method with the default values . As a side effect, the inverse methods defined for these fields are called and overwrite the value provided for the 'x_rank' fields if they are equal to '1'. If you remove the override if the create method and launch the tests you'll see the problem.

partner_manual_rank/models/res_partner.py
```diff
@@ -72,3 +72,12 @@ def _default_is_customer(self):
...
+                vals["is_customer"] = bool(vals["customer_rank"])
+            if "supplier_rank" in vals and "is_supplier" not in vals:
+                vals["is_supplier"] = bool(vals["supplier_rank"])
+        return super().create(vals_list)
```
ypapouin, line 83: Since it is too much tricky, I propose to entirely delete the `_default_*` callbacks I previously wrote and copy/paste the odoo logic from `odoo/addons/account/models/partner.py`.
That way default value is set at only one place and computed functions will be called accordingly.

```
    @api.model_create_multi
    def create(self, vals_list):
        search_partner_mode = self.env.context.get('res_partner_search_mode')
        is_customer = search_partner_mode == 'customer'
        is_supplier = search_partner_mode == 'supplier'
        if search_partner_mode:
            for vals in vals_list:
                if is_customer and 'is_customer' not in vals:
                    vals['is_customer'] = 1
                elif is_supplier and 'is_customer' not in vals:
                    vals['is_supplier'] = 1
```

ypapouin (changes requested):

PR #1854, opened 2024-10-03
[17.0][IMP] partner_email_check: improve configuration view

chaule97:
The `res_config_settings_view_form` in Odoo 17 has changed its xml style, and the `view_general_configuration` in this module was written using Odoo 16 styling. So we need to update it accordingly.


sebalix (approved):
Thank you for this update

PR #1855, opened 2024-10-04
[16.0] adds res.team model

wouitmil:


pedrobaeza:
Isn't this something similar to partner_group? Maybe it can fit your needs.

wouitmil:
> Isn't this something similar to partner_group? Maybe it can fit your needs.

Hello @pedrobaeza, Not exactly. As with this proposition, you can define one or more roles ([res.partner.role](https://github.com/OCA/partner-contact/tree/16.0/partner_contact_role)) for each member of a team

pedrobaeza:
OK then

PR #1857, opened 2024-10-07
[17.0][FW] partner_pricelist_search: port from 14.0

xaviedoanhduy:


rvalyi (approved):
I checked the 3 PRs and really there is nothing left to port

PR #1869, opened 2024-10-15
[17.0][ADD] partner_title_active

yankinmax:
Adds an ``Active`` field to the partner title model to be able to deactivate titles without deleting them


partner_title_active/README.rst
```diff
@@ -0,0 +1,78 @@
+====================
+Partner title active
```
jcoux, line 2: That file should be split for unit files (like `DESCRIPTION.md` or `CONTRIBUTORS.md`), shouldn't?

imlopes (approved):
LGTM

partner_title_active/README.rst
yankinmax: yep, but this should be generated by the pre-commit
here it wasn't the case, probably it will be after merge

partner_title_active/tests/test_partner_title.py
```diff
@@ -0,0 +1,17 @@
...
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        cls.Partner = cls.env["res.partner"]
+
+    def test_01_partner_title(self):
+        partner = self.Partner.create({"name": "A Partner"})
+        self.assertTrue(partner.active, "Partner should be active by default")
+        partner.action_archive()
+        self.assertFalse(partner.active, "Partner should be inactive after archiving")
```
vvrossem: This doesn't test whether the `title` is active or not :confused:

vvrossem (commented):
thank you for this contribution to the OCA :pray: 

TestPartnerTitle should be adapted as it doesn't assert if the title is active or not

partner_title_active/tests/test_partner_title.py
yankinmax: Why? Can you describe pls a test scenario you think is more suitable?

vvrossem: The partner that is created doesn't have a **title** set.
The scenario might be: 
1. create a partner with a title (res.partner.title)
2. test the title is active
3. archive the title, assert it is not active

yankinmax: ok, I'll adapt, thanks

yankinmax: ahaha @vvrossem , I've just get that I was archiving the partner and not the title :smiling_face_with_tear: 
good catch, shame on me!

Abimael1321:
LGTM

Abimael1321 (approved):

vvrossem:
Hello ~~@ivantodorovich~~ @simahawk , do you think this PR could be merged? :crossed_fingers:

ivantodorovich (approved):

partner_title_active/__manifest__.py
```diff
@@ -0,0 +1,19 @@
+# Copyright 2024 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+{
+    "name": "Partner Title Active",
+    "version": "17.0.1.0.0",
+    "author": "Odoo Community Association (OCA), Camptocamp",
```
simahawk: OCA should be at the end

partner_title_active/tests/test_partner_title.py
```diff
@@ -0,0 +1,23 @@
+# Copyright 2024 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo.tests.common import TransactionCase
+
+
+class TestPartnerTitle(TransactionCase):
```
simahawk: always use BaseCommon class unless you need to test mail tracking ;)

simahawk (changes requested):

simahawk (approved):

PR #1875, opened 2024-10-16
[16.0][IMP] base_partner_sequence: Allows to use unique (and distinct) generated references across all partners

rousseldenis:
In some configurations, company strategy will use a unique (and distinct) reference for all its partners. A configuration parameter will allow to enable that behavior.


AEstLo (approved):
LGTM (only code review)

lmignon (approved):
LGTM. Thank you for this improvement.

base_partner_sequence/models/res_company.py
```diff
@@ -11,3 +11,31 @@ class ResCompany(models.Model):
         string="Partner Reference Readonly",
         help="If marked, the Reference in partners will not be editable.",
     )
+    partner_generated_reference_unique = fields.Boolean(
+        compute="_compute_partner_generated_reference_unique",
+        inverse="_inverse_partner_generated_reference_unique",
+        groups="base.group_system",
+        help="This is a technical field that allows to get this from config parameter",
```
StefanRijnhart, line 18: Why is the config parameter reflected on the company? It looks misleadingly as if it can be a different value between different companies.

PR #1885, opened 2024-10-31
[FIX] Remove UserWarning: Field res.partner.industry.display_name should be declared with recursive=True

pfranck:
Add display_name with recursive=True


pfranck:
I think that should also be backported to 16.0

rousseldenis (approved):
Ok, seems fair as soon as we introduced the display_name computation that depends on parent.

flotho (approved):
LGTM

PR #1891, opened 2024-11-12
[17.0][ADD] partner_type_base

CRogos:
refactor https://github.com/OCA/partner-contact/pull/1890 in own module.

We are currently unhappy that invoice or shipping address has a firstname. Shipping addresses sometimes need an department name or other additional information, which are not person/individual like.

Therefore I would like to make the module more flexible and add an overloadable attribute is_individual, where you can control whether firstname should be visible or not. Modules could overload the _compute_contact_type method to do a different type definition of address types.

I also changed the visibility of the UI elements to is_individual and is_address_readonly instead of the address type.
The is_address_readonly we use to add another address type "contact_address" which is an individual, but also has an editiable address like address type "other". We use "other" for office addresses of a company and need to filter out these contacts for not adding these contacts to e.g. newsletters. Nevertheless we need the possibility to add individual addresses e.g. for people who are mostly working in the homeoffice and don't need their mails to the office.#

What do you think of these changes?


CRogos:
Hi @xaviedoanhduy, @rousseldenis, @houssine78
I address you, because you did the review of partner_firstname v18 
What do you think about this change?

partner_firstname/__manifest__.py
```diff
@@ -18,7 +18,7 @@
     "maintainer": "Camptocamp, Acsone",
     "category": "Extra Tools",
     "website": "https://github.com/OCA/partner-contact",
-    "depends": ["base_setup"],
+    "depends": ["base_setup", "partner_type_base"],
```
houssine78, line 21: doing it in a current version could lead to issues for database already using partner_firstname as the dependency won't install itself after pulling this new version and updating the database. People updating the code repository are not always aware of those changes. And debugging it could lead to loss of time and frustration. 

Maybe setting partner_type_base as auto_install True could avoid the issue.

what do you think ?

partner_firstname/views/res_partner.xml
```diff
@@ -4,25 +4,12 @@
         <field name="inherit_id" ref="base.view_partner_simple_form" />
         <field name="arch" type="xml">
             <xpath expr="//field[@id='individual']" position="attributes">
-                <attribute name="invisible">is_company</attribute>
-                <attribute name="readonly">not is_company</attribute>
-                <attribute name="required">type == 'contact' and is_company</attribute>
-            </xpath>
-            <xpath expr="//field[@id='company']" position="attributes">
-                <attribute name="invisible">is_company == False</attribute>
```
houssine78, line 12: those constraints don't apply anymore on the company field from what I understand

partner_firstname/views/res_partner.xml
```diff
@@ -32,54 +19,39 @@
         <field name="inherit_id" ref="base.view_partner_form" />
         <field name="arch" type="xml">
             <xpath expr="//field[@id='individual']" position="attributes">
-                <attribute name="invisible">is_company</attribute>
-                <attribute name="readonly">not is_company</attribute>
-                <attribute name="required">type == 'contact' and is_company</attribute>
-            </xpath>
-            <xpath expr="//field[@id='company']" position="attributes">
-                <attribute name="invisible">not is_company</attribute>
```
houssine78, line 40: same here

partner_firstname/__manifest__.py
CRogos: I was not aware of that. I thought if we change the version number, also the dependencies will be rechecked and installed. 

auto_install would be possible, but this also leads to unnecessary install databases which doesn't need it, correct?

I'll test it first, but an upgrade script might be the solution?

houssine78: yes auto_install True will install it on all databases. upgrade script could be a solution.

houssine78: to be precise this is the update all module that will lead to the issue. updating partner_firstname should install the module indeed. But you need to be aware of such change in the dependency.

the issue was partially fixed with the PR linked to this issue https://github.com/odoo/odoo/issues/72703 and merged on v14. It seems not having evolved in 16.0. I don't know for 17.0 and 18.0.

pedrobaeza: As already said, you shouldn't change dependencies on a stable module that is already merged in this version.

CRogos: @houssine78 thank you for the link. Very interesting.

@pedrobaeza I understand your point. The thing is, that a lot of modules are migrate 1:1 very fast under a lot of time pressure and these kind of changes will never be done/time for. 

The migration to v18 is already completed, so the next possible time is the migration to v19. So I need to develop now a feature, which we can start use in 2 years? I don't think this is working eighter. 

I totally agree, that this kind of change should be done by odoo in odoo directly with a version change, but I don't think this will ever be applied.

If this PR will not be merged, it is like it is. But I apricate your input and I'll fix the found issues soon.

pedrobaeza: You can do this change in v18, but not on a version which is already one year old.

CRogos: OK, I'll prepare a new PR for v18 and make backport for myself. I can live with this very well.

CRogos:
As discussed I've continued on the v18 branch: https://github.com/OCA/partner-contact/pull/1905

PR #1895, opened 2024-11-15
[18.0][MIG] partner_multi_relation: Migration to 18.0

BT-mchechlacz:


partner_multi_relation/models/res_partner_relation.py
```diff
@@ -92,19 +92,21 @@ def _check_partner(self, side):
...
             if (ptype == "c" and not partner.is_company) or (
                 ptype == "p" and partner.is_company
             ):
                 raise ValidationError(
-                    _("The %s partner is not applicable for this " "relation type.")
-                    % side
+                    self.env._(
+                        f"The {side} partner is not applicable for this "
+                        f"relation type."
+                    )
```
BT-cjimeno: I think the best way to format parameters into a translated string is using:

_("Example %s", params), I am not sure how it works with an fstring or if it works at all.

This also appears in other places, and if the string inside the _() changes, then I think the .po file should also be updated accordingly.

Example:
_("Warning for %s", self.name)

partner_multi_relation/models/res_partner_relation.py
```diff
@@ -92,19 +92,21 @@ def _check_partner(self, side):
...
                 )
-            category = getattr(record.type_id, "partner_category_%s" % side)
+            category = getattr(record.type_id, f"partner_category_{side}")
             if category and category.id not in partner.category_id.ids:
                 raise ValidationError(
-                    _(
+                    self.env._(
                         "The {partner} partner does not have category {category}."
                     ).format(partner=side, category=category.name)
```
BT-cjimeno: _() param formatting.

partner_multi_relation/models/res_partner_relation_all.py
```diff
@@ -234,14 +237,15 @@ def check_partner_domain(partner, partner_domain, side):
             partner_model = self.env["res.partner"]
             partners_found = partner_model.search(test_domain, limit=1)
             if not partners_found:
-                warning["title"] = _("Error!")
+                warning["title"] = self.env._("Error!")
                 if partner:
                     warning["message"] = (
-                        _("%s partner incompatible with relation type.") % side.title()
+                        self.env._("%s partner incompatible with relation type.")
+                        % side.title()
```
BT-cjimeno: _() param formatting.

partner_multi_relation/models/res_partner_relation_all.py
```diff
@@ -234,14 +237,15 @@ def check_partner_domain(partner, partner_domain, side):
...
                     )
                 else:
                     warning["message"] = (
-                        _("No %s partner available for relation type.") % side
+                        self.env._("No %s partner available for relation type.") % side
```
BT-cjimeno: _() param formatting.

partner_multi_relation/models/res_partner_relation_all.py
```diff
@@ -444,7 +450,7 @@ def create(self, vals_list):
             type_selection = self._get_type_selection_from_vals(vals)
             if not type_selection:  # Should not happen
                 raise ValidationError(
-                    _("No relation type specified in vals: %s.") % vals
+                    self.env._("No relation type specified in vals: %s.") % vals
```
BT-cjimeno: _() param formatting.

partner_multi_relation/models/res_partner_relation_type.py
```diff
@@ -169,7 +178,7 @@ def _check_no_existing_reflexive_relations(self):
             relations = relation_type._get_reflexive_relations()
             if relations:
                 raise ValidationError(
-                    _(
+                    self.env._(
```
BT-cjimeno, line 181: _() param formatting.

BT-cjimeno (approved):
LGTM, left some comments about _() param formatting.

partner_multi_relation/models/res_partner_relation.py
BT-mchechlacz: Translation work with fstrings, however this change was unnecessary here, as it would generate additional work to change it in translations. 
Regarding `_()` vs 'self.env._()` - the second option offers small performance improvements and is advised to use that while migrating the code to version 18
https://github.com/odoo/odoo/pull/174844

BT-cjimeno: Hi Marcin, sorry, maybe I didn't explain myself very well. I meant only that the proper way to parse args into a translated string, AFAIK, is by providing the args/kwargs to the _ method, it doesn't matter if it's for the old way:
"\_()" or the new "self.env.\_()".

Here is the doctstring of v18 self.env._ method:

"""Translate the term using current environment's language.

Usage:

```
self.env._("hello world")  # dynamically get module name
self.env._("hello %s", "test")
self.env._(LAZY_TRANSLATION)
```

:param source: String to translate or lazy translation
:param ...: args or kwargs for templating
:return: The transalted string
"""

self.env.\_("hello %s", "test")  <--- This is the one I was refering to, instead of doing
self.env.\_("hello %s") % "test" or .format, or an fstring... These will also work of course, but the \_ method already provides options for this, so we should use them instead IMO.

This is another example with kwargs:
self.env.\_("%(name)s's Timesheets and Planning Analysis", name=self.name)

BT-mchechlacz: My bad, I misunderstood you here. I agree with what you wrote and where it was easy fix not requiring changes in translation I went with your suggestions. Where it requires changes in translation files I didn't touch it to avoid problems with updating po files, which could easily lead to incorrect translations. I don't want to break something that is working :sweat_smile: 
Thank you for the great explanation.

BT-cjimeno (approved):
LGTM!

BT-dmoreno (approved):
👍🏻

partner_multi_relation/models/res_partner_relation.py
```diff
@@ -92,19 +92,20 @@ def _check_partner(self, side):
...
+                    self.env._(
+                        "The %s partner is not applicable for this relation type.", side
+                    )
                 )
-            category = getattr(record.type_id, "partner_category_%s" % side)
+            category = getattr(record.type_id, f"partner_category_{side}")
             if category and category.id not in partner.category_id.ids:
                 raise ValidationError(
-                    _(
+                    self.env._(
```
rousseldenis, line 108: Please use translation function arguments instead of format.

partner_multi_relation/tests/test_partner_relation_common.py
```diff
@@ -69,15 +69,15 @@ def setUp(self):
 
     def _create_relation_type_selection(self, vals):
         """Create relation type and return this with selection types."""
-        assert "name" in vals, (
-            "Name missing in vals to create relation type. Vals: %s." % vals
-        )
-        assert "name" in vals, (
-            "Name_inverse missing in vals to create relation type. Vals: %s." % vals
-        )
+        assert (
```
rousseldenis, line 72: We are in an Odoo tests context, I would have used standard unitttest functions instead

partner_multi_relation/models/res_partner.py
```diff
@@ -0,0 +1,204 @@
...
+from odoo import api, exceptions, fields, models
+from odoo.osv.expression import FALSE_LEAF, OR, is_leaf
+
+
+class ResPartner(models.Model):
+    """Extend partner with relations and allow to search for relations
+    in various ways.
+    """
+
+    # pylint: disable=invalid-name
```
rousseldenis, line 16: Do those pylint disables still applicable ?

partner_multi_relation/models/res_partner.py
```diff
@@ -0,0 +1,204 @@
...
+            "like",
+            "not like",
+            "ilike",
+            "not ilike",
+            "in",
+            "not in",
+        )
+        if operator not in SUPPORTED_OPERATORS:
+            raise exceptions.ValidationError(
+                self.env._('Unsupported search operator "%s"') % operator
```
rousseldenis, line 78: Please use translation arguments instead.

partner_multi_relation/models/res_partner.py
```diff
@@ -0,0 +1,204 @@
...
+        :rtype: str
+        """
+        self.ensure_one()
+        return "c" if self.is_company else "p"
+
+    def action_view_relations(self):
+        for contact in self:
+            relation_model = self.env["res.partner.relation.all"]
+            relation_ids = relation_model.search(
+                [
```
rousseldenis, line 182: Putting this domain in action domain will avoid a search().

partner_multi_relation/models/res_partner.py
```diff
@@ -0,0 +1,204 @@
...
+                    "|",
+                    ("this_partner_id", "=", contact.id),
+                    ("other_partner_id", "=", contact.id),
+                ]
+            )
+            action = self.env["ir.actions.act_window"]._for_xml_id(
+                "partner_multi_relation.action_res_partner_relation_all"
+            )
+            action["domain"] = [("id", "in", relation_ids.ids)]
+            context = action.get("context", "{}").strip()[1:-1]
```
rousseldenis, line 192: I would have used `safe_eval` to get the context dict, then, update it and convert to string. Which avoid all strings operations

partner_multi_relation/models/res_partner_relation.py
```diff
@@ -0,0 +1,161 @@
...
+    )
+    date_start = fields.Date("Starting date")
+    date_end = fields.Date("Ending date")
+
+    @api.model_create_multi
+    def create(self, vals_list):
+        """Override create to correct values, before being stored."""
+        context = self.env.context
+        for vals in vals_list:
+            if "left_partner_id" not in vals and context.get("active_id"):
```
rousseldenis, line 51: get `active_id`before loop to avoid all `get()` calls

partner_multi_relation/models/res_partner_relation.py
```diff
@@ -0,0 +1,161 @@
...
+        self._check_partner("right")
+
+    def _check_partner(self, side):
+        """Check partner for required company or person, and for category
+
+        :param str side: left or right
+        :raises ValidationError: When constraint is violated
+        """
+        for record in self:
+            assert side in ["left", "right"]
```
rousseldenis, line 94: We are in an Odoo context. I would have raised a `ValidationError` instead.

partner_multi_relation/models/res_partner_relation.py
```diff
@@ -0,0 +1,161 @@
...
+    @api.constrains(
+        "left_partner_id", "type_id", "right_partner_id", "date_start", "date_end"
+    )
+    def _check_relation_uniqueness(self):
+        """Forbid multiple active relations of the same type between the same
+        partners
+
+        :raises ValidationError: When constraint is violated
+        """
+        # pylint: disable=no-member
```
rousseldenis, line 135: Same here.

partner_multi_relation/models/res_partner_relation_type.py
```diff
@@ -0,0 +1,268 @@
...
+
+        If a relation has a end date in the past, then it is inactive and
+        will not be modified.
+
+        :param relations: a recordset of relations (not necessarily all active)
+        """
+        today = fields.Date.today()
+        for relation in relations:
+            if relation.date_start and relation.date_start >= today:
+                relation.unlink()
```
rousseldenis, line 91: Please build a recordset to delete and do it once.

partner_multi_relation/models/res_partner_relation_type.py
```diff
@@ -0,0 +1,268 @@
...
+
+            elif not relation.date_end or relation.date_end > today:
+                relation.write({"date_end": today})
```
rousseldenis, line 94: Same for this as today is unique value.

partner_multi_relation/models/res_partner_relation_type.py
```diff
@@ -0,0 +1,268 @@
...
+            if invalid_relations:
+                if handling == "restrict":
+                    raise ValidationError(
+                        self.env._(
+                            "There are already relations not satisfying the"
+                            " conditions for partner type or category."
+                        )
+                    )
+                elif handling == "delete":
+                    invalid_relations.unlink()
```
rousseldenis, line 155: Same as above.

partner_multi_relation/models/res_partner_relation_type.py
```diff
@@ -0,0 +1,268 @@
...
+                else:
+                    self._end_active_relations(invalid_relations)
+
+    def _get_reflexive_relations(self):
+        """Get all reflexive relations for this relation type.
+
+        :return: a recordset of res.partner.relation.
+        """
+        self.env.cr.execute(
```
rousseldenis, line 164: I'm wondering if flush with fields should be done.

partner_multi_relation/models/res_partner_relation_type.py
```diff
@@ -0,0 +1,268 @@
...
+        for relation_type in self:
+            relations = relation_type._get_reflexive_relations()
+            if relations:
+                raise ValidationError(
+                    self.env._(
+                        "Reflexivity could not be disabled for the relation "
+                        "type {relation_type}. There are existing reflexive "
+                        "relations defined for the following partners: "
+                        "{partners}"
+                    ).format(
```
rousseldenis, line 186: Please use translation function arguments instead.

partner_multi_relation/models/res_partner_relation_type.py
```diff
@@ -0,0 +1,268 @@
...
+                        relation_type=relation_type.display_name,
+                        partners=relations.mapped("left_partner_id.display_name"),
+                    )
+                )
+
+    def _delete_existing_reflexive_relations(self):
+        """Delete existing reflexive relations for these relation types."""
+        for relation_type in self:
+            relations = relation_type._get_reflexive_relations()
+            relations.unlink()
```
rousseldenis, line 196: Same as above

partner_multi_relation/models/res_partner_relation_type.py
```diff
@@ -0,0 +1,268 @@
...
+        """Allow delete of relation type, even when connections exist.
+
+        Relations can be deleted if relation type allows it.
+        """
+        relation_model = self.env["res.partner.relation"]
+        for rec in self:
+            if rec.handle_invalid_onchange == "delete":
+                # Automatically delete relations, so existing relations
+                # do not prevent unlink of relation type:
+                relations = relation_model.search([("type_id", "=", rec.id)])
```
rousseldenis, line 266: I would have done one search with records in that case `[('type_id' , 'in', self.filtered(lambda rel_type: rel_type.handle_invalid_onchange == "delete").ids)]`

partner_multi_relation/tests/test_partner_relation_common.py
```diff
@@ -0,0 +1,107 @@
+# Copyright 2016 Therp BV
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+from odoo.tests import common
+
+
+class TestPartnerRelationCommon(common.TransactionCase):
+    def setUp(self):
```
rousseldenis, line 7: please convert this to classmethod for performances reasons.

rousseldenis (changes requested):
Code review

BT-dmoreno:
This PR is going to be replaced by this other one [here](https://github.com/OCA/partner-contact/pull/1933). The reason is that the account that was used to open this one is not going to be used anymore.

PR #1898, opened 2024-11-26
[16.0][MIG] partner_noncommercial: Migration to 16.0

ntsirintanis:


partner_noncommercial/readme/CONFIGURE.rst
```diff
@@ -0,0 +1,3 @@
+To configure this module, you need to:
+
+#. This module requires no special configuration.
```
legalsylvain: I think this file can be removed.

partner_noncommercial/views/res_partner.xml
```diff
@@ -0,0 +1,56 @@
...
+            <filter name="group_country" position="after">
+                <filter
+                    name="group_city"
+                    string="City"
+                    context="{'group_by': 'city'}"
+                />
+                <filter
+                    name="group_organisation"
+                    string="Organisation"
+                    context="{'group_by': 'commercial_partner_id'}"
```
thomaspaulb: There is already an Odoo default grouping by Company: 

```<filter name="group_company" string="Company" context="{'group_by': 'parent_id'}"/>```

Wouldn't it be better to just change the string of that one?

File: partner_noncommercial/views/res_partner.xml [hunk 3]
```diff
@@ -0,0 +1,56 @@
...
+                    name="domain"
+                >[('is_company', '=', False), ('type', '=', 'contact')]</attribute>
+            </filter>
+
+            <!-- Change string of Company filter to Organisations. -->
+            <filter name="type_company" position="attributes">
+                <attribute name="string">Organisations</attribute>
+            </filter>
+
+            <!-- Add extra filters. -->
```
thomaspaulb, line 28: Odoo also has a filter where you can type a company name and then search for it:

```<field name="parent_id" domain="[('is_company', '=', True)]" operator="child_of"/>```

Maybe we should also rename that one to Organisation.

**Review by thomaspaulb — CHANGES_REQUESTED** [review]

File: partner_noncommercial/views/res_partner.xml [hunk 4]
```diff
@@ -36,18 +41,18 @@
                 />
             </filter>
 
-            <!-- Add extra groupings. -->
+            <!-- Change string of Company group to Organisations. -->
+            <filter name="group_company" position="attributes">
+                <attribute name="string">Organisations</attribute>
```
thomaspaulb: singular form seems more appropriate here

partner_noncommercial/views/res_partner.xml
```diff
@@ -7,6 +7,11 @@
         <field name="mode">extension</field>
         <field type="xml" name="arch">
 
+            <!-- Change string of parent_id field to Organisations. -->
+            <field name="parent_id" position="attributes">
+                <attribute name="string">Organisations</attribute>
```
thomaspaulb: and here as well

PR #1903, opened 2024-11-27
[18.0] [MIG] partner_contact_birthplace: Migration to 18.0

Borruso:
Migration from 17.0 to 18.0

Issue: https://github.com/OCA/partner-contact/issues/1856


test-requirements.txt
```diff
@@ -1 +1,2 @@
 odoo_test_helper
+odoo-addon-partner_contact_personal_information_page @ git+https://github.com/OCA/partner-contact.git@refs/pull/1900/head#subdirectory=partner_contact_personal_information_page
```
houssine78: I'm curious about this directive. from what I understand it allows the test to be green by taking into account an unmerged dependency. Do you remove this commit before merging or the bot take care of it ?

Borruso: I must remove this commit before merging

rousseldenis: @houssine78 The process is described here: https://github.com/OCA/maintainer-tools/wiki/Use-temporary-reference%28s%29-to-another-pull-request%28s%29

Indeed, this is used in order to:

- show tests are passing
- be able to test the PR functionally with runboat.

partner_contact_birthplace/views/res_partner.xml
```diff
@@ -0,0 +1,19 @@
...
+    <record id="view_personal_information_birthplace" model="ir.ui.view">
+        <field name="name">Birthplace field</field>
+        <field name="model">res.partner</field>
+        <field
+            name="inherit_id"
+            ref="partner_contact_personal_information_page.personal_information"
+        />
+        <field name="arch" type="xml">
+            <xpath expr="//group[@name='personal_information_group']">
+                <field name="birth_city" />
```
rousseldenis: I'm wondering why this order?

Shouldn't it country first, then state and city ?

image

rousseldenis: @SirAionTech

SirAionTech: I probably copied the same order that is used in the partner address https://github.com/odoo/odoo/blob/29405bb41516a57b635c1b8cedbfa564d626b964/odoo/addons/base/views/res_partner_views.xml#L138-L141, but it can be changed indeed if you prefer another order.

Borruso:
@pedrobaeza can reopen PR?

partner_contact_birthplace/views/res_partner.xml
legalsylvain: > I'm wondering why this order?

well, that's the same as in the previous version. Nothing changed here.

As it is the same logic as in Odoo core, I think it's valid.

Could you update your review @rousseldenis ?

legalsylvain (approved):
Diff review.

LGTM.

Thanks.

legalsylvain:
superseded by https://github.com/OCA/partner-contact/pull/2098

NL66278 (approved):
:+1: LGTM

PR #1905, opened 2024-11-28
[18.0][ADD] partner_type_base extracted from partner_firstname

CRogos:
followup in v18 of https://github.com/OCA/partner-contact/pull/1891

We are currently unhappy that invoice or shipping address has a firstname. Shipping addresses sometimes need an department name or other additional information, which are not person/individual like.

Therefore I would like to make the module more flexible and add an overloadable attribute is_individual, where you can control whether firstname should be visible or not. Modules could overload the _compute_contact_type method to change the address type behavior for each type.

I also changed the visibility of the UI elements to is_individual and is_address_readonly instead of the address type.
The is_address_readonly we use to add another address type "contact_address" which is an individual, but also has an editable address like address type "other". We use "other" mainly for office addresses of a company and need to filter out these contacts for not adding these contacts to e.g. newsletters. Nevertheless we need the possibility to add individual addresses e.g. for people who are mostly working in the homeoffice and don't need their mails to the office.

What do you think of these changes?

PS: This module currently does not adapt the base_address_extended module views.
https://github.com/odoo/odoo/blob/18.0/addons/base_address_extended/views/base_address_extended.xml
And there are also some few localizations which have expressions like "type == 'contact'".


partner_type_base/__manifest__.py
```diff
@@ -0,0 +1,21 @@
...
+# Copyright 2015 Grupo ESOC (<http://www.grupoesoc.es>)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+{
+    "name": "Partner Address Types",
+    "summary": "Base implementation to improve the address type customization.",
+    "version": "18.0.1.0.0",
+    "author": "CRogos, Odoo Community Association (OCA)",
+    "license": "AGPL-3",
+    "maintainer": "glueckkanja AG",
```
houssine78: I think that you should put your github login here and the name of your company in author

houssine78 (approved):
LGTM

CRogos:
@houssine78 I haven't tested the migration script yet, but I've used this command it during the last upgrade. What do you think about this solution?

CRogos:
The upgrade script worked for me in odoo.sh, but i am not sure if we should add a dependency to the util tool in the manifest? In my container, the library was missing and therefore the upgrade failing.

partner_firstname/__manifest__.py
```diff
@@ -18,7 +18,7 @@
     "maintainer": "Camptocamp, Acsone",
```
rousseldenis: This is not correct. The correct key is `maintainers` and value should be a list of github logins.

rousseldenis: Change also the version number if you want your script running.

partner_firstname/migrations/18.0.3.1.0/pre-migration.py
```diff
@@ -0,0 +1,10 @@
+import logging
+
+from odoo.upgrade import util
+
+_logger = logging.getLogger(__name__)
+
+
+def migrate(cr, version):
+    _logger.info("Installing dependent module 'partner_type_base' by migration script.")
+    util.force_install_module(cr, "partner_type_base")
```
rousseldenis, line 14: Why do you need this? Odoo's ORM is in charge of that.

partner_type_base/tests/test_res_partner.py
```diff
@@ -0,0 +1,59 @@
+from odoo.tests.common import TransactionCase
+
+
+class TestResPartner(TransactionCase):
+    def setUp(self):
```
rousseldenis: Use classmethod instead.

partner_type_base/__manifest__.py
```diff
@@ -0,0 +1,21 @@
...
+{
+    "name": "Partner Address Types",
+    "summary": "Base implementation to improve the address type customization.",
+    "version": "18.0.1.0.0",
+    "author": "glueckkanja AG, Odoo Community Association (OCA)",
+    "license": "AGPL-3",
+    "maintainer": "CRogos",
+    "category": "Extra Tools",
+    "website": "https://github.com/OCA/partner-contact",
+    "depends": ["base_setup"],
```
rousseldenis, line 14: Why not `base` ?

partner_type_base/models/res_partner.py
```diff
@@ -0,0 +1,46 @@
...
+                "contact",
+                "other",
+            ]
+
+            partner.can_be_parent = partner.is_company
+            partner.can_be_child = not partner.is_company
+
+    @api.model
+    def _search_can_be_parent(self, operator, value):
+        return [("is_company", operator, value)]
```
rousseldenis, line 46: Is it real ? An individual can have a child partner as delivery address for instance.

rousseldenis (dismissed):

partner_firstname/migrations/18.0.3.1.0/pre-migration.py
CRogos: I thought that too, but @houssine78 informed me that there is a bug and from my testing I can confirm that the added dependency is not installed automatically. https://github.com/odoo/odoo/issues/72703

rousseldenis: Hasn't https://github.com/odoo/odoo/pull/72661 fixed that ?

partner_type_base/models/res_partner.py
CRogos: This is totally inconsistent in Odoo.

When you go to a contact, you can add a subcontact/address:
image

When you open this other address, the contact is set as parent: 
image

But when you search for the contact to select it in this view, this is not possible:
image

In my Opinion a big weakness of Odoo, because it is very unclear to the user how to use it.
After setting a parent, you can also switch the contact to "company" and the parent_id is kept. 
Basically this could be good feature to address hierarchies, but I am not sure if all Odoo modules can work with this. 

What we are doing is, that we use a 3 level hirache Company -> "Other Address" -> Contact to bundle all people of a certain Office address. Because commerical_partner_id looks upwards until the first company, this should work. But I am not sure if it is a Odoo recommended structure. 
To have it more consistent, I added the can_be_parent attribute to make it more flexible and we overload the computation and add "Other Address" to the types which can_be_parent = True in another module.

CRogos: I had also opened an Odoo ticket on this topic 9 Months ago, and this was the answer:

> Thank you for your patience. I have been trying to find information on whether this behavior is supported or not, and in fact a lot of feedback has been created in the past suggesting we need this feature (three-level hierarchy with contacts). The reply the development team has given us in all of these cases is the behavior is indeed possible, but not entirely supported. Odoo does not support this behavior because it is a very problematic data structure, with a lot of hidden complexities, especially on the accounting aspect. This reply has been the same for feedbacks created 3 to 1 year ago, and the team does admit that if demand for this becomes very high it will be reconsidered.
> 
> Therefore, I have created a feedback ticket to shed light on the fact that you, like other customers, do need this hierarchy for your daily business flows. Depending on whether you just need to display that a contact belongs to another, the workaround proposed is to create a Studio field for it. If however this information will be used in other business flows such as accounting flows, custom code will be necessary to satisfy the need.

partner_type_base/__manifest__.py
CRogos: Good question... I've copied it from "partner_firstname" module. 

There is a module in odoo/addons/base_setup which depends on "base" and "web". It also has some minor changes on the patner kanban view. So maybe a view order issue?

A general question: 
If partner_firstname has a dependency to partner_type_base and partner_type_base depends on base, but partner_firstname also references base in a view. Shall partner_firstname reference base explicit in the manifest or should it not add the base to the depends?

partner_firstname.depends = ["partner_type_base"] or ["base", "partner_type_base"] ???
when
partner_type_base.depends = ["base"]

partner_firstname/migrations/18.0.3.1.0/pre-migration.py
CRogos: not totally as sbidoul wrote at the end of the ticket I've linked:

`Half fixed to be precise.`

partner_firstname/__manifest__.py
CRogos: I thought this is done automatically with the correct merge command? But I agree the version number needs to be 18.0.1.2.0 after the merge. I can change this.

Regarding maintainer or maintainers, it is mixed everywhere and not used very often. Is there any documentation beside this where it is without the "s": https://www.odoo.com/documentation/18.0/developer/reference/backend/module.html
I tend to remove it totally.

CRogos:
@rousseldenis could you update your review and share your thoughts regarding adding the odoo.upgrade tool to the python dependencies or not?

CRogos:
> The upgrade script worked for me in odoo.sh, but i am not sure if we should add a dependency to the util tool in the manifest? In my container, the library was missing and therefore the upgrade failing.

How should we continue on this?

CRogos:
Does this PR has a chance or shell I retract it?

SamyBrahimi (approved):

CRogos:
@rousseldenis could you please update your review?

acsonefho (approved):

CRogos:
@pedrobaeza could you add your final thoughts on this change? We [discussed this change for 17.0](https://github.com/OCA/partner-contact/pull/1891) and you've said this would be to radical for an stable version so I've moved it to v18.

pedrobaeza:
@rousseldenis do you agree?

acsonefho:
@rousseldenis :pray:

CRogos:
I don't think @rousseldenis will give further feedback. Can we continue?

CRogos:
@rousseldenis could you have a look into this PR as well?

acsonefho:
@rousseldenis :pray:

partner_type_base/__manifest__.py
legalsylvain: partner_firstname.depends = ["partner_type_base"] is enought.

xavier-bouquiaux (approved):
LGTM

xavier-bouquiaux:
@rousseldenis

NL66278:
@CRogos Please resolve conflicts (probably caused by the updates to partner_firstname module).

partner_type_base/models/res_partner.py
NL66278: For complex hierarchies you can use the partner_multi_relation module, where for instance you can setup a company to company partner relation of this 'has local office / is local office for', then link the main company/office to the local offices. Then you can setup a company to person contact type of 'has local employee / works in office' and specify which persons are working in which office. No problem for a worker to work in multiple offices. Just an example of what you can do with partner_multi_relation.

NL66278 (approved):
👍 LGTM.

CRogos:
If there are no more concerns, can we merge this?

bealdav:
In my experience when this PR merged, partner_firstname pre migration script can't be executed in my production and let it out of order.

I temporarly solved my production pinning partner_firstname to previous commits and set partner_type_base to last commits and then install partner_type_base

Even if I set all to last recent commits it fails again.

I'm sure other people 'll be in the same issue than me


EDIT

I suppose that my problem is I have not https://github.com/odoo/upgrade-util in my project

Then

`from odoo.upgrade import util` can't be executed

I've a doubt on the way to install this lib.

If i install it only with pip, is it sufficient ?

partner_firstname/models/firstname_mixin.py
```diff
@@ -26,14 +26,24 @@ def _compute_firstname_lastname_required(self):
...
+            )
+            partner.lastname_required = (
+                partner.is_individual
+                and not partner.firstname
+                or required_fields
+                in [
+                    "lastname",
+                    "firstname_lastname",
+                ]
+            )
```
legalsylvain, line 46: wow... this change is breaking the design of this mixin.

is_individual is not defined in this mixin, only in res.partner.

could you fix it ?  
Thanks !

partner_firstname/migrations/18.0.3.1.0/pre-migration.py
```diff
@@ -0,0 +1,14 @@
...
+_logger = logging.getLogger(__name__)
+
+
+def migrate(cr, version):
+    _logger.info(
+        "Installing dependent module 'partner_type_base' "
+        "by migration script for version %s.",
+        version,
+    )
+    util.force_install_module(cr, "partner_type_base")
```
paradoxxxzero, line 14: Is `upgrade-util` a hard dependency now? Because we don’t use it and this breaks our install for a migration that seems useless now.

partner_firstname/models/firstname_mixin.py
```diff
@@ -26,14 +26,24 @@ def _compute_firstname_lastname_required(self):
...
-            ]
+            partner.firstname_required = (
+                partner.is_individual
+                and not partner.lastname
+                or required_fields
+                in [
+                    "firstname",
+                    "firstname_lastname",
+                ]
+            )
```
legalsylvain, line 37: you wrote `A and B or C`  But I guess you wanted to write `A and (B or C)`. Don't you think ?

legalsylvain:
This PR brings some problems.
@CRogos. Could you take a look on my remarks.

Other note : the current implementation of is_individual is not correct. the computation of `is_individual`. is : 
```

            partner.is_individual = not partner.is_company and partner.type in [
                "contact",
                "other",
            ]
```
But, we can not guess from partner type, if the partner is an individual or a company.

Also, I'm not confortable with the new design. The objective of this module (partner_firstname) is to handle first and last name). But now, it is forcing the installation of partner_type_base, that is changing some core feature, with the can_be_parent and can_be_child. I don't say that it is not a great new module, but some people doesn't want this new feature, that is changing base module.
Indeed, odoo core is quite weird, a propose some weird possibility with companies with parent companies. But that's not the role of partner_firstname module to fix that.

It should be done in an extra module. partner_firstname_partner_type_base for exemple.

thanks

legalsylvain:
Note : 

sorry for the late regarding my comments. I'm the maintainer of the module since a few monthes, but I have not been pinged in that PR, because it is more that one year old.

regards.

NL66278:
> This PR brings some problems. @CRogos. Could you take a look on my remarks.
> 
> Other note : the current implementation of is_individual is not correct. the computation of `is_individual`. is :
> 
> ```
> 
>             partner.is_individual = not partner.is_company and partner.type in [
>                 "contact",
>                 "other",
>             ]
> ```
> 
> But, we can not guess from partner type, if the partner is an individual or a company.

@legalsylvain I think the statement is correct. Basically it is saying: if a partner is a company, it is definitely not an individual. Then the partners that are not companies are divided by partner type. Delivery and invoice addresses are also not individuals. Contacts, that are not companies are individuals, as are contacts of type "other". That last one might be debatable, but assuming individual is remaining on the safe side. Or do I make a mistake in reasoning here?

partner_firstname/models/firstname_mixin.py
NL66278: @legalsylvain I think you are right. So should this be?:
```
        partner.firstname_required = (
            partner.is_individual
            and (
                   not partner.lastname
                   or required_fields in ["firstname", "firstname_lastname"]
            )
        )
```

legalsylvain:
> Delivery and invoice addresses are also not individuals

I don't understand why ? 
is the following use case forbidden or incorrect ? 

 **partner #1 :** 
- name : "My company"
- is_company : True
- parent_id : False
- type : contact (default value)

**partner #2 :** 
- name "John Do"
- First name : john
- Last Name : Do
- is_company : False
- parent_id : partner_#1
- type: invoice
- function : Accountant, in the "My Company" Company

something like that, with demo data : 
- open azur Interior
- update "contact" Colleen Diaz, and put "Invoice type"
- save 

image

Then create a sale order to azur interior. 
when generating the invoice, the invoice is created for the accountant, colleen diaz.


image

partner_type_base/models/res_partner.py
```diff
@@ -0,0 +1,46 @@
+# Copyright 2024 Christopher Rogos (<https://www.glueckkanja.com>)
+
+from odoo import api, fields, models
+
+
+class ResPartner(models.Model):
+    """Adds last name and first name; name becomes a stored function field."""
```
Hugo-Trentesaux, line 7: This docstring does not describes what the code does. It seems to be copied from `partner_firstname` and not updated afterwards.

NL66278:
> > Delivery and invoice addresses are also not individuals
> 
> I don't understand why ? is the following use case forbidden or incorrect ?
> 
> **partner #1 :**
> 
>     * name : "My company"
> 
>     * is_company : True
> 
>     * parent_id : False
> 
>     * type : contact (default value)
> 
> 
> **partner #2 :**
> 
>     * name "John Do"
> 
>     * First name : john
> 
>     * Last Name : Do
> 
>     * is_company : False
> 
>     * parent_id : partner_#1
> 
>     * type: invoice
> 
>     * function : Accountant, in the "My Company" Company

I think we can agree that "My Company" is not an individual. As for the accountant, I think this is not how invoice addresses should be used. John Do should be a contact for partner_#1, where he can have the function Accountant, in the "My Company" Company. Then there should be an invoice address, like My Company Finance Department, PO BOX 1234 AB, email finance@mycompany.example.com. But if an Odoo instance still wants to have personal names on functional addresses', they still could enter "John Do" in the single name field shown.

partner_firstname/migrations/18.0.3.1.0/pre-migration.py
```diff
@@ -0,0 +1,14 @@
+import logging
+
+from odoo.upgrade import util
```
NL66278, line 3: There is indeed no `util` in odoo.upgrade

legalsylvain:
> As for the accountant, I think this is not how invoice addresses should be used. 

Thanks for the clarification.
Well, I can understand your point of view, and I could be agree. in a certain measure, I think it's a matter of taste / partner design. 
The question is, do we want to force all people who installed the `partner_firstname` module to have that logic ? What about existing instances (18.0 is out since more than year), that have some "John DO" settings. 
I don't understand why 
- it is now mandatory
- it should be done in the partner_firstname core OCA modules, that is widely used

> Then there should be an invoice address, like My Company Finance Department, PO BOX 1234 AB, email [finance@mycompany.example.com](mailto:finance@mycompany.example.com).

Well, I understand what you say but sometimes, it is not possible. I mean, if my customer named "My Company" says me "_send the invoices to my accountant, John Do, john.do@mycompany.example.com_", and if there is no "finance@mycompany.example.com" email. How can I design this use case ?

partner_firstname/migrations/18.0.3.1.0/pre-migration.py
NL66278, line 14: Method force_install_odoo indeed exists nowhere in Odoo @CRogos Where did you find such a method?

CRogos: Is there another way, to trigger the installation of "partner_type_base"? To my knowledge, adding it only to the manifest does not work either.

CRogos: The force_install_odoo method is from https://github.com/odoo/upgrade-util/ and referenced in the odoo documentation.
https://www.odoo.com/documentation/18.0/de/developer/reference/upgrades/upgrade_utils.html

Instead of adding the dependency to this module, we could extract what is needed from the function and add this to the migration script?

https://github.com/odoo/upgrade-util/blob/7bfebaa88a862068a964162b9f0da58cb6ec95c3/src/util/modules.py#L603-L664

CRogos:
@legalsylvain I understand your points and agree that the situation is suboptimal.

The idea of **partner_type_base** is to be as transparent as possible, and give the possibility to easily customize the behavior as needed, without changing a lot of views. (which cannot be done, if multiple modules need to change the same field attribute)

Also the implementation, when to show firstname/lastname is not consistent and configurable. 

We e.g. added other types like "support" or "office" which are no individuals but also no companies and also shouldn't have a firstname, while office should have an address and can be the parent of another individuals, but support should not have an address and cannot be a parent... and there could be a lot other use cases. Both cases need the new attributes, but not necessary a dependency on partner_firstname. Therefore I think having the dependency in partner_firstname on partner_type_base is more flexible. 

But I agree the auto upgrade must not fail. 

Basically I also agree with the "stable" policy, but I've started this PR with version 17.0 where it was to late, so I've moved the code to 18.0 where it took 1.5 years to be processed. (unfortunately the mixin change was integrated between, and an error occured while rebasing) 

I suggest:
1) to improve the migration script 
2) fix the mixin

remi-filament:
I think that this PR should be reverted before causing too much damage...

Although the intention is understandable, I think the change is too big on such a widespread module ans adding contraint on res_partner model which are not name split related.

I also think that other modules may depend on it for changing the behaviour on specific cases but we should not add a new dependency to this module.

I was thinking that maybe we can have a function determining whether split name should occur (updating a is_split_name boolean field for instance). This function would return previous behaviour by default (= not is_company) and could be overwritten in other modules. Would that cover your use case @CRogos ?

partner_firstname/models/firstname_mixin.py
CRogos: @legalsylvain do you have an explanation for  `and not partner.firstname` in the lastname_required?

Before partner_type_base, the condision was:
```
not partner.firstname
                or required_fields
                in [
                    "lastname",
                    "firstname_lastname",
                ]
```
But as soon as the firstname is set, the lastname is not required anymore?
I would understand this condition if the required field would be "lastname_or_firstname", but if "lastname" is required a firstname should not remove this requirement?

CRogos:
@remi-filament @legalsylvain I think https://github.com/OCA/partner-contact/pull/2264 fixes both issues.

PR #1916, opened 2024-12-04
[17.0][FIX] base_location_geonames_import: Fixes for re-imports

pedrobaeza:
Forward-port of #1911

Before this commit, there were 2 possible errors re-importing ZIPs:

- If city names are changed on the import process (like capitalization), the code for detecting existing entries was not taking that into account, failing to find the city.
- If there's more than one found result, using `discard` for removing the result from the ZIPs to remove gives an error.

@Tecnativa TT52020


pedrobaeza:
Fast-tracking as already merged in 16 and 18:

/ocabot merge patch

PR #1917, opened 2024-12-05
[FIX] partner_company_type: make unique constraint dependent on country

len-foss:


partner_company_type/models/res_partner_company_type.py
```diff
@@ -26,5 +26,15 @@ class ResPartnerCompanyType(models.Model):
...
 
-    _sql_constraints = [('name_uniq', 'unique (name)',
-                         "Partner Company Type already exists!")]
+    _sql_constraints = [
+        (
+            "name_country_uniq",
+            "unique (name, country_id)",
+            "Partner Company Type already exists for this country!"
+        ),
+        (
```
rousseldenis, line 35: @len-foss And why not using the EXCLUDE with all the conditions (and removing unique constraint) ?

len-foss: I generally prefer 2 simple and stupid statements to having only 1 but complicated. Unless there is a simple form to combine these?

PR #1918, opened 2024-12-06
[18.0] [MIG] partner_fax: Migration to 18.0

intero-als:
Migration from 17.0 to 18.0


pedrobaeza:
Thanks for the contribution.

Please preserve commit history following technical method explained in https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-18.0.

If the jump is between several versions, you have to modify the source branch in the main command to accommodate it to this circumstance.

intero-als:
@pedrobaeza thanks for your comment. I will open a new PR for that.

PR #1925, opened 2024-12-10
[17.0][MIG] sale_partner_address_restrict

JordiMForgeFlow:


sale_partner_address_restrict/models/sale_order.py
```diff
@@ -0,0 +1,39 @@
...
+from odoo.exceptions import ValidationError
+
+
+class SaleOrder(models.Model):
+    _inherit = "sale.order"
+
+    @api.constrains("partner_id", "partner_invoice_id", "partner_shipping_id")
+    def _check_partner_addresses(self):
+        for order in self:
+            if order.partner_id and (
```
rousseldenis, line 11: IMHO, as this changes the standard behavior, a configuration parameter should be set to activate the restriction.

sale_partner_address_restrict/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1 @@
+This module restricts the addresses to be used in the sales order form, according to the partner selected.
```
rousseldenis: In which manner?

This should be improved as from module name, we don't understand how the restriction is applied.

sale_partner_address_restrict/models/sale_order.py
JordiMForgeFlow: But you would not install the module if you don't want to have the restriction, don't you think?

rousseldenis: That has always be a debate :smile: 

But you maybe want to decouple deployment flows and activation. Not changing standard behavior by installing a module is always a good practice.

This allows also better flexibility.

And for our tests flows, that allows to avoid separate job flows for conflicting modules.

rousseldenis:
@JordiMForgeFlow Don't you think this should be moved to sale-workflow ? That should be coherent with the other module : https://github.com/OCA/sale-workflow/pull/2452

JordiMForgeFlow:
@rousseldenis makes sense, I will try to open the PR as soon as I have some time :) Thank you!

rousseldenis:
> @rousseldenis makes sense, I will try to open the PR as soon as I have some time :) Thank you!

FYI, I will do a PR on 18.0 on this basis.

JordiMForgeFlow:
closing in favor of https://github.com/OCA/sale-workflow/pull/3574

PR #1928, opened 2024-12-12
[FIX]base_location: city in res.company form view

mav-adhoc:


pedrobaeza:
Please reason this change.

PR #1930, opened 2024-12-12
[17.0][ADD] Add partner_title_for_company: extend title to companies

cvinh:
This module allows to use title for companies in addition to individuals


cvinh:
cc @vehi-invitu

yostashiro:
@cvinh I wonder if the intended use case overlaps with https://github.com/OCA/partner-contact/tree/17.0/partner_company_type?

cvinh:
> @cvinh I wonder if the intended use case overlaps with https://github.com/OCA/partner-contact/tree/17.0/partner_company_type?

Thanks for pointing it out, I did not see it... to what I can understand, partner_company_type adds a model and do not use title.
This module uses title model and is much more simple...
We will keep it for our customer...
I will add comments in README to be clear about that

cvinh:
cc @vehi-invitu can you improve readme and force push please ?

PR #1932, opened 2024-12-14
[17.0][IMP] partner_identification: Add chatter in res.partner.id_number form pop-up

vehi-invitu:
Description:
Integrate the Odoo Chatter to enable uploading attachment files through the `res.partner` pop-up form for ID Numbers.

cc @cvinh

### DIFF
--- a/partner_identification/views/res_partner_view.xml
+++ b/partner_identification/views/res_partner_view.xml
@@ -1,5 +1,5 @@
 <odoo>
-    <!-- Modification of Partner - Adding Tab for Idenification Numbers -->
+    <!-- Modification of Partner - Adding Tab for Identification Numbers -->
     <record model="ir.ui.view" id="view_partner_form">
         <field name="name">res.partner.form.id_number</field>
         <field name="model">res.partner</field>
--- a/partner_identification/views/res_partner_id_number_view.xml
+++ b/partner_identification/views/res_partner_id_number_view.xml
@@ -6,21 +6,32 @@
         <field name="arch" type="xml">
             <form string="Partner ID Numbers">
                 <sheet>
-                    <group>
-                        <field name="partner_id" />
-                        <field name="category_id" />
-                        <field name="name" />
-                        <field name="partner_issued_id" />
-                        <field name="date_issued" />
-                        <field name="place_issuance" />
-                        <field name="valid_from" />
-                        <field name="valid_until" />
-                        <field name="status" />
-                    </group>
+                    <div class="o_row">
+                        <div class="o_col">
+                            <group>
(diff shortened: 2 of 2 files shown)


cvinh:
Please fix pre-commit

partner_identification/views/res_partner_view.xml
```diff
@@ -13,8 +13,41 @@
...
+                                <div class="o_low">
+                                    <group>
+                                        <field name="place_issuance" />
+                                        <field name="valid_from" />
+                                        <field name="valid_until" />
+                                        <field name="status" />
+                                    </group>
+                                </div>
+                            </div>
+                                <separator colspan="4" string="Commentaires" />
```
rousseldenis: Code should be done in English. Thanks

partner_identification/views/res_partner_view.xml
```diff
@@ -13,8 +13,41 @@
                         colspan="4"
                         nolabel="1"
                         widget="one2many_list"
+                        mode="tree,form"
                         context="{'default_partner_id': id}"
-                    />
+                        >
+                    <form>
```
rousseldenis: Doesn't the standard behavior load the base form?

https://github.com/OCA/partner-contact/blob/227da0a6e88f16a7b325b3e662ebfdf6a7155a5c/partner_identification/views/res_partner_id_number_view.xml#L7

rousseldenis (changes requested):

partner_identification/views/res_partner_view.xml
cvinh: @vehi-invitu maybe it's only the missing display-block in the chatter that does not show the chatter in the popup
BTW I like your view reorganisation, can you apply it to the base form view ?

cvinh (approved):
LGTM

rousseldenis (approved):

cvinh:
There is an issue when we add a identification as the chatter is not usable until the identification is saved... it means that we can't add an attachment in the popup during the creation

The workaround is to save the identification and open it again... then the chatter is usable...
It's not very convenient, @rousseldenis WDYT ?

rousseldenis:
> There is an issue when we add a identification as the chatter is not usable until the identification is saved... it means that we can't add an attachment in the popup during the creation

That's the standard behavior. Record is not saved yet.

cvinh:
> > There is an issue when we add a identification as the chatter is not usable until the identification is saved... it means that we can't add an attachment in the popup during the creation
> 
> That's the standard behavior. Record is not saved yet.

Ok then... let's merge ?

dreispt (approved):

PR #1934, opened 2024-12-17
[15.0][ADD] res_partner_expense_link

Saran440:
This module add link between partner and expense sheet
Selection_004


res_partner_expense_link/models/res_partner.py
```diff
@@ -0,0 +1,29 @@
...
+
+    expense_sheet_count = fields.Integer(
+        compute="_compute_expense_count",
+    )
+
+    def _compute_expense_count(self):
+        sheet_model = self.env["hr.expense.sheet"]
+        employee_model = self.env["hr.employee"]
+        for partner in self:
+            employee_partner = employee_model.search_read(
```
rousseldenis, line 18: I don't think those searches inside the loop are performant

PR #1952, opened 2025-01-09
[16.0][FIX] partner_auto_archive: Add mail dependency to fix tracking parameter warning

mb-ife:
When installing or upgrading the module, a warning appeared about the unknown 
'tracking' parameter. This fix adds mail as a dependency to properly support 
the tracking parameter on fields, ensuring clean installation and upgrades 
without warnings.

### DIFF
--- a/partner_stage/__manifest__.py
+++ b/partner_stage/__manifest__.py
@@ -9,7 +9,7 @@
     "category": "Sales/CRM",
     "version": "16.0.1.0.1",
     "license": "AGPL-3",
-    "depends": ["contacts"],
+    "depends": ["contacts", "mail"],
     "data": [
         "security/ir.model.access.csv",
         "data/partner_stage_data.xml",


ng-ife (approved):
Finally found the issue. It's dependencies.
--edit--
It solves the issue. Still I would like to know why this problem occurs as res.partner has a chatter and should therefor be able to track.

ng-ife:
Does not occur anymore. Probably related to another module or fixed in odoo core.

ng-ife (approved):
LGTM

ng-ife:
Hello @dreispt 
I saw you already merged changes for this module before. Can you have a look here?
Thank you

partner_stage/__manifest__.py
```diff
@@ -9,7 +9,7 @@
     "category": "Sales/CRM",
     "version": "16.0.1.0.1",
     "license": "AGPL-3",
-    "depends": ["contacts"],
+    "depends": ["contacts", "mail"],
```
dreispt, line 12: Doesn't contacts already depend on mail?
Otherwise it wouldn't be capable to provide the Chatter feature.

ng-ife: @dreispt You're right. Still we have the issue that tests (e.x. on Odoo.sh fail) if we don't add this here. What about changing the dependency to mail instead of contacts. Contacts is not technically needed for the module to work.

PR #1973, opened 2025-02-02
[16.0][ADD] phone_query: Add a new module

kobros-tech:
This module allows to search and list partners through their phone numbers from the url input in browser


kobros-tech:
@wlin-kencove

kobros-tech:
@pedrobaeza 

do you think this module desrves to be merged?

pedrobaeza:
Well, it seems something very specific and with no direct user usage.

kobros-tech:
then I will close, we are not interested to have it public as it was specific as you mentioned, thanks!

PR #1976, opened 2025-02-05
[17.0][MIG] partner_country_lang: Migration to 17.0

victoralmau:
Migration to 17.0

@Tecnativa TT54939


pedrobaeza:
You should convert the onchange to compute and you won't need that dependency switch.

victoralmau:
> You should convert the onchange to compute and you won't need that dependency switch.

Changes done.

partner_country_lang/models/res_partner.py
```diff
@@ -0,0 +1,15 @@
+# Copyright 2022-2025 Tecnativa - Víctor Martínez
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import api, fields, models
+
+
+class ResPartner(models.Model):
+    _inherit = "res.partner"
+
+    lang = fields.Selection(compute="_compute_lang")
```
pedrobaeza: Always add a comment indicating that this is an override

```suggestion
    # Add the compute method on the existing field
    lang = fields.Selection(compute="_compute_lang")
```

partner_country_lang/models/res_partner.py
```diff
@@ -0,0 +1,15 @@
...
+
+    @api.depends("country_id")
+    def _compute_lang(self):
+        for item in self:
```
pedrobaeza, line 17: Be preventive if the method already exists in other module:

```suggestion
        for item in self:
            if hasattr(super(), "_compute_lang"):
                super()._compute_lang()
```

partner_country_lang/models/res_partner.py
```diff
@@ -0,0 +1,15 @@
...
+            item.lang = item.country_id.lang or item.lang
```
pedrobaeza: Isn't more declarative this way and avoid unneeded assignations?

```suggestion
            if item.country_id.lang:
                item.lang = item.country_id.lang
```

partner_country_lang/models/res_partner.py
```diff
@@ -0,0 +1,21 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import api, fields, models
+
+
+class ResPartner(models.Model):
+    _inherit = "res.partner"
+
+    # Add the compute method on the existing field
+    lang = fields.Selection(compute="_compute_lang", store=True)
```
pedrobaeza:
```suggestion
    lang = fields.Selection(compute="_compute_lang", store=True, readonly=False)
```

pedrobaeza (approved):

pedrobaeza:
Oh, wait, the CI is red.

victoralmau:
> Oh, wait, the CI is red.

Reviewing the test error happens because the existing `ir.default` for the `lang` field causes the _en_US_ language to be defined on creation without going through compute. Is it possible that not doing the compute field initially was intentional to avoid this?

pedrobaeza:
If this is not working for create, you should continue overriding the `create` as it was.

BhaveshHeliconia (approved):
LGTM

PR #1980, opened 2025-02-17
[18.0][MIG] partner_deduplicate_acl: Migration to 18.0

juancarlosonate-tecnativa:
No changes needed

@Tecnativa TT54200
@pedrobaeza @victoralmau can you check? thanks!


pedrobaeza:
You have added your name to the contributor list in the "pre-commit auto-fixes" commit, not in the migration one, which is incorrect. For fixing it, in this case, you can simply squash both commits together, but it should be named as the last one.

victoralmau (approved):

pedrobaeza (approved):
/ocabot merge nobump

PR #1981, opened 2025-02-19
[FIX] partner_contact_department: avoid accidental department creations

yajo:
Users were creating departments accidentally.

In our customers' use case, a few users create and organize departments, and the rest just use them. I think it's the more common case.

@moduon MT-8760


rafaelbn (approved):
👍🏼 👌🏼 

Thanks. 💙

Shide (approved):

fcvalgar (approved):
LGTM, thank you @yajo

PR #1982, opened 2025-02-19
[FIX] partner_contact_job_position: avoid accidental job position creations

yajo:
Users were creating job positions accidentally.

In our customers' use case, a few users create and organize job positions, and the rest just use them. I think it's the more common case.

@moduon MT-8760


Shide (approved):

rafaelbn (approved):
Perfect! 👍🏼 Thank you!

fcvalgar (approved):
LGTM. thank you @yajo

PR #1985, opened 2025-02-21
[18.0][FIX] partner_stage: Add 'order' in _read_group_stage_id to fix group…

Ricardo-MC:
Added 'order' argument with a default value to avoid errors in staging in the contacts view.


partner_stage/models/res_partner.py
```diff
@@ -14,8 +14,9 @@ def _get_default_stage_id(self):
         )
 
     @api.model
-    def _read_group_stage_id(self, states, domain, order):
-        return states.search([])
+    def _read_group_stage_id(self, stages, domain, order=None):
+        order = order or "sequence, id"
+        return stages.search([], order=order)
```
dreispt: Even simpler, sparing 1 LOC?
```suggestion
        return stages.search([], order=order or "sequence, id")
```

celm1990: I'm not sure if this change is correct.
According to the definition, the order parameter should be removed because only two parameters are expected: `groups, domain`, not `order`:
https://github.com/odoo/odoo/blob/6caa783991dad7272391ae8ea0e76f32eadc9401/odoo/models.py#L2330

The original order should be preserved, like this:
```suggestion
    def _read_group_stage_id(self, stages, domain):
        order = order or "sequence, id"
        return stages.search([], order=stages._order)
```
Additionally, for consistency, this function should now be renamed to `_read_group_stage_ids` in both the `function definition` and the corresponding `field definition`.
For example:
https://github.com/odoo/odoo/blob/6caa783991dad7272391ae8ea0e76f32eadc9401/addons/maintenance/models/maintenance.py#L405-L410

celm1990 (changes requested):

Ricardo-MC:
new PR (https://github.com/OCA/partner-contact/pull/2087)

PR #1991, opened 2025-02-25
[18.0][MIG] partner_middlename: Migration to 18.0

BhaveshHeliconia:


marcelsavegnago (approved):

BhaveshHeliconia:
@pedrobaeza : Could you please reopen this PR.

partner_middlename/tests/test_name.py
```diff
@@ -13,8 +13,9 @@ class PersonCase(TransactionCase):
     model = "res.partner"
     context = dict()
 
-    def setUp(self):
-        super().setUp()
+    @classmethod
```
NL66278, line 16: In class methods you should replace `self` with `cls`. Python does not care, but conventions do. (You could replace self with any name). 'self' is used to refer to an instance, but there is no instance yet.

BhaveshHeliconia: Thanks for pointing it out. I've updated self to cls in the class methods.

NL66278 (approved):
👍 LGTM. Thanks for your contribution and for quickly taking up remark.

PR #1983, opened 2025-02-19
[IMP] partner_contact_department: sort by name

yajo:
Current order is weird. This makes it more usable.

@moduon MT-8760

PR #2009, opened 2025-02-25
[FIX] partner_contact_department: really order by name

yajo:
First attempt in https://github.com/OCA/partner-contact/pull/1983 didn't really work because `parent_path` always contained own ID.

Now departments will really be ordered by name, or by parent path + name if they have a parent.

@moduon MT-8760


fcvalgar (approved):
Functional review,
LGTM, thank you @yajo .

image

image

rafaelbn (approved):
/ocabot merge patch

PR #2010, opened 2025-02-25
[16.0][IMP] pmr: add constraint on partner.is_company

NL66278:
Without this fix it is possible to change partner is_company field even though that might make existing connections invalid.

### DIFF
--- a/partner_multi_relation/models/res_partner.py
+++ b/partner_multi_relation/models/res_partner.py
@@ -178,7 +178,7 @@ class ResPartner(models.Model):
 
     @api.constrains("is_company")
     def _check_relation_compatibility(self):
-        """If is_company changes, check relations wether this should be allowed."""
+        """If is_company changes, check relations whether this should be allowed."""
         Relation = self.env["res.partner.relation"]
         for this in self:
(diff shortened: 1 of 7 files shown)


partner_multi_relation/models/res_partner.py
```diff
@@ -175,6 +176,29 @@ def get_partner_type(self):
         self.ensure_one()
         return "c" if self.is_company else "p"
 
+    @api.constrains("is_company")
+    def _check_relation_compatibility(self):
+        """If is_company changes, check relations wether this should be allowed."""
```
rrebollo:
```suggestion
        """If is_company changes, check relations whether this should be allowed."""
```

rrebollo (approved):
Code Review: Great work! The code looks good to me (LGTM). Thank you for your contribution! I've provided a few suggestions for your consideration—feel free to address them as you see fit.

partner_multi_relation/models/res_partner.py
NL66278: @rrebollo Thanks! I corrected spelling of `whether` everywhere.

PR #2014, opened 2025-02-27
[17.0][MIG] partner_lastname_uppercase: Migration to 17.0

BhaveshHeliconia:


marcelsavegnago (approved):

partner_lastname_uppercase/models/res_partner.py
```diff
@@ -0,0 +1,55 @@
...
+            if "lastname" in vals or "is_company" in vals:
+                individuals = self.filtered(
+                    lambda rp: rp.lastname and not rp.is_company
+                )
+                # looping in case only is_company is in vals (set to false)
+                # in that case, uppercase each lastname
+                for partner in individuals:
+                    super(ResPartner, partner).write(
+                        {"lastname": partner.lastname.upper()}
+                    )
```
rrebollo: I know it's just a migration, but couldn't it have been done in a simpler way?  
The recursive call is especially odd — I'm honestly wondering how it doesn't end up in an infinite loop.

BhaveshHeliconia:
@rrebollo : Thanks for pointing that out! I agree it could have been cleaner. I’ve revisited the logic and refactored it to simplify the approach and remove the recursive call. Please review and let me know your thoughts.

rrebollo:
@BhaveshHeliconia please fix pre-commit check complains

rrebollo (approved):
Code Review. LGTM!

partner_lastname_uppercase/models/res_partner.py
```diff
@@ -49,7 +49,6 @@ def write(self, vals):
                 # looping in case only is_company is in vals (set to false)
                 # in that case, uppercase each lastname
                 for partner in individuals:
-                    super(ResPartner, partner).write(
-                        {"lastname": partner.lastname.upper()}
-                    )
+                    if partner.lastname != partner.lastname.upper():
+                        partner.lastname = partner.lastname.upper()
```
NL66278: Use the original super(ResPartner, partner).... statement here, otherwise you will be back in the original write in a recursive call. And while this will work, because of the if, it is a lot less efficient then calling the super method.
Even more elegant would be to make the test for not already uppercase part of the lambda filtering.

NL66278 (changes requested):
👍 Just a small change and will be good to go.

BhaveshHeliconia:
> 👍 Just a small change and will be good to go.

Thanks for your suggestion. I've made the changes accordingly.

NL66278 (approved):
👍 LGTM

PR #1767, opened 2024-05-27
[17.0][MIG] partner_email_duplicate_warn: Migration to 17.0

V3st1g3:

PR #2020, opened 2025-03-03
[18.0][MIG] partner_email_duplicate_warn: mig to v18 and switch from M2O to M2M

alexis-via:
The improvement for the switch from M2O to M2M is in a separate commit.

dup_email


marielejeune:
Hi @alexis-via 
Thanks for the work! 
It seems that you started the migration based on branch 16.0. However the addon was migrated to version 17.0 and some changes were made there (some of which I don't agree with, such as changing the XMLID of the view, but that's another point). 
Could you please have a look at it?

partner_email_duplicate_warn/views/res_partner.xml
```diff
@@ -0,0 +1,21 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <record id="view_partner_form" model="ir.ui.view">
+        <field name="model">res.partner</field>
+        <field name="inherit_id" ref="base.view_partner_form" />
+        <field name="arch" type="xml">
+            <div invisible="not same_vat_partner_id" position="after">
```
MohamedOsman7:
```suggestion
            <xpath expr="//div[@name='warning_company']" position="after">
```

As @marielejeune also mentioned, could you take care of this change as part of the 17.0 migration? It comes from this commit.  (https://github.com/OCA/partner-contact/pull/1767/commits/6512f16d0e309e35348caac23620b3aea61c34be)

MohamedOsman7 (commented):
Otherwise, the code and functional review look good to me.
I also noticed a few commits that could be squashed to keep the history cleaner. 
(https://github.com/OCA/maintainer-tools/wiki/Merge-commits-in-pull-requests#mergesquash-the-commits-generated-by-bots-or-weblate)

image

CRogos (changes requested):
Could you recreate the migration starting on version 17 of this module? Because the view has been renamed from **view_partner_form_with_email_warning** to **view_partner_form**, the upgrade of this module is not working without migration script.

partner_email_duplicate_warn/views/res_partner.xml
```diff
@@ -0,0 +1,21 @@
...
+                    class="alert alert-warning"
+                    role="alert"
+                    invisible="not same_email_partner_ids"
+                >
+                  Duplicate warning: partner(s) <field
+                    name="same_email_partner_ids"
+                    widget="x2many_buttons"
+                /> has/have the same <b>Email</b>.
+                </div>
+            </div>
```
MohamedOsman7, line 18: </xpath>

MohamedOsman7:
```suggestion
            </xpath>
```

MohamedOsman7 (changes requested):
Agree with @CRogos !

Change the ending tag of 'xapth' accordingly (see below)

partner_email_duplicate_warn/views/res_partner.xml
```diff
@@ -0,0 +1,21 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <record id="view_partner_form" model="ir.ui.view">
```
CRogos, line 3:
```suggestion
    <record id="view_partner_form_with_email_warning" model="ir.ui.view">
```

This view was renamed in v17. 
Because the view does not match, and same_email_partner_id was renamed in same_email_partner_ids, I get an error during the upgrade.

CRogos (changes requested):
These can be squashed: 
image

NL66278:
Closed in favor of 2077

PR #2027, opened 2025-03-12
[16.0][ADD] partner_street_number

hbrunn:
This module extends the functionality of base_address_extended to support custom street name formats.

Further, it allows users to turn on manual entry of the components of a street name, and to bulk split/compose street names of existing partners.

I didn't pull the commits of the pre-v11 version of this because it doesn't have anything to do with the old module, apart from its name.


coleste (approved):
LGTM

partner_street_number/tests/test_partner_street_number.py
```diff
@@ -0,0 +1,75 @@
+# Copyright 2025 Hunki Enterprises BV
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl-3.0)
+
+from odoo.tests.common import Form, TransactionCase
+
+
+class TestPartnerStreetNumber(TransactionCase):
```
rrebollo, line 7: You could use `BaseCommon` instead of `TransactionCase`

rrebollo (approved):
Excellent work — truly didactic, both in design and implementation.  
It's clear you have strong mastery of Odoo. LGTM! 🚀

rrebollo:
@hbrunn btw, an obvious use case for this would be l10ns? right?

rvalyi:
@hbrunn of course the choice is yours, but did you consider releasing it under LGPL? Just thinking AGPL might be too restrictive just for customizing the address format. But if you go for AGPL so be it, I respect it too.

hbrunn:
@rvalyi my policy towards LGPL is: don't oppose it, but don't actively support it either
@rrebollo indeed, but also apart from this you might need to have separate fields for ie synchronization with external systems that require it

partner_street_number/tests/test_partner_street_number.py
hbrunn: what would be the benefit of that?

rrebollo: Mainly, it sets useful defaults and follows the OCA recommendation for version **17.0** (I know this is **16.0**).
Of course, it’s not mandatory.

NL66278 (approved):
👍 LGTM. I learned that we can use a data element as root in xml. Interesting...

PR #2030, opened 2025-03-19
[15.0][ADD] partner_address_history

Saran440:
This module similar with [partner_address_version](https://github.com/OCA/partner-contact/tree/14.0/partner_address_version)

This module will keep tracking history when you change address or name
Selection_005

Example:
1. We change name and address at time 19 Mar 2025
1. Partner change name from `Vendor` to `Vendor2`
2. Address change from `111` to `222`

But bill is confirm already at 11 Mar 2025
Normally, it will show current address and name
Selection_006

After you send context `date_change` following code
Selection_008

it will display with old address, name

Selection_007


NL66278 (changes requested):
While the module addresses a valid problem in standard Odoo, IMHO the solution it offers is not the way to do it. The problem: delivery and billing addresses have statistical and fiscal implications. If we want to know where a certain delivery went, or where  a customer was living when billed, the delivery or billing address should not change. Basically it means, once used in a delivery or invoice, those addresses should not be changed, only archived if they should no longer be used for future deliveries or invoices.

So my problem with the solution is that it puts the responsibility of showing the right address on other modules, that need to be customized to pass a date context on each use of a historical date.

Of course there is the problem of the main address of a partner. I think if this changes I think a contact address should be created (with active=False) with the old information (and an indication until which date this date was the valid address.

Odoo users should actually be discouraged from entering address information on the main contact. I remain of the opinion that is was a huge design mistake for Odoo SA to merge the Partner and Address models.

Saran440:
@NL66278 Thank you for your feedback, I agree that this module can't solved. In production we solved it by add a new field (address or name) into document instead. and change report, form to use new field instead.

NL66278:
@Saran440 Hi, nice to her that my comment was of some use, and good that you solved your requirements!

PR #2032, opened 2025-03-25
[17.0][MIG] partner_contact_weight: Migration to 17.0

vehi-invitu:
cc @cvinh


etobella:
IF it is a migration you are not keeping the history :thinking:

cvinh:
Please read the guidelines how to migrate
Can you put PR in draft, we'll make it open when it's ready

legalsylvain (commented):
I think that such feature should go in a dedicated repo. Something like vertical-health

vehi-invitu:
@legalsylvain it was already in this repo but in version 10.

legalsylvain:
> @legalsylvain it was already in this repo but in version 10.

Yes i saw. But what do you think about my proposal ?

etobella:
@legalsylvain vertical-health does not exists in OCA and vertical-medical has been dead for 7 years.

I would recommend to keep it here. But I don't have an strong opinion on this.

cvinh:
@legalsylvain partner weight is not so vertical... it can be used in fitness or sport use cases.
I don't have strong opinion neither, you decide...

legalsylvain:
hi @cvinh and @etobella 

I think that such modules ([partner_contact_weight](https://github.com/OCA/partner-contact/pull/2032),  [pos_partner_birthdate_weight_mandatory](https://github.com/OCA/pos/pull/1347), [pos_partner_weight](https://github.com/OCA/pos/pull/1344) are very specific. (pos or not).

IMO, it will be used only (or mainly) in an medical / hospital / pharmacy context. Indeed it can be used in areas related to health, such as sport, but we're still talking about biological data.

For this reason, I think we should put that modules in a dedicated medical / health repository.

I propose you :

- to use the existing repo, here : https://github.com/OCA/vertical-medical/ (It is totally empty ...). (only "dev" branches contained modules, but they have been deleted)
- to propose a PR to add your github login in this list, so you'll have some write access on this repo. https://github.com/OCA/repo-maintainer-conf/blob/master/conf/psc/vertical.yml#L102

(you work daily in medical area, and you regularly contribute to OCA, so it makes sense if you're are responsible of this repo).

What do you think ?

cvinh:
Okay

PR #2023, opened 2025-03-05
[FIX] partner_contact_department: make display name lang-aware

yajo:
Yet another iteration over https://github.com/OCA/partner-contact/pull/2009.

`display_name` wasn't translatable, but `name` was. Thus, stored record order was always on English. Now, if department names are translated and the user switches to another language, the order the user will see will match the language it is displaying.

Another fix is that now, when changing a parent department name name, children display names will change too.

I took the chance to apply some anti-flaky-tests good practices in current test suite (apart from increasing coverage to cover these new cases).

@moduon MT-8760

PR #2036, opened 2025-03-28
[FIX] partner_contact_department: proper version bump

yajo:
https://github.com/OCA/partner-contact/pull/2023 added a migration script for version 16.0.1.3.0 because it was expected to be merged as a minor bump.

However, it was merged by mistake as a patch bump in https://github.com/OCA/partner-contact/pull/2023#pullrequestreview-2697254654, resulting in the migration script not being run.

Hence, this just bumps the version appropriately to force running that script on updates.

@moduon MT-8760


yajo:
Fast-tracking.

/ocabot merge nobump

PR #2047, opened 2025-04-30
[17.0][ADD] partner_contact_telegram: New module

AlvaroRM11:
@BinhexTeam

The purpose of this PR is to add a field to enter the Telegram nickname for contacts. 
Depends on #2054


partner_contact_telegram/models/res_partner.py
```diff
@@ -0,0 +1,57 @@
...
+from odoo import _, api, fields, models
+from odoo.exceptions import ValidationError
+
+
+class ResPartner(models.Model):
+    """Partner with telegram alias and id"""
+
+    _inherit = "res.partner"
+
+    telegram_id = fields.Char()
```
christian-ramos-tecnativa: This doesn't seem to be used and is hidden, is it really needed?

pedrobaeza: Suffix `_id` is reserved for many2one fields.

arielbarreiros96: Indeed using `_id` could be confusing for a char field.
In Telegram there is also has a unique identifier for users who don't share phone number or alias for privacy reasons. While Telegram users don't have access to this identifier, the Telegram API does provide this information, and it could be helpful for retrieving users using `search()` for example.

PR #2048, opened 2025-04-30
[18.0][IMP] partner_firstname: name order in form configurable

CRogos:
This change makes the order in the form configurable in the settings.
image

I am not very happy with the **form_has_lastname_first** field in **res.partner**. Is there a better way to implement this?

### DIFF
--- a/partner_firstname/models/res_partner.py
+++ b/partner_firstname/models/res_partner.py
@@ -27,6 +27,19 @@ class ResPartner(models.Model):
         readonly=False,
     )
 
+    form_has_lastname_first = fields.Boolean(compute="_compute_form_has_lastname_first")
+
+    def _compute_form_has_lastname_first(self):
+        default_order = (
+            self.env["res.config.settings"].sudo()._partner_names_order_default()
+        )
+        self.form_has_lastname_first = (
+            self.env["ir.config_parameter"]
+            .sudo()
+            .get_param("partner_names_order", default_order)
+            != "first_last"
+        )
+
     @api.model
     def name_fields_in_vals(self, vals):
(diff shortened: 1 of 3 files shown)


amh-mw (changes requested):
What about using the existing `ir.config_parameter` `partner_names_order` instead of adding a new parameter?

CRogos:
> What about using the existing `ir.config_parameter` `partner_names_order` instead of adding a new parameter?

Because the current default is, that Lastname is above Firstname, but Firstname is before Lastname. 
When we change this, I assume many users are confused.

legalsylvain:
In my opinion, user will understand quickly.

No need to add extra fields.

When we migrate from a version to another, user face bigger changes.

CRogos:
@legalsylvain @amh-mw you prefer connecting it to the existing config parameter this way?

Or just swap it without configuration?

amh-mw:
> you prefer connecting it to the existing config parameter this way?

Yes. The user has configured how they want names displayed and honoring that choice in an additional way should not be surprising.

> Or just swap it without configuration?

No. Whatever hardcoded display style it changes to will be just as arbitrary as the current behavior.

MohamedOsman7 (approved):

amh-mw (approved):
Code review only. LGTM.

jtl-mw (approved):
Manually tested. g2g

CRogos:
@pedrobaeza could you merge?

pedrobaeza:
@legalsylvain should confirm

legalsylvain (approved):

CRogos:
@pedrobaeza legalsylvain has confirmed, could you merge?

PR #2049, opened 2025-05-01
[18.0][FIX] partner_firstname: res.partner copy() not honoring defaults

amh-mw:
This makes get_extra_default_copy_values aware of the default dict passed to copy or copy_data.

### DIFF
--- a/base_partner_sequence/models/partner.py
+++ b/base_partner_sequence/models/partner.py
@@ -25,11 +25,12 @@ class ResPartner(models.Model):
                 vals["ref"] = self._get_next_ref(vals=vals)
         return super().create(vals_list)
 
-    def copy(self, default=None):
-        default = default or {}
-        if self._needs_ref():
-            default["ref"] = self._get_next_ref()
-        return super().copy(default=default)
+    def copy_data(self, default=None):
+        vals_list = super().copy_data(default=default)
+        for record, vals in zip(self, vals_list, strict=False):
+            if record._needs_ref():
+                vals["ref"] = record._get_next_ref()
+        return vals_list
 
     def write(self, vals):
         for partner in self:
--- a/partner_firstname/models/res_partner.py
+++ b/partner_firstname/models/res_partner.py
@@ -83,23 +83,38 @@ class ResPartner(models.Model):
             ).create([vals])
         return created_partners
 
-    def get_extra_default_copy_values(self):
+    def get_extra_default_copy_values(self, default=None):
         """Method to add '(copy)' suffix to lastname or firstname, depending on name
         order configuration.
         """
-        if self._get_names_order() == "first_last":
-            return {
-                "lastname": _("%s (copy)", self.lastname)
-                if self.lastname
-                else _("(copy)")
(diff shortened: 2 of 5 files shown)


amh-mw:
@OCA/crm-sales-marketing-maintainers Would you kindly remove `stale` from this pull request?

base_partner_sequence/models/partner.py
```diff
@@ -26,10 +26,14 @@ def create(self, vals_list):
         return super().create(vals_list)
 
     def copy(self, default=None):
-        default = default or {}
-        if self._needs_ref():
-            default["ref"] = self._get_next_ref()
-        return super().copy(default=default)
+        default_ = default or {}
+        records = self.browse()
```
NL66278: You want to make copy() support multiple records, but it always runs on only one record. That is why the base method calls `self.ensure_one()`.

amh-mw: Per [BaseModel](https://github.com/odoo/odoo/blob/5fc1e34d174f7f61d692d086d0ff65fbfc72b013/odoo/models.py#L5956-L5971):
```python
@api.returns('self')
def copy(self, default: ValuesType | None = None) -> Self:
    """ copy(default=None)

    Duplicate record ``self`` updating it with default values

    :param dict default: dictionary of field values to override in the
            original values of the copied record, e.g: ``{'field_name': overridden_value, ...}``
    :returns: new records

    """
    vals_list = self.with_context(active_test=False).copy_data(default)
    new_records = self.create(vals_list)
    for old_record, new_record in zip(self, new_records):
        old_record.copy_translations(new_record, excluded=default or ())
    return new_records
```

base_partner_sequence/models/partner.py
```diff
@@ -26,10 +26,14 @@ def create(self, vals_list):
         return super().create(vals_list)
 
     def copy(self, default=None):
```
NL66278: Would it not be best to fill the reference (where needed) in a new override of copy_data? Otherwise the whole performance gain by copy() now being able to process in batch will be lost. Maybe something (not tested) like:
```
def copy_data(self, default=None):
    vals_list = super().copy_data(default=default)
    for record, default in zip(self, vals_list, strict=False):
        if record._needs_ref():
            default["ref"] = record._get_next_ref()
    return vals_list
```
And maybe also use `record._needs_ref(vals=default)` as the default might contain `is_company` or `parent_id` values that could influence the logic of_needs_ref.

base_partner_sequence/models/partner.py
NL66278: @amh-mw This indeed changed in 18.0: https://github.com/odoo/odoo/pull/154132

NL66278 (approved):
👍 LGTM. Thanks also for quickly taking up review comments.

base_partner_sequence/models/partner.py
amh-mw: Seems reasonable. I renamed `default` to `vals` to avoid closing over a method argument, but it's otherwise your implementation.

Issue #2055, opened 2025-05-07
[18.0] partner_stage: While Group by Stage


nikul-serpentcs:
## Module
partner_stage

## Describe the bug
While Group by Stage

## To Reproduce
**Affected versions**:

1. Go to Contacts
2. Open Searchbar
3.  Group by Stage

**Expected behavior**
Work with **Group By Stage**

**Additional context**
Add any other context about the problem here. (e.g. OS, Python version, ...)

![Image](https://github.com/user-attachments/assets/5df25f63-879e-4c4f-bce5-4b0dcfcc7ba4)

PR #2056, opened 2025-05-07
[FIX] [18.0] partner_stage: Groupby Stages

nikul-serpentcs:
https://github.com/OCA/partner-contact/issues/2055

### DIFF
--- a/partner_stage/models/res_partner.py
+++ b/partner_stage/models/res_partner.py
@@ -14,7 +14,7 @@ class Partner(models.Model):
         )
 
     @api.model
-    def _read_group_stage_id(self, states, domain, order):
+    def _read_group_stage_id(self, states, domain):
         return states.search([])
 
     stage_id = fields.Many2one(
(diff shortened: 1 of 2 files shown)


nikul-serpentcs:
@dreispt @rvalyi @ovnicraft @azmeuk @rousseldenis Can you please take a look?

Murtaza-OSI (approved):

nikul-serpentcs:
@pedrobaeza  can you please take a look?

pedrobaeza:
Sorry, not using this module.

partner_stage/models/res_partner.py
```diff
@@ -14,7 +14,7 @@ def _get_default_stage_id(self):
         )
 
     @api.model
-    def _read_group_stage_id(self, states, domain, order):
+    def _read_group_stage_id(self, states, domain):
         return states.search([])
```
amh-mw, line 18: Is this supposed to change to
```python
        return states.search([], order=order)
```
otherwise I don't understand what this pull request is attempting. Unit tests could be illuminating.

amh-mw (changes requested):

partner_stage/models/res_partner.py
dreispt: This seems to be fixing the wrong method argument cause the error #2055

dreispt (approved):
Tested on runboat without errors:

image

nikul-serpentcs:
@dreispt Can you please move forward?

PR #2058, opened 2025-05-07
[17.0][IMP] base_location: Add city code to be used for other purposes distinct of zip code.

sergio-teruel:
cc @Tecnativa

ping @carlos-lopez-tecnativa @carlosdauden

### DIFF
--- a/base_location/views/res_city_view.xml
+++ b/base_location/views/res_city_view.xml
@@ -12,6 +12,7 @@
             </field>
             <field name="zipcode" position="after">
                 <field name="zip_ids" widget="many2many_tags" />
+                <field name="code" />
             </field>
         </field>
     </record>
@@ -24,6 +25,7 @@
                         <field name="name" />
                         <field name="country_id" />
                         <field name="state_id" />
+                        <field name="code" />
                     </group>
                     <notebook>
                         <page name="zips" string="Zips">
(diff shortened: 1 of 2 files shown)


carlosdauden (approved):

carlos-lopez-tecnativa (approved):

pedrobaeza (approved):
/ocabot merge minor

Please add in `base_location_geonames_import` the code to fill that field.

legalsylvain:
Hi. Could you give exemple of city's code, i dont figure out code except ZIP.

Thanks !

base_location/views/res_city_view.xml
```diff
@@ -12,6 +12,7 @@
             </field>
             <field name="zipcode" position="after">
                 <field name="zip_ids" widget="many2many_tags" />
+                <field name="code" />
```
legalsylvain, line 15: For such field that are very optional, could you add optional option ?

Thanks !

pedrobaeza:
One city may have several ZIP codes. The code is something specific to identify the city. This code is in Geonames for example.

base_location/views/res_city_view.xml
pedrobaeza: There are very few fields in the tree view, so I don't think it's a problem.

PR #2060, opened 2025-05-14
[IMP] base_partner_sequence: add server action to update partners

victorvermot:
When installing this module on a large db with a lot of res.partners we want to be able to quickly add a ref on them.

This pr adds a server action that allows us to do that.

### DIFF
--- a/base_partner_sequence/models/partner.py
+++ b/base_partner_sequence/models/partner.py
@@ -6,6 +6,8 @@
 
 from odoo import _, api, exceptions, models
 
+BATCH_SIZE = 5000
+
 
 class ResPartner(models.Model):
     """Assigns 'ref' from a sequence on creation and copying"""
@@ -72,3 +74,21 @@ class ResPartner(models.Model):
         to the partner's contacts
         """
         return super()._commercial_fields() + ["ref"]
+
+    def update_partners_ref(self):
+        """Calculate ref of all partners. Called from a server action"""
+        offset = 0
+
+        while True:
+            partners = self.search(
+                [("ref", "=", False)], offset=offset, limit=BATCH_SIZE
+            )
+
+            if not partners:
+                break
+
+            for partner in partners:
+                if partner._needs_ref():
+                    partner.write({})
+
+            offset += BATCH_SIZE
--- a/base_partner_sequence/README.rst
+++ b/base_partner_sequence/README.rst
@@ -40,6 +40,8 @@ modified from the commercial partner.
 No references are assigned for contacts such as shipping and invoice
 addresses.
 
+The module also adds a server action to update all partners ref.
+
 **Table of contents**
 
 .. contents::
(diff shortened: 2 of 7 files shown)


base_partner_sequence/models/partner.py
```diff
@@ -72,3 +74,21 @@ def _commercial_fields(self):
...
+            partners = self.search(
+                [("ref", "=", False)], offset=offset, limit=BATCH_SIZE
+            )
+
+            if not partners:
+                break
+
+            for partner in partners:
+                if partner._needs_ref():
+                    partner.write({})
```
ivantodorovich, line 92: This is kind of weird.
I think I'd do this instead, which is more explicit, to-the-point and probably better performance-wise:

```suggestion
                partner.ref = partner._get_next_ref()
```

base_partner_sequence/models/partner.py
```diff
@@ -72,3 +74,21 @@ def _commercial_fields(self):
...
+        while True:
+            partners = self.search(
+                [("ref", "=", False)], offset=offset, limit=BATCH_SIZE
+            )
+
+            if not partners:
+                break
+
+            for partner in partners:
+                if partner._needs_ref():
```
ivantodorovich, line 91: Can we filter out partners that need ref directly in `search`?
That'd have a great impact in performance and memory usage during this request.

Looking at `_need_ref`, it seems we want partners having `is_company, '=', True` or `parent_id, '=', False`
So, in summary, a domain like this:

```py
@api.model
def _need_ref_domain(self):
    """Domain to retrieve partners for which we want a reference

    This method should be aligned with the pythonic implementation in
    :method:`_need_ref`.
    """
    return ["|", ("is_company", "=", True), ("parent_id", "=", False)]
```

Then, you can search partners to set ref with `self._need_ref_domain() + [("ref", "=", False)]`, or perhaps implement a `force: bool` attribute directly in `_need_ref_domain` to append that last bit (and why not also in `_need_ref`, for consistency)

base_partner_sequence/models/partner.py
```diff
@@ -72,3 +74,21 @@ def _commercial_fields(self):
         to the partner's contacts
         """
         return super()._commercial_fields() + ["ref"]
+
+    def update_partners_ref(self):
+        """Calculate ref of all partners. Called from a server action"""
+        offset = 0
+
+        while True:
```
ivantodorovich, line 82: There should always be a damn good reason to use an infinite loop.

In this case, I think you should re-implement this pattern with `odoo.tools.split_every`.
While at it, also invalidate the cache after each batch iteration, to free the in-memory cache.
Something like this:

```py
partners = self.search(domain)  # note: no limit, nor offset
for batch in split_every(BATCH_SIZE, partners.ids, self.browse):
    for record in batch:
        ...
    batch.invalidate_model()  # this will flush pending writes to the db, and clear the in-memory cache
```

ivantodorovich (changes requested):

PR #2070, opened 2025-06-04
[14.0][IMP][partner_tier_validation] ensure default state at partner …

Kev-Roche:
When creating a new partner, we can directly set a normal / confirmed state on it, without validation.
This PR avoid that by forcing the default stage on partner creation.

### DIFF
--- a/partner_tier_validation/models/res_partner.py
+++ b/partner_tier_validation/models/res_partner.py
@@ -48,3 +48,11 @@ class ResPartner(models.Model):
         if "stage_id" in vals and vals.get("stage_id") in self._state_from:
             self.restart_validation()
         return res
+
+    @api.model_create_multi
+    def create(self, vals_list):
+        default_stage_id = self._get_default_stage_id().id
+        for vals in vals_list:
+            if not vals.get("stage_id") and default_stage_id:
+                vals["stage_id"] = default_stage_id
+        return super().create(vals_list)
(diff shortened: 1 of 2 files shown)


partner_tier_validation/models/res_partner.py
```diff
@@ -48,3 +48,21 @@ def write(self, vals):
...
+        return any([self.evaluate_tier(tier) for tier in tiers])
+
+    @api.model_create_multi
+    def create(self, vals_list):
+        res = super().create(vals_list)
+        default_stage = self._get_default_stage_id().id
+        for partner in res:
+            if partner._evaluate_valid_tiers() and partner.stage_id != default_stage:
+                partner.stage_id = default_stage
+        return res
```
dreispt: I don't totally understand the purpose of  `_evaluate_valid_tiers`.

The code also prevent creating Partners in a non default stage. For example that prevents importing partenrs as alerady confirmed. Is this a good idea?

I suggest simplifying and preprocess the create values:
```suggestion
        default_stage_id = self._get_default_stage_id().id
        for vals in vals_list:
            if not vals.get("stage_id") and default_stage_id:
                vals["stage_id"] = default_stage_id
        return super().create(vals_list)
```

Kev-Roche: Hi, Thank you for your review. 
Indeed, your proposition is simplier and I applied it.

dreispt (approved):

rrebollo (approved):
Code Review. LGTM!

rvalyi:
as per reviews,

/ocabot merge patch

PR #2072, opened 2025-06-11
[16.0][IMP] partner_contact_department: add active and sequence fields

vvrossem:


partner_contact_department/views/res_partner_department_view.xml
```diff
@@ -5,11 +5,36 @@
...
+        <field name="arch" type="xml">
+            <search>
+                <field name="name" string="Department Name" />
+                <field name="parent_id" string="Parent Department" />
+                <separator />
+                <filter
+                    string="Archived"
+                    name="inactive"
+                    domain="[('active', '=', False)]"
+                />
```
SilvioC2C, line 21: Non-blocking: to achieve better UX, you may wanna add the "All" filter too:
```suggestion
                <filter
                    string="Archived"
                    name="inactive"
                    domain="[('active', '=', False)]"
                />
                <filter
                    string="All"
                    name="all"
                    domain="['|', ('active', '=', False), ('active', '=', True)]"
                />
```
Else, you'll only ever see the active records, or the archived records, never both at once.

An alternative could be:
```suggestion
                <filter
                    string="Active"
                    name="active"
                    domain="[('active', '=', True)]"
                />
                <filter
                    string="Archived"
                    name="inactive"
                    domain="[('active', '=', False)]"
                />
```
This way, if you don't select neither of the 2 filters, you'll see only active records by default, if you select only 1 of them you'll see records accordingly, if you select both you'll see both active and archived records at once (as long as there's no `<separator/>` between the 2 filters).

SilvioC2C (approved):
Overall LGTM, improvement suggested

SilvioC2C (approved):

rrebollo (approved):
Code Review – LGTM! 👍  
However, I'd suggest enriching the description — you implemented more than what the current message reflects.

NL66278 (approved):
:+1: LGTM

PR #2073, opened 2025-06-11
[16.0][FIX] partner_contact_department: zip() takes no keyword arguments

vvrossem:
`strict=True` is available only in Python 3.10+ and is incompatible with Odoo 16.0 that runs on a lower version of Python.

### DIFF
--- a/partner_contact_department/models/res_partner.py
+++ b/partner_contact_department/models/res_partner.py
@@ -47,7 +47,7 @@ class ResPartnerDepartment(models.Model):
     def name_get(self):
         """Prepend parent name to department name."""
         result = super().name_get()
-        for position, ((id_, name), rec) in enumerate(zip(result, self, strict=True)):
+        for position, ((id_, name), rec) in enumerate(zip(result, self)):
             while rec.parent_id.name:
                 name = f"{rec.parent_id.name} / {name}"
                 rec = rec.parent_id


partner_contact_department/models/res_partner.py
```diff
@@ -47,7 +47,7 @@ def _compute_display_name(self):
     def name_get(self):
         """Prepend parent name to department name."""
         result = super().name_get()
-        for position, ((id_, name), rec) in enumerate(zip(result, self, strict=True)):
+        for position, ((id_, name), rec) in enumerate(zip(result, self)):
```
SilvioC2C, line 50: I would drop the `zip` completely, and define `rec = self.browse(id_)` within the loop itself.
It would be compatible with both Py<3.10 and Py>=3.10, and it wouldn't require `result` and `self` to have the same length to work properly.

SilvioC2C (approved):
LGTM, I suggested an improvement.

However, beware: it is not true that Odoo 16.0 runs on Py<3.10. For example, Odoo SA's runbot runs w/ Py 3.10 for Odoo 16.0, and odoo.sh also runs with it.

partner_contact_department/models/res_partner.py
vvrossem: I've tried: 
```
def name_get(self):
    """Prepend parent name to department name."""
    result = super().name_get()
    for position, (id_, name) in enumerate(result):
        rec = self.browse(id_)
        while rec.parent_id.name:
            name = f"{rec.parent_id.name} / {name}"
            rec = rec.parent_id
        result[position] = (id_, name)
    return result
```
but the following assertion error is raised: 
'''
AssertionError: 'test' != 'Dpt. 1 / test'
- test
+ Dpt. 1 / test
'''

I cannot investigate further, I'm keeping `zip` without keyword arguments

yajo (approved):
/ocabot merge patch

PR #2074, opened 2025-06-12
[18.0][FIX]partner_vat_unique: Update __manifest__.py application": False,

andreampiovesana:
"application": False,

### DIFF
--- a/partner_vat_unique/__manifest__.py
+++ b/partner_vat_unique/__manifest__.py
@@ -9,7 +9,7 @@
     "website": "https://github.com/OCA/partner-contact",
     "author": "Grant Thornton S.L.P, Odoo Community Association (OCA)",
     "license": "AGPL-3",
-    "application": True,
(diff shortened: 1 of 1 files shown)


rrebollo (commented):
LGTM!!! 🎉  
But looking at the history, it seems this behavior has always been like this, so... 🤷‍♂️

marcelofrare (approved):
LGTM

andreampiovesana:
> LGTM!!! 🎉 But looking at the history, it seems this behavior has always been like this, so... 🤷‍♂️

can you approve?

MaurizioPellegrinet (approved):
LGTM

pedrobaeza (changes requested):
Incorrect commit message:

https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#commit-message

PR #2087, opened 2025-07-22
[18.0][FIX] partner_stage: Add 'order' in _read_group_stage_id to fix group

Ricardo-MC:
Added 'order' argument with a default value to avoid errors in staging in the contacts view.

This PR replaces the previous one (https://github.com/OCA/partner-contact/pull/1985), which was closed.

### DIFF
--- a/partner_stage/models/res_partner.py
+++ b/partner_stage/models/res_partner.py
@@ -14,12 +14,13 @@ class Partner(models.Model):
         )
 
     @api.model
-    def _read_group_stage_id(self, states, domain):
-        return states.search([])
+    def _read_group_stage_ids(self, stages, domain):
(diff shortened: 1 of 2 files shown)


Ricardo-MC:
Hi @dreispt, @celm1990
I have updated the method signature to follow Odoo's convention and ensured that both the field and the method use the same name.
The ordering now uses stages._order or defaults to "sequence, id" as appropriate.

Let me know if you have any further suggestions.

bosd (approved):

celm1990:
@Ricardo-MC I think this PR can be closed in favor of https://github.com/OCA/partner-contact/pull/2056, which has already been merged

PR #2088, opened 2025-07-23
[16.0] [IMP] base_partner_sequence: unique constraint for partner 'ref'.

nicolas-delbovier-acsone:
Implements a unique constraint on `res.partner.ref` to guarantee
all partner references are unique if config setting `partner_generated_reference_unique` is True.

This PR branches from #1875.


StefanRijnhart (commented):
Functionally, I would expect this module to be able to enforce that sequences are unique across commercial partners, given that sequences are shared between commercial partners and their contacts in this module. Why did you implement the feature in a way that is not aligned with how the module worked so far?

Technically, please don't update the po and pot files manually. Also, do we need the setting on the company with its intricate getter and setter methods?

NL66278:
Comments not taken up in more then a year.

PR #2113, opened 2025-08-14
[16.0][IMP] partner_mobile_duplicate_warn: make module extendable

NL66278:


ntsirintanis (approved):

partner_mobile_duplicate_warn/models/res_partner.py
```diff
@@ -15,26 +16,44 @@ class ResPartner(models.Model):
         compute_sudo=True,
     )
 
-    @api.depends("mobile", "company_id")
+    @api.depends(lambda x: x._get_same_mobile_depends())
     def _compute_same_mobile_partner_id(self):
+        empty_recordset = self.env[self._name]
```
thomaspaulb, line 21: Isn't this usually done rather with `self.browse([])` ? Anyway it shouldn't matter

thomaspaulb (approved):

partner_mobile_duplicate_warn/models/res_partner.py
```diff
@@ -15,26 +16,44 @@ class ResPartner(models.Model):
...
+        domain = [("mobile", "=", self.mobile)]
+        if self.company_id:
+            domain += [
+                "|",
+                ("company_id", "=", False),
+                ("company_id", "=", self.company_id.id),
+            ]
+        self_id = self._origin.id
+        if self_id:
+            domain.append(("id", "!=", self_id))
```
thomaspaulb, line 67: We have a small issue here - let's say that we fill in number "+312345" in partner B, which does not satisfy the domain. Partner A however also has number "+312345", and does satisfy the domain. Now the change on partner B is disallowed, but partner B him/herself was not member of the domain, so the check shouldn't have triggered at all.

I think we can solve this by rather not including `("id", "!=", self_id)` and proceeding to find ALL records that satisfy the domain. Then, before deciding there is an error, we check if length(result) >= 2 and if "self_id" is part of the results. If not, then we don't complain.

thomaspaulb (changes requested):

partner_mobile_duplicate_warn/models/res_partner.py
```diff
@@ -15,26 +16,76 @@ class ResPartner(models.Model):
...
-                same_mobile_partner_id = same_mobile_partner.id or False
-            partner.same_mobile_partner_id = same_mobile_partner_id
+        domain = [("mobile", "=", self.mobile)]
+        if self.company_id:
+            domain += [
+                "|",
+                ("company_id", "=", False),
+                ("company_id", "=", self.company_id.id),
+            ]
+        self_id = (self._origin.id or self.id) if exclude_self else False
```
ntsirintanis, line 65: Shouldn't this be the other way around?

ntsirintanis: > Shouldn't this be the other way around?

nope

NL66278:
Another impossible case where the patch is 100% covered and still the overall project coverage decreases.

partner_mobile_duplicate_warn/models/res_partner.py
NL66278: I went from include to exclude to get rid of a double negation.

thomaspaulb (approved):

marcos-mendez (changes requested):
**Automated Review -- Tests Failed**

### 1. **Root cause** of the test failure

The test failure is caused by a **missing `phone_validation` dependency** in the module manifest. The `_onchange_mobile_validation()` method is called in the test, which requires the `phone_validation` module to be installed, but it's not declared in `__manifest__.py`.

### 2. **Suggested fix**

Add `"phone_validation"` to the `depends` list in `partner_mobile_duplicate_warn/__manifest__.py:

```python
# File: partner_mobile_duplicate_warn/__manifest__.py
{
    "name": "Partner Mobile Duplicate Warn",
    "depends": [
        "base",
        "phone_validation",  # <-- Add this line
    ],
    ...
}
```

### 3. **Additional code issues**

- **Potential issue in `_find_duplicate_mobile_partner`**:  
  In the method `_find_duplicate_mobile_partner()`, the logic assumes that `self` is always included in `found_partners` when `len(found_partners) >= 2`. However, this may not hold if `_get_same_mobile_domain(exclude_self=True)` is overridden in a subclass, and the override changes behavior in a way that excludes `self` from the domain even when it shouldn't be excluded (e.g., during constraint checking). This could lead to false negatives in duplicate detection logic.

  **Location**: `partner_mobile_duplicate_warn/models/res_partner.py`, lines 63–69  
  **Suggestion**: Clarify in docstring or add a comment that `self` must be included in `found_partners` for correct behavior, or consider a more robust check.

### 4. **Test improvements**

The current test only covers basic functionality but lacks coverage for edge cases. Here are specific improvements:

#### Suggested Test Cases:

1. **Test with `exclude_self=False` in `_get_same_mobile_domain`**:
   - Create two partners with same mobile.
   - Override `_get_same_mobile_domain` to return domain without excluding self.
   - Assert `_search_same_mobile_matches_inclusive` includes both partners.

2. **Test constraint-like behavior using `_find_duplicate_mobile_partner`**:
   - Create a partner with a duplicate mobile.
   - Ensure `_find_duplicate_mobile_partner()` correctly identifies the duplicate.
   - Add a case where no duplicate exists and verify it returns empty recordset.

3. **Test override of `_get_same_mobile_depends` and `_get_same_mobile_domain`**:
   - Create a subclass that overrides these methods.
   - Ensure that the computed field `same_mobile_partner_id` reflects the override logic correctly.

#### OCA Testing Pattern Recommendation:

Use `SavepointCase` or `TransactionCase` from `odoo.tests.common` to ensure proper test isolation and database state handling. Example:

```python
from odoo.tests.common import SavepointCase

class TestPartnerMobileDuplicateWarn(SavepointCase):
    @classmethod
    def setUpClass(cls):
        super().setUpClass()
        cls.env = cls.env(context=dict(cls.env.context, tracking_disable=True))
```

This ensures consistent behavior across different environments and avoids issues with caching or transaction rollback.

--- 

Let me know if you'd like a patch file or a pull request template for this fix.

---
### ⚠️ PR Aging Alert: CRITICAL

This PR by @NL66278 has been **waiting for 213 days** — that is over **7 months** without being merged or closed.
💤 Last activity was 75 days ago.

> _Every ignored PR is a contributor who might not come back. Review time matters. ([OCA Aging Report](https://git.pop.coop/pop/oca-reviewer))_

---
### Reciprocal Review Request



Hi everyone! I found some test failures on this PR and left detailed feedback above. I am happy to discuss or help debug. In the meantime, if any of you get a chance, I would appreciate a look at my open PR(s):

**My open PRs across OCA:**
- [server-tools#3554 [MIG] datetime_formatter: Migration to 18.0](https://github.com/OCA/server-tools/pull/3554)
- [server-tools#3548 [18.0][MIG] base_kanban_stage: Migration to 18.0](https://github.com/OCA/server-tools/pull/3548)
- [hr-attendance#262 [16.0][ADD] Hr_attendance_idsecure: iDSecure (ControliD) attendance integration](https://github.com/OCA/hr-attendance/pull/262)
- [stock-logistics-workflow#2276 [16.0][ADD] stock_move_line_devaluation](https://github.com/OCA/stock-logistics-workflow/pull/2276)
- [stock-logistics-workflow#2275 [16.0][ADD] Stock move line analytic account](https://github.com/OCA/stock-logistics-workflow/pull/2275)
- [stock-logistics-workflow#2268 [16.0][ADD] stock_move_line_picking_partner](https://github.com/OCA/stock-logistics-workflow/pull/2268)
- [purchase-workflow#2694 [16.0][IMP]Purchase workflow added to review state & exception fix](https://github.com/OCA/purchase-workflow/pull/2694)

Reviewing each other's work helps the whole community move forward. Thank you!

---
_Environment via [OCA Neural Reviewer](https://git.pop.coop/pop/oca-reviewer): Minikube + K8s Job + oca-ci/py3.10-odoo16.0 | Odoo 16.0_
_Automated review by [OCA Neural Reviewer](https://git.pop.coop/pop/oca-reviewer) + qwen3-coder:30b_

BhaveshHeliconia (commented):
- Figure out one issue here:
    - Create contact `contactA` with `mobile` set and `company` not set
    - Create contact `contactB` with same `mobile` and `company` set(Make sure that company must be different from currently enabled one)
    - Go to `contactA` form and observe duplicate `mobile` error with `contactB`


**Possible Solution:**
- Removing `compute_sudo=True` parameter on `same_mobile_partner_id` field would solve this issue.
- I've already added suggestion comment regarding the same on this PR: https://github.com/OCA/partner-contact/pull/2375

PR #2115, opened 2025-08-21
[17.0] [FIX] partner_identification: pre-commit po error

primes2h:
immagine


chienandalu (approved):
Although I'm not sure if there would be an issue with the weblate synchronization.

ValentinVinagre (approved):
It is possible that there is a conflict, it would have to be changed directly in the weblate and then the bot makes the commit.
I validate it by not blocking the fix.

pedrobaeza (approved):

PR #2117, opened 2025-08-22
[18.0][IMP] partner_firstname

legalsylvain:
### Refactor 

Iintroducing a `firstname.mixin`, that will be used by partners, users, etc... removing duplicated code.

image

### Required Fields

you can configure if firstname and / or lastname are required, in "Settings > General Settings":

image


cyrilmanuel (approved):

leemannd (approved):

legalsylvain:
Thanks ! 

/ocabot merge minor

legalsylvain:
hi @oca/crm-sales-marketing-maintainers could you merge this approved PR ? 

thanks !

partner_firstname/views/res_partner.xml
```diff
@@ -18,16 +18,16 @@
                     <field
                         invisible="not form_has_lastname_first"
                         name="lastname"
-                        required="form_has_lastname_first and not firstname and not is_company and type == 'contact'"
+                        required="form_has_lastname_first and (not firstname or form_partner_names_required_fields in ['firstname_lastname', 'lastname']) and not is_company and type == 'contact'"
```
NL66278: You could simplify the required formulas in the view a lot, if you have two computed fields in the model:
```
firstname_required = form_partner_names_required_fields in ['firstname_lastname', 'lastname']) and not is_company and type == 'contact'
and
lastname_required = form_partner_names_required_fields in ['firstname_lastname', 'lastname']) and not is_company and type == 'contact'
```
So for this instance of the lastname we would get:
`required="form_has_lastname_first and (not firstname or lastname_required)"` and similar for all the other required formulas.

legalsylvain: Hi @NL66278. thanks a lot for your suggestion.

In fact, I added also firstname and lastname values in the dependency of the "*_required" fields computation. As a result, XML code is very simple now. See : 

`                            required="form_has_lastname_first and lastname_required"`

CI is green. Could you update your review ? 

thanks !

NL66278 (approved):
👍 LGTM Great to do away with duplication of code and have being required configurable. 
Just to note that sometimes I wonder whether it would not have been more simple to just add firstname and lastname as separate fields (and in other modules middlename, second lastname etc.) apart from the name field, with all the trickery and jumping through burning hoops involved in splitting the name field, or conversely creating it from the name parts.

legalsylvain:
> 👍 LGTM Great to do away with duplication of code and have being required configurable.

Thanks ! :pray: 

>  Just to note that sometimes I wonder whether it would not have been more simple to just add firstname and lastname as separate fields (and in other modules middlename, second lastname etc.) apart from the name field, with all the trickery and jumping through burning hoops involved in splitting the name field, or conversely creating it from the name parts.

I don't know ! I just use this module since 6 monthes. Indeed, it is quite complex. I'm not sure it can be avoided. The more simple is to integrate first and last name fields in Odoo core. Another story ...

partner_firstname/models/base_config_settings.py
```diff
@@ -19,6 +19,14 @@ class ResConfigSettings(models.TransientModel):
         required=True,
         inverse="_inverse_partner_names_order",
     )
+
+    partner_names_required_fields = fields.Selection(
+        selection="_partner_names_required_fields_selection",
+        help="Define which fields are required or not",
+        config_parameter="partner_firstname.required_fields",
+        required=True,
+    )
```
CRogos, line 28: @NL66278 do you also get this warning in the update logs?

**Table 'res_config_settings': unable to set NOT NULL on column 'partner_names_required_fields'** 

It is trying to set the constraint bevor the **ir_config_parameter.xml** is loaded. 

Maybe we should set a default="no" instead of adding **ir_config_parameter.xml**?

```
2025-12-01 10:47:54,525 81 INFO c4a8-18-0-integration-26098166 odoo.modules.registry: module partner_firstname: creating or updating database tables 
2025-12-01 10:47:54,637 81 INFO c4a8-18-0-integration-26098166 odoo.schema: Table 'res_config_settings': unable to set NOT NULL on column 'partner_names_required_fields' 
2025-12-01 10:47:54,664 81 INFO c4a8-18-0-integration-26098166 odoo.modules.loading: loading partner_firstname/data/ir_config_parameter.xml 
```

PR #2118, opened 2025-08-23
[18.0][ADD] partner_firstname_auth_signup

legalsylvain:
This module is a glue module installed when OCA ``partner_firstname`` module and Odoo ``signup`` module are installed.

This allows the customer to enter their first and last names in the signup form.

image

depends on : 
- https://github.com/OCA/partner-contact/pull/2117
- https://github.com/OCA/partner-contact/pull/2123

PR #2123, opened 2025-08-31
[ADD] New module: partner_is_company_auth_signup

legalsylvain:
Add a new radiobox input in signup page, to know if the user is a company, or an individual. The feature is similar to the back-office main partner page.

It avoids to have bad information, when companies are creating accounts. (is_company is unchecked by default in Odoo Core)

image


rvalyi:
interesting feature. Sure you prefer to propose it as AGPL rather than LGPL? AGPL is often a bit of a stretch for portal modules...

legalsylvain:
> interesting feature. Sure you prefer to propose it as AGPL rather than LGPL? AGPL is often a bit of a stretch for portal modules...

I don't think that a lot of modules will depends on this one. It is a quite trivial one. 
And to a certain extent, I like the contamination of the AGPL. ;-)

rvalyi:
> > interesting feature. Sure you prefer to propose it as AGPL rather than LGPL? AGPL is often a bit of a stretch for portal modules...
> 
> I don't think that a lot of modules will depends on this one. It is a quite trivial one. 
> And to a certain extent, I like the contamination of the AGPL. ;-)

well eventually l10n_br_portal may depend on it as this is company thing is something that used to cause troubles. But after we change l10n_br_base to LGPL we would like to keep l10n_br_portal as LGPL as well. So this could be a use case. But not a big deal either.

cc @marcelsavegnago

legalsylvain:
> well eventually l10n_br_portal may depend on it as this is company thing is something that used to cause troubles

Hi. I understand That is used to cause troubles. But I don't understand why you could add a dependency to that module. 

This module just add a radio input on a form. no need to depend on it, just install it, IMO. Please, elaborate if you can.

thanks !

rvalyi:
well eventually you licensed it as AGPL without thinking too much about it and was ready to change. That's why I asked. But sure enough if it stays AGPL and we aim to put l10n_br_portal as LGPL, then we wouldn't depend on it, no big deal as I said.

maisim (approved):
code review + live test from http://oca-partner-contact-18-0-pr2123-7c268faf2446.runboat.odoo-community.org

Image
👏

legalsylvain:
@hparfr, could you make a review and put a little :+1:  on this one ? it is preventing the merge of https://github.com/OCA/partner-contact/pull/2118 that you reviewed.

thanks !

hparfr (approved):

legalsylvain:
Hi some @oca maintainers could merge this one ? 
thanks !

NL66278 (approved):
👍 LGTM

PR #2127, opened 2025-09-03
[17.0][MIG] partner_auto_archive: Migration to 17.0

bizzappdev:


NL66278:
Why make archiving a partner a scheduled action using a checkbox, when you can archive a partner directly. Maybe there is a good reason, but that should be stated in the README then.

partner_auto_archive/tests/test_partner_auto_archive.py
```diff
@@ -0,0 +1,34 @@
...
+        cls.contact_3 = cls.env["res.partner"].create(
+            {
+                "name": "Partner No Archive",
+                "auto_archive": False,
+            }
+        )
+
+    def test_partner_auto_archive(self):
+        self.cron.method_direct_trigger()
+        archived_contacts = self.env["res.partner"].search(
```
BhaveshHeliconia, line 30: Looks fragile. Use this instead: `archived = (self.contact_1 | self.contact_2).filtered(lambda p: not p.active)`

partner_auto_archive/tests/test_partner_auto_archive.py
```diff
@@ -0,0 +1,34 @@
...
+            [("active", "=", False), ("name", "ilike", "Partner Archive")]
+        )
+        self.assertEqual(len(archived_contacts), 2)
+        self.assertEqual(archived_contacts[0].auto_archive, False)
```
BhaveshHeliconia, line 34: - `assertFalse` can be used instead
- Should be checked for `contact_2` as well

BhaveshHeliconia (changes requested):

GuillemCForgeFlow (approved):
lgtm code review 👍🏿

twalter-c2c (approved):
I agree with @BhaveshHeliconia rearks regarding the test setup. Looks good otherwise.

PR #2103, opened 2025-08-04
[17.0][MIG] partner_contact_in_several_companies: Migration to 17.0

kwoychesko:
Attempt 2 to migrate `partner_contact_in_several_companies` to Odoo 17.0. Previous PR was flawed.

- Updates the `search` method to handle the removal of the `count` keyword argument in core.
- Ensures all tests pass on the new version.
- Squashed previous commits where it seemed logical.

PR #2131, opened 2025-09-04
[18.0][MIG] partner_contact_in_several_companies: Migration to 18.0

bizzappdev:


BhaveshHeliconia (changes requested):
See my suggestions on this PR: https://github.com/OCA/partner-contact/pull/2324

BhaveshHeliconia:
@bizzappdev could you please check the failing test/ci?

kwoychesko:
Hi @bizzappdev, @BhaveshHeliconia ,

I've opened a fresh 18.0 migration PR for this module: #2383 

It's rebased on current 18.0 and green on CI, and it also resolves the "attached contacts also visible on the Contacts kanban" point raised here (and on #2324): on 17.0+ the web client loads records via web_search_read, which bypassed the module's search() override, so the filtering is now done there. It keeps the original authorship chain from #2103 / #2324 

Since this PR has been stale for a few months, I thought a clean PR that also fixes the open review point might help move the module forward — but I'm of course happy to defer to the PSC / maintainers on which one to take. Thanks for the original migration work here and for such careful reviewing/suggesting fixes @BhaveshHeliconia!

PR #2135, opened 2025-09-15
[ADD] Migrate partner_coc to 16.0

ddejong-therp:
The module `partner_coc` does not exist anymore in 16.0, and it allowed a `res.partner.id.number` record to be used as a 'company registry' number field.
In odoo 16 we have `company_registry` on `res.partner` in the odoo base module, which should be the reason that `partner_coc` doesn't exist anymore.
But because the module doesn't exist anymore there isn't really a nice place to put this migration query.
So now I've put it in `partner_identification`.

This is a continuation from this PR: https://github.com/OCA/partner-contact/pull/1597
Pedro recommended putting it in the migration of the base module, but I feel that is still a weird place to migrate something which is not part of the odoo core modules.


NL66278 (approved):
:+1: LGTM

PR #1347, opened 2022-09-27
[FIX] partner_company_group: module was splitted, this allows safe remove

yankinmax:
This PR is needed:
To avoid possible conflicts having both `partner_company_group` and `base_partner_company_group` with related in this repository.
I suppose the next step will be to replace `partner_company_group` by `base_partner_company_group` and related.

Depends on:
- [x] https://github.com/OCA/partner-contact/pull/1346
- [x] https://github.com/OCA/partner-contact/pull/1245
- [x] https://github.com/OCA/partner-contact/pull/1246
- [x] https://github.com/OCA/partner-contact/pull/1247

PR #1870, opened 2024-10-16
[18.0][MIG] partner_company_group: Migration to 18.0

nguyenminhchien:
Depend on:
- [x] https://github.com/OCA/partner-contact/pull/1871
- [x] https://github.com/OCA/partner-contact/pull/1872
- [x] https://github.com/OCA/partner-contact/pull/1873
- [x] https://github.com/OCA/partner-contact/pull/1874

oca-port: [blacklist PR(s) , 1243, 1257, 1657 for partner_company_group](https://github.com/OCA/partner-contact/pull/1870/commits/eca7ab3ac5561db528903e4b2d47f729183ae121)

PR #2148, opened 2025-09-29
[18.0] [REM] partner_company_group: deprecated module

ivantodorovich:
This module is a left-over from a refactor done in 15.0:
- https://github.com/OCA/partner-contact/pull/1347 -> https://github.com/OCA/partner-contact/commit/7f3f531e

That refactor has split the module in 4:
- base_partner_company_group
- account_partner_company_group
- crm_partner_company_group
- sale_partner_company_group

The purpose is to clearly allow the installation of the "Company Group" feature without pulling all the complete applications like Sales or CRM.

Then, this module was left as an easy migration path for people that had the module installed in 15.0 previous to this refactor, with XML functions to update the XML ids of the previous views to the ones in the new modules.

As such, it should not have been migrated to the newer versions.

----

Moreover, we're experiencing a bug concerning these XML functions.

In the context of a migration from `14.0` (previous to the refactor) to `18.0`, we observe:
- The old `partner_company_group` views are still there, as orphan (unreferenced by any XML-id)
- The new views no longer have the `partner_company_group` XML-id, that should be set by the XML functions

I believe it's because the "easy migration path" hasn't been thought properly.
You see, this is what happens:

1. Starting from a version of the module prior to the refactor, the views are referenced as `partner_company_group.*`
2. Upgrade the module to a version after the refactor
3. Odoo will install the dependencies first, creating duplicates of the view with new XMLids
4. Odoo will execute the XML functions, adding the `partner_company_group.*` references to the new views. The old views are now orphan -- left unreferenced.
5. Odoo will attempt to remove the `partner_company_group.*` records, because they are no longer part of this module. It can't actually remove the records because they have 2x references each, so it only removes the `partner_company_group.*` reference.

So, you're left with the old records being unreferenced (orphan), and the new records being only referenced by their new XML-ids. The XML functions end up not doing anything useful, and even harmful --> as you end up with duplicated orphan views that mess up the partner views


cyrilmanuel (approved):

TDu (approved):

mmequignon (approved):

mmequignon:
@ivantodorovich I'm not quite sure about the policy of dropping a module.
I would think that a module should be deprecated and kept, but not migrated to higher versions.

@simahawk @pedrobaeza what's your take on this ?

ivantodorovich:
> @ivantodorovich I'm not quite sure about the policy of dropping a module. I would think that a module should be deprecated and kept, but not migrated to higher versions.

That's exactly what happened in 15.0, but then I think it was migrated to 18.0 by mistake:
- https://github.com/OCA/partner-contact/pull/1870#issuecomment-3347804076

NL66278 (approved):
👍 LGTM

luisDIXMIT:
It is also present in v19.0. Should it be removed? @ivantodorovich

ivantodorovich:
Yes, unfortunately it shouldn't have been migrated :/

NL66278:
Should we not just delete the module from the 19.0 branch then? Otherwise we will have the same problem in 20.0 and beyond...

ivantodorovich:
Absolutely!

PR #2149, opened 2025-09-30
[16.0][IMP] partner_company_type: add computed field display_name

mihien:
Add the field display_name so that both the abbreviation and the full name are shown. The abbreviation is put first to make the search easier at selection.


mihien:
Equivalent changes have already been done in the most recent version of the module, which I figured a bit too late. Closing the PR.

PR #2181, opened 2025-10-22
[18.0][ADD] partner_rank_single

yankinmax:
This module introduces a constraint to ensure that a contact cannot
simultaneously be both a customer and a supplier.


partner_rank_single/models/res_partner.py
```diff
@@ -0,0 +1,47 @@
...
+                        )
+                        % (
+                            field,
+                            record.display_name,
+                            "supplier_rank"
+                            if field == "customer_rank"
+                            else "customer_rank",
+                        )
+                    )
+        return super()._increase_rank(field, n=n)
```
ivantodorovich: I think this simpler approach would also work, wouldn't it?

```suggestion
        # OVERRIDE: to check single rank
        # Because of direct SQL update in the super method
        res = super()._increase_rank(field, n=n)
        self._constrains_single_rank()
        return res
```

ivantodorovich (commented):
Thanks! Can you check the CI? 🙏🏻

partner_rank_single/models/res_partner.py
yankinmax: It wasn't possible to pass the tests. I'll double check, thanks for the review

yankinmax:
> Thanks! Can you check the CI? 🙏🏻

Done, thanks for the review.

ivantodorovich (approved):

imlopes (approved):
LGTM.

I think the module name can be changed.
The module description is perfect but reading the module name I think in a list with a single value 😆 maybe anything like `partner_rank_restriction` but this is just a detail.

Thanks for this work

SilvioC2C (approved):

NL66278:
What is the use case? Why should it be impossible for a partner to be both a supplier and a customer? I know of many cases where that happens.

yankinmax:
> What is the use case? Why should it be impossible for a partner to be both a supplier and a customer? I know of many cases where that happens.

Hello @NL66278 , thanks for your question. There are some customers which want such a fucntional behavior. The first functional request is logical: to divide partners and be confident that a supplier is not a customer and vice versa.

I've in mind that probably someone can find such module useful for their projects, that's why it's proposed to OCA.

PR #2183, opened 2025-10-22
[18.0][IMP] `partner_manual_rank`: add a setting to disable automatic ranking

yankinmax:
Introduce the setting to disable automatic ranking.


partner_manual_rank/models/res_partner.py
```diff
@@ -72,3 +72,12 @@ def _default_is_customer(self):
...
     def _default_is_supplier(self):
         return self.env.context.get("res_partner_search_mode") == "supplier"
+
+    def _increase_rank(self, field, n=1):
+        # OVERRIDE: to consider the auto rank management setting
+        parameter = self.env["ir.config_parameter"].sudo()
+        get_param = parameter.get_param
+        if not str2bool(get_param("partner_manual_rank.partner_rank_auto")):
+            # Disable automatic ranking with an empty recordset
+            self = self.browse()
```
ivantodorovich: Why not just this?

```suggestion
            return
```

partner_manual_rank/readme/DESCRIPTION.md
```diff
@@ -1,3 +1,5 @@
 This module restores the functionality of \<= 12.0 to set contacts as
 customer or vendor, and allows users to create custom filters on a more
 user-friendly manner.
+
+The module introduces also the setting to disable automatic ranking.
```
ivantodorovich: Perhaps move this to CONFIGURE.md and elaborate a bit describing what automatic ranking is

ivantodorovich (approved):

cyrilmanuel (approved):

partner_manual_rank/models/res_partner.py
```diff
@@ -72,3 +72,11 @@ def _default_is_customer(self):
...
     def _default_is_supplier(self):
         return self.env.context.get("res_partner_search_mode") == "supplier"
+
+    def _increase_rank(self, field, n=1):
+        # OVERRIDE: to consider the auto rank management setting
+        parameter = self.env["ir.config_parameter"].sudo()
+        get_param = parameter.get_param
+        if not str2bool(get_param("partner_manual_rank.partner_rank_auto")):
+            return
+        return super()._increase_rank(field, n=n)
```
SilvioC2C, line 81: Non-blocking: you could include a boolean field on the company to make this setting company-specific, and fallback to the system-wide parameter in case the partner is not linked to a specific company (`res.partner.company_id` is not a required field).

SilvioC2C (approved):

partner_manual_rank/views/res_config_settings_view.xml
```diff
@@ -0,0 +1,19 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<odoo>
+    <record id="res_config_settings_view_form" model="ir.ui.view">
+        <field name="model">res.config.settings</field>
+        <field name="inherit_id" ref="base_setup.res_config_settings_view_form" />
+        <field name="arch" type="xml">
+            <block name="contacts_setting_container" position="inside">
+                <setting
+                    help="Uncheck to disable automatic ranking"
```
ivantodorovich:
```suggestion
                    help="Automatically rank partners when sale or purchase invoices are posted"
```

partner_manual_rank/models/res_partner.py
```diff
@@ -72,3 +72,11 @@ def _default_is_customer(self):
 
     def _default_is_supplier(self):
         return self.env.context.get("res_partner_search_mode") == "supplier"
+
+    def _increase_rank(self, field, n=1):
+        # OVERRIDE: to consider the auto rank management setting
```
NL66278, line 77: Make this a regular method doc comment (with triple double quotes).

partner_manual_rank/models/res_partner.py
```diff
@@ -72,3 +72,11 @@ def _default_is_customer(self):
...
+        parameter = self.env["ir.config_parameter"].sudo()
```
NL66278: I do not like this particular coding style. For one thing the variable parameter is not a parameter, but an Odoo model. Also I think it is a strange way to first put a method in a variable and then calling through the variable, where that is not needed at all here.
So please replace with:
```
ICP = self.env["ir.config_parameter"].sudo()
if not str2bool(ICP.get_param("partner_manual_rank.partner_rank_auto")):
        return
```

NL66278 (changes requested):

partner_manual_rank/models/res_partner.py
NL66278: Or alternatively, add the boolean field to company, and as related field to res.config.settings. Then no need anymore for the ir.config.parameter. And use from either the company in the partner or lacking that from the current company in the environment.

partner_manual_rank/models/res_partner.py
ivantodorovich: That would override the original method's docstring, so we tend to avoid it.

partner_manual_rank/models/res_partner.py
ivantodorovich: Hello! Thanks for your reviews

Partners can be shared among companies, so it'd be unclear which setting to read from in such cases. I don't think the current user company would be enough.

IMHO the system parameter is a good practice for this kind of things.

partner_manual_rank/models/res_partner.py
ivantodorovich: Yeap, I agree.

@yankinmax can you update these lines?

yankinmax: Yep, thanks for the reviews

sebalix (approved):

sebalix:
Error linked to `base_location_geonames_import` tests, the job was unable to resolve `download.geonames.org`, after a manual restart it's now working. Relaunching the merge:

/ocabot merge minor

PR #2184, opened 2025-10-27
[19.0][MIG] partner_manual_rank: Migration to 19.0  

cyrilmanuel:
test removed because they failled due to the return domain with "IN" on odoo 19.0 on fields.py : 
```

 :param str search: name of a method that implements search on the field.
        The method takes an operator and value. Basic domain optimizations are
        ran before calling this function.
        For instance, all ``'='`` are transformed to ``'in'``, and boolean
        fields conditions are made such that operator is ``'in'``/``'not in'``
        and value is ``[True]``.

        The method should ``return NotImplemented`` if it does not support the
        operator.
        In that case, the ORM can try to call it with other, semantically
        equivalent, operators. For instance, try with the positive operator if
        its corresponding negative operator is not implemented.
        The method must return a :ref:`reference/orm/domains` that replaces
        ``(field, operator, value)`` in its domain.

        Note that a stored field can actually have a search method. The search
        method will be invoked to rewrite the condition. This may be useful for
        sanitizing the values used in the condition, for instance.

        .. code-block:: python

            def _search_partner_ref(self, operator, value):
                if operator not in ('in', 'like'):
                    return NotImplemented
                ...  # add your logic here, example
                return Domain('partner_id.ref', operator, value)
```


leemannd (approved):
Administrative and translations commits should be squashed

yankinmax (approved):
Technical migration seems legit, thanks

partner_manual_rank/tests/test_res_partner.py
```diff
@@ -52,8 +51,6 @@ def test_01_is_customer(self):
                 {"is_customer": True, "customer_rank": 1},
             ],
         )
-        with self.assertRaisesRegex(UserError, "Operation not supported"):
```
MiquelRForgeFlow: why remove? Shouldn't you test `NotImplemented`?

MiquelRForgeFlow (approved):

em230418:
Clicked on "runboat/build" link, {"detail":"Not Found"} is returned

PR #2188, opened 2025-10-28
[19.0][ADD] Module partner_title

alexis-via:
res.partner.title was dropped in 19.0.
This module restores title on partners.
I added "active" and "sequence" on res.partner.title, so no more need for OCA modules partner_title_active (cf 17.0) and partner_title_order (cf 18.0)

partner_title


alexis-via:
Feel free to propose a better icon than the "right hand" that I have used !

alexis-via:
I changed to use the "user" icon, it's better

new_icon

partner_title/models/res_partner.py
```diff
@@ -0,0 +1,46 @@
...
+    _inherit = "res.partner"
+
+    title_id = fields.Many2one(
+        "res.partner.title",
+        ondelete="restrict",
+        compute="_compute_title_id",
+        store=True,
+        precompute=True,
+        readonly=False,
+    )
```
AungKoKoLin1997, line 18: I think you created a compute field just to clear the title when the partner type is set to company.
I’d prefer to avoid adding multiple attributes and instead use a standard Many2one field, clearing the value through an `onchange`.

partner_title/models/res_partner.py
```diff
@@ -0,0 +1,46 @@
...
+    name_with_title = fields.Char(compute="_compute_name_with_title")
```
AungKoKoLin1997, line 20: Do you have any specific purpose for using it somewhere?
Right now, it’s not even displayed in the view.

partner_title/views/res_partner_title.xml
```diff
@@ -0,0 +1,78 @@
...
+        name="Contact Titles"
+        sequence="500"
+    />
+
+    <menuitem
+        id="res_partner_title_menu"
+        action="res_partner_title_action"
+        parent="res_partner_title_parent_menu"
+        sequence="10"
+    />
```
AungKoKoLin1997, line 77: Shouldn’t we add this menu under the Configuration section of the Contacts app, like in the previous versions, by adding `contacts` as a dependency?

partner_title/views/res_partner_title.xml
```diff
@@ -0,0 +1,78 @@
...
+                    />
+                    <group name="main">
+                        <field name="name" />
+                        <field name="shortcut" />
+                        <field name="active" invisible="1" />
+                    </group>
+                </sheet>
+            </form>
+        </field>
+    </record>
```
AungKoKoLin1997, line 27: Isn’t it enough to just have an editable tree view?

partner_title/models/res_partner.py
```diff
@@ -0,0 +1,46 @@
...
+    @api.constrains("title_id", "is_company")
+    def _check_title(self):
+        for partner in self:
+            if partner.is_company and partner.title_id:
+                raise ValidationError(
+                    self.env._(
+                        "Partner '%(name)s' is a company, so it cannot have a title.",
+                        name=partner.display_name,
+                    )
+                )
```
AungKoKoLin1997: I found an issue when creating a new record — if I change the partner type to Person, fill in the title, and save, it causes a problem.

Screenshot 2025-10-29 102041

partner_title/models/res_partner.py
```diff
@@ -0,0 +1,46 @@
...
+    )
+    name_with_title = fields.Char(compute="_compute_name_with_title")
+
+    @api.depends("is_company")
+    def _compute_title_id(self):
+        for partner in self:
+            if partner.is_company:
+                partner.title_id = False
+
+    @api.depends("is_company", "title_id")
```
AungKoKoLin1997: This will do nothing for a non-stored computed field.
```suggestion
```

partner_title/models/res_partner.py
```diff
@@ -0,0 +1,46 @@
...
+    def _compute_name_with_title(self):
+        for partner in self:
+            name = partner.name
+            if not partner.is_company and partner.title_id:
+                title = partner.title_id.shortcut or partner.title_id.name
+                name = " ".join([title, name])
+            partner.name_with_title = name
```
AungKoKoLin1997, line 35:
```suggestion
    def _compute_name_with_title(self):
        for partner in self:
            partner.name_with_title = partner.name
            if not partner.is_company and partner.title_id:
                title = partner.title_id.shortcut or partner.title_id.name
                partner.name_with_title = " ".join([title, partner.name_with_title])
```

partner_title/models/res_partner.py
alexis-via: it's very useful in reports.

partner_title/views/res_partner_title.xml
alexis-via: yep. See my PR https://github.com/OCA/partner-contact/pull/2189 that provides a glue module with contacts.

partner_title/views/res_partner_title.xml
alexis-via: I personally prefer modules that provide both a tree view and a form view.

partner_title/models/res_partner.py
alexis-via: Are you really sure about this? I don't think so.

partner_title/models/res_partner.py
alexis-via: I prefer my implementation...

partner_title/models/res_partner.py
alexis-via: We now try to avoid onchange as much as we can, especially in new code. That's what I'm doing here.

partner_title/models/res_partner.py
alexis-via: I added a comment about that to explain.

partner_title/models/res_partner.py
alexis-via: Arf !!! Looks like a 19.0 framework issue: when we enter the constraint code, neither is_company nor company_type has the proper value.
As this constraint is really not important, I removed it.

partner_title/models/res_partner.py
AungKoKoLin1997: Of course I’m sure. Why don’t you check and see?

alexis-via: Here are some examples in odoo code where they use @api.depends on non-stored fields:
https://github.com/odoo/odoo/blob/19.0/odoo/addons/base/models/res_partner.py#L403
https://github.com/odoo/odoo/blob/19.0/addons/account/models/partner.py#L1009
https://github.com/odoo/odoo/blob/19.0/addons/account/models/partner.py#L998
But, of course, the code of Odoo may be wrong.
I thought that @api.depends on non-stored field was useful to invalidate the cache.
Of course, it is not as important as on stored fields where it is vital...

AungKoKoLin1997: I checked the behavior in detail and found that without the @api.depends decorator, the compute method is triggered when changing the name field, but the name value inside the method still shows the old value before the change, resulting in an incorrect name_with_title. It also does not trigger when other fields are changed. (Seems like strange)

With the @api.depends decorator, the name_with_title field updates instantly before saving when title is changed.

Since the field isn’t displayed in the view, having or not having the decorator doesn’t affect the final result — the correct name_with_title appears after saving in both cases. Therefore, we can remove the decorator. However, if you prefer to keep it, the name field should also be included in the @api.depends.

alexis-via: dropped

alexis-via: The removal of @api.depends on _compute_name_with_title() broke the tests ! That's the best proof that @api.depends is useful on non-stored fields.
@AungKoKoLin1997 Please, don't say "Of course I’m sure. Why don’t you check and see?" when you're completely wrong. You made me loose my time.

alexis-via: I restored @api.depends => tests are green again.

vernichon (approved):

partner_title/models/res_partner.py
```diff
@@ -0,0 +1,35 @@
...
+# @author: Alexis de Lattre <alexis.delattre@akretion.com>
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import api, fields, models
+
+
+class ResPartner(models.Model):
+    _inherit = "res.partner"
+
+    title_id = fields.Many2one(
```
yostashiro, line 11: I think the field name should be changed to `title` following 18.0, or a migration script needs to be added?

alexis-via: A migration script will be needed anyway (for data), so let's put proper field names.

partner_title/models/res_partner.py
AungKoKoLin1997: I think the test failure was due to the cache not being cleared in the test case after updating `title_id` to False, which is required when removing @api.depends.
Still, keeping `@api.depends` is a better approach, as it ensures automatic cache invalidation and avoids manual handling.

CedricAtanane (approved):
Everything looks good on my side. Thanks, Alexis, for this module and also for partner_mobile, which comes at the perfect time for one of my projects.

NL66278 (approved):
👍 LGTM

alexis-via:
Thanks for the reviews and the merge. I rebased my PR for the glue module partner_title_contacts https://github.com/OCA/partner-contact/pull/2189 ; tests are green.

PR #2193, opened 2025-10-28
[19.0][ADD] Module partner_mobile

alexis-via:
Small module to restore the field "mobile" on res.partner, which was dropped in odoo 19.0

partner_mobile


vernichon (approved):

bosd (approved):

alexis-via:
Thanks for your reviews and the merge. There is also the PR for the module partner_mobile_validation, which is a glue module between partner_mobile and phone_validation https://github.com/OCA/partner-contact/pull/2194 I rebased it and tests are green, so it is ready for review/merge.

PR #2194, opened 2025-10-28
[19.0][ADD] Module partner_mobile_validation

alexis-via:
This is a glue module between partner_mobile (#2193 ) and phone_validation.


partner_mobile_validation/models/res_partner.py
```diff
@@ -0,0 +1,20 @@
...
+# @author: Alexis de Lattre <alexis.delattre@akretion.com>
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import api, models
+
+
+class ResPartner(models.Model):
+    _inherit = "res.partner"
+
+    @api.onchange("mobile", "country_id", "company_id")
```
NL66278, line 11: Why not @depends?

alexis-via: I made exactly the same implementation as phone_validation: https://github.com/odoo/odoo/blob/19.0/addons/phone_validation/models/res_partner.py#L10

NL66278 (approved):
👍 LGTM

PR #2195, opened 2025-10-31
[17.0][IMP] base_location_nuts

edescalona:
@BinhexTeam

Correcting XPath in the form view that adds NUT fields to the contact.

Problem:
When installing the `base_vat` module, the NUT fields distort the design of the contact view, displaying the VAT value lower down and making the NUT fields unidentifiable.

image

Solution:
Add the NUT fields after the address div to ensure they display correctly regardless of the installed modules.

image

I believe this is the best option without having to create another add-on just to correct this visual issue. Let me know what you think. @rafaelbn @edlopen


edescalona:
ping @edlopen @rafaelbn

rrebollo (approved):
Code Review. LGTM! In the commit message maybe "[FIX]" prefix would fit better.

edescalona:
ping @rrebollo  @edlopen @rafaelbn

jelenapoblet (approved):
Thank you for the fix, it’s working correctly now.

etobella (approved):
LGTM

/ocabot merge patch

Can you port the change to other versions? 🙏

PR #2197, opened 2025-11-03
[17.0][ADD] partner_enforce_iban_validation: new module

Mohajiro:
Add configuration to enforce strict IBAN validation that raises exceptions for invalid IBANs. This addresses the issue where Odoo's built-in IBAN validation catches exceptions silently during bulk imports, but users need strict validation in normal operations.

The module provides:
- Global switch to enable/disable strict IBAN validation
- Country-based filtering for banks and partners
- Maintains compatibility with existing Odoo IBAN validation


BlazejewskaA (changes requested):
Steps to reproduce
In the configuration, it says: "Go to 'Settings > Contacts' and activate the 'Raise exception on invalid IBAN' checkbox."

Expected behaviour

In practice, the "Raise exception on invalid IBAN" checkbox is in Settings > Statistics, so the description needs to be corrected.

BlazejewskaA (changes requested):
Steps to reproduce

go to stettings-statistics- mark checkbox " Enforce IBAN Validation"
choose a country for bank and partner- for example Poland
go to contacts- create a new one
mark country of new contact as Poland
go to invoicing tab in contact- add new bank acount
add a wrong iban- for example  DE89370400440532013000
choose alior bank- added as polish

Expected behaviour

Odoo shows an error about an invalid IBAN.- and as I understand "A validation error will be raised if the IBAN validation fails."
Current behaviour

a standard odoo notification that it might be wrong IBAN.  Exactly the same behavior in Odoo as when the "Enforce IBAN Validation" checkbox is not selected.

BlazejewskaA (approved):
functional LGTM

ivs-cetmix (approved):
Code review LGTM

ivs-cetmix:
Hey @oca/crm-sales-marketing-maintainers would appreciate your review! Looks like this is one of the long time discussed feature, so here is finally a PR 😉

partner_enforce_iban_validation/models/__init__.py
```diff
@@ -0,0 +1,5 @@
+# Copyright (C) 2025 Cetmix OÜ
```
NL66278: There should be no copyright lines in init files.

Mohajiro: @NL66278 Thank you for your review and for your proposal. The issue has been fixed. Could you please check if it is okay for you now?

partner_enforce_iban_validation/readme/CONFIGURE.md
```diff
@@ -0,0 +1,3 @@
+- Go to "Settings > Contacts" and activate the "Raise exception on invalid IBAN" checkbox.
+- Use "Bank Countries" to validate only accounts of banks located in selected countries.
+- Use "Partner Countries" to validate only partners located in selected countries.
```
NL66278, line 3: Why not only check for the bank country? An EU citizen might have a bank account in another country, like China or the USA, that does not have IBAN valuation?

Mohajiro: Yes, that's exactly how it works! The module skips country validation for non-IBAN countries like USA and China, so EU citizens with accounts there can use any valid IBAN format without country mismatch errors.

partner_enforce_iban_validation/models/res_partner_bank.py
```diff
@@ -0,0 +1,176 @@
...
+            iban_country_code: Two-letter country code from IBAN
+            country_type: 'bank' or 'partner' for error message
+
+        Raises:
+            ValidationError: country supports IBAN but doesn't match IBAN country code
+        """
+        if not country:
+            return
+
+        if not country.code:
```
NL66278: code field is required in res.country. so this check is not needed

NL66278 (changes requested):
I think what the module does is much to complicated for what it needs to accomplish: enforce IBAN validation on banks in countries that have IBAN validation. Adding validation for where the partner lives is plain wrong in my view. Having a configuration for which countries we check the bank country code is also questionable. If a country code for the banks country is in the IBAN templates, the account should be an IBAN account. I do not see a use case for enabling validation for let us say a bank in Belgium, and not enable it for banks in France.

ivs-cetmix:
Hi @NL66278 , thank you for your feedback!

> I think what the module does is much to complicated for what it needs to accomplish: enforce IBAN validation on banks in countries that have IBAN validation. Adding validation for where the partner lives is plain wrong in my view. 

We have a real case where we need such a validation. It's optional anyway, so if someone doesn't need it he can simply leave this setting empty.

> Having a configuration for which countries we check the bank country code is also questionable. If a country code for the banks country is in the IBAN templates, the account should be an IBAN account. I do not see a use case for enabling validation for let us say a bank in Belgium, and not enable it for banks in France.

Same here. Just leave the setting blank.

partner_enforce_iban_validation/models/res_partner_bank.py
Mohajiro: Thank you for catching this. The unnecessary country.code check has been removed.

For the logical aspects, please see Ivan's feedback.

NL66278:
>>Hi @NL66278 , thank you for your feedback!

I think what the module does is much to complicated for what it needs to accomplish: enforce IBAN validation on banks in countries that have IBAN validation. Adding validation for where the partner lives is plain wrong in my view.

>We have a real case where we need such a validation. It's optional anyway, so if someone doesn't need it he can simply leave this setting empty.

>>Having a configuration for which countries we check the bank country code is also questionable. If a country code for the banks country is in the IBAN templates, the account should be an IBAN account. I do not see a use case for enabling validation for let us say a bank in Belgium, and not enable it for banks in France.

>Same here. Just leave the setting blank.

But the partner validation means that if I add for instance the Netherlands to the countries to validate for partner, no Dutch partner at all can have any bank account that not has a Dutch account number. I can neither have accounts in non IBAN countries like China, nor even in neighboring countries, like Belgium. I am really curious what the usecase for this is.

NL66278:
Just did a small test on runbot. If I enable the IBAN enforcement, adding an Chinese account for a bank in Hong Kong, with a customer living in the USA results in an IBAN error message, even though I did not configure any country for either banks or partners. 
Screenshot from 2025-11-18 16-07-37

NL66278:
Actually I can no longer add a Chinese bank account to a Chinese company.

NL66278:
The closing was just me clicking the wrong button....

ivs-cetmix:
@NL66278 
> Just did a small test on runbot. If I enable the IBAN enforcement, adding an Chinese account for a bank in Hong Kong, with a customer living in the USA results in an IBAN error message, even though I did not configure any country for either banks or partners.

This is because IBAN is being check for all accounts and obviously 1234567 is not a valid IBAN account.

NL66278:
> @NL66278
> 
> > Just did a small test on runbot. If I enable the IBAN enforcement, adding an Chinese account for a bank in Hong Kong, with a customer living in the USA results in an IBAN error message, even though I did not configure any country for either banks or partners.
> 
> This is because IBAN is being check for all accounts and obviously 1234567 is not a valid IBAN account.

Well this is precisely what is wrong. Accounts for banks that are in countries not supporting IBAN should not be checked.

ivs-cetmix:
> > @NL66278
> > This is because IBAN is being check for all accounts and obviously 1234567 is not a valid IBAN account.
> 
> Well this is precisely what is wrong. Accounts for banks that are in countries not supporting IBAN should not be checked.

Thank you , now I got your point 👍

partner_enforce_iban_validation/models/res_partner_bank.py
```diff
@@ -136,8 +137,17 @@ def _check_iban(self):
                 and record.partner_id.country_id.id in partner_country_ids
             )
 
+            # Check if bank country supports IBAN (only for banks with country)
+            bank_country_supports_iban = (
+                record.bank_id
+                and record.bank_id.country
```
ivs-cetmix: There may be multiple cases where bank is not provided at all, just IBAN. This means this check will not work at all.
You can check how the same issue is already [solved](https://github.com/odoo/odoo/blob/686f4d250daf93c20462217cd3353251490fb7f0/addons/base_iban/models/res_partner_bank.py#L33) in the original module.

ivs-cetmix (changes requested):

partner_enforce_iban_validation/models/res_partner_bank.py
```diff
@@ -0,0 +1,186 @@
...
+            partner_country_match = (
+                record.partner_id
+                and record.partner_id.country_id
+                and record.partner_id.country_id.id in partner_country_ids
+            )
+
+            # Resolve country the same way as in base_iban:
+            # bank country -> partner country -> country from IBAN prefix
+            country = (record.bank_id and record.bank_id.country) or (
+                record.partner_id and record.partner_id.country_id
```
ivs-cetmix: A partner from Brazil can have an account in a German bank.

partner_enforce_iban_validation/models/res_partner_bank.py
```diff
@@ -0,0 +1,186 @@
...
+            )
+            if not country and len(iban) >= 2:
+                country = self.env["res.country"].search(
```
ivs-cetmix: 200 records will generate up to 200 `search()`
And why do we even need to search for the country? We need to check if the account prefix matches the one in the _map_iban_template.

ivs-cetmix (changes requested):

partner_enforce_iban_validation/models/res_partner_bank.py
Mohajiro: The logic no longer depends on bank_id: if there's no bank, we detect IBAN support directly from the IBAN prefix, same as in the base module.

partner_enforce_iban_validation/models/res_partner_bank.py
Mohajiro: The country is no longer taken from the partner when the bank is missing.

partner_enforce_iban_validation/models/res_partner_bank.py
Mohajiro: Fixed. All res.country.search() calls were removed.
We now check the prefix against _map_iban_template like in the original validate_iban().

partner_enforce_iban_validation/models/res_partner_bank.py
ivs-cetmix: Code tells exactly the opposite:
```python
country = (record.bank_id and record.bank_id.country) or (
                record.partner_id and record.partner_id.country_id
```

partner_enforce_iban_validation/models/res_partner_bank.py
```diff
@@ -0,0 +1,188 @@
...
+
+            country_supports_iban = (
+                country and country.code.lower() in _map_iban_template
+            )
+
+            # IBAN prefix detection (used when no bank country is available)
+            iban_country_code = iban[:2].lower() if len(iban) >= 2 else None
+            iban_country_supported = (
+                iban_country_code in _map_iban_template if iban_country_code else False
+            )
```
ivs-cetmix: Why do we need two separate variables and two separate checks that do absolutely the same thing?

ivs-cetmix (changes requested):

partner_enforce_iban_validation/models/res_partner_bank.py
Mohajiro: Fixed

partner_enforce_iban_validation/models/res_partner_bank.py
```diff
@@ -0,0 +1,185 @@
...
+                and record.partner_id.country_id.id in partner_country_ids
+            )
+
+            # IBAN prefix detection
+            iban_country_code = iban[:2].lower() if len(iban) >= 2 else None
+            iban_country_supported = (
+                iban_country_code in _map_iban_template if iban_country_code else False
+            )
+
+            if not bank_country_ids and not partner_country_ids:
```
ivs-cetmix: Why do we even need to check this if iban country is not supported?

ivs-cetmix (approved):
Code review LGTM.

@NL66278 would appreciate your functional review, thank you!

twendt-nc (approved):

partner_enforce_iban_validation/models/res_partner_bank.py
```diff
@@ -0,0 +1,185 @@
...
+                else:
+                    # No bank country → validate if IBAN prefix is supported
+                    should_validate = iban_country_supported
+            elif (
+                bank_country_match or partner_country_match
+            ) and iban_country_supported:
+                # Filters set → enforce for matching records with supported IBAN prefix
+                should_validate = True
+            else:
+                should_validate = False
```
NL66278, line 158: Although the code is convoluted and thereby difficult to follow I think that what happens here is that when you have a bank that should be validated (bank country in configured parameter), but the bank account does not start with the country code (and therefore iban_country_supported is False), you will skip validation, while this should trigger an error.

NL66278:
I would propose taking a completely different approach to get the result this module claims to have.

* Add a boolean parameter to res.country (and add to res.country view) to check whether banks in this country should enforce IBAN;
* Determine the country to use for validation, check (in that order), country_id in the bank, if that is not filled, country_id of the partner, if that is not filled country of current company;
* If the country so determined has IBAN checking set: validate the IBAN, raise ValidationException when not valid;
* If IBAN valid (and needed) check whether iban country code in accountnumber is equal to country code in country, raise Validation error if that is not the case.

As stated before, I do not see any valid reason to make this check dependent on the partner. With the setup as proposed a partner in the Netherlands can have a bank account in China without IBAN, while a Chinese partner that wants an account with a Dutch bank will have a valid IBAN enforced.

No need for ir.config parameters with methods to set and read those.

ivs-cetmix:
> I would propose taking a completely different approach to get the result this module claims to have.
> 
> * Add a boolean parameter to res.country (and add to res.country view) to check whether banks in this country should enforce IBAN;
> * Determine the country to use for validation, check (in that order), country_id in the bank, if that is not filled, country_id of the partner, if that is not filled country of current company;
> * If the country so determined has IBAN checking set: validate the IBAN, raise ValidationException when not valid;
> * If IBAN valid (and needed) check whether iban country code in accountnumber is equal to country code in country, raise Validation error if that is not the case.
> 
> As stated before, I do not see any valid reason to make this check dependent on the partner. With the setup as proposed a partner in the Netherlands can have a bank account in China without IBAN, while a Chinese partner that wants an account with a Dutch bank will have a valid IBAN enforced.
> 
> No need for ir.config parameters with methods to set and read those.

Thanks you, this is a really good idea! Some thoughts though:

1. With settings in the `res.country` how will it be possible to enforce validation for all countries that are using IBAN? And I think this would be the most common scenario. Currently this can be done by enabling the check while leaving the county list empty.
2.  Regarding the partner check, this is an optional parameter which can be simply skipped if not needed. We have at least one real case where it's being used.

PR #2200, opened 2025-11-05
[16.0][IMP] partner_multi_relation: allow to archive relation types

marielejeune:
Basic improvements to make it possible to archive relation types.

If this PR is once approved, I can forward-port it to higher versions.

NB: I've no idea why a test that is not related to my change is failing. If someone can help, feel free.

### DIFF
--- a/partner_multi_relation/models/res_partner_relation_type.py
+++ b/partner_multi_relation/models/res_partner_relation_type.py
@@ -22,6 +22,7 @@ class ResPartnerRelationType(models.Model):
 
     name = fields.Char(required=True, translate=True)
     name_inverse = fields.Char(string="Inverse name", required=True, translate=True)
+    active = fields.Boolean(default=True)
     contact_type_left = fields.Selection(
         selection="get_partner_types", string="Left partner type"
     )


NL66278 (changes requested):
👍 Thank you for your contribution, but could you also modify the views to allow archiving, unarchiving and filtering?

marielejeune:
Hi @NL66278 : archiving and unarchiving is allowed by default in Odoo v16 as soon as you add the `active` field.
Concerning the filtering, this is not done but it is easy to create a custom filter and save it in your preferences if you need it.

Note that I'll however change this PR state to draft because archived relation types are still useable on res.partner.relation records, which should be avoided. However I don't have time to work on it right now.
Feel free to contribute if you need this. Thanks

rousseldenis:
@NL66278 This was not ready

marielejeune:
@NL66278 As I told you yesterday, this is not ready since it is still possible to create, from the UI, relations with archived relation types. Hence this adding is useless. This is why I changed my PR state to draft ...

NL66278:
@marielejeune You can add the additional code later and if OK I will merge it as well.

marielejeune:
> @marielejeune You can add the additional code later and if OK I will merge it as well.

Ok but I don't see the point of merging something that is useless and doesn't work. If some people try to use the module because they see it available, they won't understand why it doesn't work ...

PR #2202, opened 2025-11-06
[FIX] parter_multi_relation: Reorganize menu hierachy

u0f:
Before this change, in the Contacts module, the menu items did not follow a consistent hierarchical structure.

Before
image

After

image

MT-12511 @moduon @yajo @rafaelbn

### DIFF
--- a/partner_multi_relation/views/menu.xml
+++ b/partner_multi_relation/views/menu.xml
@@ -3,12 +3,12 @@
     <menuitem
         id="menu_res_partner_relation"
         name="Relations"
-        sequence="3"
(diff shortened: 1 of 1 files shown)


Shide:
I think this should be a PR for Odoo

yajo (approved):
I agree with @Shide. Please try to open that Odoo PR and leave this one as draft.

If they don't merge, then we can still merge this one, as it also is a good fix.

u0f:
Odoo PR opened https://github.com/odoo/odoo/pull/234909

PR #2190, opened 2025-10-28
[18.0] [FIX] partner_property: broken partner/user export

ivantodorovich:
**Steps to reproduce:**

1. Go in the partner or users list view
2. Select a few records
3. Click on Actions > Export

**Expected:**

The Export wizard should open.

**Result:**

https://github.com/user-attachments/assets/804f6019-d845-48dd-a86a-f6658ba63f77

```
File "/odoo/src/odoo/addons/web/controllers/export.py", line 400, in get_fields
    exportable_fields.update(self._get_property_fields(fields, model, domain=domain))
                            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
File "/odoo/src/odoo/addons/web/controllers/export.py", line 321, in _get_property_fields
    field_to_get = Model._field_to_sql(Model._table, definition_record, self_subquery)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
File "/odoo/src/odoo/odoo/models.py", line 2976, in _field_to_sql
    raise ValueError(f"Cannot convert {field} to SQL because it is not stored")
ValueError: Cannot convert res.partner.properties_company_id to SQL because it is not stored
```

PR #2208, opened 2025-11-10
[18.0][FIX] partner_property: export of res.users

ivantodorovich:
Originally attempted to fix it here: 36b93fa723f020e2d0f5e7e85166f3574982ce29 (https://github.com/OCA/partner-contact/pull/2190)

But that fix only worked for res.partners due to not properly using the alias in the `_field_to_sql` method.
Then, when trying to export res.users (or any model that inherits from res.partner), we'd get the following error:

```
  File ".../odoo/addons/web/controllers/export.py", line 324, in _get_property_fields
    definition_records = target_model.search_fetch(
                         ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".../odoo/odoo/models.py", line 1781, in search_fetch
    return self._fetch_query(query, fields_to_fetch)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".../odoo/odoo/models.py", line 4240, in _fetch_query
    rows = self.env.execute_query(query.select(*sql_terms))
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".../odoo/odoo/api.py", line 993, in execute_query
    self.cr.execute(query)
  File ".../odoo/odoo/sql_db.py", line 582, in execute
    return self._cursor.execute(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".../odoo/odoo/sql_db.py", line 357, in execute
    res = self._obj.execute(query, params)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
psycopg2.errors.AmbiguousColumn: column reference "company_id" is ambiguous
LINE 1: ...TRUE) AND ("res_company"."id" IN (SELECT COALESCE(company_id...
```

Sorry for the trouble. I missed this last time

### DIFF
--- a/partner_property/tests/test_export.py
+++ b/partner_property/tests/test_export.py
@@ -3,10 +3,11 @@
 
 import json
 
-from odoo.tests import HttpCase
+from odoo.tests import HttpCase, users
 
 
 class TextPartnerExport(HttpCase):
+    @users("admin")
     def test_export_get_fields(self):
         """Text the export wizard get_fields method.
 
@@ -28,7 +29,6 @@ class TextPartnerExport(HttpCase):
                 raise ValueError(f"Cannot convert {field} to SQL because it is not stored")
             ValueError: Cannot convert res.partner.properties_company_id to SQL because it is not stored
         """  # noqa: E501
-        self.authenticate("admin", "admin")
         self.url_open(
             "/web/export/get_fields",
             data=json.dumps(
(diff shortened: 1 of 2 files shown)


partner_property/tests/test_export.py
```diff
@@ -42,3 +42,20 @@ def test_export_get_fields(self):
             ),
             headers={"Content-Type": "application/json"},
         ).raise_for_status()
+
+    def test_export_get_fields_res_users(self):
+        """Test also the export of res.users"""
+        self.authenticate("admin", "admin")
```
yankinmax: TIL: how to simulate actual login with test, thanks

yankinmax (approved):

partner_property/tests/test_export.py
victoralmau:
```suggestion
    @users("admin")
    def test_export_get_fields_res_users(self):
        """Test also the export of res.users"""
```

victoralmau: Can you also make the same change in https://github.com/OCA/partner-contact/blob/46a1cafda17eccbd620e8fac398f55617cfe464f/partner_property/tests/test_export.py#L10 ?

victoralmau (changes requested):
Functional review OK

partner_property/tests/test_export.py
ivantodorovich: sure thing, it's done!

victoralmau (approved):

ivantodorovich:
@victoralmau can we get it merged? 🙏🏻

victoralmau (approved):

PR #2212, opened 2025-11-12
[16.0][ADD] partner_archive_propagate

ntsirintanis:
This module extends the native archiving mechanism for partners.

When archiving a company or parent contact, it will also handle its descendants
according to business rules — with user control and safeguards.

Features
--------

* Adds a new **"Archive Contact and Children"** button on the Partner form.
* Shows a **wizard** listing contact-type descendants before archiving.
* Automatically skips descendants linked to active users.
* Adds a technical Many2one field (propagated_from_id) that records which parent partner caused the automatic archiving, making propagation fully traceable and reversible.
* Ensures automatic unarchive propagation.
* Includes a system setting to enforce propagation even for non-UI actions
  (imports, RPC, automated jobs, etc.).


partner_archive_propagate/views/res_partner_views.xml
```diff
@@ -0,0 +1,23 @@
...
+            <xpath expr="//sheet" position="before">
+                <header>
+                    <field name="show_prop_wizard_button" invisible="1" />
+                    <button
+                        name="action_archive_with_contacts"
+                        type="object"
+                        string="Archive Contact and Children"
+                        class="oe_highlight"
+                        attrs="{
+                            'invisible': ['|', ('active', '=', False), ('show_prop_wizard_button', '=', False)]
```
ntsirintanis, line 16: or maybe child_ids = [], and remove show_prop_wizard_button field altogether

UrzaBlock:
I tested this on runboat, and the button works nicely: 
- Archiving of Partner and its children works. You can easily leave someone out, too. 
- Unarchiving of Partner easily puts back the archived children, as well.

partner_archive_propagate/models/res_partner.py
```diff
@@ -0,0 +1,156 @@
...
+        help="Show or hide the wizard button, depending on child_ids",
+    )
+
+    def _compute_show_prop_wizard_button(self):
+        for partner in self:
+            partner.show_prop_wizard_button = bool(partner.child_ids)
+
+    def write(self, vals):
+        """Enable (un)archiving propagation"""
+        archiving = "active" in vals and vals.get("active") is False
```
NL66278: Works, but can be shortened:
`archiving = vals.get("active", True) is False`

and similar for unarchiving
`unarchiving = vals.get("active", False) is True`

partner_archive_propagate/models/res_partner.py
```diff
@@ -0,0 +1,156 @@
...
+                {"active": False}
+            )
+        archivable, unarchivable = contact_desc._split_archivable_unarchivable_user()
+        wiz = self.env["res.partner.archive.propagate.wizard"].create(
+            {
+                "partner_id": self.id,
+                "line_ids": [(0, 0, {"partner_id": p.id}) for p in archivable],
+            }
+        )
+        self._notify_skipped_partners(unarchivable)
```
NL66278: I think this notification comes to early, as the user can still cancel the whole action in the wizard.

NL66278 (changes requested):
Looks great in general and very nice to have so many unit tests. There is one remark on statements that can be more concise, and one on a message that I think comes to early.

partner_archive_propagate/models/res_partner.py
NL66278: @ntsirintanis I think you missed this remark...

NL66278:
The lacking coverage could be solved by creating a wizard for some partners that are unarchivable and then checking that in you get a message in the confirmation window: https://app.codecov.io/gh/OCA/partner-contact/pull/2212?dropdown=coverage&src=pr&el=h1&utm_medium=referral&utm_source=github&utm_content=checks&utm_campaign=pr+comments&utm_term=OCA

partner_archive_propagate/tests/test_archive.py
```diff
@@ -0,0 +1,181 @@
...
+        line_partners = Wiz.line_ids.mapped("partner_id")
+        self.assertIn(self.E, line_partners)
+        self.assertNotIn(self.B, line_partners)
+        result = Wiz.action_confirm()
+        # Confirm archives A + E, skips B,
+        # and silently archives non-contact types if active.
+        # Conditionally, the wizard may return a notification when some contacts
+        # are skipped due to active users. If so, check its content.
+        if isinstance(result, dict) and result.get("type") == "ir.actions.client":
+            self.assertEqual(result.get("tag"), "display_notification")
```
NL66278: These lines are never reached. Instead of an if statement you need two separate tests: one where you will know there will be no notification and one where you will know (because you linked some user to a partner to archive) that a notification window is returned.

partner_archive_propagate/wizard/archive_propagate_wizard.py
```diff
@@ -0,0 +1,59 @@
...
+        self.partner_id.with_context(partner_archive_propagate_ui=True).write(
+            {"active": False}
+        )
+        partners = (
+            self.line_ids.mapped("partner_id").sudo().filtered(lambda p: p.active)
+        )
+        archivable, unarchivable = partners._split_archivable_unarchivable_user()
+        archivable.write({"active": False, "propagated_archive": True})
+        self.partner_id._notify_skipped_partners(unarchivable)
+        if unarchivable:
```
NL66278, line 36: If you adapt the test - see other comment - you get to 100% coverage of the wizard.

NL66278:
@ntsirintanis Everything green, that is great. You should still delete lines 88 - 97 of the test (and then unindent 98). As this is a branch never taken.

NL66278 (approved):
👍 Great work and fully covered with tests!

partner_archive_propagate/models/res_partner.py
```diff
@@ -0,0 +1,155 @@
+# Copyright 2025 Therp BV <https://therp.nl>
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+from odoo import _, api, fields, models
+
+
+class ResPartner(models.Model):
+    _inherit = "res.partner"
+
+    propagated_archive = fields.Boolean(
```
NL66278: @ntsirintanis It occured to me that unarchiving could be made more efficient (and more traceable), when not having a boolean here, but the id of the partner that was the source of the propagation.

So replace this field with a Many2one to res.partner called `propagated_from_id`. Now when unarchiving it would be much faster to just find all the partners that where archived because this partners archiving had been propagated (and that still are archived).

The beauty of it would be that with the new module that archives through relations, nothing extra would need to be done to propagate the unarchiving to those partners as well, it would already be covered in this module.

partner_archive_propagate/models/res_partner.py
```diff
@@ -0,0 +1,155 @@
...
+    show_prop_wizard_button = fields.Boolean(
+        compute="_compute_show_prop_wizard_button",
+        help="Show or hide the wizard button, depending on child_ids",
+    )
+
+    def _compute_show_prop_wizard_button(self):
+        for partner in self:
+            partner.show_prop_wizard_button = bool(partner.child_ids)
+
+    def write(self, vals):
```
NL66278, line 30: Some extra streamlining with partner_multi_relation_archive_propagate(pmrap) is possible.

In pmrap you already selected archiving_companies and unarchiving_companies. You can do the same here, but with no selection on company,  you will have archiving_partners or unarchiving partners (never both, often none).
Then from this method:
```
            if archiving_partners:
                archiving_partners._archive_propagate()
            elif unarchiving_partners:
                unarchiving_partners._unarchive_propagate()
```
The extra functionality in pmrap to propagate archiving to related partners if the archived partner is a company can be added by overriding the _archive_propagate() method then. Then in pmrap you do not need to overwrite the write() method at all.

If you follow my suggestion to replace the boolean propagated_archive with the id of the archiving source partner), you will need to do nothing for unarchiving in pmrap,

partner_archive_propagate/models/res_partner.py
```diff
@@ -0,0 +1,192 @@
...
+        """Enable (un)archiving propagation:
+        - detect which partners are being (un)archived
+        - calls _(un)archive_propagate() on them.
+        """
+        archiving_partners = self.browse()
+        unarchiving_partners = self.browse()
+        if "active" in vals:
+            if vals["active"] is False:
+                archiving_partners = self
```
NL66278: self filtered on only active partners

partner_archive_propagate/models/res_partner.py
```diff
@@ -0,0 +1,192 @@
...
+            elif vals["active"] is True:
+                unarchiving_partners = self
```
NL66278: self filtered on only archived partners

partner_archive_propagate/models/res_partner.py
```diff
@@ -0,0 +1,192 @@
...
+            if from_wizard:
+                archiving_partners._archive_propagate(from_ui=True)
+            elif non_ui:
+                archiving_partners._archive_propagate(from_ui=False)
+        # Unarchiving, always propagate
+        if unarchiving_partners:
+            unarchiving_partners._unarchive_propagate()
+        return res
+
+    def _archive_propagate(self, from_ui=False):
```
NL66278: As the method has two independent branches, just split it in two methods, that then actualy do not need any argument (apart from self).

partner_archive_propagate/models/res_partner.py
```diff
@@ -0,0 +1,192 @@
...
+                    }
+                )
+                partner._notify_skipped_partners(unarchivable)
+
+    def _unarchive_propagate(self):
+        """Unarchive descendants archived via propagation.
+        Only touches partners where propagated_from_id points to the unarchived
+        parent, and that are currently inactive.
+        """
+        for partner in self:
```
NL66278, line 59: No need for a loop. Change the first line in the domain to `("propagated_from_id", "in", self.ids),`

NL66278 (changes requested):
Nice, but few changes still needed.

NL66278 (approved):
👍 LGTM And well tested both in unit test and IRL.

marcos-mendez (changes requested):
**Automated Review -- Tests Failed**

### 1. **Root Cause**
The test failure is caused by a database connection error (`Connection to the database failed`) during the Odoo server startup, which indicates an issue in the test environment setup or configuration, not in the module's logic itself.

### 2. **Suggested Fix**
The error occurs before any module code is executed. Ensure the test environment is properly configured with a valid database connection. This is likely due to a misconfiguration in `runboat` or CI setup, and not a code bug in `partner_archive_propagate`.

However, to make the module more robust:
- In `res_partner.py`, line ~173, `_force_non_ui()` method should handle potential `KeyError` or `ConfigParameter` access errors gracefully.
  ```python
  def _force_non_ui(self):
      """Read system setting controlling external propagation."""
      try:
          return self.env["ir.config_parameter"].sudo().get_param(
              "partner_archive_propagate.force_outside_ui", default="0"
          ) in ("1", "true", "True")
      except Exception:
          return False  # fallback to disable propagation if config is missing
  ```

### 3. **Additional Code Issues**
- **Line 130 in `res_partner.py`**: The method `_split_archivable_unarchivable_user()` is inefficient as it loops through all partners individually. It should be optimized using `sudo().read(['user_ids'])` or batch queries for better performance.
- **Missing `@api.model` decorator** in `_force_non_ui()` method — though not causing runtime error, it's semantically incorrect and may mislead developers expecting model-level behavior.

### 4. **Test Improvements**
Add the following test cases in a `TransactionCase` or `SavepointCase`:

#### Test Cases:
1. **Test propagation on archive via UI (wizard)**:
   - Create a parent partner with children (contact and non-contact).
   - Call `action_archive_with_contacts()` and verify:
     - Non-contact children are archived silently.
     - Contact children appear in wizard.
     - Active users prevent archiving of their linked partners.
     - `propagated_from_id` is correctly set.

2. **Test propagation on external archive (non-UI)**:
   - Enable `force_outside_ui`.
   - Archive a parent partner via `write({'active': False})`.
   - Verify all descendants are archived with correct `propagated_from_id`.

3. **Test unarchive propagation**:
   - Archive a parent partner with descendants.
   - Unarchive the parent.
   - Verify descendants are also unarchived and `propagated_from_id` cleared.

4. **Test default behavior (without force)**:
   - Archive a parent without enabling `force_outside_ui`.
   - Ensure no propagation happens unless triggered by UI.

Use `@tag('post_install')` for integration tests and `@tag('fast')` for unit-level tests.

These tests should be added to `tests/test_partner_archive_propagate.py`.

---
### ⏰ PR Aging Alert

This PR by @ntsirintanis has been open for **123 days** (4 months).
💤 Last activity was 33 days ago.

> _Every ignored PR is a contributor who might not come back. Review time matters. ([OCA Aging Report](https://git.pop.coop/pop/oca-reviewer))_

---
### Reciprocal Review Request



Hi everyone! I found some test failures on this PR and left detailed feedback above. I am happy to discuss or help debug. In the meantime, if any of you get a chance, I would appreciate a look at my open PR(s):

**My open PRs across OCA:**
- [server-tools#3554 [MIG] datetime_formatter: Migration to 18.0](https://github.com/OCA/server-tools/pull/3554)
- [server-tools#3548 [18.0][MIG] base_kanban_stage: Migration to 18.0](https://github.com/OCA/server-tools/pull/3548)
- [hr-attendance#262 [16.0][ADD] Hr_attendance_idsecure: iDSecure (ControliD) attendance integration](https://github.com/OCA/hr-attendance/pull/262)
- [stock-logistics-workflow#2276 [16.0][ADD] stock_move_line_devaluation](https://github.com/OCA/stock-logistics-workflow/pull/2276)
- [stock-logistics-workflow#2275 [16.0][ADD] Stock move line analytic account](https://github.com/OCA/stock-logistics-workflow/pull/2275)
- [stock-logistics-workflow#2268 [16.0][ADD] stock_move_line_picking_partner](https://github.com/OCA/stock-logistics-workflow/pull/2268)
- [purchase-workflow#2694 [16.0][IMP]Purchase workflow added to review state & exception fix](https://github.com/OCA/purchase-workflow/pull/2694)

Reviewing each other's work helps the whole community move forward. Thank you!

---
_Environment via [OCA Neural Reviewer](https://git.pop.coop/pop/oca-reviewer): Minikube + K8s Job + oca-ci/py3.10-odoo16.0 | Odoo 16.0_
_Automated review by [OCA Neural Reviewer](https://git.pop.coop/pop/oca-reviewer) + qwen3-coder:30b_

NL66278:
@UrzaBlock Could you make your remark into a review, that would give the two required approvals for a merge.

UrzaBlock (approved):
I tested this on runboat, and the button works nicely:

Archiving of Partner and its children works. You can easily leave someone out, too.
Unarchiving of Partner easily puts back the archived children, as well.

PR #2215, opened 2025-11-16
[19.0][MIG] partner_stage

bosd:
Standard migration using odoo-module-migrator /ocabot migration partner_stage .

PR #2217, opened 2025-11-17
[19.0][MIG] partner_stage_only_confirmed

bosd:
standard migration with odoo-module-migrator /ocabot migration partner_stage_only_confirmed

Depends on #2215 

Build upon #2090 

Depends on #2367 (renames res_partner.state -> stage_state; this module now filters on stage_state).


BhaveshHeliconia (commented):
@bosd the dependency has been merged. You can drop the last commit and rebase the PR.

partner_stage_only_confirmed/models/base.py
```diff
@@ -0,0 +1,65 @@
...
+                            domain = node.get("domain")
+                            if not domain:
+                                domain = "[('stage_state', '=', 'confirmed')]"
+                            else:
+                                # domain is always a string due to XML parsing
+                                if domain in ("", "[]"):
+                                    domain = "[('stage_state', '=', 'confirmed')]"
+                                else:
+                                    domain = domain[:-1]
+                                    domain += ", ('stage_state', '=', 'confirmed')]"
```
BhaveshHeliconia, line 37: Use `expression.AND`

bosd: The domain here comes from the view arch as a string and can be a dynamic domain (referencing record fields/context), so it can't be evaluated into a list for `expression.AND` (which operates on domain lists). The string append is equivalent to an implicit AND and stays correct even when the existing domain starts with `|`/`!` operators — e.g. `['|', A, B]` becomes `['|', A, B, ('stage_state', '=', 'confirmed')]` which parses as `(A OR B) AND stage_state`. So I've kept the string-based approach here.

bosd:
@BhaveshHeliconia #2367 is still open (not merged yet). The last commit here filters on the renamed `stage_state` field introduced by #2367, so it has to stay until that one lands — dropping it now would break the module. I'll rebase/drop the commit once #2367 is merged.

BhaveshHeliconia:
> @BhaveshHeliconia #2367 is still open (not merged yet). The last commit here filters on the renamed `stage_state` field introduced by #2367, so it has to stay until that one lands — dropping it now would break the module. I'll rebase/drop the commit once #2367 is merged.

Okay, got it. Thanks for explaination.

BhaveshHeliconia (approved):

PR #2219, opened 2025-11-19
[18.0][ADD] New module base_location_geolocation_import

NL66278:
Extends base_location_geonames import with the import of latitude and longitude, which is then used to set those on the partner when selecting a zip_id.


base_location_geolocation_import/wizards/geonames_import.py
```diff
@@ -0,0 +1,27 @@
...
+
+
+class CityZipGeonamesImport(models.TransientModel):
+    _inherit = "city.zip.geonames.import"
+
+    @api.model
+    def prepare_zip(self, row, city_id):
+        """Add latitude and longitude to zip vals."""
+        vals = super().prepare_zip(row, city_id)
+        if len(row) > 9:
```
BhaveshHeliconia, line 14: Should be `if len(row) > 10:` to avoid `IndexError`

base_location_geolocation_import/models/res_partner.py
```diff
@@ -0,0 +1,63 @@
...
+            [
+                ("zip_id", "!=", False),
+                "|",
+                ("partner_latitude", "in", (False, 0.0)),
+                ("geolocation_from_zip", "=", True),
+            ]
+        ).filtered(lambda r: r.zip_id.latitude != r.partner_latitude)
+        partners._compute_geolocation()
+
+    @api.depends("zip_id")
```
BhaveshHeliconia, line 48: IMO `geolocation_from_zip` and `partner_latitude` should be added as dependency

BhaveshHeliconia (changes requested):

PR #2234, opened 2025-12-08
[18.0][ADD] sale_customer_order_frequency

petrus-v:
Compute customer order frequency statistics:

* **Average order duration**: Average number of days between confirmed orders.
* **Days since last order**: Number of days elapsed since the last confirmed order.
* **Days until next order**: Estimated number of days until the next order based on the average


luisDIXMIT (approved):
Code review and LGTM!

sale_customer_order_frequency/views/res_partner_views.xml
```diff
@@ -0,0 +1,26 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <record id="view_partner_form" model="ir.ui.view">
+        <field name="model">res.partner</field>
+        <field name="inherit_id" ref="base.view_partner_form" />
+        <field name="arch" type="xml">
+            <xpath expr="///group[@name='sale']" position="inside">
```
BhaveshHeliconia:
```suggestion
            <xpath expr="//group[@name='sale']" position="inside">
```

sale_customer_order_frequency/tests/test_sale_customer_order_frequency.py
```diff
@@ -0,0 +1,77 @@
+from freezegun import freeze_time
+
+from odoo.tests.common import TransactionCase
+
+
+class TestSaleCustomerOrderFrequency(TransactionCase):
```
BhaveshHeliconia:
```suggestion
class TestSaleCustomerOrderFrequency(Basecommon):
```

sale_customer_order_frequency/tests/test_sale_customer_order_frequency.py
```diff
@@ -0,0 +1,77 @@
...
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        cls.partner = cls.env["res.partner"].create({"name": "Test Partner"})
```
BhaveshHeliconia:
```suggestion
```

sale_customer_order_frequency/tests/test_sale_customer_order_frequency.py
```diff
@@ -0,0 +1,77 @@
...
+    def _create_order(self, date_order, state="sale"):
+        return self.env["sale.order"].create(
+            {
+                "partner_id": self.partner.id,
+                "date_order": date_order,
+                "state": state,
+                "order_line": [
+                    (
+                        0,
+                        0,
```
BhaveshHeliconia:
```suggestion
                    Command.create(
```

sale_customer_order_frequency/tests/test_sale_customer_order_frequency.py
```diff
@@ -0,0 +1,77 @@
...
+
+            self.partner.invalidate_recordset()
+
+            self.assertEqual(self.partner.average_order_duration, 5.0)
+            self.assertEqual(self.partner.days_since_last_order, 9)
+            self.assertEqual(self.partner.days_until_next_order, -4)
+
+    def test_ignore_draft_cancelled(self):
+        with freeze_time("2023-01-20"):
+            self._create_order("2023-01-01", state="sale")
```
BhaveshHeliconia: Don't create sale order directly with `state` field. Instead use proper odoo methods like `action_confirm`/`action_cancel` for state transition after creation of sale order.

sale_customer_order_frequency/tests/test_sale_customer_order_frequency.py
BhaveshHeliconia: There's no need to create a `res.partner` record manually when we're already using `BaseCommon`.

BhaveshHeliconia (changes requested):

sale_customer_order_frequency/tests/test_sale_customer_order_frequency.py
petrus-v: I did it but can't see the value for the purpose of what I'm testing

sale_customer_order_frequency/tests/test_sale_customer_order_frequency.py
petrus-v: yes, base on the assumption of the previous comment ! thanks.

sale_customer_order_frequency/views/res_partner_views.xml
petrus-v: good catch thanks

sale_customer_order_frequency/tests/test_sale_customer_order_frequency.py
BhaveshHeliconia: IMO we should keep this for safety purpose to ensure that state is changed properly even with the presence of other modules functionality

BhaveshHeliconia (approved):
LGTM!

NL66278:
Module looks nice, but would it not be more on its place in the sale-workflow repository?

petrus-v:
> Module looks nice, but would it not be more on its place in the sale-workflow repository?

This module display fields on partner view, we got other module that starts by sale_* in this repo! I can move it I was not sure where is the best. Who has authority to tell which repo is best ?

NL66278:
> > Module looks nice, but would it not be more on its place in the sale-workflow repository?
> 
> This module display fields on partner view, we got other module that starts by sale_* in this repo! I can move it I was not sure where is the best. Who has authority to tell which repo is best ?

There is no authority. Just that basically any module of substance will add something to res.partner. This repo would become huge if all those modules where added here. But not blocking, it was just a question.

PR #2237, opened 2025-12-11
[18.0] [IMP] partner_rank_single: less invasive constraint on rank increase

ivantodorovich:
Instead of running the constraint check on `_increase_rank`, we now avoid increasing it if the partner is already ranked in the opposite field.

This fixes an error that occurs when installing demo data of core modules:

```python-traceback
2025-12-11 11:44:05,845 35 ERROR testdb odoo.addons.account.demo.account_demo: Error while posting demo data
  File ".../odoo/addons/account/demo/account_demo.py", line 90, in _post_load_demo_data
    move.action_post()
  File ".../odoo/addons/account/models/account_move.py", line 5426, in action_post
    self._post(soft=False)
  File ".../odoo/addons/account/models/account_move.py", line 5154, in _post
    (partner | partner.commercial_partner_id)._increase_rank('supplier_rank', count)
  File ".../partner-contact/partner_rank_single/models/res_partner.py", line 23, in _increase_rank
    "supplier_rank": "customer_rank",
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".../partner-contact/partner_rank_single/models/res_partner.py", line 15, in _constrains_single_rank
    raise ValidationError(
odoo.exceptions.ValidationError: A contact cannot be both a customer and a supplier.
```

### DIFF
--- a/partner_rank_single/models/res_partner.py
+++ b/partner_rank_single/models/res_partner.py
@@ -17,8 +17,13 @@ class Contact(models.Model):
                 )
 
     def _increase_rank(self, field, n=1):
-        # OVERRIDE: to check single rank
-        # Because of direct SQL update in the super method
+        # OVERRIDE: to ignore increasing the rank if the partner is already ranked
+        # in the opposite field
+        field_inverses = {
+            "customer_rank": "supplier_rank",
+            "supplier_rank": "customer_rank",
+        }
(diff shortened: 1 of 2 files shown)


yankinmax (approved):
Nice, thanks

PR #2244, opened 2025-12-18
[16.0][ADD] partner_report_address

AungKoKoLin1997:
This module adds a translatable Partner Address Details field on partners for use in reports.

Documents are printed in the supplier’s language, but Odoo does not provide a standard way to display a partner’s name and address in multiple languages. This module addresses that limitation.

@qrtl QT6261


partner_report_address/__manifest__.py
```diff
@@ -0,0 +1,17 @@
+# Copyright 2025 Quartile (https://www.quartile.co)
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+{
+    "name": "Partner Report Address",
+    "version": "16.0.1.0.0",
+    "author": "Quartile, Odoo Community Association (OCA)",
+    "category": "Reporting",
+    "website": "https://github.com/OCA/partner-contact",
+    "license": "AGPL-3",
+    "depends": ["base"],
```
yostashiro, line 10:
```suggestion
    "depends": ["web"],
```

yostashiro (commented):
Partial review. I tend to think this module should be in the reporting-engine repo as `report_partner_address_translatable`.

AungKoKoLin1997:
> I tend to think this module should be in the reporting-engine repo as `report_partner_address_translatable`.

@yostashiro It makes sense to me as it is related to reporting. I will create new PR in oca/reporting-engine.
Closing this PR.

PR #2057, opened 2025-05-07
[18.0][FIX] base_partner_sequence: assign a new ref when copying users

StefanRijnhart:
## Describe the bug
When copying a user that has a partner that is a company and that has a ref, the resulting new user was assigned a new partner but with the same ref.

## To Reproduce
This is an issue for instance for new users signing up, if the partner of the template user is a company and is assigned a reference.

1. Copy a user that has a partner that is a company and has a ref.

**Expected behavior**
The new user has a partner with a newly assigned reference.

PR #2245, opened 2025-12-30
[19.0][FIX] base_partner_sequence: assign a new ref when copying users

StefanRijnhart:
Port forward of https://github.com/OCA/partner-contact/pull/2057

When copying a user that has a partner that is a company and that has a ref, the resulting new user was assigned a new partner but with the same ref.

This is an issue for instance for new users signing up, if the partner of the template user is a company and is assigned a reference.

1. Copy a user that has a partner that is a company and has a ref.

**Actual behavior**
The new user has a partner with the same reference as the copied partner.

**Expected behavior**
The new user has a partner with a newly assigned reference.


rrebollo (approved):
Code Review. LGTM!

NL66278 (approved):
👍 LGTM

StefanRijnhart:
Thanks for the quick followup :star_struck:

PR #2249, opened 2026-01-01
[16.0][UPD] partner_multi_relation_function: search function

NL66278:
This has actually been running in production for a long time, but for some reason was missed for creating a PR.


NL66278:
Fast-tracked this as already running in production and was only mistakenly not included in original PR.

PR #2254, opened 2026-01-12
[19.0][MIG] partner_bank_acc_holder_name: Migration to 19.0

bizzappdev:


CRogos:
I thinks this module is not needed anymore: https://github.com/odoo/odoo/blob/19.0/odoo/addons/base/views/res_bank_views.xml#L89-L90

luisDIXMIT:
> I thinks this module is not needed anymore: https://github.com/odoo/odoo/blob/19.0/odoo/addons/base/views/res_bank_views.xml#L89-L90

You're right, I agree. This PR should be closed.

PR #2256, opened 2026-01-14
[18.0][ADD] partner_revenue

adasatorres:
@BinhexTeam T18924

This addon allows you to assign profit ranges to contacts, making it easier to categorize and segment them based on their profitability. This way, businesses can easily identify and manage their most valuable clients.


rrebollo (commented):
Code Review: Potential Feature Overlap

LGTM, but I have one concern: This module appears to have a functional overlap with the existing partner_capital addon. Both implement a turnover_range / revenue_range concept.

The key difference is scope: this module focuses solely on this feature, while partner_capital manages it as part of a broader set of functionalities.

Question for the community: Should we consolidate this feature, or is a separate, single-purpose module preferred? Let's discuss the best approach to avoid duplication.

adasatorres:
@rrebollo I am closing the PR.

PR #2261, opened 2026-01-19
[19.0][MIG] partner_address_street3

imlopes:
Updates on 19.0:

* removed `private` view here: https://github.com/odoo/odoo/commit/18fb1a66f7e4db908262fe350e14ee5baae35f7f


partner_address_street3/tests/test_street_3.py
```diff
@@ -4,10 +4,10 @@
 # @author: Alexis de Lattre <alexis.delattre@akretion.com>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
-from odoo.tests.common import TransactionCase
+from odoo.addons.base.tests.common import BaseCommon
 
 
-class TestStreet3(TransactionCase):
+class TestStreet3(BaseCommon):
```
yankinmax: Besides disabling the mail context `BaseCommon` is loading other stuff we don't need for these tests. I would rather go just with:
`cls.env = cls.env(context=dict(cls.env.context, tracking_disable=True))`

From https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-19.0

> Disable tracking by setting tracking_disable on the whole test env (eg: cls.env = cls.env(context=dict(cls.env.context, tracking_disable=True)) ) or evaluate the use of BaseCommon as base test class if you need to setup company, currency, users and groups in a special way).

yankinmax (approved):

ivantodorovich (approved):

NL66278 (approved):
👍 LGTM

PR #2262, opened 2026-01-19
[19.0][MIG] partner_email_check

imlopes:


partner_email_check/models/res_partner.py
```diff
@@ -15,7 +15,7 @@
         validate_email,
     )
 except ImportError:
-    _logger.debug(_("Cannot import 'email_validator'."))
```
ivantodorovich: IMO we can drop this `try .. except` for the import, given that validator is already added as python dependency in the module manifest. Then, we can be sure it's gonna be installed

The same applies for the conditional code implemented below (`if validate_email is None`)

ivantodorovich (approved):

leemannd (approved):

leemannd:
Translations and administrative commits should be squashed https://github.com/OCA/maintainer-tools/wiki/Merge-commits-in-pull-requests

luisDIXMIT (approved):
Tested on runboat and LGTM!

NL66278 (approved):
👍 LGTM

PR #2263, opened 2026-01-19
[18.0][FIX]+[IMP] partner_firstname : Revert part of the changes of OCA#1905

legalsylvain:
Rational:
- https://github.com/OCA/partner-contact/pull/1905 introduces a migration script to force the installation of a new module named 'partner_type_base'. The migration script doesn't work, so all production instances can not update partner_firstname for the time being.
-> removing the migration script and the dependency.
- https://github.com/OCA/partner-contact/pull/1905 makes the computed function _compute_firstname_lastname_required in 'firstname.mixin' depending is_individual field, but this field is not present in this mixin.
-> add is_individual at firstname.mixin level.
- https://github.com/OCA/partner-contact/pull/1905 consider that invoicing and delivery partner can not be individual people. That assertion can make sense, but it is changing the default odoo base module. That is not the objective of partner_firstname to introduce that change.
-> the new is_individual field introduced is now more 'agnostic'. For the time being, in `res.partner`, partner.is_individual = not partner.is_company. Note that this computation can be easely overloaded.
- Revert a change in the computation of 'firstname_required' and 'lastname_required' fields, that was incorrect. A and B or C is not the same as A and (B or C).
- Remove a default value for 'partner_names_required_fields' defined in res.config.settings, as it is useless. The default value is set in the data created.


legalsylvain:
@remi-filament : could you take a look on this one ?

partner_firstname/models/firstname_mixin.py
```diff
@@ -12,38 +12,40 @@ class FirstNameMixin(models.AbstractModel):
         " that have firstname / lastname fields."
     )
 
+    is_individual = fields.Boolean(compute="_compute_is_individual")
+
```
CRogos, line 16: I am not a fan of this solution, because is_individual has nothing to do with firstname itself.

CRogos (commented):
I prefer my PR, because it is cleaner and going forward, not backward... 

(I agree, that this kind of change would have been better 1 year ago, but I pushed as hard as I could without success)

The idea of partner_type_base is not only related to partner_firstname, as already explained in the other PR.

partner_firstname/models/firstname_mixin.py
legalsylvain: Hi. Thanks for your answer ! 

I don't get it. only individual people can have a firstname. That's exactly the reason why you introduce a dependency to partner_type_base that define is_individual.

legalsylvain:
> The idea of partner_type_base is not only related to partner_firstname, as already explained in the other PR.

Totally agree ! that's the reason why we should remove the dependency to partner_type_base. People that want partner_firstname module doesn't want allways partner_type_base .

partner_firstname/models/firstname_mixin.py
CRogos: yes, but only an indivisual can have a title, too. But an Odoo user which uses titles, does not necessary want to use firstname lastname.

We also want to send newsletter only to indivisuals, but newsletter should not have a dependency to firstname/lastname.

CRogos:
> > The idea of partner_type_base is not only related to partner_firstname, as already explained in the other PR.
> 
> Totally agree ! that's the reason why we should remove the dependency to partner_type_base. People that want partner_firstname module doesn't want allways partner_type_base .

Therefore partner_type_base is implemented a way that nothing changes, but gives a good foundation for doing changes, which are missing. So even though they don't need partner_type_base, it also should not hurt.

I totally agree that this change should have be done over 1 year ago and adding this now is not best practice. (but adding the fristname_requried fields were also not best practice)

legalsylvain:
> Therefore partner_type_base is implemented a way that nothing changes, but gives a good foundation for doing changes, which are missing. So even though they don't need partner_type_base, it also should not hurt.

I'm not agree, partner_type_base introduces can_be_parent / can_be_child and change the partner base form (hide parent_id, etc...). So "it hurts", as soon as users doesn't want that change. (that is maybe a good feature, that's not the question). Also the adress become sometime readonly. Again, maybe it's interesting, but it should be done in a dedicated module, as it is nothing to to with partner_firstname.

partner_firstname/models/firstname_mixin.py
legalsylvain: We can propose in a second time, the creation of a dedicated module `base_individual_mixin`, if you want, that adds a simple `individual.mixin` mixin with "is_individual" field. Then partner_firstname can inherit from base_is_individual_mixin, and partner_type_base can also depends on it, and other module, if you want to add modules like `mass_mailing_individual`, etc...

But for the time being, the problem is that adding new dependency in a stable version is quite problematic : 

- OCA didn't decided if it was allowed or not, but some people consider it is forbidden. (see https://github.com/OCA/odoo-community.org/issues/193) (I personally think it should be possible, but that's not a consensus).

- Technically, it look like it is not very simple to to add a dependency. https://github.com/odoo/odoo/pull/72661

CRogos:
We could discuss that partner_type_base could be split into more modules, but I definitely don't see is_individual as dependency of firstname. 

partner_base_individual
partner_base_parent

... It's getting late... see you tomorrow.

legalsylvain:
> We could discuss that partner_type_base could be split into more modules, but I definitely don't see is_individual as dependency of firstname.

Me too ! but that's the only possibility to make your change working (considering invoicing / delivering partner as not individual), without adding a new dependency to partner_firstname.

I'll not wait more before merging this PR because #1905 generated troubles in production. In a few hours : https://github.com/OCA/partner-contact/pull/1905#discussion_r2704404810 by @paradoxxxzero https://github.com/OCA/partner-contact/pull/1905#issuecomment-3767746041 by @bealdav, etc.

So reverting part of the changes introduced today, keeping `partner_type_base` for people who want to use it.

Next Step : 

1) ensure it is possible to add new dependency on stable version, closing https://github.com/OCA/odoo-community.org/issues/193
2) ensure how it is possible to add a new dependency (what is the best way to do it, because the dependency to github.com/odoo/upgrade-util/ is not OK)
3) if (1) and (2) are OK, I can propose quickly a new module base_individual to move is_individual field, no problem !

(personal feeling : I don't think that adapting is_individual will work very good, because a lot of time, odoo (and oca modules) consider that a people is an individual if is_company is False, but well, not a blocking point. IMO, all that is_individual / can_be_parent / can_be_child should be provided in Odoo core, and worked with Odoo Team.)

/ocabot merge nobump

version is bumped in the manifest.

remi-filament:
Thanks @legalsylvain, I had not enough time to review, but I think this is indeed the proper way to go.

Probably that is_individual is not the best name and needs_split_name or something like that would carry more meaning about what this field is used for but I understand you wanted to keep the naming from previous PR.

@CRogos I think your requirement to have specific fields for partner typing is valide, but it should be done in another module and if you want to have it applied to partner_firstname it should be in a third glue module between the 2 other modules. This is more the way modules are built on OCA so that core module like partner_firstname do only one thing and user/integrator can pick exactly what they need.

CRogos:
@remi-filament 
I understand that the situation was problematic, but if this creates problems in your production system, you should think about your setup.

@legalsylvain 
I don't think it is a good way to "just merge" thinks you think is the only solution. Now you've fixed your setup, but totally broke mine. I also think partner_type_base is now not usable anymore with partner_firstname because is_individual is defined on multiple places. 

In my opinion it would have been much better, to totally revert my initial commit and keep the previous implementation. Then we could have worked out an solution for 19.0 which considers more than our two opinions. But in the end it is essential that we keep the review process and don't just merge things YOU think it is right! This totally takes the joy for future contribution ...

And I still disagree that adding this dependency (in general) is the wrong way. I think implementing a base feature on top as you suggest is the wrong way. I only agree that the timeframe to do this kind of change was actually closed for 18.0 and we should have applied this for 19.0. Same for the new _required fields which came with the mixin class. 

Regarding the firstname.mixin, you could even argument, that this is also an own base module and should not be part of partner_firstname, because (theoretically) I would also want to use this elsewere e.g. 'mailing.contact' (not the best example because it has first_name already) but not in res.partner. This is currently not possible and will never be possible the way you argument.

remi-filament:
@CRogos I understand this is not ideal to rollback this change which was eventually approved and questions how decisions are made.
The problem was not that much with my setup since I have odoo upgrade utils installed, the problèmes was the addition of new rules on existing production systems which do not all share the hypothesis you take (who is allowed to have parents / children / who should inherit address or not / etc.). Odoo is not enforcing this and many users / companies have found ways to implement their need with the flexibility allowed.

legalsylvain:
Hi @CRogos 
?!? I find your last message completely unpleasant.

> Now you've fixed your setup, but totally broke mine

No. I fixed the main OCA 18.0 branch. there are thousand of instances using the module, and that can not be updated since yesterday. I did it not to correct my setup (I don't have any problem with my setup, as I didn't run any upgrade yesterday), but on a voluntary basis, because I am responsible for the module, and the OCA branch was broken.

(As a reminder: visible number of users 10k ([Odoo](https://apps.odoo.com/apps/modules/18.0/partner_firstname)))

> I also think partner_type_base is now not usable anymore with partner_firstname because is_individual is defined on multiple places.

Yes. : 
- Before : all people upgrading `partner_firstname` had a problem, since yesterday.
- Now : An instance (your) using partner_type_base will be have a problem upgrading the base code. But as you are aware, you will not do it. 
In any case, the situation today is better that then situation of yesterday.

> In my opinion it would have been much better, to totally revert my initial commit and keep the previous implementation.

Of course ! But I have not write access on partner_type_base. I'm just maintainer of partner_firstname.
If you want to propose a PR that revert totally #1905 + #2263. Please do ! 

Now it's my turn : Yesterday, I spent four hours:
- reading your PR,
- trying to understand why it broke all instances attempting to upgrade partner_firstname on a V18 instance. and specially for at least 3 OCA contributors : @remi-filament, @bealdav, @paradoxxxzero, ...
- discuss with you, trying to understand your point of view
- propose a PR to fix the current 18.0 branch

I had other things to do. So a "_little thank you for the time spent_" would be appreciated.

Now, as the main OCA branch is fixed, we can calmly discuss how to move forward ! 

thanks !

CRogos:
@remi-filament and exactly for this flexibility, partner_type_base was designed for. Instead for adapting several views, which is only possible once, now the computation methods of partner_type_base could have been overloaded by multiple other modules, adding individual functionality. (like partner_firstname, which don't need to care if the address of type "Other" should have an firstname or not. This could be easily configured/overwritten in a module) ... but with this now reversed method it is not possible anymore. 

And again, reverting would have not been the problem... coping my attributes, so that my own modules are not compatible with partner_firstname, and approving it without review process is the issue I mostly complain on... 

And sorry.... this change did not come out of the sky. If I take responsibility of a module, I first check the open PRs.

CRogos:
@legalsylvain sorry breaking your setup and wasting your time. I wasn't aware that the official upgrade tools from Odoo shouldn't be used. Unfortunately upgrade script are not tested on runboat.

I think it is not helpful that we both work on the same area at the same time... therefore I will stop on this issue. If you need any review I am happy to assist.

legalsylvain:
> @legalsylvain sorry breaking your setup

As said, you didn't broke my setup, but your PR broke the possibility for all OCA users to upgrade the module. As it was quite critical, I fast merged this PR.

>  and wasting your time.

not a bid deal. shit happens sometimes !

> And again, reverting would have not been the problem

Finally done here : https://github.com/OCA/partner-contact/pull/2267. you're right, that's the cleaner way to get off to a good start.

> I wasn't aware that the official upgrade tools from Odoo shouldn't be used. 

AFAIK, this tools is only used in migration of Odoo EE instances. I never saw an OCA module using that tool. Did you ?

I executed the following lines locally : 
```
rgrep "from odoo.upgrade" ./ --include=*.py
rgrep "from odoo.*import.*upgrade" ./  --include=*.py
```
-> 0 result in the 42 OCA 18.0 repositories.
-> a 1 single result in Odoo 18.0 CE in odoo/module/module.py (and no function is called)

If you ever saw this tools in some OCA PR, could you mention it ?

In the meantime, openupgradelib is the OCA tool widely used. In my 42 OCA repos, it is called 36 times. (OCA/Openupgrade _excluded_).

> Unfortunately upgrade script are not tested on runboat.

Yes. Could be a great improvment ! If you work on that implementation (in runboat, or directly in github actions), please ping me !

> If I take responsibility of a module, I first check the open PRs.

Check the open PRs is not mandatory to adopt add-ons in the OCA rules. If you think it should be, feel free to make a pull request in the OCA guidelines.
Personally, I'm not sure it's a good idea to impose this. For [the modules I'm responsible for](https://odoo-code-search.com/ocs/search?q=author%3AOCA+readme%3Alegalsylvain+), I don't think it would be feasible. Too much work IMO.

Thanks for the discussion.

CRogos:
> If you ever saw this tools in some OCA PR, could you mention it ?
no i did not find/look for it, and I've only tested it on odoo.sh. I assume this is why there was no error on my tests.

> In the meantime, openupgradelib is the OCA tool widely used. In my 42 OCA repos, it is called 36 times. (OCA/Openupgrade excluded).
For my migraitons I use both tools because I did not yet find an force install in OCA/Openupgrade?

> shit happens sometimes !
As you said... totally not my intent to break it, totally not my interest to fight with you. We both face issues we want to solve and looking forward to move thinks forward together. No doubt that you are a skilled programmer and that we both come to a better solution if we work together.

PR #2264, opened 2026-01-19
[18.0][IMP] partner_firstname: improve migration and mixin

CRogos:
addresses issues of: https://github.com/OCA/partner-contact/pull/1905#issuecomment-3769607118

@legalsylvain could check if this migration script is working on your system?

On the mixin I am still working on.


remi-filament:
Hi @CRogos 
I do not think you need to have such a complex and generic function for your need + since you copied code from Odoo SA you should probable keep and respect their LGPL-3 license...

Here you do not need to discover dependencies, you only need to mark state for module partner_type_base as "to_install" if not already installed. You can do that with a simple SQL update query on ir_module_module

legalsylvain:
Hi @CRogos. Thanks for your answer, and the investigation. I think I prefer that implementation https://github.com/OCA/partner-contact/pull/2263. Could you take a look ? Thanks !

CRogos:
@legalsylvain @remi-filament @NL66278 

I would prefer this PR, if you confirm that the upgrade script is working.

Please do not add a new version of is_indivisual to the partner_firstname module!

partner_firstname/migrations/18.0.4.0.1/pre-migration.py
```diff
@@ -0,0 +1,28 @@
...
+
+    modules = ["partner_type_base"]
+
+    cr.execute("SELECT state FROM ir_module_module WHERE name = %s", modules)
+    mod_state = cr.fetchone()[0]
+    if mod_state in ("installed", "to install", "to upgrade"):
+        return
+
+    cr.execute(
+        "UPDATE ir_module_module SET state = 'to install' where name = %s", modules
```
remi-filament, line 21: You van probably do that in only one request adding "AND state NOT IN 'installed', 'to install', 'to upgrade'" in this second query

partner_firstname/migrations/18.0.4.0.1/pre-migration.py
```diff
@@ -0,0 +1,28 @@
...
+    )
+
+    states = dict(cr.fetchall())
+    toinstall = [m for m in states if states[m] == "to install"]
```
remi-filament, line 25: Have you tested that one, I think you need a RETURNING in your SQL request if you want to get a proper result

PR #2267, opened 2026-01-20
[REV] Finish to revert #1905.

legalsylvain:
Begin of the revert has been done in #2263. But after comments in #2263, it looks like it's better to full revert the initial PR. (Also the other module partner_type_base, will be removed in another PR #2266).

We will see in the second time how and how implement "is_indivividual" fields.

CC : @remi-filament, @CRogos


CRogos (approved):
Code review tested LGTM.

PS: I'll create a ticket for addressing and discussing the current weaknesses/observations of partner_firstname

PR #2271, opened 2026-01-21
[ADD] partner_owner_company: soft alternative to company_id

dreispt:


partner_owner_company/views/res_partner_view.xml
```diff
@@ -0,0 +1,49 @@
...
+    <!-- Partner Form View -->
+    <record id="view_partner_form_owner_company" model="ir.ui.view">
+        <field name="name">res.partner.form.owner_company</field>
+        <field name="model">res.partner</field>
+        <field name="inherit_id" ref="base.view_partner_form" />
+        <field name="arch" type="xml">
+            <xpath expr="//field[@name='category_id']" position="after">
+                <field
+                    name="owner_company_id"
+                    options="{'no_create': True, 'no_create_edit': True}"
```
luisDIXMIT, line 13: I think if it is a company, it should be invisible, following the logic in the tree view.

luisDIXMIT (commented):
LGTM but pre-commit is failing.

partner_owner_company/views/res_partner_view.xml
luisDIXMIT: Hi @dreispt, don’t you think so?

dreispt: I don't follow - if the field is not visible, the users won't be able to the owner company on it.

BhaveshHeliconia: I agree with @luisDIXMIT. We should use `base.group_multi_company` on `owner_company_id` field same as odoo does for 'company_id' field. For only one single company, no need to show this field in view.

BhaveshHeliconia:
**I've found one issue**

- Steps to reproduce:
    - Create a `contact` with child contact. No `owner_company_id` set in both.
    - Try to search for parent contact in any contact dropdown(sale/purchase/invoice).
    - Observe that only child contact is there, not parent contact.
    
**Possible enhancements:**
- `owner_company_id` should be in sync between parent and child contacts.

partner_owner_company/views/res_partner_view.xml
dreispt: I understand now. Yes, that makes sense,

dreispt:
> **I've found one issue**

I will check.

> **Possible enhancements:**

That probably makes sense.

dreispt:
Corrections were added.

BhaveshHeliconia (commented):
Now, parent contact is there, but found following issue:
If we have one parent contact having `owner_company_id` set(Same as current company) and its one child contact having different `owner_company_id` than parent and we try to search child contact using partial name and 'Search More' option from many2one dropdown,  child contact is there. As per the docs, other company contacts should only be searched by full name/email/reference/VAT.

dreispt:
@BhaveshHeliconia 
Child contacts should "inherit" the same owner as the parent. 
So the behavior looks correct. 
Perhaps the field should be disabled/invisible for child contacts, for better UX?

BhaveshHeliconia (commented):
@dreispt I agree with you on disabling/hiding the `owner_company_id` field for child contacts. However, I think its value should still be synced from the parent contact (`parent_id.owner_company_id`) to keep the data consistent.

Also, could you please confirm whether the following behaviour is expected?

- We have two contacts with different `owner_company_id` values: `ContactA` and `ContactB`.
- Each contact has one child contact with the same `owner_company_id` as its parent: `ChildA` and `ChildB`.
- We then set `ContactA`'s `owner_company_id` to the current company and go to the Sale Order's Customer field dropdown.
- In the dropdown, we can search for `ContactA` and `ChildA` using a partial name, while `ContactB` and `ChildB` require a full name search. This seems to be the expected behaviour according to the documentation.
- However, when we click **Search More** and perform the same partial-name search, all contacts are returned.

Could you please confirm whether this is the intended behaviour of the module or if it is a bug that should be fixed?

If this is intentional, I think it would be helpful to mention this behaviour explicitly in the documentation.

dreispt:
Indeed, the implementation applies for name search only.
So that behavior is expected.

I agree it would be better for childs to inherit the parent owner_id automatically, and that the doc could be improved as you describe.

BhaveshHeliconia:
Okay, thanks for the info. Rest looks good to me. Just doc update and we're good to go.

PR #2274, opened 2026-01-28
T2593 tagged partner count field wrongly set to store field

avuff:
store=True was removed, meaning that the count is calculated in real-time for accuracy. Given the fast calculation of this field, this remains an efficient approach.


luisDIXMIT (commented):
Is this a migration PR, or why are there so many commits?

PR #2279, opened 2026-02-03
[19.0][MIG] partner_auto_archive

JasminSForgeFlow:
Standard Migration

@ForgeFlow


luisDIXMIT (approved):

AaronHForgeFlow (approved):
Functional + code

NL66278:
I think I have asked this before, but what is the use case of this module? You can just archive a partner. Why do this in a roundabout way? If there is any usecase, I think it should be explained in the readme files.

MarinaAForgeFlow:
Hi @NL66278,

I agree that the README should be updated to better illustrate the intended use cases of this module.

The partner_auto_archive module was implemented to address situations where certain partners are expected to be used only once, for a specific operation. Its purpose is to help keep the database clean by automatically archiving such partners after a defined period.

A typical use case is dropshipping. When you ship products directly to your customer’s end client, you must create a partner record for the end client delivery address. However, this contact is usually not relevant after the fulfillment of that single order and is unlikely to be reused. By marking these partners as auto_archive, the system can automatically archive them after a certain time, without requiring manual cleanup.

This reduces clutter in the partner list, avoids unnecessary data accumulation, and ensures that only relevant partners remain active in the system.

NL66278:
@MarinaAForgeFlow What you say makes sense. However in the current implementation, assuming the cron is running its default every 15 days, some partners will be deleted after 15 days, some partners the same day they have been created and anything in between. Should there not be a `delete_after` date, set by default to let us say 15 days after auto_archive has been set, and preferably being able to set to another (future) date in the user interface? Especially if drop shipping is actually done by ship ;-).

And indeed the use case should definitely be in the README.

MarinaAForgeFlow:
@NL66278 This was designed this way because the module archives records, it is not deleting them. As a result, the timing of the operation is not very critical: the data is never lost, only made inactive and hidden. We went for the simplistic approach.

Furthermore, we assumed that the partner is created and assigned to the relevant document at the same time. Therefore, even if the cron job runs on the same day the partner is created, it is not an issue, the partner will already be linked wherever it is needed, and archiving it will not affect those existing references.

NL66278 (approved):
👍 Thanks for the migration and the clarified README. I think adding a configurable period how long a partner will be kept, and the ability to update this per partner, might still expand the possible uses of the module, but this need not be part of this migration.

PR #2281, opened 2026-02-04
[ADD][18.0] partner_street_town_search : Search partner by Street and Town name

nihelgabsi-acsone:
@rousseldenis


rousseldenis:
@nihelgabsi-acsone Could you just rename the module with 'city' instead of 'town' as that is the real field name. Thanks!

rousseldenis (approved):
LGTM

BhaveshHeliconia (approved):
LGTM!

partner_street_city_search/__manifest__.py
```diff
@@ -0,0 +1,16 @@
...
+    "summary": """Enable partner search by street and city""",
+    "version": "18.0.1.0.0",
+    "license": "AGPL-3",
+    "author": "ACSONE SA/NV,Odoo Community Association (OCA)",
+    "website": "https://github.com/OCA/partner-contact",
+    "depends": [],
+    "data": [
+        "views/res_partner.xml",
+    ],
+    "demo": [],
```
NL66278, line 15: Not blocking, but we should not have empty keys in the manifest.

NL66278 (approved):
👍 LGTM

PR #2283, opened 2026-02-06
[19.0][MIG] partner_industry_secondary: Migration to 19.0

muaazsiddiq:


luisDIXMIT (commented):
Squash migration commits into a single migration commit.

partner_industry_secondary/security/partner_industry_security.xml
```diff
@@ -2,6 +2,5 @@
 <odoo>
     <record id="group_use_partner_industry_for_person" model="res.groups">
         <field name="name">Use industry for individuals</field>
-        <field name="category_id" ref="base.module_category_hidden" />
```
luisDIXMIT: In version 19, res.groups doesn’t use categories directly but can use res.group.privilege. Can you check that there isn’t a res.group.privilege similar to that category? Maybe it uses the same category.

partner_industry_secondary/__manifest__.py
```diff
@@ -6,7 +6,7 @@
 {
     "name": "Partner Industry Secondary",
     "summary": "Add secondary partner industries",
-    "version": "18.0.1.0.1",
+    "version": "19.0.1.0.1",
```
luisDIXMIT: Update it to 19.0.1.0.0

luisDIXMIT (approved):
Code review and tested on runboat, LGTM! Some minor suggestions.

bouktaibMohssin (approved):

bouktaibMohssin:
LGTM

partner_industry_secondary/security/partner_industry_security.xml
StefanRijnhart: @luisDIXMIT In core Odoo, technical groups (previously in category Hidden) were not assigned a privilege. So this seems fine to me. See https://github.com/odoo/odoo/commit/33637d137ed52f5faa

partner_industry_secondary/__manifest__.py
StefanRijnhart: @muaazsiddiq Can you fix this, resetting the minor version digits to zero?

muaazsiddiq: @StefanRijnhart, Sure

StefanRijnhart (approved):
Thanks!

/ocabot merge nobump

PR #2289, opened 2026-02-18
[19.0][MIG] partner_supplier_ref_sequence

JasminSForgeFlow:
Standard Migration

@ForgeFlow


MarinaAForgeFlow (approved):
Code review, LGTM

AaronHForgeFlow (approved):
Code LGTM

luisDIXMIT:
When I duplicate it, the copy is set as if it weren’t a supplier, but it still has a supplier reference. Is this the expected behaviour?

MarinaAForgeFlow:
> When I duplicate it, the copy is set as if it weren’t a supplier, but it still has a supplier reference. Is this the expected behavior?

I don't think this should be considered expected behavior. If the is_supplier is not copied, then the supplier reference sequence should not be set either. I will review the code and propose a fix in this migration, that can be back-ported to other versions.

luisDIXMIT (approved):

BhaveshHeliconia (approved):
LGTM!

WesleyOliveira98 (approved):

PR #2290, opened 2026-02-19
[18.0][FIX] partner_property: company context not respected when searching by computed field

cristina-hidalgo-tecnativa:
#### Error produced
```
FAIL: TestResPartnerProperty.test_properties_company_id
  File "/opt/odoo/auto/addons/partner_property/tests/test_res_partner.py", line 95, in test_properties_company_id
    self.assertEqual(
AssertionError: res.company(14,) != res.company(13,)
```
`_field_to_sql` evaluated `self.env.company` ignoring the `with_company()` context.
`_compute_properties_company_id` read `company_id` through the ORM, which filters the value under multicompany restrictions, and `with_company()` creates a new environment that doesn't share the original cache, making pending writes invisible.
`_search_properties_company_id` used raw SQL with `flush_model`, which didn't flush pending changes from other environments.

#### Fix
- `_field_to_sql`: reads `allowed_company_ids` from context instead of `self.env.company`.
- `_compute_properties_company_id`: reads `company_id` using `sudo()` to bypass multicompany filters and access pending writes regardless of the active company context.
- `_search_properties_company_id`: uses ORM search instead of raw SQL.
- `test_res_partner`: adds `invalidate_recordset()` before checking `properties_company_id` with a different company context, ensuring pending writes are flushed to the database first.

@Tecnativa TT60000
@christian-ramos-tecnativa  @pedrobaeza  could you review it please?

### DIFF
--- a/partner_property/models/res_partner.py
+++ b/partner_property/models/res_partner.py
@@ -1,8 +1,6 @@
 # Copyright 2024 Tecnativa - Víctor Martínez
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
-
 from odoo import api, fields, models
-from odoo.osv.expression import SQL_OPERATORS
 from odoo.tools.sql import SQL
 
 
@@ -30,17 +28,11 @@ class ResPartner(models.Model):
             item.properties_company_id = item.company_id or self.env.company
 
     def _search_properties_company_id(self, operator, value):
-        self.flush_model(["company_id"])
-        query = self._where_calc([])
-        query.add_where(
-            SQL(
-                "%s %s %s",
-                self._field_to_sql(self._table, "properties_company_id", query),
-                SQL_OPERATORS[operator],
-                value,
-            )
-        )
-        return [("id", "in", query)]
+        explicit = self.sudo().search([("company_id", operator, value)])
+        if operator == "=" and value == self.env.company.id:
+            implicit = self.sudo().search([("company_id", "=", False)])
+            return [("id", "in", (explicit | implicit).ids)]
+        return [("id", "in", explicit.ids)]
 
     def _field_to_sql(self, alias, fname, query=None, flush: bool = True) -> SQL:
         # OVERRIDE to allow to export the properties
--- a/partner_property/tests/test_res_partner.py
+++ b/partner_property/tests/test_res_partner.py
@@ -92,6 +92,7 @@ class TestResPartnerProperty(TransactionCase):
             partner.with_company(self.company_a.id).properties_company_id,
             self.company_a,
         )
+        partner.invalidate_recordset()  # Invalidate cache to force recomputation
         self.assertEqual(
             partner.with_company(self.company_b.id).properties_company_id,
             self.company_a,


luisDIXMIT (commented):
Hi @cristina-hidalgo-tecnativa , thanks for your work. How can I reproduce the error? It is not failing in my local database.

cristina-hidalgo-tecnativa:
> Hi @cristina-hidalgo-tecnativa , thanks for your work. How can I reproduce the error? It is not failing in my local database.

Hi @luisDIXMIT,  the error only reproduces when `partner_multi_company` (and its dependency `base_multi_company`) are installed alongside `partner_property`:

imagen

luisDIXMIT (approved):
You're right, thanks @cristina-hidalgo-tecnativa . It seems fixed, LGTM!

partner_property/models/res_partner.py
```diff
@@ -27,27 +25,28 @@ class ResPartner(models.Model):
     @api.depends_context("company")
     def _compute_properties_company_id(self):
         for item in self:
-            item.properties_company_id = item.company_id or self.env.company
+            company_id = item.sudo().company_id
```
victoralmau: This change is not necessary; it can remain as it was.: `item.properties_company_id = item.company_id or self.env.company`

partner_property/models/res_partner.py
```diff
@@ -27,27 +25,28 @@ class ResPartner(models.Model):
...
-                "%s %s %s",
-                self._field_to_sql(self._table, "properties_company_id", query),
-                SQL_OPERATORS[operator],
-                value,
-            )
-        )
-        return [("id", "in", query)]
+        company_id = self.env.context.get("allowed_company_ids", [self.env.company.id])[
+            0
+        ]
```
victoralmau:
```suggestion
```

partner_property/models/res_partner.py
```diff
@@ -27,27 +25,28 @@ class ResPartner(models.Model):
...
+        explicit = self.sudo().search([("company_id", operator, value)])
+        if operator == "=" and value == company_id:
```
victoralmau:
```suggestion
        if operator == "=" and value == self.env.company.id:
```

partner_property/models/res_partner.py
```diff
@@ -27,27 +25,28 @@ class ResPartner(models.Model):
...
+            implicit = self.sudo().search([("company_id", "=", False)])
+            return [("id", "in", (explicit | implicit).ids)]
+        return [("id", "in", explicit.ids)]
 
     def _field_to_sql(self, alias, fname, query=None, flush: bool = True) -> SQL:
         # OVERRIDE to allow to export the properties
         if fname == "properties_company_id":
+            company_id = self.env.context.get(
+                "allowed_company_ids", [self.env.company.id]
+            )[0]
```
victoralmau:
```suggestion
```

partner_property/models/res_partner.py
```diff
@@ -27,27 +25,28 @@ class ResPartner(models.Model):
...
             return SQL(
                 """COALESCE(%(company_column)s, %(env_company)s)""",
                 company_column=SQL.identifier(alias, "company_id"),
-                env_company=self.env.company.id,
+                env_company=company_id,
```
victoralmau: This change is not necessary; it can remain as it was.

victoralmau (changes requested):

partner_property/models/res_partner.py
cristina-hidalgo-tecnativa: Modified now

partner_property/models/res_partner.py
cristina-hidalgo-tecnativa: Modified now

partner_property/models/res_partner.py
cristina-hidalgo-tecnativa: Fixed

partner_property/models/res_partner.py
cristina-hidalgo-tecnativa: Modified now

victoralmau (approved):

PR #2291, opened 2026-02-20
[18.0][IMP] partner_identification: Change where the mock has to be loaded

victoralmau:
Change where the mock has to be loaded

With the changes introduced in https://github.com/odoo/odoo/commit/de056cc#diff-b37c7fd5520c97a29ddc59495779e7be61a66e15e85603928e27b3affb9dc31f, an error was occurring because the change validation on the model was being performed before tearDownClass was executed. Therefore, it has been modified so that the mock changes are applied in each test and cleaned up after each test. This way, the error will no longer appear and the test will run normally.

@Tecnativa

### DIFF
--- a/partner_category_type/tests/test_partner_category_type.py
+++ b/partner_category_type/tests/test_partner_category_type.py
@@ -1,5 +1,6 @@
 # Copyright 2023 ForgeFlow, S.L.
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+from unittest import mock
 
 from odoo.exceptions import ValidationError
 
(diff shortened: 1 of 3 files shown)


luisDIXMIT (approved):
Code review, LGTM!

JordiMForgeFlow (approved):

partner_category_type/tests/test_partner_category_type.py
```diff
@@ -65,22 +66,14 @@ def test_04_change_parent_category_type(self):
...
         )
 
-        def extended_selection(self):
-            return [("generic", "Generic"), ("test", "Test")]
-
-        original_selection = type(parent)._get_category_type_selection
-
-        try:
-            type(parent)._get_category_type_selection = extended_selection
+        with mock.patch.object(
```
NL66278, line 69: It took me some time to understand that this replaces the selection in the field definition of category_type. Maybe (probably) many people are more clever then I am, but might be worth an explanatory comment.

NL66278 (approved):
👍 If comment can be added that would be nice, but not blocking. LGTM.

pedrobaeza (approved):
/ocabot merge patch

PR #2293, opened 2026-02-25
[19.0][FIX] partner_identification: fr.po translation

therbin200:
Simply add the fr translation


luisDIXMIT:
Should translations be done via OCA Weblate?

PR #2295, opened 2026-02-28
[18.0][MIG] partner_noncommercial

hbrunn:
Usage
=====

This module changes some strings for partners to make them more
appropiate for non-commercial context, especially renaming companies to
organizations. It also adds some filters that are often used in
non-commercial settings, although they can be useful within companies


luisDIXMIT (approved):

primes2h:
@hbrunn
runboat is not active, could you force push or rebase please?

primes2h (changes requested):
Could you reduce the number of commits as per https://github.com/OCA/maintainer-tools/wiki/Merge-commits-in-pull-requests please?

partner_noncommercial/readme/USAGE.md
```diff
@@ -0,0 +1,5 @@
+This module changes some strings for partners to make them more
+appropiate for non-commercial context, especially renaming companies to
+organizations. It also adds some filters that are often used in
+non-commercial settings, although they can be useful within companies
+too.
```
primes2h, line 5: Maybe not in this PR, but it would be nice to have these filters in a separate module so they can be used independently of this. What you think?

primes2h (dismissed):

partner_noncommercial/readme/USAGE.md
hbrunn: this module is already very small, splitting it seems a bit over the top to me

hbrunn:
@primes2h squashed commits

partner_noncommercial/readme/USAGE.md
primes2h: > this module is already very small, splitting it seems a bit over the top to me

Understandable, but then the message is a bit contradictory. While these filters are useful for companies, it doesn't make much sense for a company to install the module.

partner_noncommercial/views/res_partner.xml
```diff
@@ -0,0 +1,116 @@
...
+        <field name="model">res.partner</field>
+        <field name="inherit_id" ref="base.view_partner_simple_form" />
+        <field name="arch" type="xml">
+            <!-- Change placeholders -->
+            <field name="parent_id" position="attributes">
+                <attribute name="placeholder">Organization Name...</attribute>
+            </field>
+        </field>
+    </record>
+    <record id="view_partner_form" model="ir.ui.view">
```
BhaveshHeliconia, line 67: `view_partner_form`: Avoid usage of `replace`

NL66278 (approved):
👍 LGTM

partner_noncommercial/views/res_partner.xml
NL66278: It replaces one span by another, so has none of the usual reasons to avoid replace.

PR #2258, opened 2026-01-16
[16.0] [MIG] partner_iterative_archive: Migration to 16.0

luisDIXMIT:
Standard migration.

PR #2297, opened 2026-03-02
[16.0] [FIX] fix pre-commit cache + partner_contact_age_range test

luisDIXMIT:
With this fix that I found here: https://github.com/OCA/oca-addons-repo-template/issues/334
, I'm trying to resolve the problems experienced here: https://github.com/OCA/partner-contact/pull/2258

The problem with partner_contact_age_range was that the birthdate was set to today, while the test is executed with the year 2024 frozen. When the age is computed, it also uses today, but inside the test "today" is 2024. As a result, the calculation becomes today minus birthdate (2024 vs 2026), which leads to an age of -1 (depending on the month).

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -140,7 +140,7 @@ repos:
           - --settings=.
         exclude: /__init__\.py$
   - repo: https://github.com/acsone/setuptools-odoo
-    rev: 3.1.8
+    rev: 3.3.2
     hooks:
       - id: setuptools-odoo-make-default
       - id: setuptools-odoo-get-requirements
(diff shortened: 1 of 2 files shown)


AaronHForgeFlow (approved):
LGTM!

NL66278:
@luisDIXMIT Sorry I missed that you had already made a PR. Having said that, I think my PR has the advantage, of doing the full copier update in one commit, and the failing test problem in a commit that reduces code and a dependency. What do you think? If you still prefer this PR I will merge it, as this was created first.

luisDIXMIT:
Closing this PR in favor of #2301.

PR #2299, opened 2026-03-03
[19.0][ADD] base_partner_sequence_manual_rank

MarinaAForgeFlow:
When using base_partner_sequence, the ref field is automatically set in all commercial partners based on a sequence. However, when you use partner_manual_rank and partner_supplier_ref_sequence, it does not make sense to always set the ref field for all commercial partners.

With this module, the ref field is only set for partners marked as customers.


luisDIXMIT (approved):

DavidJForgeFlow (approved):

BhaveshHeliconia (approved):

PR #2308, opened 2026-03-26
[18.0][IMP] partner_industry_secondary: search by parent name includes descendants

JordiBForgeFlow:
Override _search_display_name to use child_of when searching by name, so that searching for a parent industry also returns all its descendants at any depth, using the stored parent_path for efficient SQL resolution.


luisDIXMIT (approved):
Code review, LGTM!

BhaveshHeliconia (approved):
LGTM!

PR #2309, opened 2026-03-26
[18.0][IMP] partner_second_lastname: add name order 'Lastname, Firstname SecondLastname'

BhaveshHeliconia:
This PR adds a new name formatting option: `Lastname, Firstname SecondLastname` (key: `last_first_comma2`).
### Improvements:
- Added `last_first_comma2` to `res.config.settings`.
- Implemented computed name logic and inverse name logic in `res.partner` for the new format.
- Fixed a potential `IndexError` in `_get_computed_name` when `lastname` is missing.
- Updated documentation in `CONFIGURE.md`.
- Added comprehensive tests in `test_config.py`, `test_multiple_names.py`, and `test_name.py`.


partner_second_lastname/models/res_config_settings.py
```diff
@@ -14,7 +14,10 @@ def _partner_names_order_selection(self):
             "last_first_comma": "Lastname SecondLastname, Firstname",
             "first_last": "Firstname Lastname SecondLastname",
         }
-        return [(k, new_labels[k]) if k in new_labels else (k, v) for k, v in options]
+        result = [(k, new_labels[k]) if k in new_labels else (k, v) for k, v in options]
+        # Separate format where only the first lastname is followed by a comma
+        result.append(("last_first_comma2", "Lastname, Firstname SecondLastname"))
```
luisDIXMIT, line 19: Why didn't you add it directly to the new_labels dict?

BhaveshHeliconia: @luisDIXMIT Here, if we add it to 'new_labels' dict, then it won't get added in 'result' as we're looping on 'options' and 'options' don't have 'last_first_comma2', so we've added it separately afterwards.

luisDIXMIT: Oh, thanks for explaining me!

luisDIXMIT (approved):
Tested on runboat and LGTM!

legalsylvain (changes requested):
If we add new key, we should propose changes in related OCA modules. don't you think ? 
At least : 
./hr/hr_employee_firstname
./hr/hr_employee_second_lastname/
./partner-contact/partner_second_lastname/

Thanks !

BhaveshHeliconia:
> If we add new key, we should propose changes in related OCA modules. don't you think ? At least : ./hr/hr_employee_firstname ./hr/hr_employee_second_lastname/ ./partner-contact/partner_second_lastname/
> 
> Thanks !

Yes, you're right. This should also be handled in the hr_employee_second_lastname module. I'll take a look and propose the necessary changes there as well.

BhaveshHeliconia:
@legalsylvain here is the PR for hr_employee_second_lastname: https://github.com/OCA/hr/pull/1554

PR #2312, opened 2026-03-31
[19.0][MIG] partner_email_duplicate_warn: Migration to v19

MohamedOsman7:
Migration of partner_email_duplicate_warn from 18.0 to 19.0.

- Bumped module manifest version to 19.0.1.0.0.
- Updated partner form inheritance xpath to match 19.0 base partner view structure.
- Regenerated addon README artifacts via pre-commit hooks.

Notes:
- The original anchor warning_company no longer exists in 19.0; warning_tax is used as the new insertion point.
- I added a **x2many_links** widget to avoid depending on account. The x2many_buttons widget only works if account is installed, otherwise it falls back to the default table. So this keeps the inline duplicate links without adding extra dependencies.


luisDIXMIT (approved):
Tested on runboat, LGTM!

CRogos (approved):
Tested + Code LGTM

CRogos:
@pedrobaeza can you merge?

PR #2316, opened 2026-04-07
[19.0][MIG] partner_deduplicate_by_website

CRogos:


luisDIXMIT (approved):

partner_deduplicate_by_website/tests/test_crm_deduplicate_by_website.py
```diff
@@ -0,0 +1,71 @@
...
+        wizard.action_start_manual_process()
+        found_match = False
+        for line in wizard.line_ids:
+            match_ids = safe_eval(line.aggr_ids)
+            if self.partner_1.id in match_ids and self.partner_2.id in match_ids:
+                found_match = True
+                break
+        self.assertTrue(found_match)
+
+    def test_deduplicate_by_website_and_is_company(self):
```
MohamedOsman7, line 42: @CRogos 
1. Shouldn't this test be named "test_deduplicate_by_website_and_email" instead of "test_deduplicate_by_website_and_is_company", since "is_company" is not actually used here? Or was the intention to test the behavior with 2 company type contacts?

2- Do you think we should add more test cases, like individual vs. company, individual vs. individual, and company vs. company?

They seem to work fine in the UI and are quite similar, but just wanted to confirm.

Otherwise LGTM

CRogos: I've added is_company now, but the tests are very similar. Not sure if this tests is really needed or should be removed.

MohamedOsman7 (approved):

MohamedOsman7:
@pedrobaeza could you merge?

PR #2324, opened 2026-04-16
[MIG][17.0] mig partner_contact_in_several_companies

NL66278:
Attempt to solve the problems encountered here: https://github.com/OCA/partner-contact/pull/2103


NL66278 (commented):
👍 LGTM. Note that I am not the real author of this PR, just recreated it to solve a problem with the pre-commit in the previous PR.

Also note that IMHO a lot of the functionality of this module is covered by partner_multi_relation.

NL66278:
@kwoychesko You might want to have a look at this, so your contribution to this repository can be moved forward.

luisDIXMIT (approved):

partner_contact_in_several_companies/models/ir_actions.py
```diff
@@ -0,0 +1,24 @@
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import models
+
+
+class IRActionsWindow(models.Model):
+    _inherit = "ir.actions.act_window"
+
+    def read(self, fields=None, context=None, load="_classic_read"):
```
BhaveshHeliconia, line 9: Remove invalid `context` from args

partner_contact_in_several_companies/views/res_partner.xml
```diff
@@ -0,0 +1,279 @@
...
+            </xpath>
+            <xpath
+                expr="//page[@name='personal_information_page']/group[@name='personal_information_group']"
+                position='after'
+            >
+                <p invisible="not contact_id"> To see personal information
+                    about this contact, please go to to the this person form: <field
+                        name="contact_id"
+                        class="oe_inline"
+                        domain="[('contact_type','!=','attached')]"
```
BhaveshHeliconia, line 241: Make domain consistent with py model domain(`("is_company", "=", False), ("contact_type", "=", "standalone")`)

BhaveshHeliconia (changes requested):

BhaveshHeliconia:
`Attached` type contacts are also visible on `contacts` kanban view. Only `Standalone` type contacts should be visible.

kwoychesko:
Hi @NL66278, @BhaveshHeliconia — and thanks @NL66278 for keeping this moving.

On the open review point (attached contacts also showing on the Contacts kanban): the root cause is that since 17.0 the web client loads kanban/list records through `web_search_read` → `search_fetch` → `_search`, which bypasses the public `search()` that this module overrides to hide attached contacts. So the override never runs for the views, and the attached contacts leak through.

The fix is to apply the same domain rewrite in `web_search_read` (and keep it on `search` for direct ORM calls). It's important not to move it into `_search`: that method also backs record-rule access checks and one2many prefetches, so filtering there raises an `AccessError` when opening a contact form and empties out `other_contact_ids`.

For v17.0:

```python
def _search_show_all_positions_domain(self, domain):
    flag = self.env.context.get("search_show_all_positions", {})
    if not (flag.get("is_set") and not flag.get("set_value")):
        return domain
    domain = expression.normalize_domain(domain)
    attached_contact_domain = expression.AND(
        (domain, [("contact_type", "=", "attached")])
    )
    # Disable the filter for this inner search to avoid recursion.
    attached_contacts = self.with_context(
        search_show_all_positions={"is_set": False}
    ).search(attached_contact_domain)
    return expression.OR(
        (
            expression.AND(([("contact_type", "=", "standalone")], domain)),
            [("other_contact_ids", "in", attached_contacts.ids)],
        )
    )
  
@api.model
def search(self, domain, offset=0, limit=None, order=None):
    return super().search(
        self._search_show_all_positions_domain(domain),
        offset=offset,
        limit=limit,
        order=order,
    )

@api.model
def web_search_read(
    self, domain, specification, offset=0, limit=None, order=None, count_limit=None
):
    return super().web_search_read(
        self._search_show_all_positions_domain(domain),
        specification,
        offset=offset,
        limit=limit,
        order=order,
        count_limit=count_limit,
    )
```

This replaces the old `search()` override. Note: on 18.0 `web_search_read` additionally carries `@api.readonly`; on 17.0 it's `@api.model` only, as above.

I've applied and verified this same approach on the 18.0 migration (#2383, now green on CI), including a regression test that exercises `web_search_read` and one that checks an attached contact stays readable under the hide context. Happy to send it as a patch or a PR against this branch if that's easier — just say the word.

PR #2328, opened 2026-04-22
[16.0][ADD] partner_address_change_lock

Kev-Roche:
Simple module to track adress / country changes on partners. 

Block partner address or country changes when unpaid invoices exist. The lock scope (country only vs. full address) is configurable per company in accounting settings. Address changes are tracked in an history log on the partner form.


partner_address_change_lock/views/res_config_settings.xml
```diff
@@ -0,0 +1,31 @@
...
+                    <div class="o_setting_left_pane">
+                    </div>
+                    <div class="o_setting_right_pane">
+                        <div id="partner_address_lock_mode">
+                            <label
+                                    string="Address Lock Mode"
+                                    class="o_light_label"
+                                    for="partner_address_lock_mode"
+                                />
+                        </div>
```
emiliesoutiras, line 30: you can set icon for company-specific, like [this sample](https://github.com/OCA/OCB/blob/61abde439610bc7f9fe8fde03c64f0bc99d21363/addons/hr_timesheet/views/res_config_settings_views.xml#L82)

```suggestion
                            <span class="fa fa-lg fa-building-o" title="Values set here are company-specific." aria-label="Values set here are company-specific." groups="base.group_multi_company" role="img"/>
                        </div>
```

emiliesoutiras (approved):
Everything works fine, just a display adjustment

luisDIXMIT (commented):
There is a .po file, but there isn’t a related .pot file. Is it correct?

partner_address_change_lock/models/res_partner.py
```diff
@@ -0,0 +1,123 @@
...
+        self.ensure_one()
+        unpaid = self.env["account.move"].search_count(
+            [
+                ("partner_id", "=", self.id),
+                (
+                    "move_type",
+                    "in",
+                    ["out_invoice", "out_refund", "in_invoice", "in_refund"],
+                ),
+                ("payment_state", "not in", ["paid", "reversed", "cancel"]),
```
BhaveshHeliconia: Remove non-existing `cancel` state from `payment_state` check

partner_address_change_lock/models/res_partner.py
```diff
@@ -0,0 +1,123 @@
...
+                    vals, changed
+                ):
+                    raise ValidationError(
+                        _(
+                            "You cannot modify the address of partner "
+                            '"%s" because they have unpaid invoices'
+                        )
+                        % (rec.name)
+                    )
+            self._append_address_history()
```
BhaveshHeliconia, line 127: `_append_address_history()` must be after super()

BhaveshHeliconia (changes requested):

PR #2341, opened 2026-04-30
[19.0][FIX] partner_department: fix Contacts tab and Departments tab position

MiquelRForgeFlow:
Exclude department-type partners from the Contacts tab child_ids and move the Departments tab to appear right after the Contacts tab.


AaronHForgeFlow (approved):
👍

MarinaAForgeFlow (approved):

PR #2342, opened 2026-05-05
[18.0][FIX] partner_contact_address_default: use allowed_company_ids in depends_context

cristina-hidalgo-tecnativa:
Using `@api.depends_context("company")` on a Binary field causes a stale cache key when `env.company` is switched mid-transaction, resulting in:

`ValueError: Compute method failed to assign res.partner.partner_delivery_domain`

Replace with `@api.depends_context("allowed_company_ids")`, which reads directly from context and keeps the cache key consistent.

@Tecnativa TT59807
@pedrobaeza


pedrobaeza (approved):

carlos-lopez-tecnativa (approved):
LGTM

PR #2343, opened 2026-05-11
[19.0][IMP] partner_department: improve UX and enforce department integrity

MiquelRForgeFlow:
Switch the Members tab to kanban mode aligned with the Contacts tab. Add inline form dialogs to Members and Departments tabs so the partner type is pre-set via context when creating records. Change the child_ids department exclusion domain by a constraint: partners cannot have a department as parent_id.

### DIFF
--- a/partner_department/models/res_partner.py
+++ b/partner_department/models/res_partner.py
@@ -2,6 +2,7 @@
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
 
 from odoo import api, fields, models
+from odoo.exceptions import ValidationError
 
 
 class ResPartner(models.Model):
@@ -20,6 +21,13 @@ class ResPartner(models.Model):
     department_member_ids = fields.One2many(
         "res.partner",
         "department_id",
+        domain=[("type", "!=", "department")],
+        string="Direct Members",
+    )
+
+    department_all_member_ids = fields.Many2many(
+        "res.partner",
+        compute="_compute_department_all_member_ids",
         string="Members",
     )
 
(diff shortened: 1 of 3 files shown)


luisDIXMIT (commented):
Should be contacts page hiden on department contacts? It has no sense be able to create a contact from there when a department contact can't have child contacts.

partner_department/models/res_partner.py
```diff
@@ -48,3 +49,11 @@ def action_view_department_members(self):
             "domain": [("department_id", "=", self.id)],
             "context": {"default_department_id": self.id},
         }
+
+    @api.constrains("parent_id", "type")
+    def _check_parent_not_department(self):
```
DavidJForgeFlow, line 79: Why this restriction? One employee could be child of Sales Department in a company right? I don't think this makes sense here

MiquelRForgeFlow: "contacts" should be child of companies, always. they have `department_id` field to indicate if they belong to a department, but not they are not child of departments.

DavidJForgeFlow: This is an specific restriction. You could want to do it another way and restricting that behavior in a OCA generic module I think it's not the solution. This kind of constraints should be in a custom module and keep here only technical or super standard constraints.

MarinaAForgeFlow: The thing is, if you move the constrain out, you either move the department_id too or the field becomes useful. From my understanding, the initial solution proposed using the department_id field as the relation between a contact and a department, not using then the parent_id one. Therefore because of this field, the constrain makes kind of sense.

MarinaAForgeFlow (approved):

MiquelRForgeFlow:
> Should be contacts page hiden on department contacts?

No, departments can have some childs ("invoice", "delivery", "other")

JordiBForgeFlow (approved):

PR #2344, opened 2026-05-12
[19.0][IMP] partner_supplier_ref: show supplier_ref in partner list view

JordiBForgeFlow:
Add the supplier_ref field to the res.partner list view as an optional column hidden by default, placed before the name column.


luisDIXMIT (approved):
Code review and tested locally because runboat is broken. LGTM!

PR #2345, opened 2026-05-14
[19.0][MIG] partner_address_format_domestic: Migration to 19.0

AungKoKoLin1997:
@qrtl QT6658


partner_address_format_domestic/models/res_country.py
```diff
@@ -11,6 +11,8 @@ class ResCountry(models.Model):
     is_relevant_for_company = fields.Boolean(compute="_compute_is_relevant_for_company")
 
     def _compute_is_relevant_for_company(self):
-        company_countries = self.env["res.company"].search([]).mapped("country_id.id")
+        company_countries = (
+            self.env["res.company"].search([], limit=251).mapped("country_id.id")
```
luisDIXMIT: Why limit to 251?

luisDIXMIT (commented):
Code review and tested locally because runboat is broken, LGTM justo one minor question

partner_address_format_domestic/models/res_country.py
AungKoKoLin1997: If we don't add `limit`, there is pre-commit issue. I added the `limit` attribute and use `251` because there are `251` country records in odoo. So, it is safe even there is a company for each country.

luisDIXMIT: Nice, thanks for the explanation.

luisDIXMIT (approved):
LGTM!

partner_address_format_domestic/tests/test_partner_address_format_domestic.py
```diff
@@ -23,7 +23,9 @@ def setUpClass(cls):
         cls.partner_jp = cls.env["res.partner"].create(
             {"name": "Yamada Taro", "country_id": cls.country_jp.id}
         )
-        cls.env.company = cls.company_us
+        cls.env = cls.env(
+            context=dict(cls.env.context, allowed_company_ids=[cls.company_us.id])
+        )
```
yostashiro: Can we move this right after cls.company_us, so partners are created with the right env (we might get hiccups at some point otherwise)?

partner_address_format_domestic/models/res_country.py
```diff
@@ -11,6 +11,8 @@ class ResCountry(models.Model):
...
+        )
         for country in self:
             country.is_relevant_for_company = country.id in company_countries
```
yostashiro: Applying `limit=251` is a bit random. Besides it's applied to the wrong target. It shouldn't be on companies.

```py
        for country in self:
            country.is_relevant_for_company = self.env["res.company"].search_count([("country_id", "=", country.id)], limit=1)
```
should circumvent the pre-commit issue.

yostashiro (changes requested):

partner_address_format_domestic/models/res_country.py
AungKoKoLin1997: Updated!

partner_address_format_domestic/tests/test_partner_address_format_domestic.py
AungKoKoLin1997: Done!

BhaveshHeliconia (approved):
Code review LGTM!

yostashiro (approved):
Code review. LTGM.

partner_address_format_domestic/tests/test_partner_address_format_domestic.py
```diff
@@ -0,0 +1,36 @@
...
+        cls.country_jp = cls.env["res.country"].search([("code", "=", "JP")], limit=1)
+        cls.country_us.address_format = "%(state_code)s\n%(zip)s\n%(country_name)s"
+        cls.country_us.address_format_domestic = "%(state_code)s\n%(zip)s"
+        cls.country_jp.address_format = "%(zip)s\n%(state_code)s\n%(country_name)s"
+        cls.country_jp.address_format_domestic = "%(zip)s\n%(state_code)s"
+        cls.company_us = cls.env["res.company"].create(
+            {"name": "Test Company", "country_id": cls.country_us.id}
+        )
+        cls.env = cls.env(
+            context=dict(cls.env.context, allowed_company_ids=[cls.company_us.id])
```
nobuQuartile, line 21:
```suggestion
            context=dict(cls.env.context, tracking_disable=True, allowed_company_ids=[cls.company_us.id])
```

> Disable tracking by setting tracking_disable on the whole test env (eg: cls.env = cls.env(context=dict(cls.env.context, tracking_disable=True)) ) or evaluate the use of BaseCommon as base test class if you need to setup company, currency, users and groups in a special way).

Migration to version 19.0
https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-19.0#:~:text=Disable%20tracking%20by%20setting%20tracking_disable%20on%20the%20whole%20test%20env%20(eg%3A%20cls.env%20%3D%20cls.env(context%3Ddict(cls.env.context%2C%20tracking_disable%3DTrue))%20)%20or%20evaluate%20the%20use%20of%20BaseCommon%20as%20base%20test%20class%20if%20you%20need%20to%20setup%20company%2C%20currency%2C%20users%20and%20groups%20in%20a%20special%20way).

partner_address_format_domestic/models/res_country.py
```diff
@@ -0,0 +1,17 @@
...
+    _inherit = "res.country"
+
+    address_format_domestic = fields.Text(string="Domestic Address Format")
+    is_relevant_for_company = fields.Boolean(compute="_compute_is_relevant_for_company")
+
+    def _compute_is_relevant_for_company(self):
+        for country in self:
+            country.is_relevant_for_company = self.env["res.company"].search_count(
+                [("country_id", "=", country.id)], limit=1
+            )
```
nobuQuartile:
```suggestion
    def _compute_is_relevant_for_company(self):
        ids = set(self.env["res.company"].search([]).mapped("partner_id.country_id.id"))
        for country in self:
            country.is_relevant_for_company = country.id in ids
```

Error Condition
search_count([("country_id", "=", country.id)]) on res.company raised ValueError: Cannot convert res.company.country_id to SQL because it is not stored. This occurred when opening any country form in Odoo 19.0.

Reason
res.company.country_id is a computed (non-stored) field that delegates to partner_id.country_id. The Odoo ORM cannot convert non-stored fields to SQL — this is enforced in fields.py via:
if not self.store or not self.column_type:
    raise ValueError(f"Cannot convert {self} to SQL because it is not stored")

https://github.com/odoo/odoo/commit/8b8c7287913cffe9d6efd661f0a5b2a149caf62c#diff-956d895aa67961bac940841f7c3d1e10eb8ecabec82ef017803c4a6a3bb7cd22R1126-R1127
https://github.com/odoo/odoo/blob/c51938f92180abd150ef381bd867c39cc5467e54/odoo/orm/models.py#L5369
https://github.com/odoo/odoo/blob/19.0/odoo/orm/fields.py#L1218

Fix
Collect all company country IDs upfront via mapped("partner_id.country_id.id")

AungKoKoLin1997: Please see the fix.

partner_address_format_domestic/tests/test_partner_address_format_domestic.py
AungKoKoLin1997: I use `Basecommon`.

nobuQuartile: OK. Thank you.

nobuQuartile (approved):
LGTM: Code review and functional review

PR #2348, opened 2026-05-15
[16.0] [ADD]partner_multi_relation_archive_propagate

ntsirintanis:
This module extends the behaviour of
`partner_archive_propagate` to also propagate archiving and
unarchiving **via partner_multi_relation relations**.

When a company is archived:

* all related partners linked through relation types with
  ``propagate_archive = True`` are considered for propagation;
* related partners without active users are archived and marked with
  ``propagated_from_id`` pointing to the company;
* related partners linked to active users are skipped and listed in a
  message on the company.

When the company is unarchived, all partners that were archived because
of it – whether through the partner hierarchy or through
multi-relations – are restored automatically by the core
`partner_archive_propagate` logic.

Non-company partners do not propagate via this module, even if they are
involved in relations with ``propagate_archive = True``.


ntsirintanis:
@NL66278

luisDIXMIT:
Can you squash commits into one unique add commit?

ntsirintanis:
@NL66278 @luisDIXMIT

NL66278 (approved):
👍 LGTM, covered by extensive tests and succesfully running in production at one of our clients.

luisDIXMIT (approved):

luisDIXMIT:
Must the change on module partner_archive_propagate be in a separated PR as improvement? Maybe I was wrong and that change must be in a separated commit. What do you think @NL66278 ? It seems to me rare.

ntsirintanis:
> Must the change on module partner_archive_propagate be in a separated PR as improvement? Maybe I was wrong and that change must be in a separated commit. What do you think @NL66278 ? It seems to me rare.

nah, no need to do that. The change is minimal and fully justified

NL66278:
> Must the change on module partner_archive_propagate be in a separated PR as improvement? Maybe I was wrong and that change must be in a separated commit. What do you think @NL66278 ? It seems to me rare.

In general I prefer to have changes to different modules in different commits. In this case the changes to partner_archive_propagate make that module easier to extend, which is then used in the new module. Could have been two commits, but for me not blocking.

partner_archive_propagate/models/res_partner.py
```diff
@@ -166,20 +175,26 @@ def action_archive_with_contacts(self):
           allowing the user to exclude some.
         """
         self.ensure_one()
-        descendants = self._get_descendants().filtered(lambda p: p.active)
-        contact_desc = descendants.filtered(lambda p: p.type == "contact")
-        if not contact_desc:
+        candidates = self._get_archive_propagation_candidates().filtered(
```
BhaveshHeliconia, line 178: One possible enhancement: 
- Let say we've two company contacts with following child contacts: 
   - CompanyA: Contact1, Contact2
   - CompanyB: No Child
- Contact1 and Contact2 are of contact type
- Also, one relation exists including these two company contacts with Propagate archive enabled in its relation type
- Now, try to archive CompanyB contact using Archive Contact and children button
- Only one contact(CompanyA) is shown on wizard line. It should show Contact1 and Contact2 as well.

NL66278: This might indeed be a nice addition in a future enhancement. Use case: we have registered a company as a partner in a collaboration, we have registered several departments of the partner company, including some of their contacts. However for the moment I think the module is good to go as is.

BhaveshHeliconia: Okay, then proceed with the current functionality.

BhaveshHeliconia (approved):

PR #2349, opened 2026-05-17
[19.0][FIX] partner_contact_address_default: invalidate parent cache on archive

### LINKED ISSUE #2347: [17.0] partner_contact_address_default: stale default address cache after archiving child
### Steps to reproduce
1. Install `partner_contact_address_default`.
2. Install `auditlog`.
3. Create a parent with child contacts.
4. Set child contacts as `partner_delivery_id` / `partner_invoice_id` / `partner_contact_id`.
5. Archive those child contacts.
6. Call `address_get(["delivery", "invoice", "contact"])`.

### Current behavior
`address_get()` may return stale/unexpected ids.

### Expected behavior
Archived child defaults should be ignored and fallback values should be returned.

### Root cause
Parent default fields are cleared, but cache may still keep stale values in some
module interaction contexts.

### Proposed fix
After clearing these fields on parent partners, explicitly invalidate:
- `partner_delivery_id`
- `partner_invoice_id`
- `partner_contact_id`

fsmw:
Closes #2347

When a child contact is archived, the parent partner fields partner_delivery_id, partner_invoice_id and partner_contact_id are cleared.  Without explicit cache invalidation, stale values can persist in multi-module contexts and be returned by address_get.

Changes:
- Split parent search and cache invalidation into explicit steps
- Call invalidate_recordset on aggregated parent set for all three fields
- Add test to verify no stale cached id after archiving child

Run pre-commit locally: passed.

### DIFF
--- a/partner_contact_address_default/tests/test_partner_contact_address_default.py
+++ b/partner_contact_address_default/tests/test_partner_contact_address_default.py
@@ -66,6 +66,15 @@ class TestPartnerContactAddressDefault(common.TransactionCase):
         self.assertEqual(res["invoice"], self.partner.id)
         self.assertEqual(res["contact"], self.partner.id)
 
+    def test_contact_address_archived_cache(self):
+        self.partner.partner_delivery_id = self.partner_child_delivery2
+        # Prime the cache by reading the field before archiving the child
+        _ = self.partner.partner_delivery_id.id
+        self.partner_child_delivery2.write({"active": False})
+        res = self.partner.address_get(["delivery"])
+        # Ensure no stale cached id from the archived child is returned
+        self.assertEqual(res["delivery"], self.partner_child_delivery1.id)
+
     def test_partner_domains(self):
         self.partner._compute_partner_domains()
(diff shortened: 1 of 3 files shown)


luisDIXMIT (approved):
Code review, LGTM!

BhaveshHeliconia (approved):

partner_contact_address_default/tests/test_partner_contact_address_default.py
```diff
@@ -66,6 +66,15 @@ def test_contact_address_archived(self):
         self.assertEqual(res["invoice"], self.partner.id)
         self.assertEqual(res["contact"], self.partner.id)
 
+    def test_contact_address_archived_cache(self):
```
yostashiro, line 69: This test passes even without proposed changes on res.partner.write(), and I wasn't sure the fix was addressing a "false" issue. Does the originating issue https://github.com/OCA/partner-contact/issues/2347 (seems to be on 17.0) still reproduce with the latest auditlog which uses ThrowAwayCache?

PR #2351, opened 2026-05-18
[18.0][FIX] partner_firstname: fallback for empty names on create

micheledic:
## Problem
Any inbound email with a malformed `From:` header (helpdesk,project, CRM, mass mailing, plain mail thread) routes through`name_create`, which can build a partner with `name=False` when`parse_contact_from_email` extracts neither a display name nor avalid email. The empty name skips `_get_inverse_name`,`firstname`/`lastname` stay empty, and `_check_name` raises`EmptyNamesError`. The whole mail flow aborts.
## Fix
`res_partner` post-processes the result of `_get_inverse_name`.  When both parts come back empty, `create` falls back to  `vals['email']` and `_inverse_name` falls back to `record.email`, or to a translatable placeholder (`Unknown`) when no email is  available. The `_get_inverse_name` signature stays unchanged, so submodules that override it (`partner_middlename`,  `partner_second_lastname`) keep working without changes.


BhaveshHeliconia (approved):
LGTM!

NL66278:
The fallback to email resolves a real problem where we are fetching emails, and do not want them to be lost because of a missing name. However the fall-back to "unknown" in all other cases, basically does away with the requirement to have a name filled. If we really do not want this check, it would be easier to remove it then to circumvent it.

I think the check for some kind of name filled is still very useful. But we need to prevent the mail test to succeed, which can easily be done by not running our tests when in test mode (unless we are testing the check itself, which can be done by checking for a context key passed by the test).

micheledic:
> The fallback to email resolves a real problem where we are fetching emails, and do not want them to be lost because of a missing name. However the fall-back to "unknown" in all other cases, basically does away with the requirement to have a name filled. If we really do not want this check, it would be easier to remove it then to circumvent it.
> 
> I think the check for some kind of name filled is still very useful. But we need to prevent the mail test to succeed, which can easily be done by not running our tests when in test mode (unless we are testing the check itself, which can be done by checking for a context key passed by the test).

changed the logic, now lastname = email if name is empty!

NL66278 (approved):
👍 Nice solution! LGTM!

NL66278:
@micheledic I think adding this case to the unit test will solve the coverage problem

micheledic:
> @micheledic I think adding this case to the unit test will solve the coverage problem

https://github.com/OCA/partner-contact/pull/2351/changes#diff-264b619001c131ef415278c2b2bd3fafdc3f714a0aa311db34cdacdb0e655e7a

i already added the test case

micheledic:
@legalsylvain can you check ?

NL66278:
@micheledic It is line 181 that is not covered, the fallback on email:
```

    def _inverse_name(self):
        """
        Try to revert the effect of :meth:`._compute_name`.

        NOTE:
        Field write order matters here. Each assignment triggers ``write()`` and
        ``_check_name``. Writing falsy values first can cause transient
        ``EmptyNamesError`` and previously made the ``mail`` module fail to install
        unless it was installed before ``partner_firstname``.

        We therefore use update() to prevent this issue.
        """
        for record in self:
            if (
                not parts.get("lastname")
                and not parts.get("firstname")
                and record.email
            ):
                parts["lastname"] = record.email    **<=== This is not covered**
            record.update(parts)
```

micheledic:
> @micheledic It is line 181 that is not covered, the fallback on email:
> 
> ```
> 
>     def _inverse_name(self):
>         """
>         Try to revert the effect of :meth:`._compute_name`.
> 
>         NOTE:
>         Field write order matters here. Each assignment triggers ``write()`` and
>         ``_check_name``. Writing falsy values first can cause transient
>         ``EmptyNamesError`` and previously made the ``mail`` module fail to install
>         unless it was installed before ``partner_firstname``.
> 
>         We therefore use update() to prevent this issue.
>         """
>         for record in self:
>             parts = record._get_inverse_name(record.name, record.is_company)
>             if (
>                 not parts.get("lastname")
>                 and not parts.get("firstname")
>                 and record.email
>             ):
>                 parts["lastname"] = record.email    **<=== This is not covered**
>             record.update(parts)
> ```

I removed the _inverse_name fallback. It only kicks in when someone writes name='' on an existing partner, which isn't what the bug is about: the bug is the mail gateway creating partners with empty names.
Before that partner_middlename and partner_second_lastname both override _inverse_name without super() so the fallback never runs whenthose modules are installed.

The create() fix is enough to cover the actual case. If someone deliberately wipes the name later, letting _check_name raise is the right behaviour.

leob-cl:
Does this behaviour also need to apply when creating contacts manually via the UI?
As far as I understand it, this was to solve a problem when contacts are created from emails with bad `From:` headers, right?

Recently we updated our OCA modules, and noticed this behaviour when we create new invoice addresses under a company.
Usually we would leave the name fields empty, so it would just take the company name on invoices.

But with this behaviour, the last name is then set to the email address, which then shows up on new invoices sent to the contact. Naturally we can remove it manually each time, but it's just an extra step that shouldn't be necessary IMO

Maybe by excluding child addresses or certain address types..

Thanks for your hard work guys 😋lmk

PR #2353, opened 2026-05-18
[18.0][FIX] res_partner: skip company default when install_mode is active

jqbeltran2:
# [FIX] res_partner: skip company default when install_mode is active

During module installation, Odoo sets `install_mode=True` in the context when loading XML data. In this scenario,
`_default_company_id` was assigning `env.company` to every partner created without an explicit company_id, regardless of which company the record was intended for.

This causes a ValidationError when a localization module (e.g. l10n_ca) creates bank accounts for a demo company different from the one active at install time, as the partner ends up with the wrong company and fails the journal's bank account company check.

Skip the default when `install_mode` is present in context, since XML data loading could remain company-neutral.

### DIFF
--- a/partner_company_default/models/res_partner.py
+++ b/partner_company_default/models/res_partner.py
@@ -16,6 +16,7 @@ class ResPartner(models.Model):
         context = self.env.context
         if (
             context.get("creating_from_company")
+            or context.get("install_mode")
             or config["test_enable"]
             and not context.get("test_partner_company_default")
         ):


jqbeltran2:
@fernandahf

partner_company_default/tests/test_partner_company_default.py
```diff
@@ -44,3 +44,12 @@ def test_partner_company_default(self):
             .create({"name": "Test Partner 2"})
         )
         self.assertEqual(partner.company_id, company_fr)
+
```
pedrobaeza: This is not adding too much value, as it's not a real installation situation, just simulating it through the context. If tomorrow, the context key changes in Odoo, this will still pass. Better to remove it.

pedrobaeza (approved):

partner_company_default/tests/test_partner_company_default.py
NL66278: @pedrobaeza  I do not know whether I quite agree. The functionality itself depends on context key 'install_mode'. Unlikely that Odoo will change that in the lifetime of 18.0. If they do in a future version it should be taken up in the migration.

NL66278 (approved):
👍 LGTM

partner_company_default/tests/test_partner_company_default.py
pedrobaeza: Yeah, but the aim of a test is to detect if something changes, and this test is not going to detect anything, as it's testing the code itself, not the process.

NL66278: I think testing the code has value in itself. If testing the process can be done, this would of course be even better.  Do you have an idea how to check whether module installation still sets the context key install_mode?

pedrobaeza: No for this case, but my point is adding lines and lines of code with pretty low value doesn't worth, and this is the case IMO.

luisDIXMIT (approved):
LGTM!

BhaveshHeliconia (approved):
LGTM!

partner_company_default/tests/test_partner_company_default.py
jqbeltran2: @pedrobaeza I see your point. I removed it

PR #2357, opened 2026-05-20
[19.0][IMP] partner_identification: Respect ID expiry in compute/inverse/search

### LINKED ISSUE #2196: partner_identification should allow multiple identifications of the same category when not overlapping
## Module
partner_identification

## Describe the bug
identifications can, and often will have, temporary validity. However when entering a new identification through a specialize number and the _inverse_identification method, where one already exists for the partner, an error will be thrown, even when the existing identification in the same category has already expired. Also the compute and search methods should take expiry into account.

fsmw:
Forward-port of #2356 to 19.0.

When an identification has expired (active=False, status=close), the compute, inverse, and search methods should not consider it as the current active ID for the partner.

**Changes:**
- Add `_get_active_id_numbers()` helper filtering by `active=True` and `status != 'close'`
- `_compute_identification`: use helper to skip expired IDs
- `_inverse_identification`: use helper to allow creating new ID when existing ones are expired
- `_search_identification`: exclude inactive/closed IDs from search domain

Closes #2196

@NL66278

### DIFF
--- a/partner_identification/models/res_partner.py
+++ b/partner_identification/models/res_partner.py
@@ -19,6 +19,14 @@ class ResPartner(models.Model):
         string="Identification Numbers",
     )
 
+    def _get_active_id_numbers(self, category_code):
+        """Return active (non-expired) ID numbers for the given category."""
+        return self.id_numbers.filtered(
+            lambda r: r.category_id.code == category_code
+            and r.active
+            and r.status != "close"
+        )
+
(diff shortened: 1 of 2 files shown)


NL66278 (approved):
👍 Same as the 18.0 branch. So once more thank you and LGTM.

BhaveshHeliconia (approved):
LGTM!

PR #2358, opened 2026-05-20
[19.0][MIG] partner_property: Migration to 19.0

dnplkndll:
Port of `partner_property` from 18.0 to 19.0. Follows the [OCA migration guide](https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-19.0).

## Non-mechanical adaptations worth flagging

- **`_search_properties_company_id`** — 19.0's Domain optimizer normalises `(field, '=', value)` to `(field, 'in', {value})` before reaching the search method, so the `operator == '='` guard never fired and partners without an explicit `company_id` were silently excluded from the implicit-match branch. Now accepts both shapes.
- **`_search_partner_properties_definition_{company,person}`** — `Domain.TRUE` is *falsy* in bool context (sentinel: `__bool__` returns `not self.is_true()`), so `cond and Domain.TRUE or Domain.FALSE` always evaluated to `Domain.FALSE`. Rewrote with an explicit ternary + XOR on `(empty, negate)`. Also accepts the `in`/`not in` shapes post-optimizer-normalisation.

## 19.0 relevance

`res.partner` now inherits `properties.base.definition.mixin` (single global `properties` field). This module adds **per-company** `properties_type_{company,person}` fields keyed off `env.company`, plus the person/company-type split — distinct from core. Open question for reviewers: expose all three properties fields, or hide the core one.


dnplkndll:
@ivantodorovich @pedrobaeza @victoralmau — this is ready for review.

pedrobaeza:
On 19, there's already properties in the contacts. Why migrating this?

dnplkndll:
> **per-company** `properties_type_{company,person}`

is adds per company. I could not find any real specific cases to know if that was of value. just getting back into practice doing MIGs 

will close if there is not a use case for the partner property in company context.

pedrobaeza:
Well, the value of this module is to have different properties per contacts of type company and of type contact, so if that's not respected upstream, then please go ahead and continue with the migration, but add in CONTEXT.md the reason of this module.

dnplkndll:
> Well, the value of this module is to have different properties per contacts of type company and of type contact, so if that's not respected upstream, then please go ahead and continue with the migration, but add in CONTEXT.md the reason of this module.

added the verified impact in CONTEXT.md

BhaveshHeliconia (changes requested):
In Odoo 19, properties added from the Contacts form are displayed on all contact records, regardless of any other criteria.

This module extends that behavior by allowing properties to be associated with a specific `company_type`. As a result, a property created for contacts with `company_type = 'company'` will only be displayed on contacts of that type and not on contacts with a different `company_type`.

### Possible enhancements

- The Contacts form currently shows two **Add Property** buttons: the standard Odoo button and the one introduced by this module. Making them more clearly distinguishable would improve usability.
- Likewise, properties created through the module-specific button could be visually differentiated from standard Odoo properties to make their origin and behavior more obvious to users.

partner_property/views/res_partner_views.xml
```diff
@@ -0,0 +1,48 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<odoo>
+    <record id="view_partner_form" model="ir.ui.view">
+        <field name="name">res.partner.form</field>
+        <field name="model">res.partner</field>
+        <field name="inherit_id" ref="base.view_partner_form" />
+        <field name="arch" type="xml">
+            <notebook position="before">
```
BhaveshHeliconia: Can we add a new notebook tab for this?

BhaveshHeliconia (approved):
LGTM!

PR #2371, opened 2026-05-28
[19.0] [MIG] partner_external_map : Migration to 19.0.

uncannycs:
[19.0] [MIG] partner_external_map : Migration to 19.0.


BhaveshHeliconia (commented):
We should use latest URL formats provided by map service providers.

PR #2374, opened 2026-05-31
[19.0][IMP] partner_email_check: clearer duplicate-email message + multi-company scope option

cliffkujala:
## Summary

This improves the duplicate-email handling of `partner_email_check` in two ways: a clearer, access-aware error message, and a new option to scope uniqueness per company instead of globally.

### Clearer, access-aware duplicate message

The duplicate check is an `@api.constrains`, which always runs as superuser, so uniqueness was (and still is) enforced across **all** records regardless of the acting user's record rules. Previously a collision raised a generic `Email 'x@y.com' is already in use.`, which is confusing when the user cannot see the conflicting partner (record rules / multi-company): they are blocked from using an address that appears nowhere they can look.

After detecting a conflict, the constraint now re-checks whether the acting user can actually read the conflicting record (`self.sudo(False)`) and tailors the message:

- **Readable** → identifies the record so the user can reuse it:
  `Email address x@y.com is already in use by <name> (ID: 123). Please input another email address or use the existing record.`
- **Not readable** → explains the block without disclosing the record:
  `Email address x@y.com is already in use by a record you do not have access to. Please input a different email address, or contact your system administrator to request access.`

### Multi-company duplicate scope option

New company setting `partner_email_check_duplicate_scope` (exposed in Settings, shown only when duplicate filtering is enabled):

- **Across all companies** (`global`, default) — unchanged behavior: an address may be used only once in the whole database.
- **Within the same company** (`company`) — the same address may be reused by a partner of another company, but must stay unique inside each company (`company_id = False` is treated as its own bucket).

## Compatibility notes

- The new setting defaults to `global`, preserving the previous behavior.
- The success-path message is now more verbose and, for readable records, discloses the conflicting partner's name and ID (previously a generic string). This is intentional and is the point of the change.

## Tests

Adds coverage for: per-company scope allowing the same address in another company, per-company scope still blocking within one company, global scope blocking across companies, and the access-restricted message path (duplicate hidden by a record rule is still blocked, without leaking the record).

### DIFF
--- a/partner_email_check/tests/test_partner_email_check.py
+++ b/partner_email_check/tests/test_partner_email_check.py
@@ -2,6 +2,7 @@
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 
 
+from odoo import Command
 from odoo.exceptions import UserError, ValidationError
 from odoo.tests.common import TransactionCase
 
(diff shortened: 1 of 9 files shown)


partner_email_check/tests/test_partner_email_check.py
```diff
@@ -100,6 +100,111 @@ def test_duplicate_addresses_allowed_by_default(self):
...
+        # trip on unrelated access checks, e.g. base_partner_sequence reading
+        # ir.sequence) but blocked from reading the conflicting record.
+        groups = self.env.ref("base.group_user") | self.env.ref(
+            "base.group_partner_manager"
+        )
+        restricted_user = self.env["res.users"].create(
+            {
+                "name": "Restricted",
+                "login": "restricted_user",
+                "group_ids": [(6, 0, groups.ids)],
```
BhaveshHeliconia:
```suggestion
                "group_ids": [Commad.set(groups.ids)],
```

BhaveshHeliconia (approved):
Code and functional review LGTM!

partner_email_check/tests/test_partner_email_check.py
cliffkujala: Thanks for the review! Good call on `Command.set` — much cleaner than the `(6, 0, ...)` tuple. Small heads-up: the suggestion as written has a typo (`Commad.set`), so I applied a corrected version manually: added `from odoo import Command`, switched this line to `Command.set(groups.ids)`, and also converted the `ir.rule` `groups` line below to `Command.set([])` for consistency. Pushed in the latest commit.

cliffkujala:
@BhaveshHeliconia do you know what next step is to get this across the finish line for merge?  Is it just a matter of time, or is it being passed up because of the codecov/project failing check?  

I think this PR is clean and all functional checks are green (test with Odoo, test with OCB, pre-commit, runboat). The only red mark is codecov/project for just a -0.03% diff in coverage, which is whole-repo coverage noise from other network-dependent tests elsewhere in the repo.  codecov/patch is green for this particular PR (the new code is fully covered). 

Could a maintainer merge when convenient? Thanks!

BhaveshHeliconia:
@cliffkujala you need one more review. After that, a maintainer can merge the PR.

cliffkujala:
> @cliffkujala you need one more review. After that, a maintainer can merge the PR.

@BhaveshHeliconia How does that happen?  Just time?

BhaveshHeliconia:
> > @cliffkujala you need one more review. After that, a maintainer can merge the PR.
> 
> @BhaveshHeliconia How does that happen? Just time?

Yes, pretty much. Another reviewer needs to approve the PR. Once you have the required reviews, a maintainer can merge it when they have time. If you'd like to speed things up, you can politely ping reviewers or maintainers after giving them some time to review.

PR #2375, opened 2026-05-31
[19.0][IMP] partner_email_duplicate_warn: access-aware duplicate banner

cliffkujala:
## Summary

`partner_email_duplicate_warn` shows an on-change banner listing other partners that share the same email, each rendered as a clickable link. The matching partners are computed with `compute_sudo`, so the banner could list — and link to — records the current user is not allowed to read. Clicking such a link (or, depending on the view, merely rendering it) raises the standard Odoo `AccessError` dialog, with no explanation of why the record is off-limits.

This makes the banner access-aware: it links only to duplicates the acting user may actually read, and summarises the rest as a count instead of leaking links to forbidden records.

## Changes

- `_compute_same_email_partner_ids` now filters the matched partners through the acting user's read access (`_filtered_access("read")`), so `same_email_partner_ids` only ever contains readable records.
- New computed field `same_email_inaccessible_count`: the number of same-email partners the user cannot read.
- The banner view shows the readable duplicates as links (unchanged behavior for those) plus, when `same_email_inaccessible_count` is set, a short message noting that N additional record(s) the user cannot access also use this address.

## Behavior notes

- No `AccessError` is raised by the banner anymore; unreadable duplicates are surfaced as a count rather than a link.
- Detection itself is unchanged (still finds all duplicates); only what is *disclosed* to the user respects record rules / multi-company access.

## Tests

Adds `test_partner_duplicate_inaccessible`: a duplicate hidden from the acting user by a record rule is counted in `same_email_inaccessible_count`, is not present in `same_email_partner_ids`, and reading the computed fields raises no `AccessError`.

### DIFF
--- a/partner_email_duplicate_warn/__manifest__.py
+++ b/partner_email_duplicate_warn/__manifest__.py
@@ -4,7 +4,7 @@
 
 {
     "name": "Partner Email Duplicate Warn",
-    "version": "19.0.1.0.0",
+    "version": "19.0.2.0.0",
     "category": "Partner Management",
     "license": "AGPL-3",
     "summary": "Warning banner on partner form if other partners have the same email",
--- a/partner_email_duplicate_warn/models/res_partner.py
+++ b/partner_email_duplicate_warn/models/res_partner.py
@@ -14,11 +14,17 @@ class ResPartner(models.Model):
         string="Partner with same e-mail",
         compute_sudo=True,
     )
+    same_email_inaccessible_count = fields.Integer(
+        compute="_compute_same_email_partner_ids",
+        string="Partners with same e-mail you cannot access",
+        compute_sudo=True,
+    )
 
     @api.depends("email", "company_id")
     def _compute_same_email_partner_ids(self):
         for partner in self:
             same_email_partner_ids = []
+            inaccessible_count = 0
             if partner.email and partner.email.strip():
                 partner_email = partner.email.strip().lower()
                 domain = [("email", "=ilike", "%" + partner_email + "%")]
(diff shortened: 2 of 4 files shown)


BhaveshHeliconia (commented):
@cliffkujala Could you guide me no how I can reproduce this issue? I've tried with some ways, but it is not reproducible as of now.

BhaveshHeliconia:
- While checking this PR https://github.com/OCA/partner-contact/pull/2374, I reproduced the issue. 

- Test case: 
    - Enable companyA
    - Create a contact with company and email set
    - Enable companyB
    - Try to create another contact with company and same email as the previous one
    - Observe 'Access Rights' error

- This PR addresses this issue.

- Also, in odoo, `same_vat_partner_id` and `same_company_registry_partner_id` fields of `res.partner` don't have `compute_sudo` parameter, instead it is handled directly inside compute method(`_compute_same_vat_partner_id`). So, IMO we should follow the same pattern as odoo follows for these fields.

BhaveshHeliconia (approved):
Functional review LGTM!

cliffkujala:
Thanks for the review and the approval, @BhaveshHeliconia!

On reproduction: the key trigger is the acting user not having **read access to the matching duplicate** partner. Multi-company (your companyA/companyB steps) is one clean way to hit it; a record rule that hides the existing partner from the current user (e.g. a salesperson restricted to their own contacts) is another. In either case, before this PR the same-email banner would compute the match as superuser and then render a link to a record the user can't read, raising the raw `AccessError` dialog with no explanation. With this change the banner links only to readable duplicates and reports the rest as a count.

On the `compute_sudo` vs. core pattern point: agreed, mirroring how Odoo core handles `same_vat_partner_id` / `same_company_registry_partner_id` (no `compute_sudo`, escalating inside the compute method) is a reasonable and consistent approach. It's a small, behavior-equivalent refactor. Since it touches the privilege model of a maintained module and isn't required for correctness, I'd defer that call to the module maintainer (@alexis-via) — happy to apply it in this PR if preferred, or leave it for a follow-up.

NL66278 (approved):
👍 LGTM

cliffkujala:
Thank you for the merge approval.

PR #2377, opened 2026-06-02
[16.0][FIX] partner_archive_propagate: Adapt standard archive action, update USAGE and tests

ntsirintanis:
Archiving a contact via the Archive dropmenu would archive the contact and its children. With this commit, only the contact is archived, not the children.

 Updated USAGE and expanded unit tests.


ntsirintanis:
@NL66278

NL66278 (approved):
👍 LGTM. Very nice.

PR #2384, opened 2026-06-08
[19.0][ADD] partner_display_ref

JasminSForgeFlow:
This module prefixes the customer reference (`res.partner.ref`) to a partner's
`display_name` whenever it is rendered inside an opted-in view — typically as
`[C00123] Acme Corp` in an autocomplete dropdown.

The decoration is a generic mechanism: it activates only when a view injects
the `partner_display_ref_field` context key naming the field to prefix on a
`res.partner` field. Contacts, CRM,
Invoicing and any other view that does not opt in continue to see the plain
partner name. Companion modules (e.g. `sale_partner_display_ref`) wire this
mechanism into specific apps.

@ForgeFlow


ChristianSantamaria (approved):
Only review code LGTM 👍

BhaveshHeliconia (approved):
LGTM!

partner_display_ref/models/res_partner.py
```diff
@@ -0,0 +1,23 @@
...
+
+class ResPartner(models.Model):
+    _inherit = "res.partner"
+
+    @api.depends("ref")
+    @api.depends_context("partner_display_ref_field")
+    def _compute_display_name(self):
+        super()._compute_display_name()
+        ctx = self.env.context
+        field_name = ctx.get("partner_display_ref_field")
```
rousseldenis, line 15: Maybe a little explanation on why computing the value on context value in this case ?

JasminSForgeFlow: For now, this case https://github.com/OCA/purchase-workflow/pull/3094

and for the same kind of case where the field name is different, to handle it with any particular app

luisDIXMIT (approved):
Code review and tested locally, LGTM!

PR #2394, opened 2026-06-22
[FIX] partner_email_check: fix external dependency name for email-validator

iglesias-andres:
The external_dependencies key used "email-validator" (with hyphen),
which is not a valid Python import name. Odoo checks dependencies by
importing them, so the correct name is "email_validator" (with underscore).

### DIFF
--- a/partner_email_check/__manifest__.py
+++ b/partner_email_check/__manifest__.py
@@ -12,7 +12,7 @@
     "installable": True,
     "application": False,
     "license": "AGPL-3",
-    "external_dependencies": {"python": ["email-validator"]},
+    "external_dependencies": {"python": ["email_validator"]},
     "data": ["views/base_config_view.xml"],
     "demo": ["demo/res_company_demo.xml"],
 }


luisDIXMIT:
Can you add "[19]" to the beginning of the title and fix the pre-commit issue?

cliffkujala:
Thanks for picking this up. Before it goes in, could you say what you actually observed — an install failure, or a warning in the log?

I ask because I can't reproduce a problem with the current value, and I don't think the reasoning in the description holds for 19.0. `Manifest.check_manifest_dependencies` calls `check_python_external_dependency`, which doesn't import first:

```python
try:
    version = importlib.metadata.version(requirement.name)
except importlib.metadata.PackageNotFoundError as e:
    try:
        # keep compatibility with module name but log a warning instead of info
        importlib.import_module(pydep)
        _logger.warning("python external dependency on '%s' does not appear to be a valid PyPI package. Using a PyPI package name is recommended.", pydep)
```

It resolves the PyPI *distribution* name first and only falls back to importing, warning that a PyPI name is preferred. `email-validator` with the hyphen *is* the PyPI name — `importlib.metadata.distribution("email-validator").metadata["Name"]` returns exactly that — so the manifest is already in the recommended form, and the underscore spelling is the one that warning discourages.

The change also looks like a no-op in either direction. `importlib.metadata` normalizes lookups (PEP 503 normalization, plus dashes treated as underscores), so both spellings reduce to the same key `email_validator`. On Python 3.12.3 with email-validator 2.3.0 both resolve through the metadata path with no warning; and in the awkward case of a distro-packaged install with no dist-info, both miss the metadata lookup and both succeed via the import fallback. I can't find an environment where the two differ.

If you did hit a real failure, I'd like to understand the environment — there may be something else going on that's worth fixing properly.

PR #2400, opened 2026-07-02
[19.0][ADD] partner_hide_admin_contact: hide administrators' contacts from regular users

mikecolangelo:
## New module: `partner_hide_admin_contact`

In deployments where regular internal users shouldn't be aware of who administers the system (multi-tenant platforms, service providers hosting many businesses in one database), a system administrator's contact card is still visible to every internal user through Contacts and relational fields.

This module adds a global record rule that hides any contact linked to a `base.group_system` user from regular internal users:

- Administrators keep seeing every contact, including other administrators.
- Everyone keeps seeing their own contact, even if they are an administrator.
- A settings toggle (General Settings) deactivates the rule for deployments that want the stock behaviour back, e.g. to let users add an administrator as a follower.
- Implemented via a stored computed flag (`is_system_admin_contact`) so the rule's domain stays cheap to evaluate.

## Companion module: `mail_thread_followers_access_fix`

`mail.thread`'s `message_partner_ids` compute/inverse `browse()` follower partners directly, assuming every follower of a document is universally readable. Any restrictive record rule on `res.partner` (this module's, but also per-company scoping rules and the like) breaks that assumption: the moment a hidden partner follows a document, merely opening that document raises an `AccessError`.

The companion module rebuilds the follower list through `search()` instead, which enforces record rules and silently omits the followers the current user cannot see, keeping subscribe/unsubscribe semantics intact for the visible ones. It only depends on `mail` and is useful on its own for any deployment with restrictive partner visibility rules.

Note: this companion is also bundled in OCA/multi-company#1024 (draft), which needs it for the same reason. It only needs one home — happy to drop it from either PR as reviewers prefer, or move it to a more appropriate repository (e.g. OCA/social) if that's the better fit.

## Testing

- 7 tests (5 + 2), green on a fresh vanilla Odoo 19 install with only these two modules — no other OCA module involved.
- Validated manually end to end in a live multi-company deployment sandbox (admin hidden from regular users including same-company ones, colleagues still visible, own contact always visible, settings toggle verified both ways, chatter regression with an administrator follower verified).
- Pre-commit green.


legalsylvain (changes requested):
slope review / no test.

PR #2403, opened 2026-07-15
[19.0][FIX] partner_firstname: always keep explicit name

pierre-halleux:
When a partner is created from a m2o field, the search term is passed as ``default_name`` in the context and inverted into ``lastname`` by ``default_get``. Finishing the name in the create dialog does not refresh ``lastname`` so the search term ends up being saved instead.

With this change, an explicit ``name`` in ``vals`` overrides the inverted parts.


legalsylvain (approved):
lgtm. Thanks ! 
/ocabot merge patch

PR #2280, opened 2026-02-03
[18.0][FIX] partner_firstname: name inverse avoid constraint error

KristianSolkaer:
## Fix inverse name computation

This PR fixes a constraint issue in the inverse computation of the computed
`name` field.

### The issue
Each field assignment in `_inverse_name()` triggers a separate `write()` and
constraint check. Writing falsy values first could temporarily put the record
in an invalid state and raise `EmptyNamesError`.

This caused the `mail` module installation to fail when
`partner_firstname` was installed first (during the update of
`base.partner_root` / OdooBot).

### What is changed
- Write truthy inverse values first to avoid transient constraint violations

### Result
- `mail` can be installed after `partner_firstname`
- Safer inverse name computation

PR #2404, opened 2026-07-17
[17.0][FIX] partner_firstname: stabilize name inverse to avoid constraint errors

JulienMartinez:
## Fix inverse name computation

Backport of #2280 to 17.0.

Each field assignment in `_inverse_name()` triggered a separate `write()`
and constraint check. Writing falsy values first could temporarily put the
record in an invalid state and raise `EmptyNamesError`.

### DIFF
--- a/partner_firstname/models/res_partner.py
+++ b/partner_firstname/models/res_partner.py
@@ -236,11 +236,20 @@ class ResPartner(models.Model):
         return {"lastname": parts[0], "firstname": parts[1]}
 
     def _inverse_name(self):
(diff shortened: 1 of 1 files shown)


DorianMAG (approved):
LGTM

partner_firstname/models/res_partner.py
flotho, line 51: comared to https://github.com/OCA/partner-contact/blob/18.0/partner_firstname/models/res_partner.py#L62 there are small differences in the v18 version.
could you report the differences in the v17 release ?

flotho (changes requested):

PR #2408, opened 2026-07-27
[19.0][FIX] partner_shipping_policy: Update dependencies for picking policy computation

imlopes:
Using only `partner_id` on the dependency is OK but the compute isn't called when updating ONLY `partner_shipping_id` and this one is the first used on the method.


imlopes:
https://github.com/OCA/partner-contact/pull/2409

PR #2416, opened 2026-08-04
[19.0][IMP] partner_email_check: allow bypassing the checks per operation

cliffkujala:
## What

Two context keys that turn a check off for a single operation:

- `partner_email_check_skip_deliverability` — skips the DNS lookup, and keeps validation and normalization.
- `partner_email_check_skip_syntax` — skips both, exactly as disabling the company setting does.

The company settings are untouched and nothing changes for callers that do not opt out.

## Why

The checks run on every `res.partner` create and write. That is right for an address a user typed, and wrong for one the code did not collect from a user.

Odoo's incoming mail gateway creates a partner for the sender of every message it accepts, and bulk senders — HubSpot, Mailchimp, SendGrid — routinely send from a subdomain that publishes no MX record. With **Check email deliverability** enabled, storing that sender raises `ValidationError`. `mail.thread.message_route` reads any exception out of `message_new` as a misconfigured alias, so one such message produces:

- a discarded message,
- an alias flagged **Invalid**, which refuses every later message to it as well,
- and a bounce aimed at an address that by definition cannot receive it, retried until the relay gives up.

We hit that in production: a HubSpot campaign sent from `39698511m.em.knapheide.com` took a working inbound alias out of service until someone noticed and cleared the flag. Nobody's data was improved by refusing the address.

The only workaround available today is disabling the company setting, which gives up the check for exactly the addresses it was written for — the ones staff type in. A caller that knows it is storing a sender rather than user input can say so instead.

## Usage

```python
# Store the address even though its domain accepts no mail.
partner.with_context(partner_email_check_skip_deliverability=True).write(vals)

# Store it without validating or normalizing it at all.
partner.with_context(partner_email_check_skip_syntax=True).write(vals)
```

## Tests

Four added, covering that the deliverability bypass still normalizes (mixed case in, lowercase out), that it does not quietly take the syntax check with it, that it does not persist onto the record's own environment, and that the syntax key works while the company setting is on.

Documented in `readme/USAGE.md`.

### DIFF
--- a/partner_email_check/models/res_partner.py
+++ b/partner_email_check/models/res_partner.py
@@ -61,12 +61,32 @@ class ResPartner(models.Model):
         return result.normalized.lower()
 
     def _should_check_syntax(self):
+        """Whether to validate the address, and so also normalize it.
+
+        ``partner_email_check_skip_syntax`` in the context turns this off for a
+        single operation, for the same reason as
+        ``_should_check_deliverability``. It skips normalization too, exactly as
+        disabling the company setting does.
+        """
+        if self.env.context.get("partner_email_check_skip_syntax"):
+            return False
         return self.env.company.partner_email_check_syntax
 
     def _should_filter_duplicates(self):
         return self.env.company.partner_email_check_filter_duplicates
 
     def _should_check_deliverability(self):
+        """Whether to require that the address' domain accepts mail.
+
+        ``partner_email_check_skip_deliverability`` in the context turns this
+        off for a single operation. Code that stores addresses it did not
+        collect from a user needs that: the incoming mail gateway creates a
+        partner for the sender of every message, and bulk senders routinely send
+        from a subdomain that publishes no MX record, so a hard check there
+        discards the message rather than improving anyone's data.
+        """
+        if self.env.context.get("partner_email_check_skip_deliverability"):
+            return False
         return self.env.company.partner_email_check_check_deliverability
 
     @api.model_create_multi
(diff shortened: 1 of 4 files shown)


partner_email_check/tests/test_partner_email_check.py
```diff
@@ -121,3 +121,39 @@ def test_invalid_email_addresses_allowed(self):
...
+        partner = self.test_partner.with_context(
+            partner_email_check_skip_deliverability=True
+        )
+        partner.email = "Cezrik@ACOA.nrdkt"
+        self.assertEqual(partner.email, "cezrik@acoa.nrdkt")
+
+    def test_syntax_still_checked_when_deliverability_skipped(self):
+        """Bypassing one check must not quietly bypass the other."""
+        self.check_deliverability()
+        with self.assertRaises(ValidationError):
```
ivantodorovich, line 141: better to use assertRaisesRegex to ensure the exception is actually what you expect

partner_email_check/models/res_partner.py
```diff
@@ -61,12 +61,32 @@ def _normalize_email(self, email):
...
+
+        ``partner_email_check_skip_deliverability`` in the context turns this
+        off for a single operation. Code that stores addresses it did not
+        collect from a user needs that: the incoming mail gateway creates a
+        partner for the sender of every message, and bulk senders routinely send
+        from a subdomain that publishes no MX record, so a hard check there
+        discards the message rather than improving anyone's data.
+        """
+        if self.env.context.get("partner_email_check_skip_deliverability"):
+            return False
```
ivantodorovich, line 89: Maybe it's a personal take, but IMO it's better to use positive flags

```suggestion
        if not self.env.context.get("partner_email_check_deliverability", True):
            return False
```

Meaning: by default it's True, but you can set it to False explicitly

PR #2417, opened 2026-08-06
[IMP] partner_store: Move "opening_hours" from geospatial/website_geo…

LINTOANTONY007:
Upon Merging this PR,

1) Field `opening_hours` in `res.partner` model is moved from  `geospatial/website_geoengine` module to this module,  as `opening_hours` is a store related property field. So I think it must be in this module.

Please Refer https://github.com/OCA/geospatial/pull/474


LINTOANTONY007:
@pedrobaeza , Can you review this PR?

pedrobaeza:
I think that field is also unrelated to `partner_store`. You may have opening hours without stores. You may use https://github.com/OCA/delivery-carrier/tree/18.0/partner_delivery_schedule

LINTOANTONY007:
@tarteo , Can you review this PR?

tarteo (changes requested):
Makes sense to me, currently this field is only used by the geolocation store locator. It certainly makes more sense to have this field here. 

I noticed this field was never included in the `res.partner` view anyways, it was probably never used 😅.  IMO we could just remove this field, or otherwise add it to the store (`res.partner`) view

LINTOANTONY007:
IMO if `opening_hours` is removed ,  it will make a negative impact on users who are using (if any ) this feature.  So I suggest to keep the field in this module.

@tarteo WDYT ?

tarteo:
@LINTOANTONY007 Yes, we can keep it, but we need a view  for it as there is none atm, otherwise the user can't use it.

LINTOANTONY007:
> @LINTOANTONY007 Yes, we can keep it, but we need a view for it as there is none atm, otherwise the user can't use it.

Got it.

LINTOANTONY007:
@tarteo , Added view for the field,  `opening_hours`.

image

image

PR #2419, opened 2026-08-10
[19.0][MIG] partner_identification_unique_by_category: Migration to 19.0

kalpeshgajera-uncannycs:


tonygalmiche (approved):
Tested on Odoo 19.0:

- Duplicate ID number rejected within the same category when "Enforce unicity" is enabled
- Same ID number accepted across different categories (uniqueness correctly scoped per category)
- Enabling "Enforce unicity" on a category that already contains duplicates is correctly rejected

No issues found. Works as expected.

PR #2420, opened 2026-08-12
[19.0][FIX] properly display phone2 in partner views

NICO-SOLUTIONS:
Before:
before_list
before_kanban
before_form
before_simple_form

After:
after_list
after_kanban
after_form
after_simple_form

Here, the mobile number extension is activated as well, so please don’t get irritated by it.

### DIFF
--- a/partner_phone_secondary/README.rst
+++ b/partner_phone_secondary/README.rst
@@ -60,7 +60,8 @@ Authors
 Contributors
 ------------
 
-- Iván Todorovich <ivan.todorovich@gmail.com>
+-  Iván Todorovich <ivan.todorovich@gmail.com>
+-  Nils Coenen <nils.coenen@nico-solutions.de>
 
 Maintainers
 -----------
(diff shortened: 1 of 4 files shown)


ivantodorovich:
Thanks!

Could you share a screenshot of before vs after in the PR description? 🙏🏻

NICO-SOLUTIONS:
> Thanks!
> 
> Could you share a screenshot of before vs after in the PR description? 🙏🏻

@ivantodorovich 

moved it to the first PR comment...

ivantodorovich:
Thanks!

Could you add a `priority` to the view, so that when `partner_mobile` is also installed the `phone2` lands directly after `phone`, and before `mobile` ?

NICO-SOLUTIONS:
> Thanks!
> 
> Could you add a `priority` to the view, so that when `partner_mobile` is also installed the `phone2` lands directly after `phone`, and before `mobile` ?

@ivantodorovich 
Done. Since partner_mobile doesn't define a priority, it defaults to 16. I set the priority of these views to 20, so they are applied afterwards and phone2 ends up directly after phone and before mobile. I also updated the "after" screenshots in the first commit comment.

ivantodorovich (approved):
Thanks!!

PR #2421, opened 2026-08-12
[18.0][FIX] partner_firstname: honor is_company in create values

marcos-mendez:
`create()` decides whether the partner being created is a company only from `vals.get("company_type")`. A create call that passes `is_company=True` together with a `name` — a perfectly valid combination, and what modules that don't know about `company_type` naturally send — has the name split into `firstname`/`lastname` as if the partner were a person:

```python
partner = env["res.partner"].create({"name": "dummy company", "is_company": True})
partner.firstname  # "dummy"
partner.lastname   # "company"
```

This was found while migrating the cooperator module family to 18.0: `get_partner_company_vals()` passes `is_company` and company partners came out with split names.

The fix makes `create()` also honor `is_company` in the values. Two regression tests included (one for `is_company` in vals, one confirming the existing `company_type` path still works).

Module test suite after the change: 51 tests, 0 failures.

### DIFF
--- a/partner_firstname/models/res_partner.py
+++ b/partner_firstname/models/res_partner.py
@@ -59,7 +59,9 @@ class ResPartner(models.Model):
         created_partners = self.browse()
         for vals in vals_list:
             partner_context = dict(self.env.context)
-            is_company = vals.get("company_type") == "company"
+            is_company = vals.get("company_type") == "company" or vals.get(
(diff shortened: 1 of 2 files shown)


legalsylvain (approved):
LGTM.

Thanks.

/ocabot merge patch

PR #2427, opened 2026-08-29
[FIX] partner_multi_relation: clear the menu left naming the action 19.0 removed

moylop260:
Upgrading a database that predates 18.0 leaves the backend blank for every user. No apps, no menus, and in the browser a SyntaxError about HTML where JSON was expected, because /web/webclient/load_menus answers 404:

    Record does not exist or has been deleted. (Record: ir.actions.act_window(803,))

The menu is menu_res_partner_relation and the action is action_res_partner_relation_all.

Three things have to line up for this, and each one is reasonable on its own.

The menu has been a plain container since 18.0. It is declared with a name and a parent, no action, and the two menus that do carry actions hang underneath it:

https://github.com/OCA/partner-contact/blob/7e3e10eca258adf40a1f245b539401d77d644798/partner_multi_relation/views/menu.xml#L3-L14

But a <menuitem> without an action attribute emits no such field, and _load_records only writes the fields it is given, so the value an older version stored is never rewritten:

https://github.com/odoo/odoo/blob/27391b2f112039e52198ddf590548e7baa5968dc/odoo/tools/convert.py#L339-L353

It sits in the column, valid, claimed by no data file.

19.0 is what makes it dangle. In 18.0 the action lived in views/res_partner_relation_all.xml and targeted res.partner.relation.all:

https://github.com/OCA/partner-contact/blob/7e3e10eca258adf40a1f245b539401d77d644798/partner_multi_relation/views/res_partner_relation_all.xml#L71-L78

That whole file is gone in 19.0, and the model with it: migrations/19.0.1.0.0/pre-migration.py drops the SQL view the action was there to show.

https://github.com/OCA/partner-contact/blob/bce3eb448e580ad58eeb43c714d0dfa87710cb08/partner_multi_relation/migrations/19.0.1.0.0/pre-migration.py#L13

The replacement is a different record over a different model, action_res_partner_relation on res.partner.relation, so the old one is deleted rather than renamed:

https://github.com/OCA/partner-contact/blob/bce3eb448e580ad58eeb43c714d0dfa87710cb08/partner_multi_relation/views/ir_actions_act_window.xml#L8-L11

That deletion is correct and this patch does not touch it. The action cannot be kept or renamed; its model is gone.

And nothing connects the two. ir_ui_menu.action is a reference column holding '<model>,<id>' as plain text, with no foreign key, so no cascade fires and no constraint complains. The menu keeps naming a record that no longer exists, and one such menu is enough to take the whole backend down.

Measured on a production database upgraded from 18.0: the action exists before the upgrade and after the Odoo Upgrade Service, with no dangling menu on either side; it is gone after the module updates, with menu_res_partner_relation left pointing at it while its two siblings resolve.

The action is cleared, not repointed. Guessing which action the menu meant, by matching it against whichever one the module declares now, is wrong exactly when it matters: a menu that silently opens the wrong action is worse than one that opens nothing. The XML already says what this menu is -- a container.

The sweep is scoped to this module's own menus. A missing reference here is a class of damage rather than one bad row, and any renamed action can produce another, so it looks for what does not resolve instead of the known id; but cleaning menus another module owns would be reaching outside what this migration can reason about.

It runs post-migration and not pre-migration. The records are deleted while this module's data reloads, so a pre-migration would find the action still alive and nothing to clean.

The test walks every menu this module owns and resolves its action, since one stale reference is all it takes.

### DIFF
--- a/partner_multi_relation/__manifest__.py
+++ b/partner_multi_relation/__manifest__.py
@@ -2,7 +2,7 @@
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
 {
     "name": "Partner Relations",
-    "version": "19.0.1.1.1",
+    "version": "19.0.1.1.2",
     "author": "Therp BV,Camptocamp,Odoo Community Association (OCA)",
     "website": "https://github.com/OCA/partner-contact",
     "complexity": "normal",
--- a/partner_multi_relation/migrations/19.0.1.1.2/post-migration.py
+++ b/partner_multi_relation/migrations/19.0.1.1.2/post-migration.py
@@ -1,0 +1,56 @@
+# Copyright 2026 Vauxoo <https://www.vauxoo.com>.
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+import logging
+
+from openupgradelib import openupgrade
+
+logger = logging.getLogger(__name__)
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    """Clear this module's menus that still name an action 19.0 removed.
+
+    ``menu_res_partner_relation`` has been a plain container since 18.0, but
+    databases coming from older versions still carry the action it used to have,
+    ``action_res_partner_relation_all``. A ``<menuitem>`` without an ``action``
+    attribute emits no such field, and ``_load_records`` only writes the fields it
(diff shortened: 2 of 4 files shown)


BhaveshHeliconia (approved):
Looks good to me thanks!

BhaveshHeliconia:
could you please check the failing pre-commit?

PR #2436, opened 2026-09-17
[15.0][IMP] partner_data_vies_populator: backport address parsing from v17

hbrunn:
I also added a commit to patch away the remote calls during tests, which should be cherry-picked to the other versions to avoid failing tests when VIES is overloaded, which happens regularly, amongst others because of CI doing it wrong like here


dreispt (approved):
Looks great, thank you!

luisDIXMIT (approved):
Code review, LGTM!

